PR #93, opened 2014-11-06
Stay on Tab that was opened in Detailview, when clicking Edit-Button

gunnicom:
If you got Tabs on your Detailview, this patch opens the tab you were looking at, when clicking the Edit-Button.


gunnicom:
Any chance to get that included?

gunnicom:
I will add new Pull Requests, as i found that my implementation was incorrect. It did only work in developer mode due to template caching. Found the correct way to do now.

PR #98, opened 2014-12-02
Avoid warnings for and detect the problem: Warning: json_encode(): Inval...

JoshuaQ:
...id UTF-8 sequence in argument occurred

### DIFF
--- a/include/JSON.php
+++ b/include/JSON.php
@@ -68,7 +68,11 @@ class JSON
      */
     public static function encode($string, $addSecurityEnvelope = false, $encodeSpecial = false)
     {
-        $encodedString = json_encode($string);
+        try{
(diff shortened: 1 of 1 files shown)


samus-aran:
Hi @JoshuaQ 

This commit does not provide any functionality - unless you have a test case this works?  I have tested this using non UTF-8 strings and no Exception is raised.  This is because json_encode returns boolean false on 'error' so there is no exception to catch.  

However, you have raised a point that we need to consider this.  Perhaps include the json_last_error to confirm the error message.  I'm closing this pull request but hopefully you can contribute more to this issue.

Oh, and just to request that all new PR are to be made to 'hotfix' branch than the master branch.  Thanks.

PR #111, opened 2015-01-15
- Fixed an issue with a custom button does not get displayed in sub panel

menshinobi:
If you have a custom module and a custom button in the subpanel of that module, the custom button won't be displayed because of this if function.

### DIFF
--- a/include/ListView/ListView.php
+++ b/include/ListView/ListView.php
@@ -504,10 +504,12 @@ function process_dynamic_listview($source_module, $sugarbean,$subpanel_def)
                             $button_contents[] = $_content;
                             unset($_content);
                         }
-                        else
-                        {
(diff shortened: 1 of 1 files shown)


samus-aran:
Hi @menshinobi 

Thanks for supplying this fix.  I can confirm from my own testing that this is the case and would be happy to merge this into the project.

However, a couple of things regarding to the PR is that we require that any bug fixes are to be merged into 'hotfix' branch rather than 'master'.  Unfortunately Github does not have a way of changing this so a new PR must be made.

Second point is that the CLA must be signed to provide.  You can do it before your next PR - its a matter of clinking through to the Details of this failed check here and accepting the agreement.

Again, thank you for your contribution and let us know if you can not make a new PR and we can make a commit for you.

PR #128, opened 2015-02-24
Update controller.php

dbphoto84:
Fixed a major bug where report checkboxes where not checked.


lewisje1991:
Hi dbphoto84,

This fix is in the wrong place it should be done when the pdf content is generated.

Also please note make your requests against the hotfix branch.

Thanks
Lewis

PR #137, opened 2015-03-03
Fix HTTP 500 with +SymLinksIfOwnerMatch.  Email Open Notification (MDN/Read receipt).

### LINKED ISSUE #169: HTTP 500 error on upgrade or new install of 7.2
Hosting control panel has tightened up security.  +FollowSymLinks is forbidden, replaced by +SymLinksIfOwnerMatch.  Upgrade to 7.2 or new install of 7.2 overwrites this change.

See also

https://www.drupal.org/node/1269780
https://my.bluehost.com/cgi/help/search?sort=&search=SymLinksIfOwnerMatch
https://my.hostmonster.com/cgi/help/htaccess
http://www.virtualmin.com/node/24260

Proposed solution:
Change to +SymLinksIfOwnerMatch in core installer and upgrade code.

More info:
http://blog.sucuri.net/2013/05/from-a-site-compromise-to-full-root-access-symlinks-to-root-part-i.html

Shared hosting is still a mass market and as more and more sharehosters disable +FollowSymLinks for good reasons.
So in fact
- +SymLinksIfOwnerMatch is the evolving option you can safely rely on
- +FollowSymLinks you can not rely on and if you need it you know what to do.

Here is the PR to fix this issue.
https://github.com/salesagility/SuiteCRM/pull/137

chris001:
1. Fixes #169  +FollowSymLinks is deprecated and causes HTTP 500 Internal Server Error crash when enforced by secure server control panels such as VirtualMin.  
   Use of +SymLinksIfOwnerMatch instead is required by current security standards.
2. Fixes issue, where user asked for return receipts on email messages sent out by the CRM system.  Add enhancement. Email Open Notification (also called MDN or "read receipt").  With one setting in config_override.php, all your emails that you write in the Email module will request an "open notification" from the receiver.  If the receiver accepts, you will receive a message back from their email system when they open or delete the email message.  They can always refuse to provide the email open notification (due to privacy rules etc).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,3 +84,5 @@ branches:
     - hotfix
     - develop
     - /feature.*/
+    - /patch-*/
+    
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,86 @@
+language: php
+
+php:
+  - 5.3
+  - 5.4
+  - 5.5
+  - 5.6
+  - hhvm
+  - 7.0
+  - 7.1
+
+sudo: false
+
+cache:
+  directories:
+    - $HOME/.composer/cache
+    - $HOME/.app/cache
+
+env:
+  global:
+    - deps="no"
+    - SYMFONY_VERSION=""
+    - PHP_CS="yes"
+    - SYMFONY_DEPRECATIONS_HELPER=weak
+
+matrix:
+  fast_finish: true
+  include:
+    - php: 5.3
+      env: COMPOSER_FLAGS="--prefer-lowest"
+    - php: 5.6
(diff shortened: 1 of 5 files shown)


horus68:
@chris001 can you please commit this PR to the proper branch (hotfix)
Probably this is why its not merged since 2015!

horus68:
@mattlorimer this needs an hand on clahub agreement (probably the old bug on validation from them)

Dillon-Brown:
Original issue has been resolved for the hotfix-7.8.x branch. The second part of this PR seems to be an unrelated enhancement and will need to be redone for the develop branch.

PR #193, opened 2015-04-03
Replace all "SELECT TOP $count" with "DECLARE @topcount INT SET @topcoun...

klou:
...t=$count SELECT TOP (@topcount).  This eliminates tablescans (for MSSQL) and significantly improves performance.

On my system, this dropped some query times from >30s to <1s.

### DIFF
--- a/include/database/MssqlManager.php
+++ b/include/database/MssqlManager.php
@@ -454,7 +454,7 @@ class MssqlManager extends DBManager
                         if (!empty($orderByMatch[3])) {
                             $selectPart = array();
                             preg_match('/^(.*)(\bFROM\b.*)$/isU', $matches[2], $selectPart);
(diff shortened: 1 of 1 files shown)


chris001:
Have a look at this.
Looks like a brilliant way to get full support for all databases, Microsoft SQL, Postgres, Oracle, IBM DB2, etc.
http://www.slideshare.net/jakubpas/implementation-of-doctrine-orm-in-sugar-crm-75
Slide presentation on adding Doctrine to SugarCRM 7.5.  
Explains how SugarQuery is a nice addition to Sugar 7.5, but lacks 7 important features compared to Doctrine.  
It's optional, and requires no changes to the core SuiteCRM/SugarCRM code.  
8 advantages including 20-400x faster speed, transactions that make failed commits simple to undo, etc.

Issue #175, opened 2015-03-17
Create Project from Project Templates


Mausino:
If i created Project from Project Templates on my localhost, i saw this:

Warning: Invalid argument supplied for foreach() in C:\xampp\htdocs\EEE\crm\custom\modules\ProjectTask\updateDependencies.php on line 23 Warning: Invalid argument supplied for foreach() in C:\xampp\htdocs\EEE\crm\custom\modules\ProjectTask\updateDependencies.php on line 23 Warning: Invalid argument supplied for foreach() in C:\xampp\htdocs\EEE\crm\custom\modules\ProjectTask\updateDependencies.php on line 23 Warning: Invalid argument supplied for foreach() in C:\xampp\htdocs\EEE\crm\custom\modules\ProjectTask\updateDependencies.php on line 23 Notice: Undefined offset: 2 in C:\xampp\htdocs\EEE\crm\modules\AM_ProjectTemplates\controller.php on line 106 Notice: Undefined offset: 2 in C:\xampp\htdocs\EEE\crm\modules\AM_ProjectTemplates\controller.php on line 107 Fatal error: Call to a member function format() on boolean in C:\xampp\htdocs\EEE\crm\modules\AM_ProjectTemplates\controller.php on line 108

Also same issue is on my live version, and it created the project. If i going ot project, it created me Project but only 2 Project Tasks (in Template was 4).

I tried instal fresh new 7.2.1 version, but issue persist

I searched for answer in forum, but it wasn't there any solution

https://suitecrm.com/forum/suitecrm-7-0-discussion/4022-project-task-error#15921

SAME ISSUE on demo site of official SuiteCRM page !!!

http://screencast.com/t/wPcOhKQSE4Q

Miky

PR #202, opened 2015-04-10
Fix for #175 fatal error for project task end dates

aromasko:
Include end date of project tasks in array in the case when there is more than one project task in the template.

https://github.com/salesagility/SuiteCRM/issues/175


mattlorimer:
merged to hotfix-7.2.2

Issue #204, opened 2015-04-10
My Project Tasks Dashlet doesn't show tasks created using Project Task Template


aromasko:
When using a Project Task Template to create a project the tasks do not show up in the My Project Tasks dashlet after creation.

They will show up, however, when the task PERCENT_COMPLETE (% Cpl) has a value.

\modules\AM_TaskTemplates\vardefs.php sets 'percent_complete' type as varchar but in \modules\ProjectTask\vardefs.php type is set to int

PR #205, opened 2015-04-10
Fix for #204 TaskTemplate percent_complete type mismatch

aromasko:
When using a Project Task Template to create a project the tasks do not show up in the My Project Tasks dashlet after creation.

They will show up, however, when the task PERCENT_COMPLETE (% Cpl) has a value (which should be an int).

\modules\AM_TaskTemplates\vardefs.php sets 'percent_complete' type as varchar but in \modules\ProjectTask\vardefs.php type is set to int


mattlorimer:
merged into hotfix-7.2.2

PR #209, opened 2015-04-17
Replacing ¿4294967295? with '-1' ins spanish automatic/non refresh/manual checks

arraintxo:
Spanish dashlets on home page where broken when "do not refresh automatically" was selected on preferences popup.
App was unusable as dashlets with that options selected where continuously refreshing.


mattlorimer:
Merge into hotfix-7.2.2

PR #223, opened 2015-04-29
Update AOW_WorkFlow controller.php

mdemaree99:
Dynamic Dropdown does not behave like standard Dropdown field.
This update will allow user to select multiple values as needed.

See Bug 222:  Workflow doesn't work correctly with dynamic dropdowns bug


mdemaree99:
Please reevaluate.   CLA for salesagility/SuiteCRM has been signed.

Issue #78, opened 2014-09-26
Error with spanish translation


ErMaqui:
Hi,

I have lot of cases when use the CRM who i see a "undefined" message when i'm using CRM on Spanish. As i can see, the js file with strings lacks some of them (They have one array empty).

Tested on latest version.

cmarrero:
Hi,

The "undefined" still appears even if you set to 775 the permision for the cache folder.


![2015-01-20 00-52-15](https://cloud.githubusercontent.com/assets/212819/5812741/dc7304f0-a03e-11e4-92d8-f95065b9ea85.png)
![2015-01-20 00-51-21](https://cloud.githubusercontent.com/assets/212819/5812742/dc75fe3a-a03e-11e4-937d-156ef00b085c.png)
![2015-01-20 00-51-10](https://cloud.githubusercontent.com/assets/212819/5812740/dc6e0432-a03e-11e4-84f3-557452da5c10.png)

fgpalacios:
edit the config.php of sugar and replace


'default_permissions' =>
array (
'dir_mode' => 493,
'file_mode' => 420,
'user' => '',
'group' => '',
)

cmarrero:
Hi,
I try the solution about permissions, rebuilt the languages in repair and still the same problem
Regards

cmarrero:
Hi,

I see this error in the console every time a page is access:

SyntaxError: expected expression, got ')'
SUGAR.language.setLanguage('app_strings', );
es_es.j...zimvnow (línea 1, col 42
Route to file= cache/jsLanguage/es_es.js?v=CbTah9NLNnDAATDzimvnow

Regards

cmarrero:
I have found this on the forums 

https://suitecrm.com/forum/bug-tracker/2559-calendar-form-empty-with-no-english-language

"
I've fouun the reaons why of this problem ... but not the solution. The problem (seems) caused by the following statement
SUGAR.language.setLanguage('app_strings', );
which is repeated twice at the beginning of the file..cache\jsLanguage\it_it.js.
I've tried to remove this double statement and everything works fine.
"

In some part SuiteCRM is inserting for no english lenguaje  this  line: SUGAR.language.setLanguage('app_strings', ); 

If you see the file in the cache folder "cache/jsLanguage/es_es.js" has no the label "LBL_LOADING_PAGE" but that is include in the file "include/language/es_es.lang.php"

Regards

cfiorillo:
Hi,
[...]

PR #226, opened 2015-05-01
Fix security groups language files (PHP 5.5 compatibility)

JanSiero:
When using SuiteCRM in the spanish language, the JavaScript language files are not compiled correctly, causing the jscalendar to show english titles. (english month names instead of the spanish ones)

This is caused by stray characters in Latin1 encoding in UTF8 language files. the JSON encoder stumbled on these characters, because it only accepts UTF8 as of PHP version 5.5.

These Latin1 characters are present in the Security Groups language files of es, fr, it and pt.

The diffs in Github don't show the changes, because apparently github adapts to the character encoding of each line. However, when you inspect the previous version of the file (showed as latin1 in Github) and the new file (showed as UTF8), the difference is shown.


JanSiero:
fixes #78

PR #244, opened 2015-05-12
allow group dates by day,month or year on reports

guillermofr:
Reports module does not allow group dates by month, day or year. This change allow it, but use this option only with date fields =)


horus68:
@guillermofr can you take a new look int this PR and submit to the proper branch?
Note: you shouldn't include translations.

guillermofr:
i dont rememeber this commit xD

gunnicom:
Such a good enhancement, and now lost because suite developers concentrate on a new theme, not on issues and pull request :(

mattlorimer:
@gunnicom  this was already achieved in a slightly more flexible way, to work in more scenarios,

gunnicom:
@mattlorimer : Then communication should be improved, because "can you take a new look int this PR and submit to the proper branch?" does not sound like it is implemented. And if its implemented, and the developer does a new pull request, then the pull request would be declined leading to more frustration to people who want to participate.

PR #251, opened 2015-05-20
Fix Post-Auth RCE vulnerability

0x27:
Hotfix for the post-auth RCE vulnerability disclosed here: https://github.com/XiphosResearch/exploits/tree/master/suiteshell

Obviously, its not a real fix - the uploader needs to be completely rewritten to use whitelisting instead, amongst other things, but I am not a PHP developer with time to do this.

### DIFF
--- a/include/utils.php
+++ b/include/utils.php
@@ -330,7 +330,7 @@ function get_sugar_config_defaults() {
 	return_session_value_or_default('translation_string_prefix', false),
 	'upload_badext' => array (
 	'php', 'php3', 'php4', 'php5', 'pl', 'cgi', 'py',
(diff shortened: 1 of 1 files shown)


mattlorimer:
thanks for highlighting this issue

0x27:
For the future, does SalesAgility (as the SuiteCRM mantainer) have a security contact and GPG key? Could you publish such information on your main site? As mentioned in the disclosure timeline I made a couple of efforts to get this fixed before its deadline expired, but seemingly I never was put in touch with whoever is responsible for the security of the project and all that. Its possible more bugs are found in future (and there have been previous RCE issues due to unserialize) so it would be helpful :)

PR #258, opened 2015-05-25
Remove orphan language files

horus68:
Those language files are not included in any language pack. They has less then 15 files by language, usually found in custom folders. There are also wrong language code used as for German ge_ge (should be de_DE that also have files in here!) and Ukrainian ua_ua (should be uk-UA)
They need to be installed as a module (preference as a full pack).
Language contents are uploaded to https://www.transifex.com/projects/p/suitecrm to allow full language pack creation.
Only en_us , es_es and ru_ru will be available in SuiteCRM zip master download and as installed languages.
Note: SalesAgility should consider to decouple spanish and russian languages too as an instalable module!


horus68:
bump! Can you merge this pull request before V7.3 ? Its an easy one to include and then it will not create issues with language packs that users will install!

horus68:
Should already been removed in this PR:
Language changes by johnM2401 · Pull Request #395 - https://github.com/salesagility/SuiteCRM/pull/395

PR #264, opened 2015-05-26
A few words: Spanish translation

leonardobarrientosc:
Theres is a lot of Failures related to Contributor License Agreement, what should we do?


horus68:
working on the translation and fixing thing in github will never get you in a good translation with people helping out. Non english languages should be created as installable packs and created from Transifex system https://www.transifex.com/projects/p/suitecrm/

PR #272, opened 2015-06-04
Fixing javascript error in campaign wizard non-emailed based campaign

aaronhagopian:
Found this error in the wizard when you try to add target lists either new or from the pop-up, you would get a javascript error.


mattlorimer:
Merged into hotfix-7.2.3 thanks

Issue #228, opened 2015-05-04
AOW: undefined index (50% bug, 50% filling up the logs)


jobst:
Hi

In my php logfile I have loads (and I mean loads) of
`PHP Notice:  Undefined index: module in /PATH_TO_SUITECRM/modules/AOW_WorkFlow/AOW_WorkFlow.php on line 134`
It stems from an incorrect IF statement
`if($_REQUEST['module'] != 'Import'){`
it should be
`if( isset($_REQUEST['module']) && $_REQUEST['module'] != 'Import'){`

Jobst

PR #322, opened 2015-07-28
I have taken out the extra error lines within the log for schedulers

darthtang:
#228 has been seen to and fixed. No more extra error lines in the log


mattlorimer:
This would prevent the workflow running if $_REQUEST['module'] is not set

Issue #321, opened 2015-07-28
Hitting enter in the password input saves the user but not the password


JimMackin:
In the edit user screen if you enter a new password for a user it will save the user but not the password. You need to hit "Save" in order to save the password (and then get the confirmation popup).

This should probably either behave the same in both cases or enter shouldn't cause the submit in the password field.

darthtang:
Also note:

When filling in fields on the 'User Profile' tab when editing a user. And then if you hit space to save, the system WILL NOT save! The system will only save anything if you click on the 'Save' Button at the top. 

I also suggest maybe an 'Are you sure you want to save?' confirmation box before each save! This will add to a user friendly environment :+1:

Dillon-Brown:
@Mausino I can replicate the same, will put up a fix.

Mausino:
Great.. the patch is working 👍 @Dillon-Brown should it get some milestone ?

Mac-Rae:
Confirmed in latest 7.10.x 

Doesn't save however does open user email settings located on the other page :/

Mausino:
@Dillon-Brown why wasn't this solution merged yet? it needs something?

Dillon-Brown:
@Mausino Hmm, not sure. I might have just missed it. Either way, it will need re-based to hotfix-7.10.x and the code-review comment completed. I'll take a look when I get a chance. Thanks!

PR #323, opened 2015-07-28
Saving fields when user hits 'Enter' when editing a CRM-User

darthtang:
Saving fields when user hits 'Enter' when editing a CRM-User

solves issue #321


mattlorimer:
think you need to remove the console log

PR #339, opened 2015-08-08
PHP 5.5 compatibility

JanSiero:
Includes 2 commits that concern the compatibility to PHP 5.5

 - Issue with strtolower for non-latin characters: they lead to characters that cannot be parsed by JSON. JSON doesn't accept non UTF-8 since PHP 5.5
 - Issue with Module Builder/Studio concerning the creation of fields: Vardefs are being written and reread. The Zend op-cache causes the cached version to be read, instead of the newly written file

Please ignore the first two commits.


mattlorimer:
Thanks merged into the next Hotfix

Issue #342, opened 2015-08-10
td width truncates label in resource chart


2007H3:
in Chrome Version 44.0.2403.125 m

The Project> Resource Chart has truncated the top-left label, as shown below.

![reschart](https://cloud.githubusercontent.com/assets/12518763/9185211/9367288c-3f7f-11e5-810f-6cf8721ee8c0.jpg)

digitalmcgrath:
Pull request for this issue has been submitted. https://github.com/salesagility/SuiteCRM/pull/343

PR #343, opened 2015-08-11
Fixes Issue #342

digitalmcgrath:
Removed left-padding from project resource chart so "previous month" displays correctly.


mattlorimer:
Thanks, merged into the next Hotfix

PR #355, opened 2015-08-17
Annoying php notice in log file

gunnicom:
Suppresses php notice that fills apache log files.


mattlorimer:
Thanks, Merged into the next hotfix

PR #382, opened 2015-08-20
 Fix SugarCRM Bug and add Security Group support to listviewACLHelper

eggsurplus:


mattlorimer:
Thanks merged into the next hotfix

Issue #364, opened 2015-08-19
Single click on X (close task) in task listview causes HTTP 404 error


chris001:
Create a new task, if there is not at least one already listed, and fill in minimal information necessary to make the new task.
Save it.
Single-click on the X icon (close task) in the task ListView, 
A popup asks "Do you want to close this task? Yes No", 
Before you can respond, the page instantly redirects to an invalid URL, "(suitecrm base url/undefined" which rightfully produces a HTTP 404 error, since that URL does not actually exist.

ghermans:
I have the same issue to close a call.

Mausino:
Hi,
we had same issue, and my programmer fixed it in this way (same issue was for calls also)
sorry i am not programmer, i can only post it, but sombody can fix it in code :) :1234: 
Here is message from him:

-------------------------------------
the bug was introduced by Inline edit feature
and appears when inline edit is activated on list view
so to fix the bug i changed a suitecrm file, not a custom file

the file changed is include/InlineEditing/inlineEditing.js

around line 63 I've added this code:   

   // Fix issue when click a close ( X ) button
   if (typeof linkUrl === "undefined") {
    return false;
   }

after line:
e.preventDefault();

--------------------------------------------

Miky

PR #397, opened 2015-08-24
Fix for #364 Click X redirects to undefined 404

chris001:
Proposed fix for issue #364 (click X icon to close redirects to bad url /undefined causing a 404 http error.


mattlorimer:
thanks merged into the next hotfix

PR #404, opened 2015-08-26
Skip .git folder on file system checks

gunnicom:
.git folder does not need to be checked for Upgrades.


mattlorimer:
thanks merged into the next release

PR #408, opened 2015-08-26
Only display Menu Buttons if there is at least one button

gunnicom:
If a Module defines 'buttons' => array() in detailviewdefs, suitecrm displays the list without buttons (If "Display actions within menus" is active ) This gives a minor layout error.


gunnicom:
This does not work anymore in 7.5.1 like intended. Have to look into it again.

PR #421, opened 2015-08-31
ERROR:Reports with parameter doesn't work in Dashlet

mm-senguel:
The Reports-Parameter-Value is in Array Index 1


mattlorimer:
This makes no sense if you set the parameter_value to a hard coded '1' it will not work correctly, it will only work for the second parameter.

Issue #400, opened 2015-08-25
Alerts check caught in loop


JimMackin:
The alerts check periodically checks for new alerts. 
The result of this isn't checked. This means that if the users session expires this page will be loaded into the alerts div which loads the alert check which starts a chain of calls which slows down the browser.

The check shouldn't load the page if it's a login redirect. Even better would be to have a single instance of the update variable so that if the alerts javascript is accidentally loaded more than once the calls still only happen at the specified period.

danieljsamson:
I will take a look at it.

danieljsamson:
I added the fix into multiple alerts for 7.4.

danieljsamson:
I have made a merge request for 7.3.2

PR #424, opened 2015-09-01
issue #400 fix

danieljsamson:
Alerts now looks for the login form then stops updating the alert manager.


danieljsamson:
Also you will need to rebuild the js from js source...

PR #447, opened 2015-09-07
Fix for issue 446

bdokimakis:
Inline editing does not respect link target attribute.

https://github.com/salesagility/SuiteCRM/issues/446


mattlorimer:
thanks, merged into the next hotfix

Issue #449, opened 2015-09-07
City field missing in lead conversion


vic-t:
When converting a lead to contact, the city field is not presented in the form. Still, the city is correctly taken over to the contact (if there was one in the lead).

PR #459, opened 2015-09-11
Adding the City field to the Lead Convert form in the Contact section.

digitalmcgrath:
This fixes issue #449


mattlorimer:
Clanhub still seems to be having issue, so I will manually include this, but thanks for your input

PR #469, opened 2015-09-16
Issue 360

vladbar:


mattlorimer:
are these now in other individual pull request

PR #470, opened 2015-09-16
Scrolling in Edit of Layout in Studio/ModuleBuilder not possible

gunnicom:
Same bug and fix like here, just other theme:
https://github.com/salesagility/SuiteCRM/commit/c0b3fd5764e937f59ebaa094949237b52e05107e


mattlorimer:
Thanks, Merged into the next hotfix

Issue #278, opened 2015-06-11
suitecrm version 7.2.2 - studio: file name length


selimerunkut:
In studio => module => create field: 
if the fieldname is too long, its cut and no user message is displayed.

It would be nice if the field name length could be longer and if its too long that a user message is displayed.

Mausino:
as wrote @samus-aran here #2597

> I have begun testing this but I am unable to retrieve the max length error. The input field itself only allows 25 characters max so why require an validate message? Can you demonstrate this validation to myself to pass testing.

I think this can be closed and only problem can be if will somebody move other CRM to SuiteCRM. In this case i have not any knowledge how will SuiteCRM behave. (if system will cut after 25 letter other letters or we get error)

PR #479, opened 2015-09-21
Fix for issue #278

vladbar:


mattlorimer:
should update the jssource

Issue #380, opened 2015-08-20
Double save required for Studio Layouts


johnM2401:
When saving a studio layout it seems to revert back to how it was before you changed it, so you have to change then Save&Deploy twice. Upon the first save&Deploy, it does update the Layout but simply shows you the previous one.

darthtang:
I am unable to reproduce bug here. @willrennie have you been able to reproduce this bug?

willrennie:
Yes, all studio issues are verified hence flagged with High Priority and Bug labels.

PR #485, opened 2015-09-21
Fix for issue #380

vladbar:


vladbar:
Hi, I retested this code and it works fine. It actually requires less steps, so I  submitted a new fix.

PR #493, opened 2015-09-22
Vardefs definitions and custom data from view to JS DOM

Abuelodelanada:
When developing some complex JS functionality is very useful to have module vardefs definitions in the DOM so you can read it from JS.

view edit php

accounts_edit

view edit php 2

accounts_edit_2

### DIFF
--- a/include/MVC/View/SugarView.php
+++ b/include/MVC/View/SugarView.php
@@ -84,6 +84,8 @@ class SugarView
     var $fileResources;
 
     private $settings = array();
+    private $_js = array();
+
     /**
      * Constructor which will peform the setup.
      */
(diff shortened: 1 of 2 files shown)


Abuelodelanada:
Hi! 
do you need some modification or explanation in order to aprove this pull request?

horus68:
@Abuelodelanada I'm not on the dev team but can you commit this PR to the proper branch (hotfix)?
Hope then it gets proper attention!

Abuelodelanada:
@horus68 ok, just give me some days, and I'll do it.!

Abuelodelanada:
@horus68 Here you have :D https://github.com/salesagility/SuiteCRM/pull/2440

Issue #413, opened 2015-08-27
Popup Search View - Relationships are not working


selimerunkut:
Example:
Studio => add "Product Category [aos_product_category_name]" (aos_product aos_product_categories relationship) to  Products > Layouts > PopupView > Popup Search

Quote  EditView => "Line Items" => click the arrow to open "Popup Search"
"Product Category" field should be visible but upon selecting a product category nothing happens.

willrennie:
Related to #116

shogunpol:
@selimerunkut , this issue has been tested on recent version of SuiteCRM(7.7.4), and it not appear.
Can you replicate the issue in recent version of SuiteCRM?
Thanks for your contribution.

ijdavie:
Due to no activity to this issue, we will close this if more information is given and the issue still exists this can be reopened to be resolved. 
Thanks

PR #494, opened 2015-09-23
Fix for issue #413

vladbar:


mattlorimer:
This seems too specific, there must be a more general revolution.

Issue #448, opened 2015-09-07
AOR Ignoring Roles


willrennie:
If a user has Export Rights set to 'None' for the Reports module, they can still Export a report and Download the Report as a PDF.

PR #500, opened 2015-09-27
Fix for issue #448

vladbar:


JimMackin:
Fixes should go against the hotfix branch.

PR #504, opened 2015-09-29
CSS Fixes

johnM2401:
A few changes that have come from the Forums.


mattlorimer:
you can't change the link on the home button to directly link to the 'Home' module, as the home module is configurable and may not be 'Home'

PR #509, opened 2015-09-29
FIX #508 - Now we can load multiple JS files from the constructor of the view and get only one minifyed

### LINKED ISSUE #508: Add multiple JS files to a view in only one minifyed JS file.

editviewdefs_accounts

- Each file is not minifyed.

view edit php_accounts

Abuelodelanada:
The minifyed JS file is saved in:
cache/themes/<Theme>/js/

See #508 for details

### DIFF
--- a/include/MVC/View/SugarView.php
+++ b/include/MVC/View/SugarView.php
@@ -83,6 +83,7 @@ class SugarView
     var $responseTime;
     var $fileResources;
 
+    private $_js = array();
     /**
      * Constructor which will peform the setup.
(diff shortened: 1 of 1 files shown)


Abuelodelanada:
Hi! 
do you need some modification or explanation in order to aprove this pull request?

horus68:
@mattlorimer  Wrong branch. This should be closed 
User resubmitted to hotfix here https://github.com/salesagility/SuiteCRM/pull/842

PR #616, opened 2015-11-02
Escaping issues

horus68:
Escaping issues on translations for multi-lines with ""


horus68:
duplicated from https://github.com/salesagility/SuiteCRM/pull/781

Issue #436, opened 2015-09-03
Duplicated language key - LBL_EDIT_FIELDS


horus68:
Duplicated language key but different value will be override when used:
https://github.com/salesagility/SuiteCRM/blob/master/modules/ModuleBuilder/language/en_us.lang.php

lines 379/381
'LBL_EDIT_FIELDS'=>'Edit Custom Fields',
'LBL_EDIT_FIELDS'=>'Edit Fields',

likhobory:
There are much more duplications. See attachment:
https://yadi.sk/d/G782bzupir7Vb

PR #617, opened 2015-11-02
Language duplicated strings

horus68:
Removed some language duplicated strings from english language files
They are duplicated key on the same language file.
This causes additional work to translators even if system only uses the first key so its better to be removed!


horus68:
Try to solve the issue https://github.com/salesagility/SuiteCRM/issues/436

Issue #610, opened 2015-11-01
Count query failing for calls/meetings in 7.4


mattlorimer:
There appears to be an issue with the count query for calls/meeting in 7.4, on the My calls dashet it shows '1 to 10 of 0'

This also shows in the log:-

Query Failed: SELECT count(**DISTINCT .id**) c FROM calls LEFT JOIN calls_users c_u on c_u.call_id = calls.id where ((calls.status IN ('Planned')
)) AND calls.deleted=0 AND (calls.assigned_user_id = '1' OR (c_u.user_id = '1' AND c_u.deleted = 0 )) : MySQL error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ') c FROM calls LEFT JOIN calls_users c_u on c_u.call_id = calls.id where ((c' at line 1

willrennie:
Resolved in 7.4.1 and later.

PR #630, opened 2015-11-06
7.4 hotfix strict

gymad:


gymad:
(commit "strict error fixed on listview" - 2859634 contains a fix for #610)

PR #646, opened 2015-11-11
This closes #540.  As Jim had mentioned in the call, the SELECT state…

### LINKED ISSUE #540: Removing a contact from a meeting prevents readding
If there is an entry in the meetings_contacts table for a contact where the relationship has been marked deleted attempting to add the contact again from the meeting edit view silently ignores the contact and doesn't add them.
Looking in the meetings form handler it may just be a case of adding "deleted = 0" to the query.

May also be an issue for other invitee types (users, e.t.c.)

paulgodley:
…ments simply needed to have "deleted=0" added to the query.  This was the case for the user, contact and the lead.  I have tested each of these and now the entry is added for the appropriate user, contact and lead, regardless of whether they had been deleted previously.  This holds true for clicking both the "save" and the "save and send invites" buttons.


paulgodley:
The destination project was incorrect.

Issue #604, opened 2015-10-29
& --> symbol issues


Mausino:
For example, 

if will i use this symbol in case update, it doesn't save the update text after & symbol, 

same issue in reports if are you filter via parameter (if company or another module record have & symbol in name of record)

Miky

paulgodley:
This requires a call to ecodeURIComponent before the text containing the special characters are sent to the server.  Ideally all client data being passed to the server in a URI should be encoded in this way.  I will see where this needs to be called across the project, rather than just fixing a single issue with this.

PR #651, opened 2015-11-11
Bug #604 

paulgodley:
This addresses the issue of case updates not showing any text after special characters such as &-->.  This adds uri encoding to the case update string before it is passed up to the server, resolving this issue.


JimMackin:
Manually merged.

PR #658, opened 2015-11-12
This closes #598.

### LINKED ISSUE #598: quotes, error in product_qty
hello, please check my Post and comments in:

https://suitecrm.com/forum/suitecrm-7-0-discussion/6799-quotes-error-in-product-qty

thanks.

paulgodley:
The original code was stripping off the zero part of 10, 100, 1000, ... of the list item quantity when the currency significant digits field was set to 0.  This update will only remove the fractional part and the decimal point (whatever it has been specified to be in the locale settings) if the fractional part is composed purely of zeros.


paulgodley:
Hi @willrennie , as per @mcarop's query above, can you please confirm if this is in version 7.4.3?

PR #661, opened 2015-11-12
closes #639 - editing opportunities amount and amount_usdollar via in…

### LINKED ISSUE #639: Amount not updating via Inline Editor module Opportunities
When you navigate to the Opportunities ListView, click on the Inline Editor for Amount.  Enter a different amount to the current amount and click the inline save (or return).  The amount field will display the previous amount, be missing decimal points if the number was .00 and does not display the currency selected.

samus-aran:
closes #639 - editing opportunities amount and amount_usdollar via inline editor, displaying associated currency and apply conversion when edited

### DIFF
--- a/include/InlineEditing/InlineEditing.php
+++ b/include/InlineEditing/InlineEditing.php
@@ -373,6 +373,24 @@ function formatDisplayValue($bean, $value, $vardef, $method = "save", $view)
 
     }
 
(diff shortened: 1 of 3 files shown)


Dillon-Brown:
PR no longer necessary. Original issue has been resolved.

PR #701, opened 2015-11-24
closes #680 - value was being unecessarily reformatted which loses th…

### LINKED ISSUE #680: product_discount generated wrong in template
When I use 'Print as PDF' with, for example, a quote, the $aos_products_quotes_product_discount variable can be shown wrongly in certain cases.

This happens if you use an amount, not a percentage, and this amount is higher than '999'.
When you, for example, give a discount of '1.234', the resulting value in the generated template is '1,23'.
It doesn't matter if the decimal separator used is a comma or a dot; I've seen this with both configurations.
I've seen this in the separate plug-in v5.3.3 and, just tried it, in the public SuiteCRM demo, which is on SuiteCRM 7.3.1.

samus-aran:
…e non standard decimal and thousand separators.  Also corrected percentage not being shown by default.

### DIFF
--- a/modules/AOS_PDF_Templates/templateParser.php
+++ b/modules/AOS_PDF_Templates/templateParser.php
@@ -55,14 +55,19 @@ class templateParser{
 				if (strpos($name,'product_discount')>0){
 					if($value != '' && $value != '0.00')
 					{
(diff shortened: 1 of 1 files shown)


Dillon-Brown:
No longer relevant.

PR #718, opened 2015-11-26
Stay on Tab when going from DetailView to EditView

gunnicom:
When on a Tab on Detailview and going to EditView, always the first Tab is open again.
This Patch stays on the Tab that was open on DetailView.
One Commit of the patch is a minor Bugfix, but is needed for the Complete Patch.
The other three patches are the ones that add the enhancement.

Invalidates pull request #93 that did only work in development mode because of template caching.

### DIFF
--- a/include/EditView/EditView.tpl
+++ b/include/EditView/EditView.tpl
@@ -53,9 +53,9 @@ class="yui-navset"
     {{counter name="tabCount" start=-1 print=false assign="tabCount"}}
     <ul class="yui-nav">
     {{foreach name=section from=$sectionPanels key=label item=panel}}
-        {{counter name="tabCount" print=false}}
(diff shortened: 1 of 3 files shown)


Dillon-Brown:
Appears to no longer work correctly. Please change branch to hotfix-7.8.x

Dillon-Brown:
Closing for now. This would need to be re-made for hotfix-7.8.x.

gunnicom:
That is a good way to get rid of PRs *sarkasmoff*

Dillon-Brown:
@gunnicom There's nothing stopping you from re-opening if you want to continue this. I don't see much point in having a 2 year old non-working PR up though.

gunnicom:
It is not working _anymore_ because of salesagility, not because of me.

Issue #712, opened 2015-11-26
Sort PDF Templates when generating Letter


JimMackin:
When generating PDFs from say an account the user is prompted with a list of templates. By default this is unsorted. Sorting these alphabetically would help users find templates.

ghermans:
+1

EZS-JD:
+1

amariussi:
Confirmed in 7.8.2
![templates_not_sorted](https://cloud.githubusercontent.com/assets/8026129/23472094/0bcb3b1e-feac-11e6-85fd-eb156f7f0ffd.png)

samus-aran:
![](https://github.trello.services/images/mini-trello-icon.png) [Sort PDF Templates when generating Letter](https://trello.com/c/SeaIpFm4/133-sort-pdf-templates-when-generating-letter)

samus-aran:
We have now set up a new home for suggestions at Trello.  All github issues that were labeled 'suggestion' have been moved and will be closed.  Certain ones will be progressed within the new Suggestion Box and may be re-opened.

Announcement of moving Suggestions:
https://suitecrm.com/forum/suggestion-box/13691-moving-suggestions-from-github-to

New SuiteCRM Suggestion Box
https://trello.com/b/Ht7LbMqw/suitecrm-suggestion-box

PR #720, opened 2015-11-26
#712 order templates alphabetically

sgaved:
Orders templates alphabetically as suggested in enchancement #712


samus-aran:
[Sort PDF Templates when generating Letter](https://trello.com/c/SeaIpFm4/133-sort-pdf-templates-when-generating-letter)

PR #744, opened 2015-12-02
2 values wrongly formated

horus68:
Language parsing errors on line breaking with ''


horus68:
duplicated from https://github.com/salesagility/SuiteCRM/pull/780

PR #750, opened 2015-12-03
closes #715 Radio button without a 'blank' value in it's list will be…

### LINKED ISSUE #715: Custom Required Field not flagging
If you create a custom field and set as  required. Place it in a module.

fill out the record and DO NOT fill in the required field then it will not flag and warn the user that a required field is not filled

samus-aran:
… flagged up as a required field upon save.

issue #715
possible related issue #1280


Dillon-Brown:
Superseded by #4268

PR #763, opened 2015-12-07
Strict Standards

gidsil:
Non-static method CLASS::METHOD() should not be called statically in file.php on line X

Added static to function statement to avoid PHP error when the function is called statically.


samus-aran:
Please view our updated [Coding Standards](https://suitecrm.com/wiki/index.php/Coding_Standards) and [How to Contribute to SuiteCRM Project](https://suitecrm.com/wiki/index.php/Contributing_to_SuiteCRM) guidelines.  Thanks again for your contribution!

PR #770, opened 2015-12-10
Fix typo in events

sgaved:
Change a ready to already


mattlorimer:
I think the real issue is the hard coded language strings

PR #798, opened 2015-12-18
Commented hardcoded css lines

dgrudinin:
CSS <link rel="stylesheets"> include hardcoded CSSs from SuiteR theme, but can be used in other themes. This breaks other themes styles & colors.


samus-aran:
@dgrudinin We have established a new pull request process which any new pull requests should be made against salesagility:hotfix rather than master.  

Could you also supply examples of where the existing code breaks other themes?  So we can test your new pull request efficiently.  Thanks.

Issue #583, opened 2015-10-20
Navigation module highlighting


jobvector:
SuiteCRM currently does not highlight the module you're currently using in the navigation.

![bildschirmfoto 2015-10-20 um 12 33 47](https://cloud.githubusercontent.com/assets/15199551/10604807/59d92950-7727-11e5-94ba-381437881904.png)

improved usability:
![bildschirmfoto 2015-10-20 um 12 37 05](https://cloud.githubusercontent.com/assets/15199551/10604840/993e69e8-7727-11e5-986a-0bba0d07f892.png)

SuiteCRM Version: 7.3.1

jobvector:
the highlighting is availibale for the Suite 7 theme but not for Suite R

ghermans:
+1

simbanic:
+1

jobvector:
For now we simply impemented this feature ourselves:
1. create the following javascript file 

$(document).ready(function() {
   $("li.topnav:has(span.currentTab)").addClass("currentSelectedModule");
   $("li.topnav span.currentTab a").addClass("currentSelectedModuleLink");
});

2. create the corresponding CSS classes
.currentSelectedModule {
    background-color: #5880BA;
}
/* optional: reimplement the link underline to support the visibility */ 
.currentSelectedModuleLink {
    text-decoration: underline !important;
}

3. include both files in /custom/themes/SuiteR/tpls/_head.tpl
<link rel="stylesheet" type="text/css" href="custom/themes/SuiteR/css/highlighting.css">
<script type="text/javascript" src="custom/themes/SuiteR/javascript/highlighting.js"></script>

ghermans:
@jobvector can you make a pull request for this?

pgorod:
I came here to start an issue about the Top Navigation, but since this one is here already I will just add to the discussion.

This issue is more important than just a small visual detail. I find that the initial impact of SuiteCRM is considerably hindered by this, because when people first start looking at the program and trying to understand it, this unnoticeable-shifting-menu-structure is a real pain. 

I am in favor of (urgently) adding some visual differentiation, but on a more fundamental level I question whether that is the best place to have that sort of contextual menu. That is not a good place to be moving things around, because that is where users start reading the screen to look for what they need. I would say that the basic menu structure should be fixed, and any contextual, changing option would either need to be very different visually, or placed elsewhere (e.g. on the right). Possibly a differentiated text could also help ("Accounts actions" instead of just "Accounts").

[...]

PR #812, opened 2015-12-22
Navbar module highlighting

jobvector:
Implemented highlighting of the currently active module.

Referring to & Fixing:
https://github.com/salesagility/SuiteCRM/issues/583

### DIFF
--- a/themes/SuiteR/css/navhighlight.css
+++ b/themes/SuiteR/css/navhighlight.css
@@ -1,0 +1,7 @@
+.current-selected-module {
+  background-color: #5880BA;
+}
+/* optional: reimplement the link underline to support the visibility */
+.current-selected-module-link {
+  text-decoration: underline !important;
+}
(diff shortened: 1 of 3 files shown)


jobvector:
Why is this a invalid enhancement, is their any other solution implemented to the suitecrm that has the same feature?

samus-aran:
@jobvector - There is a planned new theme coming out in beta that would overwrite this.  That is why it was set to invalid.

pgorod:
@samus-aran I have a question about closing this. Is the new theme is coming out with a different name (not SuiteR)? If so, I would still do something to SuiteR to fix it, because it's not going to disappear, people will still use older versions, older themes (sometimes due to some technical difficulty), and that theme would really benefit from having this improved. If the LTS version is going to have SuiteR on it, it should have a fixed SuiteR on it...

(BTW, I suggested a different visual solution in a comment on the Issue #583 .)

jobvector:
for us any visual solution would be ok

PR #813, opened 2015-12-22
Removing commented duplicated lines from Installation

horus68:
The commented lines are duplicated but with opposite content.
When translating online, Transifex do not display them but fill the contents with the 1st string translation.
So removing this lines makes it fail-proof when using Transifex to translate


mattlorimer:
7.5beta2 is not valid branch to create a pull request for,please use hotfix for issues in the current release, or if you have a fix for a specific version, please specify and create a pull request to the appropriate hotfix branch

PR #823, opened 2015-12-30
fixed Connector: Modify Mapping and Connectors displaying across all Themes

shep517:
Correctly extended the custom Connectors controller to include functions from original controller. Fixed Connectors to display properly on all themes by adding a check for the destroy function on the display object (necessary because theme SuiteR return a display object of html code causing js error)


shep517:
I believe I should have created a request to a hotfix branch...first time.

Issue #825, opened 2016-01-04
CSV export creates partial file when export_excel_compatible is set


jari-m:
Running CSV export, such as on an index view, sets the http Content-Length header wrong (to approximately half of the correct size) when export_excel_compatible is set.
This can or will result in only a partial file getting delivered to the browser.
The current code calculates Content-Length before converting the encoding of the content to UTF-16LE when it should calculate after having converted the encoding. Conversion doubles the content size.

This is the current end of /export.php
```php
header("Expires: Mon, 26 Jul 1997 05:00:00 GMT" );
header("Last-Modified: " . TimeDate::httpTime() );header("Cache-Control: post-check=0, pre-check=0", false );
header("Content-Length: ".mb_strlen($transContent, '8bit'));
if (!empty($sugar_config['export_excel_compatible'])) {
    $transContent=chr(255) . chr(254) . mb_convert_encoding($transContent, 'UTF-16LE', 'UTF-8');
}
print $transContent;
sugar_cleanup(true);
```
And this one works correctly. I have moved the line with the Content-Length calculation.
```php
header("Expires: Mon, 26 Jul 1997 05:00:00 GMT" );
header("Last-Modified: " . TimeDate::httpTime() );header("Cache-Control: post-check=0, pre-check=0", false );
if (!empty($sugar_config['export_excel_compatible'])) {
    $transContent=chr(255) . chr(254) . mb_convert_encoding($transContent, 'UTF-16LE', 'UTF-8');
}
header("Content-Length: ".mb_strlen($transContent, '8bit'));
print $transContent;
sugar_cleanup(true);
```

PR #828, opened 2016-01-04
Corrects wrong Content-Length - Fix for issue #825 - CSV export creates partial file

jari-m:
This one fixes salesagility/SuiteCRM #825 - CSV export creates partial file when export_excel_compatible is set.

Content-Length was calculated before converting the encoding of the content to UTF-16LE. It should be calculated after conversion of encoding because conversion changes content size.


mattlorimer:
merged to hotfix

PR #842, opened 2016-01-12
FIX #508 - Now we can load multiple JS files from the constructor of the view and get only one minifyed 

### LINKED ISSUE #508: Add multiple JS files to a view in only one minifyed JS file.

Abuelodelanada:
The minifyed JS file is saved in:
cache/themes//js/

See #508 for details

### DIFF
--- a/include/MVC/View/SugarView.php
+++ b/include/MVC/View/SugarView.php
@@ -83,6 +83,7 @@ class SugarView
     var $responseTime;
     var $fileResources;
 
+    private $settings = array();
     /**
      * Constructor which will peform the setup.
      */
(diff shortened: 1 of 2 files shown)


Abuelodelanada:
When developing some complex JS functionality is very useful to have module vardefs definitions in the DOM so you can read it from JS.

26970c0c-615f-11e5-85e6-3fe7d90d4079

cb8035c2-615f-11e5-881a-9d884867c346


3a82d978-6162-11e5-89eb-62d584c0317e


4ad4bf1c-6162-11e5-9b00-6bf161999e12

vjaro:
+1

PR #867, opened 2016-01-25
Formating with html tags to allow translation  [Language]

horus68:
file Include/en_us.notify_template.html: 
Including tags to translate with Crowdin. 
It requires some tags to use this file as html and break lines


pgorod:
:+1:

PR #906, opened 2016-02-02
Contacts Subpanels Bug

adamjakab:
Corrected errors in subpanelsdef keys causing js errors in Contacts module details view (no panel dragging) and who knows where else.
Look like errors left behind by some automatic replacement


mattlorimer:
Merged into hotfix

PR #908, opened 2016-02-03
Updated the API methods to the new REST layout

paulgodley:
This is the initial migration of the REST API methods into the new layout.


lewisje1991:
Hi Paul,

Can you update this to include the authorisation?

Thanks
Lewis

vjaro:
+1 looks promising

Issue #927, opened 2016-02-08
7.5.1 - Unable to Change or Add Dashboard Tabs


johnM2401:
User is unable to change or add Dashboard tabs.

When they click "Add Tab", they are met with the error:

_ReferenceError: addDashboardForm is not defined_

If they click on an existing tab, they are met with the error:

_ReferenceError: retrievePage is not defined_

Tested on a fresh install and an Upgraded instance, occurs on both

PR #936, opened 2016-02-10
Closes bug 927.  The path to the js file had not been updated since t…

paulgodley:
…he move from custom.  Updated the path and the "add tab" action now fires appropriately. closes bug #927


phreddyrose:
Awesome!  Glad it works!

PR #947, opened 2016-02-11
Case state pair - Opened/Closed

horus68:
First value pair from 'case_state_dom' changed to be the opposite of the second value. So Opened/Closed
related: Just to make it easy to find: Moving 'case_state_default_key'
to be above its array 'case_state_dom'


mattlorimer:
'Open' makes more sense than 'Opened' in this case

horus68:
If so "case_state_default_key" can't have the same value then  'case_state_dom' or it will be translated with Crowdin when its not supposed to be translated.

PR #988, opened 2016-02-22
For long usernames search box will be partly hidden by alert-nav

gunnicom:
When the text field resizes when clicking in the field, it resizes and hides the beginning unter the alert-nav. With margin the resize stops just behind the alert notification.


gunnicom:
Found out that this will not be enough to fix that. I try to fix it in some other way.

gunnicom:
Superseded by https://github.com/salesagility/SuiteCRM/pull/995

PR #1003, opened 2016-02-24
Fixed #707 added an if statement which stops script from leaving to another view if action is Clone

### LINKED ISSUE #707: Studio - Clone a field doesn't work
To reproduce this issue please go to:
http://example.com/index.php?module=ModuleBuilder&action=index&type=studio

*Studio > Calls > Fields*

After click on any field you want, when **Clone** button is pressed inside **Edit Field** tab, it brings a floating div with the message:

`Result`
`This operation is completed successfully`

But instead, it closes the **Edit Field** tab and then returns to **Edit Fields** tab
When you pressed the **X** on the floating div, it closes doing nothing.

__*Note:*__ I tried on Accounts, Contacts and Opportunities modules as well and it fails too.

Regards,

Attached image, after press the **Clone** button.

suitecrm-studio-clone-button-issue

vladbar:
Fixed #707 added an if statement which stops script from leaving to another view if action is Clone

### DIFF
--- a/modules/ModuleBuilder/javascript/ModuleBuilder.js
+++ b/modules/ModuleBuilder/javascript/ModuleBuilder.js
@@ -633,41 +633,44 @@ if (typeof(ModuleBuilder) == 'undefined') {
 					urlVars[urlVar[0]] = urlVar[1];
 				}
 
-				// check where we are and do it if we are in field editor in module builder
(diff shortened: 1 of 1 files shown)


JimMackin:
Seems to cause issues when saving list view layouts.

PR #1041, opened 2016-03-03
Prevent creating invalid image references

tfreier:
If you use a quote or invoice templates that included the product image you have to have a product image for every product otherwise it will show an image load error since the URL in the img-tag is invalid.

This code change checks if there is an image for the product before creating the img-tag.


ghermans:
Hey thanks for the contribution, however PR requests on the master branch are not accepted by Salesagility.
The correct branch for PR's is hotfix

tfreier:
Thanks @ghermans, I will submit another request for that branch.

PR #1047, opened 2016-03-04
Fix for parse error, double quotes instead of single

vjaro:
fix for parse error, double quotes instead of single


mattlorimer:
A fix has already been supplied for this issue, but thanks for your input

PR #1055, opened 2016-03-09
substituted 'die()' in sugar_die method with \Exception

adamjakab:
It is very difficult to check for errors and react to them - especially when working on tests - when application is killed.
Is there any reason why this shoul not be a valid alternative?


chris001:
This is a good PR @adamjakab you're right, `sugar_die` should absolutely throw an `Exception` instead of calling `die()`, Production professional PHP applications running on real user data should never call `die()`.  Why did you close this?

PR #1059, opened 2016-03-09
ModuleBuilder - removing duplicated language strings  [Language]

horus68:
Removing duplicated languages strings on same array
string  'LBL_CONFIRM_FIELD_DELETE' = Making actual multinline into a single line for translations


modules/ModuleBuilder/language/en_us.lang.php
```diff
@@ -320,7 +320,6 @@
             'layoutsHelp' => 'Select a <b>Layout to edit</b>.<br/<br/>To change the layout that contains data fields for entering data, click <b>Edit View</b>.<br/><br/>To change the layout that displays the data entered into the fields in the Edit View, click <b>Detail View</b>.<br/><br/>To change the columns which appear in the default list, click <b>List View</b>.<br/><br/>To change the Basic and Advanced search form layouts, click <b>Search</b>.',
             'subpanelHelp' => 'Select a <b>Subpanel</b> to edit.',
             'searchHelp' => 'Select a <b>Search</b> layout to edit.',
-            'labelsBtn' => 'Edit the <b>Labels</b> to display for values in this module.',
```
samus-aran: Why delete this label - this is the label that is used currently - would it not be best to delete the other duplicate?

samus-aran: Should this one be deleted?

horus68:
Agree with the switch. I was not aware of what line was used

modules/ModuleBuilder/language/en_us.lang.php
horus68: ok for me. changed the file

PR #1084, opened 2016-03-12
Fix undefined index for Facebook and Twittter issue#1080

LEAP-nishit:
$currModule['label'] = $module was being overridden if $module is Facebook or Twitter.

Re-arranged the conditional to ensure it was set earlier if condition is met and avoid the Undefined Index in $currModule['label'] = $GLOBALS['app_list_strings']['moduleList'][$module];


JimMackin:
Thanks for the contribution.

PR #1092, opened 2016-03-15
Fixed #1091 if action Clone is undefined then allow ListView to save

### LINKED ISSUE #1091: Studio -> ListView -> save and deploy doesn't work
Studio -> ListView -> save and deploy: does not save fields which were moved from Hidden/Available to Default .

vladbar:
Fixed #1091 if action Clone is undefined then allow ListView to save


mattlorimer:
I think you have branched from develop instead of master/hotfix

PR #1094, opened 2016-03-15
Convert value to unescape ampersand and comparision for audit works

gunnicom:
If you have a audited field that contains ampersands, they always get written to audit log because ampersands are not escaped in $after_value
Example:
$before_value = "Company GmbH & Co. KG"
$after_value = "Company GmbH & amp; Co. KG"   // space only to show the amp here correctly
So it seems unequal for audit log, but is saved equal because from_html is calles before update query.
Calling from_html like it is done before the update query fixes this.


gunnicom:
I cannot reproduce this bug anymore. Needs more investigation from me why this happened.

PR #1100, opened 2016-03-16
DevelModule

adamjakab:
I made this module because I noticed there is a very high query execution count when there are reminders present in database (i think there must be a bug).
This module (you can enable it from Configuration on the bottom) unfortunatelly cannot work as an add-on module since it requires modifications to core files.
the module is usable but not by all means complete(ACL/cleanup/etc) - it you think it is worth to be part of the core - it could become a good tool to help to develop modules and it could also be a helper tool for testing.
Once enabled, look for the orange 'cog' icon in the bottom right corner of your page.


adamjakab:
sorry - something went wrong.

PR #1102, opened 2016-03-16
Devel module

adamjakab:
I have created this module to be able to check on an abnormally high sql query execution that was/is happening in sugarcm when there are reminders in the database(i think it is still an isuue).
You can enable this module on the bottom of the configuration page, and once enabled, you will have a little orange 'cog' icon in the bottom right corner of your pages (for now showing query count/memory/exec time) and link to the Devel module where you can inspect each request.
Unfortunatelly, this module cannot be a standalone module since it requires modifications to the core files. 
This could be a good tool to use for developing/debugging and could also be a helper tool to testing.


adamjakab:
NEEDS FIX! Early sesssion_start() breaks rest authentication.
will fix soon

adamjakab:
I am closing this PR until travis test are correctly implemented in develop branch

PR #1107, opened 2016-03-16
Fixed #1105 - link field editing confirmation message.

### LINKED ISSUE #1105: Inline editing - link field editing confirmation message.
When the user edits a link field and clicks away (without adding any change to the field), they will get a prompt: 
"You have clicked away from the field you were editing without saving it. Click ok if you're happy to lose your change, or cancel if you would like to continue editing" . 

This is incorrect - they should only get a prompt if they added a change and then clicked away. 

All other non-link fields have correct/consistent inline editing - users would only get a prompt if they added a change and then clicked away. If they did not add a change and clicked away, no prompt is displayed.

vladbar:
Fixed #1105 In case of a link inline editing if(user_value != output_value) { would always be true as output_value contained HTML+value, even though user_value gets just textbox value without HTML


JimMackin:
@likhobory Well spotted. I've created an issue to raise this (https://github.com/salesagility/SuiteCRM/issues/1160).

PR #1114, opened 2016-03-17
Composer

jakubpas:
- Introduce installation by composer:

```composer create-project jakubpas/suitecrm your_application_name```

- introduce PS-4 namespaces for SuiteCRM
- moving internal and abandoned classes to namespaced /src (currently SugarLogger and XTemplate)
- moving external  libraries to composer's /vendor (currenty Smarty as example)
- some minor refactoring

The jakubpas/suitecrm can be replaced by official SuiteCRM packagist location as soon as it will be set up.


mattlorimer:
you may wish to check out from/ merge with the develop branch where composer is partially set up

jakubpas:
Ok thanks. There are actually two branches with composer partially setup "RestAPI" and "develop". I did not know which one is better to start.

mattlorimer:
develop

PR #1129, opened 2016-03-21
Tests/Travis Fix

adamjakab:
Fix phpunit test runs for travis so that we can evaluate future PRs


adamjakab:
Got here too late? In this PR - all tests are passing - but it seems like you have disabled integration with travis on this branch.

adamjakab:
Ok, tests are now passing. Since tests on php 5.4 are important (now allowed to fail) we need to downgrade phpunit to 3.7 so that dependency on symfony/yaml can pass. This however means that all 'assertNotFalse' calls needs to be changed to 'assertTrue' I guess.

adamjakab:
OK, tests on php 5.4 are now NOT failing due to composer dependency problems so now we need to remove it from allowed_failures.

mattlorimer:
5.4 is not important, 5.5+ is

adamjakab:
well, now php54 is tunning happily - I considered php54 to be important because in check_php_version in utils.php states `$min_considered_php_version = '5.3.0'` - so maybe even php53 should be added to the tests.

we still have:

**php7** - as i commented in .travis.yml - this is not a php7-ready project - it has tons of warnings about the deprecated constructors: `Deprecated: Methods with the same name as their class will not be constructors in a future version of PHP...` - but at the end it fails on a class not found fatal error: `Fatal error: Uncaught Error: Class 'SecurityGroup' not found in ...`. Could it be the autoloader?
Anyhow - I'd remove this test alltogether.

**hhvm** - funnily enough phpunit exits with 0 even though it never completes the tests due to `ZIP Error(11): Status(0): Arhive(cache/zipTest.zip): Directory(cache)` error - I did not have time to check on this - my guess is that there is a die() call somewhere in there.

mattlorimer:
min_considered_php_version will become 5.5, constructors are being fixed, 5.5, 5.6, 7 should be required with HHVM in allowed_failures.

Do you mean remove the php7 test all together or a specific test?

adamjakab:
well, you are right, I guess there is no reason why tests for php7 should be removed - but for the time being, at least, i'd put php7 under allowed_failures matrix (someone should look into why it fails on php7- i've got no idea).
And yes, if we change `min_considered_php_version` to 5.5 then we can remove php5.4 alltogether.

I'd like to see tests working on this branch becasue i have a pending PR #1100 which is failing (not because the PR is failing (maybe) but becuase the tests were not configured correctly) - now with this PR and from here on finally we/you can see if a PR is breaking the application or not without having to look at the code. I also have a few pending ideas/enhancements I have not PRed yet because of until now tests were failing.

This is why i'd put php7 (until test problems get fixed under allowed_failures matrix.

Thanks

mattlorimer:
> This is why i'd put php7 (until test problems get fixed under allowed_failures matrix.

yes I agree, it would make sense.

PR #1134, opened 2016-03-21
SugarFeed content cut results in invalid characters

Szellszi:
Consider the following scenario:
You want to save an Opportunity with a name exactly 28 bytes long and the Account name is exactly 35 bytes long. The default currency is Euro with it's sign (€) in the config and the amount of the opportunity is 2016,03.

With the text of the SugarFeed, this will be more than 255 bytes long (since the € sign takes 3 bytes) so Suite gonna try to cut it after 255 bytes. Since this is at the "middle" of the 3 byte € sign, it will cut it half which results in two invalid characters, which can give the database a hard time.

If we count the UTF-8 characters instead of the bytes of the string, and cut the string as UTF-8, then the problem disappears.

### DIFF
--- a/modules/SugarFeed/SugarFeed.php
+++ b/modules/SugarFeed/SugarFeed.php
@@ -253,9 +253,9 @@ class SugarFeed extends Basic {
         }
         $text = strip_tags(from_html($text));
 		$text = '<b>{this.CREATED_BY}</b> ' . $text;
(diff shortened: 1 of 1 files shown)


horus68:
@Szellszi can you please commit this PR to the proper branch (Hotfix)?

PR #1143, opened 2016-03-22
Avoid creating img tag if there is no product image

tfreier:
If you use a quote or invoice templates that included the product image you have to have a product image for every product otherwise it will show an image load error since the URL in the img-tag is invalid.

This code change checks if there is an image for the product before creating the img-tag.


JimMackin:
Thanks for the fix. Now that it's merged it should be in the next release.

PR #1152, opened 2016-03-24
Php7 - All tests are passing

adamjakab:
Renamed all constructor methods and added some minor code changes.
4 Additional tests are marked Incomplete because core code returns different results in php5 and php7


adamjakab:
I admit I still have problems moving between branches - if it is a problem that commits that are not mine are showing up here I'll redo the PR . thanks

adamjakab:
i am redoing this correctly (hopefully) ;)

PR #1153, opened 2016-03-24
Php7 - all tests are passing

adamjakab:
1) renamed all class constructor methods to comply with php7
2) added some minor code fixes
3) had to mark 1/2 tests incomplete beacause core code returns different values in php5 and php7


adamjakab:
sorry - still having problems with conflicts

PR #1154, opened 2016-03-24
Php7 - all tests are passing

adamjakab:
1) renamed all class constructor methods to comply with php7
2) added some minor code fixes
3) had to mark 1/2 tests incomplete beacause core code returns different values in php5 and php7


mattlorimer:
Amazing Work, Thanks!!

ChangezKhan:
Hello @adamjakab, here in this commit, in `include\utils.php` file, in method named `sugar_die`, this line is commented out `echo $error_message;` and because of that it is not showing any error message on UI, if the record is not available in the CRM or whenever this method executes. Kindly refer this bug #2971. 
I am not able to understand the need of commenting out this line. If uncommenting out this line wouldn't cause any kind of impact, then it would be better if we uncomment it. Kindly explain.

(Sorry, I don't know how to contact you, so that's why I am mentioning this here.)

PR #1159, opened 2016-03-24
Fixed #682 Theme settings for supbanels menu + icons

### LINKED ISSUE #682: Theme settings for supbanels menu + icons
1. We can't set the color (in theme settings) of icons for subpanels, because if i set it like white, it will not see the pencil for inline editing or other icons 

2. We can´t configure the color of menu for subpanel dropdowns (+ dropdown in listview), because it's using color of buttons menu in theme settings.

accounts

recci:
The Action menu buttons and the icons should actually be two separate issues. This fixes the action menu drop colour selection issues. Suggest opening a separate bug for the icon colour issues.


recci:
I fixed the issue as part of a client project. See screen shots of fix below.

Added new theme setting:
screenshot from 2016-03-24 11 22 00

Fix in action:
screenshot from 2016-03-24 11 21 48

PR #1163, opened 2016-03-24
Fix iCal car encoding

chandon:
Fix iCal problem with Cyrilic charset

### DIFF
--- a/modules/iCals/iCal.php
+++ b/modules/iCals/iCal.php
@@ -554,7 +554,7 @@ class iCal extends vCal {
 
         $str .= vCal::create_ical_string_from_array($ical_array,true);
 
-        return $str;
+        return htmlspecialchars_decode(preg_replace("/&#([0-9]+)\\\\;/",'&#$1;',utf8_decode($str)));
     }
 
 }


horus68:
@chandon can you please submit this PR to the proper branch (Hotfix)?

glaz123:
That change breaks cyrillic character in iCAL in 7.10.7. utf8_decode converts any UTF8 character to ISO8859-1, and therefore any non ISO8859-1 characters will be shown as question marks.

bukbr:
Hi! I did exactly as directed, however, the .ics file continues to be generated without the accentuation of the Portuguese language. Could anyone help? What should I do to get the characters accented? Version 7.10.7

glaz123:
What solve the problem with cyrillic for me, is change the string in this commit from:
        return htmlspecialchars_decode(preg_replace("/&#([0-9]+)\\\\;/", '&#$1;', utf8_decode($str)));
to 
        return htmlspecialchars_decode(preg_replace("/&#([0-9]+)\\\\;/", '&#$1;', $str));

bukbr:
> What solve the problem with cyrillic for me, is change the string in this commit from:
> return htmlspecialchars_decode(preg_replace("/&#([0-9]+)\\;/", '&#$1;', utf8_decode($str)));
> to
> return htmlspecialchars_decode(preg_replace("/&#([0-9]+)\\;/", '&#$1;', $str));

Thanks very much, it worked! :-)

timo12357:
Works for me, too. When can this be included in a release, @Dillon-Brown ?

Dillon-Brown:
Hi @timo12357, thanks for highlighting this. I'll make a PR with this change and if it works as expected then it should go into the next release.

Dillon-Brown:
Fix here: #7691

PR #1190, opened 2016-04-01
Update AOR_Report.php

mugge6:
In case of accessing a custom field, $table_alias has to be used instead of $old_alias. Otherwise the wrong relation will be used.


samus-aran:
Hi @mugge6  . As rule of thumb all bug fixes should be set against the 'hotfix' branch rather than the 'master'. Please create a new PR to link to 'hotfix'.

PR #1201, opened 2016-04-05
fix for #1200: changing emails parent relation leads to email be rela…

### LINKED ISSUE #1200: Emails: save() does not recognize parent_(bean|id) modifications in fetched_row
SuiteCRM: 7.0.2 (but does not matter, that error even was in SugarCE)

**Description:**
When Editing an Email one can modify the parent relation. Let's say I am about to store an Email to an Account. I chose Account as parent-type and type in the accountname ("account A"). Saving the Email saves the parent relation correctly. That Email ist related to to "account A".

Now i edit that Email egain, but now choose another account ("account B") as parent. 
- Saving that Email shows the Email in "Account A" and "Account B".
- There are multiple undeleted entries in table "emails_beans" for that email-bean.
- The code in /Emails/Email.php, save(), line ~1065 checks for modified parent-type and/or parent-id by comparing $bean->parent_type and $bean->fetched_row['parent_type']... But that particular code is AFTER parent::save() and thus $bean->fetched_row is not set anymore (with the beans contents before saving) and cannot be compared. Thus there are no modifications, the old parent relation is not deleted but the new parent relation is added. 

**Expected Behaviour:**
- when changing the emails beans parent id/type, old parent relations must be softdeleted and only one (the new one) relation must exist.
- One email can only be related to one parent bean (but one Account can have multiple different emails).

PR follows.

croessler:
…ted to both parent beans, not only one.

- updated save() method, fixed logical error when accessing bean->fetched_row AFTER parent::save(). Doing so leads to bean->fetched_row not to contain the fields before modifying and thus the email is in state "nothing has been modified". 
- new: save checks the fetched_row contents correctly. modifications to the parent type/id are being recognized, the old parent-relation is deleted, the new one is added.
- a bit of code styling applied so one can read that stuff.

CLA has been signed.
Fixes #1200

### DIFF
--- a/modules/Emails/Email.php
+++ b/modules/Emails/Email.php
@@ -1060,24 +1060,26 @@ class Email extends SugarBean {
                  }
 			}
 
-			parent::save($check_notify);
-
-			if(!empty($this->parent_type) && !empty($this->parent_id)) {
-                if(!empty($this->fetched_row) && !empty($this->fetched_row['parent_id']) && !empty($this->fetched_row['parent_type'])) {
-                    if($this->fetched_row['parent_id'] != $this->parent_id || $this->fetched_row['parent_type'] != $this->parent_type) {
+            // croessler: has the parent-bean or its id been modified? compare via fetched-row...
+            // if so, delete the old relation and add the new one. one email can only be related to one parent, not multiple parents.
+            if (!empty($this->parent_type) && !empty($this->parent_id)) {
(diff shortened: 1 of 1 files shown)


samus-aran:
Hi @croessler . As rule of thumb all bug fixes should be set against the 'hotfix' branch rather than the 'master'. Please create a new PR to link to 'hotfix'.

PR #1218, opened 2016-04-07
Bug in "Compose email" in ListViews 

cfiorillo:
When using internal mail client and inline editing is disabled


samus-aran:
Hi @cfiorillo 

As rule of thumb all bug fixes should be set against the 'hotfix' branch rather than the 'master'. Please create a new PR to link to 'hotfix'.  

PR #1220, opened 2016-04-07
fixed installation warnings - now installs silently

adamjakab:
modules: Reminders and Reminders_Invitees declare PK resulting in multiple PK definition


adamjakab:
@mattlorimer - Halleluja! - I'm glad we have functioning tests!
However if you check on the php7 log output we still have some `PHP Deprecated:  Methods with the same name as their class will not be constructors in a future version of PHP`  warnings.

adamjakab:
I have also fixed the unwanted output during tests like that 'VCARD' stuff, the flow_bar output and the 'NOOP: could not create folder' output (there is also a '01' output coming from the Emails module from somewhere. 
Now, I can see them again. Is that on purpose?

mattlorimer:
01 def not on purpose, deprecated method, looks like it from a 3rd party add-on that should probably be stripped out, as I don't think it used anymore, I'll look into, but thanks for the fixes

mattlorimer:
NOOP: could not create folder' and VCARD are not on purpose, where/when did you fix that?

PR #1228, opened 2016-04-10
Patch 1123

ghermans:


pgorod:
This change was reverted when PHPMailer was upgraded to new version.

We shouldn't be doing changes inside the 3rd party package, we have our own file to override stuff in this class.

So 
`include/phpmailer/class.phpmailer.php` is 3rd party
but
`include/SugarPHPMailer.php` is ours

I don't know if this needs to be put back in, I haven't heard people complaining...

PR #1255, opened 2016-04-20
Fixed #1253 - added a switch statement.

### LINKED ISSUE #1253: Reports "Equal to period" doesn't work as expected
If I set a date field as a parameter and have it as, for example "Equal to this quarter" I would expect this to return any records in April, May or June. However it creates the following:
`field = "2016-04-01 00:00:00"`
Which returns no results because no field has exactly this value.

vladbar:
Fixed #1253 - added a switch statement which adds a period instead of a plain date (if condition has Equal to preiod).

### DIFF
--- a/modules/AOR_Reports/AOR_Report.php
+++ b/modules/AOR_Reports/AOR_Report.php
@@ -1324,8 +1324,40 @@ class AOR_Report extends Basic {
                         $value = "{$value} OR {$field} IS NULL";
                     }
 
-                    if (!$where_set) $query['where'][] = ($tiltLogicOp ? '' : ($condition->logic_op ? $condition->logic_op . ' ': 'AND ')) . $field . ' ' . $app_list_strings['aor_sql_operator_list'][$condition->operator] . ' ' . $value;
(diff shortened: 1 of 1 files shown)


modules/AOR_Reports/AOR_Report.php
```diff
@@ -1324,8 +1324,40 @@ function build_report_query_where($query = array()){
...
+                        if (array_key_exists($condition->value, $app_list_strings['date_time_period_list'])) {
+                            $params = $condition->value;
+                        } else {
+                            $params = base64_decode($condition->value);
+                        }
+                        $value = '"' . getPeriodDate($params)->format('Y-m-d H:i:s') . '"';
+
+                        switch ($app_list_strings['aor_sql_operator_list'][$condition->operator]) {
+                            case "=":
+                                if (!$where_set) $query['where'][] = $field . ' BETWEEN ' . $value .  ' AND ' . '"' . $date . '"';
```
mattlorimer: I don't think you need to check if for !where_set on every line, you should check this before you enter this if/else

PR #1257, opened 2016-04-20
Fixed #1253 - added a switch statement.

### LINKED ISSUE #1253: Reports "Equal to period" doesn't work as expected

vladbar:
Fixed #1253 - added a switch statement which adds a period instead of a plain date (if condition has Equal to preiod).

### DIFF
--- a/modules/AOR_Reports/AOR_Report.php
+++ b/modules/AOR_Reports/AOR_Report.php
@@ -1324,7 +1324,40 @@ class AOR_Report extends Basic {
                         $value = "{$value} OR {$field} IS NULL";
                     }
 
-                    if (!$where_set) $query['where'][] = ($tiltLogicOp ? '' : ($condition->logic_op ? $condition->logic_op . ' ': 'AND ')) . $field . ' ' . $app_list_strings['aor_sql_operator_list'][$condition->operator] . ' ' . $value;
+                    if(!$where_set) {
+                        if ($condition->value_type == "Period") {
+                            $date = '"' . getPeriodEndDate($params)->format('Y-m-d H:i:s') . '"';
+                            if (array_key_exists($condition->value, $app_list_strings['date_time_period_list'])) {
+                                $params = $condition->value;
+                            } else {
+                                $params = base64_decode($condition->value);
+                            }
+                            $value = '"' . getPeriodDate($params)->format('Y-m-d H:i:s') . '"';
+
+                            switch ($app_list_strings['aor_sql_operator_list'][$condition->operator]) {
+                                case "=":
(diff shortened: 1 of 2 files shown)


modules/AOR_Reports/AOR_Report.php
```diff
@@ -1324,7 +1324,40 @@ function build_report_query_where($query = array()){
...
+                            $value = '"' . getPeriodDate($params)->format('Y-m-d H:i:s') . '"';
+
+                            switch ($app_list_strings['aor_sql_operator_list'][$condition->operator]) {
+                                case "=":
+                                    $query['where'][] = $field . ' BETWEEN ' . $value .  ' AND ' . '"' . $date . '"';
+                                    break;
+                                case "!=":
+                                    $query['where'][] = $field . ' NOT BETWEEN ' . $value .  ' AND ' . '"' . $date . '"';
+                                    break;
+                                case ">":
```
mattlorimer: > < >= <= all have the same action can they not be grouped together

modules/AOR_Reports/AOR_Report.php
```diff
@@ -1324,7 +1324,40 @@ function build_report_query_where($query = array()){
...
+                                    break;
+                                case ">=":
+                                    $query['where'][] = $field . ' ' . $app_list_strings['aor_sql_operator_list'][$condition->operator] . ' ' . $value;
+                                    break;
+                                case "<=":
+                                    $query['where'][] = $field . ' ' . $app_list_strings['aor_sql_operator_list'][$condition->operator] . ' ' . $value;
+                                    break;
+                            }
+                        } else {
+                            $query['where'][] = ($tiltLogicOp ? '' : ($condition->logic_op ? $condition->logic_op . ' ': 'AND ')) . $field . ' ' . $app_list_strings['aor_sql_operator_list'][$condition->operator] . ' ' . $value;
```
mattlorimer: do the previous factor in the $tiltLogicOp check, should maybe be a more generic check if within !where_set

modules/AOR_Reports/aor_utils.php
```diff
@@ -205,6 +205,80 @@ function getPeriodDate($date_time_period_list_selected)
 }
 
 /**
+ * getPeriodDate
+ * @param $date_time_period_list_selected
+ * @return DateTime
+ */
+function getPeriodEndDate($date_time_period_list_selected)
+{
+    if ($date_time_period_list_selected == 'today') {
```
mattlorimer: better as a switch statement

PR #1293, opened 2016-04-27
Added new 'Copy emails from WorkFlow Module' option to Workflow's 'Cr…

adriangibanelbtactic:
…eate Record' action

That means that all emails from the workflow module will be copied to the new created record on the action if the option is checked.

By default this option is not checked.


samus-aran:
@adriangibanelbtactic Thanks for your contribution. Usually with enhancements we ask you make a PR to develop rather than master. Can this be amended and we will consider all PR for the next major release?

adriangibanelbtactic:
Thank you @samus-aran for pointing out my mistake (first time doing a PR for Suitecrm).
The new pull request can be found at: https://github.com/salesagility/SuiteCRM/pull/1348 .

I close this pull request as it no longer applies.

PR #1316, opened 2016-05-01
Fix for unchmodabble file systems

antonioribeiro:
When you are on an hybrid filesystem, for instance, a Windows NTFS mounted on a Linux host, your filesystem may not be chmoddable, but the folder may be still writable. In those cases we get an error while trying to install a module. This fix prevents that.

### DIFF
--- a/include/utils/sugar_file_utils.php
+++ b/include/utils/sugar_file_utils.php
@@ -69,7 +69,7 @@ function sugar_mkdir($pathname, $mode=null, $recursive=false, $context='') {
 	}
 
 	if($result){
-		if(!sugar_chmod($pathname, $mode)){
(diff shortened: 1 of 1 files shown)


samus-aran:
Hi @antonioribeiro 

As rule of thumb all bug fixes should be set against the 'hotfix' branch rather than the 'master'. Please create a new PR to link to 'hotfix'.  We also prefer that each 'fix' you provide that it is isolated in a branch on your repo i.e. branch 'issue_123' so that when you make multiple PRs that it only includes the commits relevant to the fix rather than your 'master'.

antonioribeiro:
Done: https://github.com/salesagility/SuiteCRM/pull/1407

PR #1332, opened 2016-05-04
Fixed #1320 "One of" condition for multi select does not working

### LINKED ISSUE #1320: "One of" condition for multi select does not working in Reports module in suiteCRM - 7.6
Hi,
When I try to create new report in suiteCRM-7.6 for account module I'm setting the condition like this

> Accounts type equal to one of "multi select dropdown of account type " AND 
> Accounts industry equal to one of "multi select dropdown of account industry".

When I run this report it is displaying nothing and if I see in suitecrm.log then it is printing wrong query.So query is failing.

Query generated for report is like this:

"SELECT `accounts`.id AS 'accounts_id', `accounts`.name AS 'Name0', `accounts`.account_type AS 'Type1', `accounts`.industry AS 'Industry2' FROM `accounts`  WHERE ( (`accounts`.account_type = 'Analyst') `accounts`.account_type = (`accounts`.account_type = 'Analyst') OR (`accounts`.industry = 'Apparel') ) AND  accounts.deleted = 0  LIMIT 0,20

: MySQL error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '`accounts`.account_type = (`accounts`.account_type = 'Analyst') OR (`accounts`.i' at line 1"

Any help will be appreciated.

Thanks in advance. 
suitecrmreportsbug

604media:


LewMegas:
Is this supposed to work for custom fields? 

We have a multi-select field called Categories and this feature is not working for us

Issue #1289, opened 2016-04-27
Unable to send email campaign SuiteCRM7.6RC & SuiteCRM7.6.1


chandan999:
Hi All,
       I am not able to send email campaign in Suitcrm 7.6 RC.Even test campaign emails are not going.Its getting stuck in queue.But same thing is working fine suiteCRM 7.5.3.

samus-aran:
Hi @chandan999 - does the logs say anything with regards to the queue?

chandan999:
After recent changes in code in suiteCRM 7.6.1

if (!$test && $lock_count!= 1) { 
			$GLOBALS['log']->fatal("Error acquiring lock for the emailman entry, skipping email delivery. lock status=$lock_count " . print_r($row,true));
			continue;  //do not process this row we will examine it after 24 hrs. the email address based dupe check is in place too.
		}

But still i am facing issue of sending emails manually from email queue.It is getting stuck "in process" state.Log says "Error acquiring lock for the emailman entry, skipping email delivery"

chandan999:
appendErrorMessage($error_message).This function is not working in suitecrm 7.6.1

samus-aran:
Thanks for the response @chandan999 We have assigned this accordingly.

dodexahedron:
Processing of the emailman table always fails when using mysqli, if the storage engine for the emailman table is innodb and autocommit is on (which it is, by default).

This is because the mysqli_affected_rows function returns only the number of rows affected by the LAST query, even if that query was implicit. If the table is innodb and autocommit is on, the last query was COMMIT, which affects 0 rows, leading to ALL outbound email processing failing.

Potential fixes:
Force storage engine of emailman table to something other than InnoDB (I hate this idea).
Make MysqliManager.php turn autocommit off on its connections and get the count before committing (seems reasonable).

dodexahedron:
Why is this "lock" checking code there, in the first place?  I think that's the more important question to be asking, here.

My proposed solution at least "fixes" it and the solution proposed by @gymad also looks like it would work around it, too, but both are hacks and the original "lock" code looks like a hack to get around something specific that may have been happening in the past, anyway.

If it's there to make sure that a SQL command actually did happen, then WOW we really aren't trusting MySQL, are we?
If it's there because MyISAM was not ACID, then...well...using InnoDB is the solution, there.
Either way, I honestly think the code just needs to go away.  Then, none of these hacky workarounds will be necessary in the first place.

[...]

PR #1355, opened 2016-05-06
Fix for #1289

dodexahedron:
This fixes https://github.com/salesagility/SuiteCRM/issues/1289

The MysqliManager::query function got a new optional parameter to control whether a query is auto-committed or not.
Then, EmailManDelivery calls with that parameter set to false, and commits the query after the count of rows affected is gotten.

### DIFF
--- a/include/database/MysqliManager.php
+++ b/include/database/MysqliManager.php
@@ -160,6 +160,14 @@ class MysqliManager extends MysqlManager
 
 		return $result;
 	}
+	
+	/**
+	 * Commits all uncommitted queries
+	 */
+	public function commit()
+	{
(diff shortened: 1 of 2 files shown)


samus-aran:
@dodexahedron 

As rule of thumb, we do not merge fixes into the 'master' branch.  You will need to create a new PR to merge to 'hotfix'.

Thanks for your contribution.

dodexahedron:
No problem. Will do.
Do you want it created on my local hotfix and then the PR made to the upstream hotfix?
Or do you not care at that point, so long as it merges to the upstream hotfix branch?

PR #1392, opened 2016-05-12
Fixed #441 Add totals to CSV export of reports

### LINKED ISSUE #441: AOR Totals not Exported
If a user adds a total(or multiple totals) to a report and then Exports that report, totals are not shown in the CSV. These are however shown in the PDF Template that can be downloaded.

emcrobert:
Fixed #441 Add totals to CSV export of reports

### DIFF
--- a/modules/AOR_Reports/AOR_Report.php
+++ b/modules/AOR_Reports/AOR_Report.php
@@ -833,6 +833,74 @@ class AOR_Report extends Basic {
         return $html;
     }
 
+    function getTotalCSV($fields,$totals) {
+        global $app_list_strings;
+
+        $currency = new Currency();
+        $currency->retrieve($GLOBALS['current_user']->getPreference('currency'));
+
+        $csv = "";
+        $delimiter = getDelimiter();
+
+        foreach($fields as $label => $field){
+            if(!$field['display']){
+                continue;
+            }
+            if($field['total']){
+                $totalLabel = $field['label'] ." ".$app_list_strings['aor_total_options'][$field['total']];
+                $csv .= $this->encloseForCSV($totalLabel);
+                $csv .= $delimiter;
+            }else{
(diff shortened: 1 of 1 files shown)


Dillon-Brown:
Hi @emcrobert, would you be able to either fix the conflicts or tick the checkbox to enable us to do so? thanks.

Dillon-Brown:
Closed due to inactivity

pgorod:
@samus-aran @Dillon-Brown 

I've been looking at this with greater attention and I would recommend:

- reopen this PR, it has value
- label it "community contribution" and "needs updates"
- it will require being re-made manually, since the file has changed a lot. 

For this work, I recommend:

- it seems this PR duplicates an existing function, `getTotalHTML` and renames it to `getTotalCSV`, making some changes to alter the output string. But meanwhile the original `getTotalHTML` has evolved quite a bit, so to update this PR someone should follow the most current version of `getTotalHTML` and apply the new changes to `getTotalCSV`. 
- optionally, to avoid code duplication, build a single function that does both things, receiving an argument specifying type (CSV or HTML) and handling the markup, delimiters and line breaks accordingly.

Dillon-Brown:
@pgorod Not sure I see much value in re-opening. We wouldn't be able to update the PR ourselves as they haven't ticked the checkbox and as you say the functions have changes significantly since this PR was created.

pgorod:
I am sure it can be replicated because if there is no code in place to add the totals data to the CSV, it won't be there in the file :)

I realize we wouldn't use the code here directly, this would just serve as a reminder that the problem exists, and where things need to be changed to fix it. Another option is to reopen the original issue, or open a new one, and leave a link to this closed PR. Whatever you prefer.

Dillon-Brown:
@pgorod Agreed, I'll re-open the original issue.

Issue #1406, opened 2016-05-14
Module install error on unchmodabble filesystems


antonioribeiro:
When you are on an hybrid filesystem, for instance, a Windows NTFS mounted on a Linux host, your filesystem may not be chmoddable, but the folder may be still writable. The result will the known error message while trying to install a module: 

```
Specified directory '' for zip file 'file.zip' extraction does not exist.
```

pgorod:
:+1: +1 for the concept of _unchmoddability_ of a file system... : - )

PR #1407, opened 2016-05-14
Fix for unchmodabble file systems

antonioribeiro:
When you are on an hybrid filesystem, for instance, a Windows NTFS mounted on a Linux host, your filesystem may not be chmoddable, but the folder may be still writable. In those cases we get an error while trying to install a module. This fix prevents it to happen.

Fix for https://github.com/salesagility/SuiteCRM/issues/1406

### DIFF
--- a/include/utils/sugar_file_utils.php
+++ b/include/utils/sugar_file_utils.php
@@ -69,7 +69,7 @@ function sugar_mkdir($pathname, $mode=null, $recursive=false, $context='') {
 	}
 
 	if($result){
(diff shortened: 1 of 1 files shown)


Dillon-Brown:
Hi @antonioribeiro, just working through the backlog here. Would you be able to fix the conflict or tick the box allowing us to do it?

Thanks for the contribution!

ijdavie:
closed as referenced pull request addresses issue.

PR #1411, opened 2016-05-16
Branch issue 1396 - AOR Conditions from custom fields from a related module do not work

604media:


shogunpol:
The PR solves only part of the issue, it work as described, however when you select in:
reports -> editView -> conditions the Parameter box, the solution it doesn't work(i.e.work only when Parameter is not ticked).
@604media , can you try solve this issue?

Dillon-Brown:
Closed due to inactivity.

gunnicom:
Should this really be closed as shogunpol said "The PR solves ... part of the issue"
So, there is an issue, and the PR fixes something.

Dillon-Brown:
@gunnicom In this case it appears that this code has already been merged in at some point along with the original issue being closed. In general I would say that even if that were not the case it should still be closed. We should be using the GitHub issue's to keep track of bugs as opposed to the pull requests themselves, if the original PR owner has no interest in updating the PR so it can be reviewed/merged or giving us access to make changes then there is not much point in leaving it up.

PR #1422, opened 2016-05-17
Changed MarkerClusterer.IMAGE_PATH value

### LINKED ISSUE #1405: Missing icons in events and maps
In events in "Select Delegates" there is a missing icon and in maps the "Cluster" icon will not be displayed.

-- Andreas

Vinedine:
Fixes #1405;  Changed value to new location

### DIFF
--- a/modules/jjwg_Maps/views/view.map_markers.php
+++ b/modules/jjwg_Maps/views/view.map_markers.php
@@ -193,7 +193,7 @@ var markerClusterer = null;
 var markerClustererToggle = null;
 var clusterControlDiv = null;
 // Clusterer Images - Protocol Independent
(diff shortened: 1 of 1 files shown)


PeterTheOne:
This will work, but I don't recommend fixing it like this. If the URL changes once more you will have the same troubles. Also Github is not a cdn.

Instead you can download and include the images from https://github.com/googlemaps/js-marker-clusterer/tree/gh-pages/images starting with `m`. And linking to those locally.

Also I recommend not setting the `IMAGE_Path` variable but the imagePath option on initializing the Clusterer, like this:

    var options = {
        imagePath: 'images/m'
    };

horus68:
@ijdavie maybe this is to be closed too? See related https://github.com/salesagility/SuiteCRM/issues/1438

horus68:
@shogunpol  maybe this is to be closed? 
See related https://github.com/salesagility/SuiteCRM/issues/1438

PR #1454, opened 2016-05-20
Feature: New Operator AOR_Reports "Not Contains"

gunnicom:
I need this operator, and more. This will be the first new Operator and i will add more if wanted.


LewMegas:
This would be a perfect solution to our problems, specifically regarding the creation of target lists from reports.

We have a custom multi-select field called Categories where a contact can be listed as a Current Employee,  a Contractor and an Ex-Employee for example.

We have followed the above steps and Not Contains is displaying as an Operator, however it does not appear to be excluding undesirable Categories when we create the report. 

Is this feature unlikely to work with custom multi-select fields?

horus68:
@gunnicom can you submit this PR to the proper branch (Hotfix)?

gunnicom:
Changed branch: https://github.com/salesagility/SuiteCRM/pull/2324

PR #1502, opened 2016-05-27
Fixed #1489 duplicating a report with a chart removes the chart from the original report

### LINKED ISSUE #1489: Duplicating report with a chart removes the chart from the original report
As per title, if you duplicate a report that has a chart, the chart gets removed from the original report, though everything looks correct in the duplicated report.

Originally reported by TaufiqueAhmed on the forums and I can replicate the issue myself using 7.6.2.

emcrobert:
Fixes issue #1489. Previously code checked if we had chart id and edited existing chart. This caused an issue as a chart can only be related to one report. Added check so if a report is being duplicated we create a new chart object.

### DIFF
--- a/modules/AOR_Charts/AOR_Chart.php
+++ b/modules/AOR_Charts/AOR_Chart.php
@@ -71,11 +71,12 @@ class AOR_Chart extends Basic {
     }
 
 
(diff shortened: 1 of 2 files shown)


samus-aran:
This was good, but it also required to match the new parameter in AOR_ChartTest.php

PR #1507, opened 2016-05-27
Fixed #1506 - IE11 cannot display generic 'application/octet-stream'.

### LINKED ISSUE #1506: IE11 does not display uploaded images
IE11 cannot display generic 'application/octet-stream' (which is defined as "arbitrary binary data" in RFC 2046).

vladbar:
Fixed #1506 - IE11 cannot display generic 'application/octet-stream' (which is defined as "arbitrary binary data" in RFC 2046).

### DIFF
--- a/modules/Contacts/views/view.detail.php
+++ b/modules/Contacts/views/view.detail.php
@@ -53,6 +53,17 @@ class ContactsViewDetail extends ViewDetail
 	public function display(){
 		global $sugar_config;
 
+		// Fix for issue 1506: removed picture is still cached in the Detail View.
+		$js=<<<EOQ
+		<script type='text/javascript'>
+			if(!window.location.hash) {
+				window.location = window.location + '#loaded';
+				window.location.reload();
+			}
+		</script>
+EOQ;
+		echo $js;
+
 		$aop_portal_enabled = !empty($sugar_config['aop']['enable_portal']) && !empty($sugar_config['aop']['enable_aop']);
 
 		$this->ss->assign("AOP_PORTAL_ENABLED", $aop_portal_enabled);
--- a/download.php
+++ b/download.php
@@ -170,7 +170,12 @@ else {
             $doQuery = false;
         }
 
-        $mime_type = 'application/octet-stream';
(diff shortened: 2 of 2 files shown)


modules/Contacts/views/view.detail.php
```diff
@@ -53,6 +53,17 @@ class ContactsViewDetail extends ViewDetail
 	public function display(){
 		global $sugar_config;
 
+		// Fix for issue 1506: removed picture is still cached in the Detail View.
```
mattlorimer: this will only work for contacts and may not be the best approach to solving this issue, I think  a global solution for images is needed

vladbar: I agree. Besides its not only images - there are also problems with 
Firefox and Chrome caching language strings for dropdowns.

This bit is too specific:
         // Fix for issue 1506: removed picture is still cached in the 
Detail View.
         $js=<<<EOQ
         <script type='text/javascript'>
             if(!window.location.hash) {
                 window.location = window.location + '#loaded';
                 window.location.reload();
             }
         </script>
EOQ;
         echo $js;

for:

issue 2: Firefox and Google Chrome display cached image.
Steps to reproduce:
use image field in the Contacts module.
Add an image in Edit View, then save.
Go back to the edit view and and add another image and save again.
The detail view still displays the first image.



But this bit at least worked for
issue 1: IE11 does not display image saved through the field of type Image.

// Fix for issue 1506: IE11 cannot display generic 
'application/octet-stream' (which is defined as "arbitrary binary data" 
in RFC 2046).
$mime_type = mime_content_type($local_location);
if($mime_type == null || $mime_type == '') {
     $mime_type = 'application/octet-stream';
}

I should've created two pull requests as these are two different problems.

Do you want me create a new pull request with just:
// Fix for issue 1506: IE11 cannot display generic 
'application/octet-stream' (which is defined as "arbitrary binary data" 
in RFC 2046).
$mime_type = mime_content_type($local_location);
if($mime_type == null || $mime_type == '') {
     $mime_type = 'application/octet-stream';
}

- I tested it works for IE11.

mattlorimer: yes it would be better to keep it separate, so we can pull it in, with any other fix getting in the way

Issue #1519, opened 2016-05-31
7.6.4 AOR - Group checkbox stoped working


mmadey:
After upgrade to 7.6.4 the grouping checkbox in AOR stopped working. 
I had a report that did grouping per employee, adding all of their opportunity values and counting how many where there. After upgrade to 7.6.4 for each opportunity there is a separate line. 

In previous
![group1](https://cloud.githubusercontent.com/assets/18421032/15674503/60bb4fd4-273d-11e6-9668-f69c46cd35f1.png)
![group2](https://cloud.githubusercontent.com/assets/18421032/15674504/60bd0e00-273d-11e6-80b8-a01601299cd3.png)

 versions (7.6.x and 7.5.x) it worked fine.

PR #1527, opened 2016-06-01
Branch issue 1519

604media:


pgorod:
@604media just a tip: if you use the expressions `close, closes, closed, fix, fixes, fixed, resolve, resolves, resolved` in the title or description, Github will automatically reference the PR in the issue, which is useful to navigate around.

(It would also close the issue automatically if the PR was against the default branch, but that won't apply here, if I understand correctly. See [here](https://help.github.com/articles/closing-issues-via-commit-messages/))

So in this issue you could write `Fixes #1519` and Github would automatically auto-fill the title of the issue (that's also useful!) and cross-reference it from the issue to the PR.

And thanks for your contributions!

604media:
@pgorod Thanks for your comment.

mattlorimer:
doesn't work with mssql

PR #1542, opened 2016-06-02
Fixes 1535 - Campaigns - Create Person Form

604media:


samus-aran:
@604media - Ref comment here:
https://github.com/salesagility/SuiteCRM/pull/1543#issuecomment-227682585

PR #1554, opened 2016-06-06
Develop 7.7

gymad:


pgorod:
Are you sure a 29MB text file should be in this commit?
https://github.com/salesagility/SuiteCRM/pull/1554/commits/640f270fcfd50e39081bd7bb96eb22d68d8dcb7f

pgorod:
@samus-aran the 29Mb text file is now published in the installer for the 7.7 Beta. It looks like a mistake to me, but I might be wrong.
It's in themes/default/images/cmpwzrd/permission.txt
and it's coming from the commit referenced in the above comment.

daniel-samson:
Thanks for pointing this out. I have made a new pull request pr #1611 to correct this issue.

pgorod:
@daniel-samson I'm not sure how the installer scripts handle file deletions, but it would be nice for people that installed the 7.7 Beta, and later upgrade, to have that file deleted from their local installations Thanks.

Issue #1249, opened 2016-04-19
Reminders Module - There is no action by that name


AsimPervez93:
Whenever I click on the Reminders module, I get this error:

![screenshot_6](https://cloud.githubusercontent.com/assets/17923501/14641098/854dc92a-063c-11e6-8d74-877c4bcc391d.jpg)

I've tried disabling the module in AJAX but this did not fix it.

ShadMickelberry:
As far as I know there is no interface for Reminders. They are used by the Activities modules (Tasks, Calls, and Meetings) for notifications. You should probably just remove them in the Display Modules and Subpanels in Admin.

samus-aran:
Agreed - We should remove the Reminders as a navigational tab.

shogunpol:
The described issue has been tested on recent version of SuiteCRM(7.7.2), and it does not appear as is expected.
@AsimPervez93 If you agree can you close this issue.

ijdavie:
Due to no activity no this issue, We will close this issue unless more information can be given. 
Thanks.

PR #1555, opened 2016-06-06
Fixed: #1249 - Hiding Reminders for the navigation but keep them in the moduleList

daniel-samson:
Related to issue #1249. Hiding Reminders for the navigation.


daniel-samson:
This change prevents normal users from setting up reminders.

PR #1573, opened 2016-06-08
Fixed #1568 - Removed JS alert messages from the Alerts Manager

### LINKED ISSUE #1568: Remove Alert Manager Message
In the manager it says "Click OK to view this call or click Cancel to dismiss this message". This doesn't make sense in this context.

daniel-samson:


samus-aran:
Note - Requires a reMinifying JS and Group JS

PR #1574, opened 2016-06-08
Merging changes from harris

daniel-samson:


lewisje1991:
I believe the branch should be hotfix.

PR #1584, opened 2016-06-10
Fixed#1570 - Removing uninvited from reminders

daniel-samson:
(cherry picked from commit 4d467a4)


samus-aran:
Works if working using one method via editview - will raise github issue for other scenarios

PR #1593, opened 2016-06-13
Fixed #1549 - Case status should default to Closed if its State is Closed but it has the following statuses: New, Assigned or Pending Input

### LINKED ISSUE #1549: Mass update - Case state/status values
Hi,

When you mass update the cases, and you will change the state value from open to close, you will still see them (the cases) as open, because, there was not changed the status value... same issue with status value, if you change the status value via mass upadte, the state value still shows as open.

Same issue you get if you using the inline edit :) 

case

vladbar:
Fixed #1549 - Case status should default to Closed if its State is Closed but it has the following statuses: New, Assigned or Pending Input


mattlorimer:
to specific

PR #1614, opened 2016-06-17
Fixed #1558 - AOR Related Fields With Link Option

### LINKED ISSUE #1558: 7.6.4 - AOR - If row has "Link" checked, the ID is missing from the link
This seems to be an issue on 7.6.4
If you create a report, and select "Link" for a row,  the returned results' links don't have the ID, thus the link doesn't work.

screenshot from 2016-06-07 09 40 53

Found via this forum post:
https://suitecrm.com/forum/suitecrm-7-0-discussion/9596-reports-bug-or-am-i-doing-something-wrong#32998

daniel-samson:


HeiHyd:
Hey,
the Link works perfect now, but the scheduled report is broken.

bildschirmfoto 2016-07-08 um 06 46 47

samus-aran:
@HeiHyd Is this is caused by this fix for scheduled reports?  I wouldn't think it would.  Can you confirm scheduled reports worked prior to this fix?

daniel-samson:
Added a fix for the parent module fields. I have tested it on both MySQL and MS SQL, they both seem to work.

modules/AOR_Reports/AOR_Report.php
```diff
@@ -692,8 +692,18 @@ function build_report_html($offset = -1, $links = true, $group_value = '', $tabl
             foreach($fields as $name => $att){
                 if($att['display']){
                     $html .= "<td class='' valign='top' align='left'>";
-                    if($att['link'] && $links){
-                        $html .= "<a href='" . $sugar_config['site_url'] . "/index.php?module=".$att['module']."&action=DetailView&record=".$row[$att['alias'].'_id']."'>";
+                    if(!empty($att['link']) && $links){
+                        if(
+                            !empty(array_key_exists($att['field'], $field_bean->field_name_map)) AND
```
samus-aran: I also get that write context error: PHP Fatal error:  Can't use function return value in write context.
Change the function:
array_key_exists($att['field'], $field_bean->field_name_map) !== FALSE

modules/AOR_Reports/AOR_Report.php
```diff
@@ -1097,32 +1121,68 @@ function build_report_query_select($query = array(), $group_value =''){
...
+     * @param $table_name - (string) table name
+     * @param $query - (array) used to build up query
+     */
+    function injectParentModuleIDField ($table_name, &$query) {
+        $needle = $this->db->quoteIdentifier($table_name).".id";
+        $found = false;
+
+        // find if the key already exists
+        foreach($query['select'] as $select) {
+            if(!empty(stristr($select, $needle))) { $found = true; break; }
```
samus-aran: I get another write context error: PHP Fatal error:  Can't use function return value in write context.
Change the function:
if(stristr($select, $needle) !== FALSE) { $found = true; break; }

Issue #1563, opened 2016-06-08
Scaling on barcharts incorrect


gunnicom:
![chart_axis](https://cloud.githubusercontent.com/assets/550251/15888561/9dc2e6be-2d67-11e6-865a-dabf261774e1.PNG)
For small y-axis max values that are not dividable by 5 the axis scaling is faulty.
In the graph you can see its 1,2,4,5,6 so 3 is missing and scaling is bad (2 shows as 1.7)
For me it works better if skipping the function calculateMaxYForSmallNumbers() completely, so rgraph handles the scaling better.

```
modules/AOR_Chart.php   
                ymax:calculateMaxYForSmallNumbers($chartDataValues)      [position 392:22]  
                ymax:calculateMaxYForSmallNumbers($chartDataValues)      [position 434:22]  
                ymax:calculateMaxYForSmallNumbers($chartDataValues)      [position 467:22]  
                ymax:calculateMaxYForSmallNumbers($chartDataValues),      [position 546:22] 
```

```
include/SuiteGraphs/RGraphIncludes.php  
                function calculateMaxYForSmallNumbers(dataPoints)      [position 28:26] 
```

paulgodley:
Hi there.  The calculateMaxYForSmallNumbers was added to resolve bug #876.  A compromise would be to make the maxYForSmallNumbers set to 10, as this should allow for appropriate scaling of the smaller numbers?

gunnicom:
That would not change the problem, as it will then again occur with eleven. I think it maybe good to have the scale dividable by 5. I will add a patch for that.

gunnicom:
Just for reference, there is something on that on rgraph. But got no real solution:

http://www.rgraph.net/forum/message.html/y-axis-labels-are-incorrect-when-all-data-values-are-1-ok-if-one-or-more-20140529132751.html

gunnicom:
Closed first pull request because i like the new one more.

PR #1616, opened 2016-06-17
Fix for https://github.com/salesagility/SuiteCRM/issues/1563

gunnicom:
If max is less than 5 return 5, else return null so rgraph handles the scaling.
Fix for #1563

### DIFF
--- a/include/SuiteGraphs/RGraphIncludes.php
+++ b/include/SuiteGraphs/RGraphIncludes.php
@@ -31,8 +31,9 @@ $chart = <<<EOD
                     if(dataPoints !== undefined && dataPoints.length > 0)
                         largest = Math.max.apply(Math,dataPoints);//http://stackoverflow.com/a/14693622/3894683
 
-                    if(largest === null || largest < maxYForSmallNumbers)
-                        return maxYForSmallNumbers;
+                    if (largest === null || largest < maxYForSmallNumbers) {
+                      return maxYForSmallNumbers;
+                    }
                     return(null);
                 }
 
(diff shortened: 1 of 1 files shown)


horus68:
@gunnicom I'm not on the dev team but can you commit this PR to the proper branch (hotfix)?

gunnicom:
I wont do that until i know if its worth the work. Waiting months, then when asked for change the branch. Done that within minutes, than the pull request is not wanted. That is totally demotivating.
See: https://github.com/salesagility/SuiteCRM/pull/2324

horus68:
I do understand your point I also have some PR from me waiting (or ignored!). 
My request was a personal one and after been browsing some old PR loking for fixes I could use for myself.

Sometimes a PR could not be approved by a particular reason but it could be picked up later in a new direction. **It can also be used by other users for their own installations.**
And maybe this one gets better luck and be merged!

That's why its relevant to have it on the Hotfix branch and avoid it be totally ignored!
Note: there has been many PR merged these days (since 7.4.4) so I hope the backlog will be reduced.

gunnicom:
Many PRs are merged ... mostly SuiteP theme PRs, because instead of fixing old bugs suite adds a complete new theme with many new bugs instead of fixing old ones.
One more thing that holds my motivation from rising.
Enough said on this topic. Maybe one day i will start doing something again, but not now.

Dillon-Brown:
@gunnicom If you can change this to branch hotfix-7.8.x we can get this merged. Thanks.

PR #1644, opened 2016-06-24
Fix #1472 - strip HTML tags from notification e-mail Subjects

### LINKED ISSUE #1472: HTML tags visible in notification email Subject
From what I gather, every single notification email for **every SuiteCRM installation that is using a Crowdin translation** is currently being sent with this sort of subject line (as viewed by end-users):

Record assignment:
`<p>Contacto CRM - Francisco Silva</p>`

Call Reminder:
`<p>Chamada de CRM - Perguntar ao cliente</p>`

Despite not being a critical bug, it _is_ quite visible and doesn't make SuiteCRM look good.

Those HTML tags are a consequence of the way the Crowdin translation platform handles the template file `include/en_us.notify_template.html` which is not a real HTML file, though it pretends to be one. I've checked with @horus68 and there is no easy fix for this from the Crowdin side of things.

So I will propose a PR here to fix this, by stripping HTML from all email notification's Subject lines. I argue that this makes sense anyway because we are using bits of HTML from a template file, but subject lines should never include HTML tags, according to the official specs.

This is my first PR, let me know if I'm doing something wrong...

pgorod:

### DIFF
--- a/data/SugarBean.php
+++ b/data/SugarBean.php
@@ -2765,7 +2765,7 @@ class SugarBean
         $xtpl->parse($template_name . "_Subject");
 
         $notify_mail->Body = from_html(trim($xtpl->text($template_name)));
-        $notify_mail->Subject = from_html($xtpl->text($template_name . "_Subject"));
+        $notify_mail->Subject = strip_tags(from_html($xtpl->text($template_name . "_Subject")));
 
         // cn: bug 8568 encode notify email in User's outbound email encoding
         $notify_mail->prepForOutbound();


samus-aran:
@pgorod - just need you to re-sign the CLA (since it had changed hands) :+1:

pgorod:
@samus-aran I had already signed it a while back. Anyway, I just went into that CLA site and signed it again. Now it says "You have signed this CLA" but the PR still shows the failed check.

Can you somehow tell Github to re-check?

pgorod:
@samus-aran ... or, if you can't force a re-check, just close this PR, and when I get notified of that, I will do a new one. Hopefully the check will pass now, and it's not that much work recreating the PR...

pgorod:
I'll try closing and redoing the PR to see if that clears the CLA check

PR #1648, opened 2016-06-24
Bug Fix for audit fields not working with AOW any change.

lewisje1991:
This fixes bugs with AOW not picking up on any change when a field is audited.

### DIFF
--- a/data/SugarBean.php
+++ b/data/SugarBean.php
@@ -5495,27 +5495,36 @@ class SugarBean
         return $this->create_new_list_query($order_by, $where, array(), array(), 0, '', false, $this, true, true);
     }
 
-
     /**
+     * Checks auditing is enables and then carrys out an audit on the current bean.
+     *
      * @param $isUpdate
      */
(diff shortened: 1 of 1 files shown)


data/SugarBean.php
```diff
@@ -5515,4 +5494,28 @@ public function create_export_query($order_by, $where)
...
+                foreach ($auditDataChanges as $change)
+                {
+                    $this->db->save_audit_records($this, $change);
+                    $this->fetched_row[$change['field_name']] = $change['after'];
+                }
+            }
+        }
+        else
+        {
+            $GLOBALS['log']->debug('Auditing: Retrieve was not called, audit record will not be created.');
```
mattlorimer: should only log it if $isUpdate is true and !isset($this->fetched_row), maybe split $this->is_AuditEnabled() && $isUpdate into a separate if statement

vijaykanugonda:
Still issue persists and code changes are present in sugarbean in 7.8.3

PR #1654, opened 2016-06-25
Strip HTML tags from notification e-mail Subjects (reprise)

pgorod:

### DIFF
--- a/data/SugarBean.php
+++ b/data/SugarBean.php
@@ -2765,7 +2765,7 @@ class SugarBean
         $xtpl->parse($template_name . "_Subject");
 
         $notify_mail->Body = from_html(trim($xtpl->text($template_name)));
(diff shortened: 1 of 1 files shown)


pgorod:
@samus-aran I'm sorry, I've done everything I could. 
You'll have to check that clahub test, it's not working properly...

samus-aran:
Hey @pgorod - We'll check the integration but it most likely its an issue on the Clahub side.  We'll raise an issue with them if we can't see anything obvious.  But unfortunately for the time being we can't progress then further.

pgorod:
@samus-aran 
Ok!
Since this is such a small edit, if you prefer to do it yourself in a new PR, go ahead.
I will still be able to ["bask in the warm fuzzy glow"](http://www.jsmackin.co.uk/suitecrm/contributing-to-suitecrm/) (quoting Jim Mackin) of helping the project.

PR #1680, opened 2016-06-30
Fixed #1679 - Emails module broken

### LINKED ISSUE #1679: Emails module broken
When translation file contains single quotes then Emails module is not working and throws JavaScript errors in browser. Single quotes should be escaped when exported to JavaScript variables.

Confirmed on version 7.6.4 with polish translation.

krzyc:
Escape single quotes when creating translation strings for JavaScript.
#1679

### DIFF
--- a/modules/Emails/EmailUI.php
+++ b/modules/Emails/EmailUI.php
@@ -441,6 +441,7 @@ eoq;
 		$lang = "var app_strings = new Object();\n";
 		foreach($app_strings as $k => $v) {
 			if(strpos($k, 'LBL_EMAIL_') !== false) {
+				$v = str_replace("'", "\'", $v);
 				$lang .= "app_strings.{$k} = '{$v}';\n";
 			}
 		}


krzyc:
I do not understand why clahub check failed, because on CLHHub I have "You have signed this CLA." message.

samus-aran:
@krzyc There is an issue at the moment with clahub and github where the fails check if you make a change via the web interface.

krzyc:
@samus-aran Should I take any other action (e.g. make commit from true git) or just wait for fix and commit will then be accepted?

gymad:
Hi, I had same issue in the develop for different reason. I think your fix doesn't correctly cause the escaped variable could be both of string and array and I think may it is because the str_replace escaping removed from this part. I recommend apply a json_encode on the value instead of str_replace as you can see in c83808e commit.

krzyc:
@gymad Your fix looks great. And if it is so flexible than maybe $email_mod_strings and $ieModStrings should be treated same way instead of str_replace?

Dillon-Brown:
Closing this as a duplicate of already merged PR #1642.

PR #1684, opened 2016-06-30
Html code typo in URL

horus68:
some extra clean up


samus-aran:
@horus68 There is currently an issue with clahub and that users who submit code changes via the github interface fail the check.  We unfortunately are awaiting for the fix at clahub end to be deployed but currently we advise (if possible) for users to submit code changes via their development environment rather than github interface.

PR #1694, opened 2016-07-04
Only add Bean to array if it really exists

gunnicom:
Otherwise if the bean does not exist php will throw "Trying to get property of non-object"


Dillon-Brown:
Please fix conflicts and update branch to hotfix-7.8.x.

Dillon-Brown:
Looks like this has already been fixed here: #2629

gunnicom:
@Dillon-Brown An this is another good example why these long standing PRs are a problem.
My fix was from July 2016, and the #2629 was from November 2016 ... two people doing the same work, because nobody at salesagility cared. Maybe mine is inferior, but that does not make it better.

Dillon-Brown:
@gunnicom I think we can all agree that the longstanding PR's need addressed, just a matter of time/resources. As you can see I've started to focus on the PR's from the back of the list and getting them assessed / ready to merge. Thank you for all your contributions so far.

PR #1695, opened 2016-07-04
Fix #1693 - Failing tests

### LINKED ISSUE #1693: Tests are failing in devel branch
Dear @JimMackin ,

I am addressing this issue to you because (to my great happiness) you have created the first test suite.
IMHO, testing for a project of this size is not an option but a must and the automated tests for each PR or commit is absolutely necessary.

As you will be aware, the actual extent of our tests cover around 7% of the code base but it is much better than nothing.

The issue here is that I have a feeling that testing and CI integration is not taken seriously and PRs and commits make their way to branches with failing tests.

I have just checked out the devel branch with 6 failing tests.
How's one meant to develop for SuiteCrm in a safe manner if the branches contain failing tests?

I feel really discouraged to contribute in this manner.
Can't we work something out?

adamjakab:


adamjakab:
These tests on the sql strings are making me crazy - column orders are different in sql statement on my devel environment and Travis.

PR #1699, opened 2016-07-05
Test fixes for random custom field order

adamjakab:
AccountTests should now pass

### DIFF
--- a/modules/DynamicFields/DynamicField.php
+++ b/modules/DynamicFields/DynamicField.php
@@ -169,7 +169,7 @@ class DynamicField
         }
 
         $GLOBALS['log']->debug('rebuilding cache for ' . $module);
-        $query = "SELECT * FROM fields_meta_data WHERE $where deleted = 0";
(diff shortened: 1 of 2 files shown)


adamjakab:
@samus-aran ,

ok - this should be ok now - I have introduced SuiteCRMUnitTest and SuiteCRMFunctionalTest classes because we should be doing both (at the moment they are mixed together.) but I will do that in separate PR.

I have not found the cause of the inversed order fields in the sql statements. 
On Travis they were added in inversed alphabetical order:

- jjwg_maps_lng_c
- jjwg_maps_lat_c
- jjwg_maps_geocode_status_c
- jjwg_maps_address_c

and on my local development they were the other way around.

Anyhow, now this PR makes sure that they are always in alphabetically increasing order.

chris001:
This is a worthy PR from @adamjakab .  
It's been almost a year while it's been in "Review" status.
@mattlorimer what are you reasons for not merging this PR ?
What else do you need?
Please respond.

PR #1698, opened 2016-07-04
notify_template.html - include proper html tags to allow translation

horus68:
Formating include/language/en_us.notify_template.html
with html tags to allow translation on Crowdin.

This cause an issue on email subject filed but its solved by a related PR #1472 and #1654 that should be implemented at same time.

This PR is a recreation from #867 because I had to delete my repo.

PR #1702, opened 2016-07-05
Strip HTML tags from notification e-mail Subjects

samus-aran:
## Description
This is in reference to a PR (https://github.com/salesagility/SuiteCRM/pull/1654) that was made by @pgorod which needed to be merge via git itself.

## Motivation and Context

## How To Test This
First - pull down horus68 PR (https://github.com/salesagility/SuiteCRM/pull/1698) to have HTML tags within Assigned To notification emails.  Then Assigned a record whch actives this email.  The received email should not receive HTML within the subject.

## Types of changes

### Final checklist

### DIFF
--- a/data/SugarBean.php
+++ b/data/SugarBean.php
@@ -2762,7 +2762,7 @@ class SugarBean
         $xtpl->parse($template_name . "_Subject");
 
         $notify_mail->Body = from_html(trim($xtpl->text($template_name)));
-        $notify_mail->Subject = from_html($xtpl->text($template_name . "_Subject"));
+        $notify_mail->Subject = strip_tags(from_html($xtpl->text($template_name . "_Subject")));
 
         // cn: bug 8568 encode notify email in User's outbound email encoding
         $notify_mail->prepForOutbound();


samus-aran:
Ok guys (@horus68 and @pgorod ) - this needs more work that just your two PRs.  Testing this gives me an email but shows the html tags as the email sent is in text/plain content type.  Even if I set that notification email as default HTML using the IsHTML(true) function it now breaks the link to display the record - I haven't even tried it on a Case description yet.

So a little more work is needed - so this will need to go up for review internally.

pgorod:
@samus-aran 

I suppose my fix didn't create any problems, just removing tags from Subject field. You didn't get any problems in the subject field, did you? I had tried that fix in my system, both in English and in Portuguese, and the emails were sent correctly.

Remember his fix is dependent on mine, but mine isn't dependent on his. And (if it is in fact working) it does fix every single notification email sent in every translated SuiteCRM instance.

The problems you saw (in the email body, I assume), which language were you using? English, or some translation from Crowdin?

It's possible that @horus68 also has to do some work on the Crowdin side of things, for this all to fit in...

samus-aran:
Hey @pgorod 

You said the email sent correctly yeah? Are you talking about your fix only or did you test with html in the body and subject in the notification template? In theory with just your fix, if you had html in body and subject only the subject will be stripped and you will physically see the html tags inside the body as well. Is that an accurate test?

pgorod:
On my system I never added horus68's fix, so **using just SuiteCRM code** and a translation pack I got this:

email1
Notice that link won't work, there are a few problems there, but whatever they are, they are unrelated. This is just a test system and I need to configure it more carefully (notably, to remove the trailing `/` from my hostname, to avoid that duplication `//`). But I don't know where that forward slash `/` after `index.php` comes from, and I don't think it can possibly work... the HTML is:
```
<<a href='parse.pl?redirect=http%3A%2F%2F10.0.0.200%2F%2Findex.php%3Fmodule%3DContacts%26action%3DDetailView%26record%3Dd3ea5e5a-7d77-b5ad-953a-56a0090aeb36' target='_blank'><font color='red'>http://10.0.0.200//index.php\?module=Contacts&action=DetailView&record=d3ea5e5a-7d77-b5ad-953a-56a0090aeb36</font></a>>
```
That doesn't look good at all... double `<<` in `a` tag, forward slash in link... but the same link is generated if I send the mail in English!

But the relevant thing for my fix are only those `<p>` and `</p>` on the subject line (on the top, where it says `Assunto`). The test done after my fix basically leaves everything the same except it removes those.

On the same system, in English, no tags show on the subject line (both before and after the fix).

I never touched any options for Text or HTML email, where do I find that option?

samus-aran:
Thanks for additional comment - I'm getting the idea now. I'll test with a language pack specifically to see what html is included. There are no options for notification emails - they only send in plain text. So if there was html - you could confirm that by looking at the source of your email - the system wouldnt convert it to display html in the mail client anyway and all you'll see is html tags (like you do in your subject without the fix).

pgorod:
One problem here is that we are probably looking at previous bugs (the weird link), that were simply going by unnoticed. This could be confirmed on a clean system with no fixes and no language packs.

Another thing that can be confusing is that the file that horus68 changed - notify_template.html - is not really an html file. It should probably be renamed to .tpl extension, because it is just a list of strings that the SuiteCRM code uses to piece together the emails.

However, those strings are separated with some HTML paragraph tags, and that's how Crowdin knows how to break up the file so that each string can be translated individually.

In the future, it would be useful to be able to have more functional templates for emails, so that better looking, full HTML could be generated (keeping the strings apart for translation). These text emails are really ugly. But that's the future, for now it would be nice just to get all this working. Thanks.

horus68:
@samus-aran  any updates on this?
The problem is not solved yet.
The alternative is to not have this file translated and that should not be the way to do!

pgorod:
**The easy part**: merge PR #1702 with the simple fix for the HTML tags in the subject line. It's such a simple fix, and I've tested it ok. _As long as it is tested without the other fix_, it's really easy to approve this. Just get an experienced PHP developer to look at it and conclude it is harmless. This will solve one  problem that is visible to end-users, so it's worth doing right away.

**The hard part**: getting the translation system to work correctly with `include/language/en_us.notify_template.html`. This will solve the problem of the extra work necessary to manually generate translation files. This is a relevant problem because horus68 now maintains 53 languages!

I'm willing to restart my efforts with testing and correcting this second problem. @horus68, how can I test this? It's a code change that goes together with a language file change, right? I have to change both at the same time and try a notification email? Where do I get the correct language file for this? Thanks.

horus68:
The file uploaded to Crowdin is an edited file to include html tags. Crowdin uses it as it is and include the file in the pack with no extra work. 


or as a direct link - The uploaded html:
https://dl.dropboxusercontent.com/u/441710/Suitecrmissues/uploaded%20to%20crowdin%20en_us.notify_template.html
If you use a diff tool you can see the differences from the english one included in the core pack or sse the PR here:
See the changes here: https://github.com/salesagility/SuiteCRM/pull/1698

pgorod:
I worked on this during the weekend, here's what I found about **the current state** of things, without any PR's applied:

- **we're sending the link as text** but most Email clients will "linkify", making it clickable. But we're not sending a full HTML link as in  `<a href="...">`
- if we **put HTML tags into the template** file en_us-notify.html (this file extension seems to encourage us), it will go into the final email. In this case, we're still sending email in text, but the fact is that some HTML tags might be recognized by specific Email clients. So we can be sending `<p>` and `<br>` and they will be interpreted as paragraphs and breaks by some webmail clients, but not in Gmail, where they show literally.
- working with this requires **testing with several Email clients**. Gmail is a specially tricky one and deserves special attention. It helps to have one client that will show you the source of the email (e.g. Thunderbird).
- if we ever want pretty notification emails, with a more complete HTML that works, we need to move to multipart email, where we send both as text and as **proper HTML**, with the appropriate MIME type.
- I am almost sure there are currently **bugs with the link generation** (the one you click on the email to access the assigned record in SuiteCRM). Links don't work for me, but seem to be working for other people. I will simply avoid this issue, and leave the link as it is. If there is a bug, I won't make it better or worse with any changes on the current PR.

To get horus68's new template file to work, code changes are needed. Basically, since sending the tags with the text email seems unadvised, I would lean towards simply **removing these two tags** `<p>` `<br />` from the email body before sending. I can do this easily, I'm just not sure it is the best solution.

I would turn this (in SugarBean.php):
```
        $notify_mail->Body = from_html(trim($xtpl->text($template_name)));
        $notify_mail->Subject = from_html($xtpl->text($template_name . "_Subject"));
```
into something like this:
```
        $tempBody = from_html(trim($xtpl->text($template_name)));
        // Crowdin translation system requires some tags in the template; we remove them here:
        $someTagsToStrip = array(
            "<p>" => "",
            "</p>" => "",
            "<br />" => "",
            "<br/>" => "",
            "</br>" => "",
            "<br>" => "");
        $notify_mail->Body = strtr($tempBody, $someTagsToStrip);
        // now strip _all_ HTML tags from subject where they are never allowed:
        $notify_mail->Subject = strip_tags(from_html($xtpl->text($template_name . "_Subject")));

```
@samus-aran is this acceptable? It will work for the current templates (they have no tags, so stripping `<p>` and `<br>` tags will keep them the same). It will also work for the new templates with tags.

pgorod:
@horus68 please confirm that these will be the only two tags you will be adding: `<p>`  and `<br>`.

Or does Crowdin require anything else?

horus68:
@pgorod  The only tags used in the file are:
`<p></p>`
`<br />`
Note: the br format is not required to be strictly like that but I had made it all the same in the file.

I think we can also use Divs if you need!!

pgorod:
I would love to have some input from samus-aran, but she hasn't been active, I think she's on holidays, or doing something else. 

I've been reading about PHP Mailer, the class we use to send mail, and I still have to do some more tests. Maybe getting an HTML version of the email isn't too hard, and it would be nice to be able to make these emails better looking.

But the basic approach is already taking shape: en_us-notify.html will include exclusively `p` and `br` tags (we'll allow for all possible forms or variations of opening and closing tags) and these will be stripped in text versions of the email. 

It's not a very elegant solution (replacing HTML tags as text replacements, instead of using some nice and complete HTML parsing class) but it would work, and it isn't _less_ elegant than what is currently in the code... : - )

pgorod:
@horus68 I'm sorry, I only get a few hours every week to look at this and it's turning out to be more complicated than I initially thought. 

Anyway, I think I have it working now, with nicer links, and nicer looks, but I would like to check with you if it's appropriate for Crowdin. Can you please try these blocks in Crowdin to see if they are handled appropriately (i.e. easy for translators)?

**For Accounts:**
```
`<!-- BEGIN: Account_Subject -->
<p>SuiteCRM Account - {ACCOUNT_NAME}</p>
<!-- END: Account_Subject -->
<!-- BEGIN: Account -->
<p><b>{ASSIGNER}</b> has assigned an Account to <b>{ASSIGNED_USER}</b>.</p>
<p>
Name: {ACCOUNT_NAME}<br />
Type: {ACCOUNT_TYPE}<br />
Description: {ACCOUNT_DESCRIPTION}
</p>
<p>You may <a href={URL}>review this Account</a>.</p>
<!-- END: Account -->
```
The novelties are some extra tags (like `<b>`) for better graphic effect and proper links with a text to click on, instead of clicking directly on a URL.

**Here's another one, more complex, for Meetings:**
```
<!-- BEGIN: Meeting_Subject -->
<p>SuiteCRM Meeting - {MEETING_SUBJECT}</p>
<!-- END: Meeting_Subject -->
<!-- BEGIN: Meeting -->
<p>To: {MEETING_TO}</p>
<p><b>{ASSIGNER}</b> has invited you to a Meeting</p>
<p>
Subject: {MEETING_SUBJECT}<br />
Start Date: {MEETING_STARTDATE}<br />
End Date: {MEETING_ENDDATE}
</p>
<p>
<!-- BEGIN: Meeting_External_API -->
Meeting URL: {MEETING_URL}
</p>
<!-- END: Meeting_External_API -->
<p>Description: {MEETING_DESCRIPTION}</p>
<p><a href={ACCEPT_URL}&accept_status=accept>Accept this meeting</a>.</p>
<p><a href={ACCEPT_URL}&accept_status=tentative>Tentatively Accept this meeting</a>.</p>
<p><a href={ACCEPT_URL}&accept_status=decline>Decline this meeting</a>.</p>
<!-- END: Meeting -->
<!-- BEGIN: MeetingSpecial_Subject -->
<p>SuiteCRM Meeting - {MEETING_SUBJECT}</p>
<!-- END: MeetingSpecial_Subject -->
<!-- BEGIN: MeetingSpecial -->
<p><b>{ASSIGNER}</b> has assigned a Meeting to <b>{ASSIGNED_USER}</b>.</p>
<p>
Subject: {MEETING_SUBJECT}<br />
Status: {MEETING_STATUS}<br />
Start Date: {MEETING_STARTDATE}<br />
End Date: {MEETING_ENDDATE}<br />
Description: {MEETING_DESCRIPTION}
</p>
<p>You may <a href={URL}>review this Meeting</a>.</p>
<!-- END: MeetingSpecial -->
```
If you could please try pasting these blocks in your en_us.notify_template.html (replacing the corresponding blocks that exist there already) and loading it into Crowdin. Then tell me if it's ok for translation purposes. Thanks.

horus68:
It works just fine 👍 
step2_en_us notify_template html

pgorod:
I made a new PR #2401 with all the changes together. Let's hope everything works fine and that @samus-aran or someone can merge it soon. Thanks to everyone involved and sorry for my delays, I'm only starting to learn my way through PHP and Github.

PR #1751, opened 2016-07-14
Suitep editview pr

daniel-samson:
Styling Edit Views

## Motivation and Context
SuiteP theme
## How To Test This
Test the edit views for each module.

## Types of changes
Changed to the editview.tpl's.
There is a change in the detailview.tpl which fixes the header issue.

## Also
New tpls must be created for each module that has custom header/footer code and place in the suitep theme folder


themes/SuiteP/css/style.css
```diff
@@ -806,23 +807,30 @@ th, td {
     margin-bottom: 2px;
 }
 
-#EditView_tabs {
-    padding: 8px;
-    background: #fff;
-}
+/*#EditView_tabs {*/
```
samus-aran: Don't comment out - remove

themes/SuiteP/css/style.css
```diff
@@ -3805,37 +3871,211 @@ fieldset[disabled] .btn-primary.active {
...
+    background-color: #D66C60;
 }
 
 #EditView_tabs .edit input, #EditView_tabs .edit textarea {
     margin-top: 5px;
     margin-bottom: 5px;
 }
 
+.col-sm-6 .edit-view-field input[type="text"] {
+    width: 60% !important;
```
samus-aran: Are these !important going to cause issues with the 1/2 column widths?

daniel-samson: The do not effect the column widths. They effect the text fields inside of the column widths.

There is a issue with making the fields display as the same widths. So that they look like the design. The important forces the text field width.

Most of the !important parameters are fixing inline / hardcoded styles that come from the fields tpls and the js files that have hard coded content in them. This is especially evident in the use of tables inside SuiteCRM. We need to move towards using using divs with css classes. As that would give us much more control over the appearance of SuiteCRM.

We could fix this by creating a tpl for each field type, moving hard coded content into tpls and then overriding them. This would allow us to add css classes to each field type. However, this is a task in itself. One thing we could do is make the relate field buttons be [inline](http://getbootstrap.com/components/#input-groups-buttons-multiple). We could also make dropown as [http://getbootstrap.com/components/#input-groups-buttons-segmented](segmented fields) for the fields have elements next to the inputs.

PR #1777, opened 2016-07-21
Fixed problem with markers clusterer. 

daniprados:
The google code repository of marker clusterer library has been removed.  With this pull we change the link to a the github repo.
## Description

Changed   "//google-maps-utility-library-v3.googlecode.com/svn/trunk/markerclustererplus/images/m" to  "//raw.githubusercontent.com/googlemaps/js-marker-clusterer/gh-pages/images/m"

You can reproduce the error using maps module and having a lot o georeferenced  points.

## Types of changes

- [\* ] Bug fix (non-breaking change which fixes an issue)
### Final checklist

- [\* ] My code follows the code style of this project found [here](https://suitecrm.com/wiki/index.php/Coding_Standards).
- [\* ] I have read the [**How to Contribute**](https://suitecrm.com/wiki/index.php/Contributing_to_SuiteCRM) guidelines.

…een removed.  A better solution could be adding js-marker-clusterer to the SuiteRepository


daniprados:
signed contributor license agreement

horus68:
@daniprados  the user @comunicatek also needs to sign the clahub license as he is the committer to your hotfix branch

daniprados:
Both users have signed now.

horus68:
see your requests @shogunpol

chris001:
Why is this not merged.  Image icons are broken in Maps module without it.

An even better fix would to include the icon images in the suitcrm repo, and let the user load the icons from the suitecrm server directly, since many high security crm instances and users (bank, etc) are blocked from direct internet access, and cannot load images from the internet, only from the crm server.

daniprados:
In the last commit I added images in  SuiteCRM folder removing external dependencies.
It's a better solution and  removes one innecesary external dependency.

daniprados:
This problem is still present in current version of suitecrm.

horus68:
@Dillon-Brown maybe this is duplicated in https://github.com/salesagility/SuiteCRM/pull/3089

gody01:
Yes, this is duoplicate of: #3089
But in #3089 images are included on local server, not served from github.
I believe that used resources should be always part of software package and not served from remote.
This can be problem even in case of strict security policy and firewall configuration.

Dillon-Brown:
@gody01 It looks like initially this PR was using GitHub for the images but was then updated to be the same as #3089.

Dillon-Brown:
Issue reference: #4134

daniprados:
Now  images are added in the module.  Solving this problem is a must.  Every time I update SuiteCRM I have to apply the patch manualy when it is so easy to fix the problem.  Use the PR you prefer but use one of them.

timo12357:
This fix does not seem to work in 7.8.5. Cluster icons are still míssing even if they are locally copied and referenced.

PR #1798, opened 2016-07-27
Fixed #1309 - Reports export to csv correctly when date parameter is …

### LINKED ISSUE #1309: AOR Reports wont generate a PDF or CSV when a date parameter is selected
If you have a date condition on a report and check the parameter checkbox then you get an SQL error when you try to view the report as a PDF OR CSV.

Below is an example I encountered on an opportunities report
 `opportunities`.date_entered = DATE_ADD(opportunities., INTERVAL   )

If you uncheck the parameter checkbox the same report exports without any issues.

This issue exists in 7.5.1 and 7.6. Not tested any other versions.

MikeyJC:
Fixed #1309 - Reports export to csv correctly when date parameter is selected.
## Description

Added the same code from the report.tpl that retrieves each variable and appends them into the parameter_value array to the download_csv_button_old function in AOR_Report.js. Was previously only grabbing a single field.
## Motivation and Context

Allows the user to export a csv file when a date field parameter is in use.
## How To Test This

Create a report that uses date type condition and set it to become a parameter. Check to make sure that the report is displaying the correct results, and then select the export option. If working, it should have the same data as on the report detail view.
## Types of changes

### Final checklist

…selected

### DIFF
--- a/modules/AOR_Reports/AOR_Report.js
+++ b/modules/AOR_Reports/AOR_Report.js
@@ -38,6 +38,21 @@ $(document).ready(function() {
             var fieldType = $('#aor_conditions_value_type\\['+ln+'\\]').val();
             _form.append('<input type="hidden" name="parameter_type[]" value="'+fieldType+'">');
             var fieldInput = $('#aor_conditions_value\\['+ln+'\\]').val();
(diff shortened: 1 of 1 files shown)


Dillon-Brown:
Original issue resolved.

Issue #1812, opened 2016-07-28
AOR Reports: Conditions with NULL value Bug


MikeyJC:
Any condition that follows after a condition with a NULL or blank value is ignored.
#### Issue

In AOR Reports, when a condition has a NULL or blank value (e.g. an unchecked box), then all of the conditions that come afterwards will not be parsed correctly and do not filter the report data correctly. The condition that causes this issue still works correctly. This is only an issue when the value is empty (a checked box doesn't cause this issue).
#### Expected Behavior

All conditions are taken into account and the data represented is correct.
#### Actual Behavior

Data filtering is correct up until the condition with the NULL/blank value, at which point it fails to filter down any further.

Dillon-Brown:
Duplicate of #3375

PR #1813, opened 2016-07-28
Reports with NULL value conditions fixed

MikeyJC:
Fixed an issue in reports where conditions after a condition with it's value set to NULL would be ignored.
## Description

In AOR_Reports.php, the build_report_query_where function builds the query for the condition with an empty value in a certain way. However, originally it would not use parenthesis for this part of the query, meaning the OR operator would affect the rest of the query. The change places the brackets around the specific section of the query for the NULL value so that it is contained.  

Issue link: #1812
## Motivation and Context

Allows the user to build a report with multiple NULL value conditions and place them in any order.
## Types of changes

### Final checklist


Dillon-Brown:
Duplicate of merged PR #3421

Issue #1042, opened 2016-03-03
Inline edit - date format issue + issue of replacing the orig. date by actual date


Mausino:
Hi, If are you inline editing the date in any module, it has 2 issues:

First... it is replacing your saved date with your actual date... and it is confusing, because if i want change some date only for one day forward, it show me actual date and i must search which date was there... in editview it is working without issue...

look on video: http://screencast.com/t/7oL4DAkZq

Second: after inline editing you will see the date in default US format... after refresh you will see date in format which you set up in your settings

![date](https://cloud.githubusercontent.com/assets/5773777/13489386/fd0228c0-e126-11e5-826a-64051e239cbe.png)

Miky

shogunpol:
@Mausino , the both issues described above seems to not appear in recent version of SuiteCRM(7.7.8), can you confirm this, if yes i can close an issue.

Mausino:
@shogunpol Yes, i agree 👍 i am closing those 2 issues :)

PR #1848, opened 2016-08-02
Fixed if that was preventing date getting set properly. Fixed #1783

### LINKED ISSUE #1783: Inline editing of date fields resets date to current date
#### Issue

When using the inline edit feature for a date field, a date selection box pops up, but doesn't use the date in the date field, but the current date.

#### Expected Behavior

The date selector should be set to the value of the date field.
#### Actual Behavior

The date selector in an inline editing date field always defaults to the current date. When saving, the content of the date field will be overwritten with the current date.
#### Possible Fix

We are using German date format (dd.mm.yyyy) - could this be the reason?
#### Your Environment

* SuiteCRM Version used: 7.6.2 and newer

emcrobert:
Fixes issue #1783 with inline editing. This is also one of the 2 issues mentioned in #1042. 
## Description

Fixes the if statement in inline editing as before it was always setting date fields to the current date rather than the date that was previously set
## Motivation and Context

Stops date fields defaulting to the current date rather than the previous date set by the user when inline editing
## How To Test This

Set a date in any module then use inline editing to update it. The initial value shown should be the date that was previously set
## Types of changes

### Final checklist


Dillon-Brown:
This change is already merged into 7.8.x

PR #1933, opened 2016-08-12
Fixed #1029 - prevent user from adding or editing field if the fields name already exists. This fix also works for custom fields.

### LINKED ISSUE #1029: Adding a new field with a new dropdown list results in core field edit
An issue occurs when a user adds a new drop down field with a new drop down list for the field. For example, if a user adds a new 'employees' field with a drop down list 'employees_list' and saves this, the existing core 'employees' field is modified, instead of a new 'employees_c' field being generated. This could have further implications on existing functionality/data.

vladbar:
## Description
Fixed #1029 - prevent user from adding or editing field if the fields name already exists. This fix also works for custom fields.

## How To Test This
Go to admin -> Studio -> Add field -> change Data Type to dropdown, Radio or MultiSelect -> Edit -> if the field name already exists, then it should prevent you from editing and display an error message.

## Types of changes

### DIFF
--- a/modules/ModuleBuilder/controller.php
+++ b/modules/ModuleBuilder/controller.php
@@ -915,5 +915,14 @@ class ModuleBuilderController extends SugarController
         }
     }
 
+    function action_getModuleFields ()
+    {
+        $bean = BeanFactory::getBean('Accounts');
+        $field_defs = $bean->getFieldDefinitions();
+        $keys = array_keys($field_defs);
+
+        echo json_encode($keys);
+    }
+
 }
 ?>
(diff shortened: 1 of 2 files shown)


modules/ModuleBuilder/controller.php
```diff
@@ -915,5 +915,14 @@ function action_savetablesort ()
         }
     }
 
+    function action_getModuleFields ()
+    {
+        $bean = BeanFactory::getBean('Accounts');
```
JimMackin: Should this be hardcoded to Accounts?

PR #1955, opened 2016-08-15
Quote MSSQL column names when creating tables

OttoG:
## Description
When upgrading to 7.7.1 and using MS SQL Server (at least with version 2005), the upgrade fails in the last step. This is due to the column name 'open' in the 'aobh_businesshours' table. The identifier 'open' needs to be quoted with square brackets, since 'open' is a reserved word.

## Motivation and Context
The change fixes upgrades to current SuiteCRM versions with SQL Server. Presumably, the same problem exists during fresh installs, and if so, should likewise be fixed by this change.

Also, square-bracket-quoting identifier names in computer-generated SQL code is always good practice with MS SQL Server.

## How To Test This
In a SuiteCRM installation with MSSQL, delete the 'aobh_businesshours' table if it exists. Then upgrade to version 7.7.1 with the upgrade wizard, or if the system is already current, do a quick repair and rebuild. The upgrade/repair should work with this fix.

## Types of changes

### Final checklist

Upgrading to SuiteCRM 7.7.1 fails in the last step with SQL Server, due to the unquoted aobh_businesshours column name 'open', which is a reserved word. Quoting all column names fixes the problem.

### DIFF
--- a/include/database/MssqlManager.php
+++ b/include/database/MssqlManager.php
@@ -1765,6 +1765,9 @@ EOQ;
 		// always return as array for post-processing
 		$ref = parent::oneColumnSQLRep($fieldDef, $ignoreRequired, $table, true);
 
+		// Quote the column name (fixes problems with names like 'open', as found in aobh_businesshours)
+		$ref['name'] = $this->quoteIdentifier($ref['name']);
+
 		// Bug 24307 - Don't add precision for float fields.
 		if ( stristr($ref['colType'],'float') )
 			$ref['colType'] = preg_replace('/(,\d+)/','',$ref['colType']);


samus-aran:
Hi @OttoG I tried to test this (with 2008) and it is not proceeding through the upgradeWizard with the fix provided.  What logs do you get to confirm that is the issue?

samus-aran:
Hi @OttoG That is it merged but please refer to the 'How to Contribute' Guide as it will show you how to only include the correct commits and have the Pull Request point to the right branch.

https://suitecrm.com/wiki/index.php/Contributing_to_SuiteCRM#Make_a_Bug_Fix_Pull_Request

PR #2003, opened 2016-08-22
Fixed #1700 - convert parametarised dates correctly if the user date format changes. saveOptions() function from AORReportsDashlet.php needs to convert dates to the db format in order for comparison to work.

### LINKED ISSUE #1700: AOR Report Dashlet date parameters wont work
#### Expected Behavior

If you put a report in a dashlet with date parameter you should be able to change them.
#### Actual Behavior

If you change the parameter in the dashlet they wont be taken. Sometimes the result is empty after changing, sometimes it displays all rows without filter.
#### Possible Fix

Dont know.
#### Steps to Reproduce

Make a report.
Add one or more date field Parameters.
Add this Report as Dashlet to homepage.
Change date parameter value in dashlet.
#### Context

Reporting is totally broken not only in dashlets.
#### Your Environment
- SuiteCRM Version used: 7.6.4, 7.7beta2
- Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): FF 47
- Environment name and version (e.g. MySQL, PHP 7): PHP 5.6
- Operating System and version (e.g Ubuntu 16.04): Ubuntu 15.10

All in all it is not good that suitecrm concentrates on a new theme or a new reporting tool, or a new issue template, and leaves hundreds of bugs lying around for month. The template makes more work for bug reporters, and then the bugs wont get fixed, and then the frustration is higher because of the more work for open a new issue.

vladbar:
Convert parametarised dates correctly if the user date format changes. saveOptions() function from AORReportsDashlet.php needs to convert dates to the db format in order for comparison to work.

relates to #1700

### DIFF
--- a/modules/AOR_Reports/Dashlets/AORReportsDashlet/AORReportsDashlet.php
+++ b/modules/AOR_Reports/Dashlets/AORReportsDashlet/AORReportsDashlet.php
@@ -134,6 +134,21 @@ class AORReportsDashlet extends Dashlet {
 //            $req['parameter_value'][1] = $firstValue;
 //        }
         $allowedKeys = array_flip(array('aor_report_id','dashletTitle','charts','onlyCharts','parameter_id','parameter_value','parameter_type','parameter_operator'));
+
+        // Fix for issue #1700 - save value as db type
+        for($i = 0; $i < count($req['parameter_value']); $i++) {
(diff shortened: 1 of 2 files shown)


samus-aran:
Good Work @vladbar

PR #2037, opened 2016-08-25
Fixed #1138 AOW_WorkFlow - execution on 'New Records' fails when user…

### LINKED ISSUE #1138: AOW_WorkFlow - execution on 'New Records' fails when user's dateformat is not strtotime compatible
A workflow defined to run on: 'On_Save' + 'New_Records' is not executed when user has a date format defined in profile as 'd/m/Y' (and most probably others).
The method `check_valid_bean` in modules/AOW_WorkFlow/AOW_WorkFlow.php has:

```
$beanDateEnteredTimestamp = strtotime($timedate->asUser(new DateTime($timedate->fromDb($bean->date_entered))));
$beanDateModifiedTimestamp = strtotime($timedate->asUser(new DateTime($timedate->fromDb($bean->date_modified))));
$thisDateEnteredTimestamp = strtotime($this->date_entered);
```

and `strtotime` will return (FALSE|-1) if the output is not a correctly formatted string:

[PHP.NET](http://php.net/manual/en/function.strtotime.php)

> 'strtotime — Parse about any English textual datetime description into a Unix timestamp ... Returns a timestamp on success, FALSE otherwise. Previous to PHP 5.1.0, this function would return -1 on failure....Dates in the m/d/y or d-m-y formats are disambiguated by looking at the separator between the various components: if the separator is a slash (/), then the American m/d/y is assumed; whereas if the separator is a dash (-) or a dot (.), then the European d-m-y format is assumed. If, however, the year is given in a two digit format and the separator is a dash (-, the date string is parsed as y-m-d.'

and because `$timedate->asUser` WILL also return dates formatted differently.

The use of strtotime here is inappropriate and in some cases will lead to the skipping of the workflow
In some cases because if the date is 
- '3/3/2016' then the date conversion to unixtimestamp will work 
- '3/4/2016' - maybe works but date is converted erronously to the equivalent of 4th of march if 'd/m/Y' is used because as stated before when slash is used in string the American format is assumed(m/d/y)
- '13/3/2016' - will NOT WORK and workflow will be skipped because of the above reason will identify month===13 which clearly does not exists

...PR coming on hotfixes.. ;)...

shogunpol:
Fix #1138 AOW_WorkFlow - execution on 'New Records' fails when user's dateformat is not strtotime compatible
## Description

Using the proposition of @adamjakab ( https://github.com/salesagility/SuiteCRM/pull/1146/files ),
Thanks @adamjakab for contribution.

- Create WorkFlow with module:accounts, Run-on: New Records
   condytions: Module:Accounts, Field:Date Created, Operator: greater or equal then, Type: value,   value (some date)
- Create Account, watch is date fail
- Repeat for different time formats

References issue #1138

## How To Test This

Repeat steps from replicate an issue
## Types of changes

### Final checklist

…'s dateformat is not strtotime compatible

### DIFF
--- a/modules/AOW_WorkFlow/AOW_WorkFlow.php
+++ b/modules/AOW_WorkFlow/AOW_WorkFlow.php
@@ -457,10 +457,11 @@ class AOW_WorkFlow extends Basic {
 
         if($this->flow_run_on){
 
-            // database time correction with the user's time-zoneqq
-            $beanDateEnteredTimestamp = strtotime($timedate->asUser(new DateTime($timedate->fromDb($bean->date_entered))));
-            $beanDateModifiedTimestamp = strtotime($timedate->asUser(new DateTime($timedate->fromDb($bean->date_modified))));
-            $thisDateEnteredTimestamp = strtotime($this->date_entered);
+            // database time correction with user's time-zone (using user's date format)
+            $userDateFormat = $timedate->get_date_time_format();
+            $beanDateEnteredTimestamp = date_create_from_format($userDateFormat,$timedate->asUser(new DateTime($timedate->fromDb($bean->date_entered))))->getTimestamp();
+            $beanDateModifiedTimestamp = date_create_from_format($userDateFormat,$timedate->asUser(new DateTime($timedate->fromDb($bean->date_modified))))->getTimestamp();
+            $thisDateEnteredTimestamp = date_create_from_format($userDateFormat,$this->date_entered)->getTimestamp();
 
             switch($this->flow_run_on){
                 case'New_Records':
--- a/modules/AOW_WorkFlow/AOW_WorkFlow.php
+++ b/modules/AOW_WorkFlow/AOW_WorkFlow.php
@@ -459,9 +459,20 @@ class AOW_WorkFlow extends Basic {
 
             // database time correction with user's time-zone (using user's date format)
             $userDateFormat = $timedate->get_date_time_format();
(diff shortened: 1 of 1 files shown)


modules/AOW_WorkFlow/AOW_WorkFlow.php
```diff
@@ -457,10 +457,11 @@ function check_valid_bean(SugarBean &$bean){
 
         if($this->flow_run_on){
 
-            // database time correction with the user's time-zoneqq
-            $beanDateEnteredTimestamp = strtotime($timedate->asUser(new DateTime($timedate->fromDb($bean->date_entered))));
-            $beanDateModifiedTimestamp = strtotime($timedate->asUser(new DateTime($timedate->fromDb($bean->date_modified))));
-            $thisDateEnteredTimestamp = strtotime($this->date_entered);
+            // database time correction with user's time-zone (using user's date format)
+            $userDateFormat = $timedate->get_date_time_format();
+            $beanDateEnteredTimestamp = date_create_from_format($userDateFormat,$timedate->asUser(new DateTime($timedate->fromDb($bean->date_entered))))->getTimestamp();
```
mattlorimer: getTimestamp will cause a fatal if the datetime is null see the unit test results

mattlorimer:
already fixed here https://github.com/salesagility/SuiteCRM/commit/0f2f884f88cd488b5d763de79cb10ef13457820c

PR #2152, opened 2016-09-07
Refactoring: removed ineffective code

masadwin:
$login_image is being overwritten after an if-else condition which
renders the if-else ineffective


mattlorimer:
please review the https://github.com/salesagility/SuiteCRM/wiki about making a pull request

Issue #2082, opened 2016-08-30
Notifications not working


bcowman:
I have SuiteCRM version 7.7.3 but cannot get Notifications to work.  Also the Desktop Notification Count never changes.

I have read all the community posts on this subject and tried all suggestions:
          I have tried IE11 and the latest versions of Chrome and Firefox.
          I have enabled Desktop Notifications for each browser.
         Cron jobs as running.
         The time zone in php.ini is correct.

My environment is IIS8, mySQL 5.626, PHP 5.6

The fact that the Desktop Notification Count never changes leads me to think this is a bug.

samus-aran:
Hi @bcowman Are you meaning perhaps the Alerts/Reminders (for when meetings and calls are coming up)? Probably just our terminology, but I believe you mean the alert counter up at the top right of the instance?

bcowman:
Yes, Alerts/Reminders for calls and meetings and the Alert Counter at the top right.

KhyberPass:
Hi @bcowman, alerts are working for me with 7.7.4 - but are not very reliable as there are a few things that can break them. You need to ensure the following
- Notifications are enabled under your Profile -> Advanced. When you click on the Enable button you should get a notification
- The meeting or call has the popup checked, a time set and a person invited - like this
![image](https://cloud.githubusercontent.com/assets/5484293/18434884/42bbf752-7933-11e6-8ed4-66fcf9c8c4b5.png)

- The browser tab is active and not inactive ie. you have Suite open on the browser 

Also, you can use the following to see the alert time counting down. Paste into a Chrome console window.
console.log("Alerts:", alertList.length, "-", alertList[0].time, "Time:", secondsSinceLoad);
This assumes that you only have 1 alert pending. If the length shown is more than 1 then you may need to adjust the [0] to [1] or more.
The alert time is fixed and the secondsSinceLoad counts up in seconds. When secondsSinceLoad is greater than the alert time the notification will trigger. When a page is loaded the times are reset to new values.
Nick

bcowman:
Hi Nick,

Thanks for taking the time to reply to my query...

I have upgraded to 7.7.4.

- Notifications are enabled
- The Call has popup checked, a time set and person invited.  If I also include the Email Invitees then I WILL receive the email.
- Browser tab is active

I have created only one alert.  If I post the code you supplied into Chrome Console I get the following error:

[...]

Issue #766, opened 2015-12-09
Timezone issue for Meeting reminders


vjaro:
Steps to reproduce:
1. Change user timezone settings to GMT+8
2. Create a meeting
3. Add 4 reminders with 15,10,5,1 minutes prior popup reminders.

Expected result 
the desktop popup notifications will appear.
Actual result
Waited for all the reminders to popup but no reminders were displayed.

Now when i update my user timezone settings to UTC and did the previous steps, the reminders started displaying correctly as per defined on the popup reminder time.

KhyberPass:
As of 7.4.3 reminders for Calls and Meetings do not work correctly unless the user’s time zone is set to UTC. If the user’s time zone is set to say Europe (UTC+1) the reminder will trigger 1 hour later.

Some details, as part of the Reminder enhancements for 7.4.3 it looks like code was copied from include/javascript/jsAlerts.php to modules/Reminders/Reminder.php, however the code in jsAlerts queried date/time directly from the database where time is UTC, whereas Reminder was changed to use the Bean which returns time in the user’s time zone.

This problem is also present in 7.5.1. A patch fix will be posted soon.

armadillotec:
In the 7.5.1 doesn't take timezone from php.ini

PR #2167, opened 2016-09-08
Fixed alerts not counting down

daniel-samson:
## Description

These fixes make the alerts count down so that the user gets the notification reminder.

relates to issue #2082 
## Motivation and Context

When a user set up a reminder on calls or meetings they do not receive a desktop notification.
## How To Test This

Create a popup reminder in calls and meetings.
Wait for the notification.

If you want to check the counting down you can run the following in the browser console:

> setInterval(function() { console.log(alertList[0].time) }, 1000);
## Types of changes

### Final checklist


KhyberPass:
Hi @daniel-samson, I was having some issues with alerts and came across your proposed fix.
I did some work on alerts some time ago when they were not working for me in #766 so have looked into this before.
In general alerts are working for me in 7.7.4 but have some issues.
Looking at your proposed fix to the current issue I am not too sure if it is fixing anything, but could be missing something. Instead of using the setTimeout() to count up a single variable and compare against the time in each alert the proposal counts down each alert.
The change does make it easier to see what is going on and debug. Does it fix any issues?

The biggest issue with the current way the alerts work is setTimeout(), in Chrome when the tab is inactive then setTimeout() does not trigger every second, this causes the alert to popup at the wrong time or not at all. If you use the debug below, switch to another tab for a bit and come back you will see that the count up (secondsSinceLoad) is not correct.
console.log("Alerts:", alertList.length, "-", alertList[0].time, "Time:", secondsSinceLoad);
I am looking into a better way of doing this to make it more reliable.
Not too sure if you are doing the same too?
Nick

daniel-samson:
@KhyberPass Sorry for the delay. The this fix is to try and force the time left to count down. So that if it passes the threshold it displays the notification. Perhaps we should change the setTimeout to setInterval instead.

KhyberPass:
Hi,
Yes, I can see where you are going with this but not too sure if it will change anything.

I have looked at alerts as they do not always work for me. One of the main problems is to do with setTimeout() that calls checkAlerts() every second.

setTimeout() or setInterval () is not very reliable. On Chrome, for example, if the tab is not selected setTimeout() does not run with the same priority and therefore the countdown loses time and alerts show late or not at all. There are a number of ways to fix this that I have read about, the best one would be to have the alert time as epoch sent from the server rather than a countdown and check this on the browser side with a calibrated current time.

Also there are some improvements in Reminder.php that can be made. Currently a check is done on the event time not taking into account the reminder time i.e. an event happens at 10:30 with a 15 min reminder. The check is done on 10:30, this means that between 10:15 and 10:30 a negative countdown time is sent to the browser javascript code.

There is also a very small possibility that alerts can be missed with the current implementation. IF an alert in the browser if just about to trigger and a page is loaded by the time the server side query is done the reminder time could have just ticked over and the alert would not be generated.

I am currently doing a fix to Reminders for #2246. To do this I have written some unit tests for Reminders. When this is done I was going to look at improving Alerts.

Nick

PR #2220, opened 2016-09-14
Fixed #1624 - AOR Reports: Date parameter

### LINKED ISSUE #1624: AOR Reports: Date parameter only accepts format 2016-06-20
If the user has a different date format like "20.06.2016" and tries to change a date parameter at the detailview or in a dashlet the report wont work with the new parameters.

If you save a report for the first time, the date format is no issue, so editview seems to handle the date format.
But if you try to edit the parameter "on-the-fly" and click update, or in a dashlet, the date is invalid.

vladbar:
Additional fixes for issues raised in #1624 but this time for reports run in mssql environment


shogunpol:
This issue it appears to be fixed in previous commits.

PR #2313, opened 2016-09-26
Fixed #2153 - added image generating link to the template parser.

### LINKED ISSUE #2153: Images in PDFs not showing
#### Issue

PDF images showing up in templates are not created when PDF is generated.
#### Expected Behavior

Create PDF template with a local image (located in custom/themes/default/images/).
Create PDF through invoice/quotes etc.
PDF created with images embedded.
#### Actual Behavior

PDF created with image placeholder (the square with the cross).
#### Other Testing Steps

I also tried linking to external images, hosted on a different platform - same error.
Checked all permissions and all are fine (and not changed since update).

Tested in Demo and that seems to work.
#### Context

This is directly effecting our ability to invoice our clients.
#### Your Environment
- SuiteCRM Version used 7.7.4 (was working in 6.7.x).
- All browsers tested (Chrome, Firefox, Internet Explorer, Edge).
- Environment name and version: MySQL, PHP 7.
- Windows 10 & 7 pro tested.

vladbar:
## Description

references issue #2153
Added a conditional statement which checks whether field is of type image. 
If the field is of type image then a link pointing to the location of the image is generated.
## Motivation and Context

PDF images showing up in templates are not created when PDF is generated.
## How To Test This
1. Create a new field on invoice of type image
2. Add the new field to the edit screen of invoice
3. Create an invoice record and upload a jpg image to the new field.
4. Create a pdf template on invoice and add the image field to it.
5. Select Print as PDF on the created invoice record.
6. Select the PDF template that you added the image field to.
7. Examine the generated PDF.
## Types of changes

### DIFF
--- a/modules/AOS_PDF_Templates/templateParser.php
+++ b/modules/AOS_PDF_Templates/templateParser.php
@@ -52,7 +52,7 @@ class templateParser
 
     function parse_template_bean($string, $key, &$focus)
     {
-        global $app_strings;
+        global $app_strings, $sugar_config;
         $repl_arr = array();
 
         foreach ($focus->field_defs as $field_def) {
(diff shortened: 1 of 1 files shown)


cuffy3141:
Just one suggestion, it would be nice if the width and height could be picked up from the field definition.

PR #2409, opened 2016-10-12
Fixed #1561 - Create and save notes which have related sent email ids…

### LINKED ISSUE #1561: Email Attachments Sent By Workflow Do Not Show In Related Activity
When sending emails manually, if there's an attachment(s) on the email template, the related activity created in List View will show a paper clip to indicate there's an attachment, and in Detail View, there are links below the email to the attachments.

This does not happen when email templates with attachments are sent by the workflow.  It does appear to send the attachments, but there is NO paperclip on the related activity in List View and NO links indicating any attachments in Detail View.

Version is 7.5.2

vladbar:
## Description

references issue #1561 
Every time an email is sent by the Workflow module if there are some attachments, new Notes need to be created with unique ids and sent email id as a parent id.
Also a new file needs to be created based on the original attachment file. This file must have the newly generated Note id as its name, as the modules/Emails/DetailView.php generates a link which points to this file.
## Motivation and Context

Email Attachments Sent By Workflow Do Not Show In Related Activity
## How To Test This
1. Create an email template which has an attachment.
2. Create a workflow which sends emails to Contacts and uses the above mentioned template.
3. Activate the Scheduler action which processes workflows.
4. Expected results on the Contact Detail View: the related activity created in the History sub-panel List View will show a paper clip to indicate there's an attachment, and in Detail View, there are links below the email to the attachments.
5. Actual results on the Contact Detail View: no paper clip and no attachment link inside the related Email record Detail View.
   ## Types of changes
6. [x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/modules/AOW_Actions/actions/actionSendEmail.php
+++ b/modules/AOW_Actions/actions/actionSendEmail.php
@@ -429,6 +429,27 @@ class actionSendEmail extends actionBase {
             $emailObj->status = 'sent';
             $emailObj->save();
 
+            // Fix for issue 1561 - Email Attachments Sent By Workflow Do Not Show In Related Activity.
+            foreach($attachments as $attachment) {
+                $note = new Note();
+                $note->id = create_guid();
+                $note->date_entered = $attachment->date_entered;
+                $note->date_modified = $attachment->date_modified;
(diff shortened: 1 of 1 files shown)


samus-aran:
Works as expected.

PR #2539, opened 2016-11-01
Fix #2246 - Notifications / Alerts do not trigger depending on some user date formats

### LINKED ISSUE #2246: Notifications / Alerts do not trigger depending on some user date formats
#### Issue

The user profile can be used to configure the displayed date format.
When a user has set the date format to dd/mm/yyyy and the reminder module assumes mm/dd/yyyy it will not add an alert. Other date formats may also not work.

Setting the date format to yyyy-mm-dd or mm/dd/yyyy works ok.

Also see #2082
#### Expected Behavior

Alert should be triggered according to the reminder date time.
#### Actual Behavior

No alert is triggered.
Using the console enter the following
`console.log("Alerts:", alertList.length, "Time:", secondsSinceLoad);`
The output will show that the alert length is 0.
#### Possible Fix

See below
#### Steps to Reproduce

1. Using a users profile set the date format to dd/mm/yyyy
2. Create a new Call and set the start date to 30 minutes away, set the popup reminder to 15 minutes
3. Use the console to display the alerts as detailed above, there will be no alerts
4. A notification will not happen.
5. Change the user date format to yyyy-mm-dd and repeat the above
6. The notification will happen
#### Context

Reminders do not occur.
#### Your Environment

- Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Chrome Version 52.0.2743.116 m 64bit
- Environment name and version (e.g. MySQL, PHP 7): MySQL 5.5.48-37.8, PHP 5.4.45
- Operating System and version (e.g Ubuntu 16.04): Hostgator CentOs 6.8

KhyberPass:
When a user has set the date format to dd/mm/yyyy reminders will not trigger. Setting the date format to yyyy-mm-dd or mm/dd/yyyy works ok.

## Description

Fix for #2246 
The reminders module looks at the list of reminders, when the date and time of the reminder gets close it will trigger an alert count down to start.
Problem was that strtotime() was converting the textual start datetime read from the database into  timestamp. If the server is using mm/dd/yyyy and the input string is dd/mm/yyyy errors will occur. E.g. for the 5th of Jan 2016 the timestamp will be for the 1st of May 2016 and for the 21st Jan 2016 a null timestamp will be returned as 21 is an invalid month. In the code strtotime($relatedEvent->date_start)

$relatedEvent->date_start is from the bean and is in the User timezone and date time format. In this case if the user date format is set to dd/mm/yyyy the calculated timestamp will be incorrect.

Note that email reminders work, the code for this is 
`$remind_ts = $timedate->fromUser($eventBean->date_start)->modify("-{$reminderBean->timer_email} seconds")->ts;`

In this case the bean date time string is converted to a DateTime object using the fromUser() function that takes into account the user settings and therefore operates correctly. This method should be used.
The following testing was performed with the user timezone set to Sydney Australia +11, and the server timezone as USA CET.

## Motivation and Context

Reminders do not work with some user date formats

## How To Test This

Set user date format to d/m/y and create a call with a reminder. Ensure a notification triggers.

I have tested using the following user date formats

- Y-m-d OK
- m-d-y OK 
- d-m-y OK
- y/m/d OK
- m/d/y OK
- d/m/y OK
- y.m.d OK
- m.d.y OK
- d.m.y OK 

And with these date time formats
- d/m/y  h:ia
- d/m/y  h:iA
- d/m/y  h:i a

Also unit tests were written for the Reminder module. Currently no unit test exists for this module.  Technically it is probably more functional testing rather than unit testing. This work is in my GitHub at https://github.com/KhyberPass/SuiteCRM in branch [IssueReminderTest](https://github.com/KhyberPass/SuiteCRM/tree/IssueReminderTest/tests/tests/modules/Reminders)
I need to do a bit more testing on it before submitting this.

## Types of changes

### Final checklist

### DIFF
--- a/modules/Reminders/Reminder.php
+++ b/modules/Reminders/Reminder.php
@@ -56,8 +56,6 @@ class Reminder extends Basic
     var $importable = false;
     var $disable_row_level_security = true;
 
-    var $assigned_user_id;
-
     var $popup;
     var $email;
     var $email_sent = false;
@@ -294,21 +292,9 @@ class Reminder extends Basic
             foreach ($popupReminders as $popupReminder) {
                 $relatedEvent = BeanFactory::getBean($popupReminder->related_event_module, $popupReminder->related_event_module_id);
 
-if ($relatedEvent) 
-{
-//	print("\nDate1 " . $relatedEvent->date_start);
-//	print("\nDate2 " . $timedate->fromUser($relatedEvent->date_start));
-//	print("\nNow " . $dateTimeNow);
-//	print("\nDate " . strtotime($relatedEvent->date_start));
-//	print("\nNow  " . strtotime(self::unQuoteTime($dateTimeNow)));
-//	print("\nEnd  " . strtotime(self::unQuoteTime($dateTimeMax)));
-//    print("\nDelay " . $app_list_strings['reminder_max_time']);
-//    print("\nNow Db " . $timedate->getNow(true)->asDb(true));
-}
-
                 if ($relatedEvent && isset($relatedEvent->date_start))
                 {
-                    // Start date from the bean is in user timezone and format
+                    // Start date from the bean is a string in user timezone and format
                     // convert using user format to a GMT timestamp
                     $dateTimeStart = $timedate->fromUser($relatedEvent->date_start)->ts;
                 }
@@ -317,7 +303,6 @@ if ($relatedEvent)
                     // The start date is not set so default it to the current time
                     $dateTimeStart = $dateTimeNow;
                 }        
-//print("\nTimes " . $dateTimeStart . " " . $dateTimeNow . " " . $dateTimeMax);
 
                 if ($relatedEvent &&
                     (!isset($relatedEvent->status) || $relatedEvent->status == 'Planned') &&
--- a/modules/Reminders/Reminder.php
+++ b/modules/Reminders/Reminder.php
@@ -56,6 +56,8 @@ class Reminder extends Basic
     var $importable = false;
     var $disable_row_level_security = true;
 
+    var $assigned_user_id;
+
     var $popup;
     var $email;
     var $email_sent = false;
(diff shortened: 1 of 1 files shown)


KhyberPass:
Hi @shogunpol Just wondering what is happening to this fix?
There have been a number of fix releases since this was ready to merge on the 1 Nov 2016 but this was never included. Is there something wrong with it that is stopping it?
I have a number of other fixes and enhancements that I can do but wont bother contributing if they are never going to be included.

shogunpol:
@KhyberPass , no its nothing wrong with your solution, is just need to pass last stage of code verification, and after this can be merged. Sometimes is just amount of other PRs, time and limited resources disallows us to put some of PRs.

modules/Reminders/Reminder.php
```diff
@@ -297,17 +291,30 @@ public static function addNotifications(jsAlerts $alert, $checkDecline = true)
         if ($popupReminders) {
             foreach ($popupReminders as $popupReminder) {
                 $relatedEvent = BeanFactory::getBean($popupReminder->related_event_module, $popupReminder->related_event_module_id);
+
+                if ($relatedEvent && isset($relatedEvent->date_start))
```
kdjmonaghan: Please keep your code to PSR-2 standard: http://www.php-fig.org/psr/psr-2/

If statements are a bit off as well as some other sections of the new code you've written, otherwise the fix works as expected. :+1:

samus-aran:
Whther @KhyberPass is able to adjust - if not, @kdjmonaghan is your Commit superseeds the above?

kdjmonaghan:
@samus-aran - Essentially, the same solution within my commit, just adjusted to the standards. If @KhyberPass doesn't update his commit, this commit superseeds his. :) :+1:

wardluke:
Have we any idea when this fix can be merged please?

Dillon-Brown:
@wardluke I'll take a look at fixing the conflicts so we can get this code-reviewed. If you want to speed up the process it would be helpful to leave a comment confirming that this fix fixes the original issue for you. Thanks!

wardluke:
@Dillon-Brown I'll went to test the fix in a Bitnami VM and noticed the problem did not exist anymore, in version 7.10.5.

My problem version is 7.10.4.

Has the conflicting fix/change been merged?  Could you reference it here please?

PR #2562, opened 2016-11-03
Fixed #1366 - Using parse_template_with_bean instead of parse_template_bean

### LINKED ISSUE #1366: Using "E-Mail" from the action menu does not include variables from the templates
After implementing the fix from pull request #1168, the emails themselves will send, but the don't pull the variables mentioned in the email templates  (e.g. Instead of "John", the email is addressed to "prospect_first_name").

daniel-samson:
## Description

In relation to #1366, this change fixes when you try to email a target (prospect) from the list view. I have created a new method which is not dependant on the a specific module. It should still work even when say accounts/leads are missing. It will also replace $contact_ fields when the contact module is not present. Which is a requirement for email campaigns

## Motivation and Context

Resolves the issue.

## How To Test This

- Create a email template with fields from prospects, and contacts/leads/prospects.
- Try emailing the template from the propects list view.

Also test email campaigns.

## Types of changes

### Final checklist

Ensures that leads / prospect fields are correctly parsed.

### DIFF
--- a/modules/EmailTemplates/EmailTemplate.php
+++ b/modules/EmailTemplates/EmailTemplate.php
@@ -737,6 +737,67 @@ class EmailTemplate extends SugarBean
         return $string;
     }
 
+    public static function parse_template_with_bean($template, $bean) {
+        // Get the properties of the bean.
+        $properties = get_object_vars($bean);
+
+        // When the $bean->module_name is in the following array,
+        // replace the fields beginning with '$contact_" with the values from $bean.
+        //
+        // This ensures that the user can select a single template for campaign emails.
+        $replace_contact_fields_with_module_fields = array(
+            'Leads',
+            'Prospects'
+        );
+
+        foreach($properties as $property_name => $property_value) {
+            // When the property of the $bean is a drop down value
+            // translate the value using options specified in the $bean->field_defs.
+            if( !empty($property_value)
+                AND isset($bean->field_defs[$property_name])
+                AND !empty($bean->field_defs[$property_name])
+                AND strtolower($bean->field_defs[$property_name]['type']) == "enum" ) {
+
+                $translated = translate($bean->field_defs[$property_name]['options'], $bean->module_name, $property_value);
+
+                if(is_array($translated) AND isset($translated[$property_value])) {
+                    $property_value = $translated[$property_value];
+                } else {
+                    $property_value = $translated;
+                }
+            }
+
+            // Replace the template fields with the values of the $bean fields.
+            switch(gettype($property_value)) {
+                case "object":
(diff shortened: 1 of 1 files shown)


LuminisIndia:
Though I am not sure because I am not a coder/expert about this, I guess the above solution is working, but not 100%.  

As we have followed this and solution worked absolutely fine with normal email sending via Activities Menu.  But when I tried to use the same template for sending emails via workflow, it failed to convert the datetime fields correctly.  Please refer to discussion at URL: 

https://www.suitecrm.com/forum/advanced-opensales-workflow-portal-support/11983-emails-sent-via-workflow-showing-only-date-but-not-time-of-meeting

With thanks,
RK 
(Luminisindia)

LuminisIndia:
Though I am not sure because I am not a coder/expert about this, I guess the above solution is working, but not 100%.  

As we have followed this and solution worked absolutely fine with normal email sending via Activities Menu.  But when I tried to use the same template for sending emails via workflow, it failed to convert the datetime fields correctly.  Please refer to discussion at URL: 

https://www.suitecrm.com/forum/advanced-opensales-workflow-portal-support/11983-emails-sent-via-workflow-showing-only-date-but-not-time-of-meeting

With thanks,
RK 
(Luminisindia)

LuminisIndia:
After placing the previous comment, I realized that the above solution has nothing to do with the template parsing when sending email via AOW.  AOW uses very different code for doing so.

In fact, we are so encouraged by the accuracy of results via above mentioned Fixed #2562 , that we wished to have the same approach of template parsing for AOW too.

I also have raised an issue regarding the same, #2659 that they should replicate the method suggested by 	[daniel-samson](https://github.com/daniel-samson) for AOW email template parsing too.  Unfortunately, they have simply marked it as a suggestion.  I don't know now, how many more months it would take them to really consider this suggestion.

With thanks,

RK
(Luminisindia)

EZS-JD:
I did some quick tests of this today. It seems to work very well. Only noticed one little shortcoming:

**multi-enum fields** are parsed exactly as written in the database field, i.e. e.g. 
`^item1^, ^item2^, ^item3^`

(enum fields work fine!)

It would be much nicer to have the language file entries shown here instead!
I believe global variable $app_list_strings needs to be used in order to read out the language file entries for the respective multi-enum items. But not sure.

chris001:
What would brilliant is to enhance this PR to save the organization the embarrassment of sending out unparsed garbage.  Search all outgoing email with a regular expression, for unreplaced variables: any word starting in `$` which matches a field variable from any module.  When found, notify the admin with an urgent alery, and defer sending that email.

PR #2604, opened 2016-11-10
scrm-39 -  Replace X templates on SubPanels with Smarty

daniel-samson:
## Description

I order to make the sub panels fit the SuiteP design. So that the column names are at the top. I have change the sub panels so that they now support smarty templates. 

This means you can also override the **include/SubPanel/tpls/SubPanelDynamic.tpl** file in the themes. Which could be quite useful. I have create a new view called **include/ListView/ListViewSubPanel.php** which is extends the standard ListView.

 I haven't completely gotten rid of the x templates. I have left some of the custom views which currently are not being used anyway. I have also left some hard coded stuff like the pagination in the ListViewSubpanel. As some of the processing is handed here. Perhaps in the future we might move some of this code out into a template.

## Motivation and Context

- Makes the subpanels fit the SuiteP design.
- Keeps the Suite7 and SuiteR the same.

## How To Test This

- Check each sub panel in SuiteCRM.
- Ensure that each function works like it did before.

## Types of changes

### Final checklist


include/ListView/ListViewSubPanel.php
```diff
@@ -0,0 +1,890 @@
+<?php
+    if (!defined('sugarEntry') || !sugarEntry) {
```
samus-aran: Move to below the Licence

include/ListView/ListViewSubPanel.php
```diff
@@ -0,0 +1,890 @@
...
+        die('Not A Valid Entry Point');
+    }
+    /*********************************************************************************
+     * SugarCRM Community Edition is a customer relationship management program developed by
```
samus-aran: Update License

samus-aran:
Tested on Chrome/Firefox

- When reducing the window size the subpanel fields are not be subsequently hidden similar to the ListView.  When window is reduced the fields in the subpanel should be hidden depending on how many can fix in the screen size.

daniel-samson:
Closing for staging. Please refer to https://github.com/salesagility/SuiteCRM/pull/2657

PR #2646, opened 2016-11-15
Fixed #2644 created an entry point to allow callback.php authorisation, added correct path to config.php

### LINKED ISSUE #2644: Twitter authentication process error.
#### Issue
The following error message is received when trying to log in to Twitter:
Error: cannot access include/social/twitter/twitter_auth/callback.php

#### Steps to Reproduce
- Setup Twitter App
  1. https://apps.twitter.com/app/

  2. Settings tab
  Name * - any 
  Description * - any
  Website * - http://example.com/twitteroauth
  Callback URL - http://example.com/twitteroauth/callback.php

  3. Keys and Access Tokens tab 
  Application Settings -> need to "Reganerate Consumer Key and Secret" !!! do it every time smth is changed
  Your Access Token -> might need to "Regenerate My Access Token and Token Secret"

  4. Permissions tab
  leave on "Read only"

- Setup Connector in your SuiteCRM instance
  1. Admin -> Connectors -> Set Connector Properties -> Twitter tab

  2. Consumer Key:  * -> enter your "Consumer Key (API Key)" from Twitter

  3. Consumer Secret:  * -> enter your "Consumer Secret (API Secret)" from Twitter

- if Home page is already opened Ctrl F5 on the Home page - so the new token will be generated for the "Log into Twitter" button

- click on the "Log into Twitter" button

#### Expected Behavior
- you will be redirected back to the home page

#### Actual Behavior
The following error message is received
Error: cannot access include/social/twitter/twitter_auth/callback.php

vladbar:
references issue #2644

## Description
The following error message is received when trying to log in to Twitter:
Error: cannot access include/social/twitter/twitter_auth/callback.php

The problem is that callback.php is not an entry point so the .htaccess prevents twitter from accessing this file.

This issue can be fixed by creating an entry point to allow callback.php authorisation, also $url variable in twitter_helper.php needs to be changed in order to reflect that we call an entry point.

Another fix is to call config.php by specifying the correct path - include_once('../../../../../config.php');

## How To Test This
- Setup Twitter App
  1. https://apps.twitter.com/app/

  2. Settings tab
  Name * - any 
  Description * - any
  Website * - http://example.com/twitteroauth
  Callback URL - http://example.com/twitteroauth/callback.php

  3. Keys and Access Tokens tab 
  Application Settings -> need to "Reganerate Consumer Key and Secret" !!! do it every time smth is changed
  Your Access Token -> might need to "Regenerate My Access Token and Token Secret"

  4. Permissions tab
  leave on "Read only"

- Setup Connector in your SuiteCRM instance
  1. Admin -> Connectors -> Set Connector Properties -> Twitter tab

  2. Consumer Key:  * -> enter your "Consumer Key (API Key)" from Twitter

  3. Consumer Secret:  * -> enter your "Consumer Secret (API Secret)" from Twitter

- if Home page is already opened Ctrl F5 on the Home page - so the new token will be generated for the "Log into Twitter" button

- click on the "Log into Twitter" button

## Types of changes

### DIFF
--- a/include/MVC/Controller/entry_point_registry.php
+++ b/include/MVC/Controller/entry_point_registry.php
@@ -86,6 +86,7 @@ $entry_point_registry = array(
 	'campaignTrackerSave' => array('file' => 'modules/CampaignTrackers/Save.php', 'auth' => true),
 	'emailMarketingList' => array('file' => 'modules/EmailMarketing/List.php', 'auth' => true),
 	'setCampaignMarketingAndTemplate' => array('file' => 'modules/Campaigns/WizardCampaignSave.php', 'auth' => true),
+	'twitter_callback' => array('file' => 'include/social/twitter/twitter_auth/callback.php', 'auth' => true),
 );
 
 ?>
--- a/include/social/twitter/twitter_auth/callback.php
+++ b/include/social/twitter/twitter_auth/callback.php
@@ -5,7 +5,7 @@
  * Verify credentials and redirect to based on response from Twitter.
  */
 
-include_once('../../../../../config.php');
+include_once('../../../../config.php');
 
 global $sugar_config;
 
--- a/include/social/twitter/twitter_helper.php
+++ b/include/social/twitter/twitter_helper.php
@@ -80,7 +80,7 @@ function duplicate_check($db, $text, $date)
 function check_auth($url)
 {
 
-    $url = $url . "/include/social/twitter/twitter_auth/callback.php";
+    $url = $url . "/index.php?entryPoint=twitter_callback";
 
     $config = '';
 


chris001:
@vladbar  
Tried configuring Twitter OAuth Connector as you describe.  Problem is 
Website * - http://mycompany.com/crm/twitteroauth  This URL gives a 404, doesn't exist on SuiteCRM.
Callback URL - http://mycompany.com/crm/twitteroauth/callback.php  Also gives a 404, doesn't exist either.

Aren't these URLs supposed to be real, valid pages?

vladbar:
Hi Chris,

thank you for the raised concern.

http://example.com/twitteroauth/callback.php is just a placeholder (this is what Twitter API suggests to put there) - the SuiteCRM will call an appropriate file anyway.

You can also change http://example.com/twitteroauth to http://example.com as far as I know.

chris001:
@vladbar I tested this and it appears to allow a SuiteCRM user to login to their own personal Twitter account, from the SuiteCRM "My Activity Stream" dashlet, using  the button "Login to twitter".  Once the CRM user logs in to their twitter account, they can authorize the Twitter app, which you created for the SuiteCRM Twitter Connector, and from that point onward, when they write their status updates in the Activity Stream dashlet, it'll post the status update for both users of the CRM who happen to be logged in, to Twitter for the world to see, and possibly also post the status update to Facebook, G+, and IBM Connections, if the CRM user has logged into those.

Question is, how to make this Twitter connector *read and save* the tweets of the Leads, Contacts, Account, Opportunities, etc, for whom the Twitter account username is saved with the Lead/Contact/Account/Opportunity record.   And present these tweets in a summarized format, in relevant context, such as during a Call, Email, in a Detail view of their record, in a Suite Spot report on top keywords they are talking about online.  This is a valuable information  because it helps to getting to know your customer - Social CRM.

samus-aran:
Not quite sure the benefit of this functionality. Perhaps keeping it broken (TODO remove it entirely) is the current path at the moment.

isleshocky77:
@samus-aran Just ran across this issue as well as a [forum post](https://suitecrm.com/suitecrm/forum/installation-upgrade-help/890-7-1-facebook-and-twitter-config) from two years ago. 

According to that post, these were bugs which were going to be fixed in the next release.

Is it now of your opinion that this functionality is useless and should just removed? I was really interested in looking at the connectors and seeing if I could get Facebook/Twitter/LinkedIn feeds from Leads/Contacts/Accounts displayed in a useful manner.

This Authorization workflow is currently the only way to authorize the SuiteCRM installation to be able to use the Twitter API. 

As an add on, I got to this point when I happened to cross [this forum post](https://www.ibm.com/developerworks/opensource/library/x-sugarcrmcon/index.html) describing how to make a new connector with relation to an old linked connector. A working linkedin connector would be extremely useful.

Thoughts on using this issue to fix the broken authorization workflow for the twitter connector?

PR #2683, opened 2016-11-21
Feature scrm 28 quick search

gymad:
Pull request for Quick Search features (SCRM-28, US 5.6 - Quick Search)

## Description

PR contains the following changes and features:
- Search icon on top/bottom of ListView
- Search Form Popup for ListView (pull in basic search)
- style for Popup From
- Search popup link added to 'no results' screen
- visual indication that a filter has been applied to the records and display current search criteria(s)
- Show search forms in popup checkbox added to System Settings and Config

## How To Test This

(rebuild grouping js files need for testing)

Add the following lines to config_override.php for e.g:

$sugar_config['use_legacy_search'][0] = 'Accounts';
$sugar_config['use_legacy_search'][1] = 'Users';
$sugar_config['use_legacy_search'][2] = 'Employees';

## Types of changes

### Final checklist


daniel-samson:
**Issues:**

In the following screen, when I click on the tick or the button (instead of the icon), nothing happens. This will cause problems for tablet and mobile users.  This is also true with the other buttons in the list view toolbar. It also occurs in all of the themes. As a user you should be able to anywhere on the button. 

quick_search_tick

The only exception is when the button has a caret or is a drop down.


On the same screen I get an error in the console:
quick_search_error_message

When I am using the SuiteR theme. The style of the select and bulk action buttons have to much padding on the the left and right.

quick_search_suite_r


**Consideration:**
Though there is a clear button in the popup. When I click on it it doesn't refresh the page. I have to click search in order to see any results again. Perhaps there should be a reset link when no results are given.

include/ListView/ListViewSmarty.php
```diff
@@ -1,11 +1,14 @@
 <?php
-if(!defined('sugarEntry') || !sugarEntry) die('Not A Valid Entry Point');
-/*********************************************************************************
+if(!defined('sugarEntry') || !sugarEntry) {
```
samus-aran: This should be moved to underneath the license

install/install_utils.php
```diff
@@ -1,11 +1,14 @@
 <?php
-if(!defined('sugarEntry') || !sugarEntry) die('Not A Valid Entry Point');
-/*********************************************************************************
+if(!defined('sugarEntry') || !sugarEntry) {
```
samus-aran: Same Again - move under licence

samus-aran:
Okay Dokay.

1. We are not hiding the current search functionality as the new filter will be introduced gently.  So can we have the normal search functionality in the layout again aswell.

2. Reason also for the above is because we haven't discussed how to show saved searches which just now the PR doesn't currently do.

Apart from that it looks good.

samus-aran:
Updated:

Need to make this module configurable so remove the system setting checkbox and adjust the functionality to be module specific.  Please see task SCRM-90 assigned :)

PR #2696, opened 2016-11-23
Fixed #2689 - inline editing removes related parent from Calls by assigning id to accounts_id

### LINKED ISSUE #2689: Inline Edit make you loose Parent just on Calls module, Detail View
#### Issue
When you edit in line any field of the Calls module in Detail View, when the record is saved, the Parent record is not showed anymore.

#### Possible Fix
**I found that on the db the parent_id is still there. It's the parent_type that get lost.**

#### Steps to Reproduce
1. http://demo.suiteondemand.com/
2. go to Calls module
3. enter into detail view of a call
4. edit in line any field 
5. hit enter to save
6. the parent record link disappear

#### Your Environment
* SuiteCRM Version used: 7.6.5, 7.6.6, 7.7.6, 7.7.8
SuiteR and SuiteP theme
English and Italian language

vladbar:
## Description
The issue is that after inline editing bean is saved with parent relationship field having null value. It can be fixed by assigning $bean->parent_id to this parent relationship field.

references issue #2689

## Motivation and Context
Inline editing of any field on the Detail View of the Calls module removes related Account.

## How To Test This
1. go to Calls module
2. enter into detail view of a call
3. inline edit any field
4. hit enter to save
5. the parent record link disappear

## Types of changes

### DIFF
--- a/include/InlineEditing/InlineEditing.php
+++ b/include/InlineEditing/InlineEditing.php
@@ -340,6 +340,13 @@ function saveField($field, $id, $module, $value)
             }
         }
 
+        // Fix for gitHub issue 2689 - inline editing removes related Account from Calls
+        if(isset($bean->parent_id) && $bean->parent_id != null && $bean->parent_id != '') {
+            if($bean->account_id == null || $bean->account_id == ''){
+                $bean->account_id = $bean->parent_id;
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


vladbar:
will need to add some additional checks as the parent field for Calls can be different (not just account_id)

Issue #146, opened 2015-03-04
Additional Details pop-up issues


willrennie:
Overflows off page on modules on list view.

Also creates Calendar issues/overflows off page if there is a lot of spacing/padding in a record when clicking the 'i' on the record within the Calendar month/shared view.

willrennie:
@daniel-samson you should look to address this as part of the new theme, if you have time. Low priority item but would be good to fix.

PR #2721, opened 2016-11-29
fixed - #146 works for all themes

shogunpol:
fixed - #146 works for all themes
## Description

fixed - #146 works for all themes
References issue #146

## Types of changes

### Final checklist

### DIFF
--- a/themes/Suite7/css/style.css
+++ b/themes/Suite7/css/style.css
@@ -5821,6 +5821,7 @@ table tr#delegates_search {
 /* fix schedule bar popup on calendar */
 .ui-front, .ui-dialog {
     z-index: 1100 !important;
+    left: 1531px !important;
 }
 
 .panel-heading {
--- a/themes/SuiteP/css/style.css
+++ b/themes/SuiteP/css/style.css
@@ -7212,11 +7212,7 @@ hr {
     border: 1px solid transparent;
     border-radius: 4px;
 }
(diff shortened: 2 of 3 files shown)


samus-aran:
@shogunpol We can not go ahead using !importants.  In the 7.8 beta 1 release this .ui-dialog is now not using !important.  Can you adjust to remove the !important and still work as expected?
Also the left is very specific.... would it be the same if you reduced the size of the browser as @CamoMacdonald states?

themes/Suite7/css/style.css
```diff
@@ -5821,6 +5821,7 @@ table tr#delegates_search {
 /* fix schedule bar popup on calendar */
 .ui-front, .ui-dialog {
     z-index: 1100 !important;
+    left: 1531px !important;
```
samus-aran: Left is very specific but surely this would change depending on the side of the window and number of columns on the list view?

samus-aran:
The solution is not to use CSS but resolve the offset of the dialog call in getAdditionalDetails 
https://github.com/salesagility/SuiteCRM/blob/master/jssource/src_files/include/javascript/sugar_3.js#L3283

^^^ That line of code is causing all the grief.  However I not sure why it needs to be appended to the content div.  But I suggest you start there and make the dialog pop up at an offset of the element within that content div.

Dillon-Brown:
Closed due to inactivity.

PR #2766, opened 2016-12-06
Fix #2765 - Saving EditView from Studio breaks if top row has only one field

### LINKED ISSUE #2765: Saving EditView from Studio breaks if top row has only one field
#### Issue
Modify an EditView inside Studio. Change the top row in any panel to display only one field instead of two. Now click "Save & Deploy". An error pops up and the view is cut off at the row with one field.

#### Possible Fix
I found the bug and will submit a pull request.

#### Your Environment
* SuiteCRM Version used: 7.6.10

dfstrauss:
## Description

Fix for bug #2765 

Fix the view metadata parser (GridLayoutMetaDataParser) to properly initialize the "(filler)" and the "(empty)" dummy field definitions.

Rendering of the template breaks if these values are not set properly.

## How To Test This
Modify an EditView inside Studio. Change the top row in any panel to display only one field instead of two. Click  "Save & Deploy".

Before the fix this would break; now it works.

## Types of changes

### Final checklist

### DIFF
--- a/modules/ModuleBuilder/parsers/views/GridLayoutMetaDataParser.php
+++ b/modules/ModuleBuilder/parsers/views/GridLayoutMetaDataParser.php
@@ -142,9 +142,8 @@ class GridLayoutMetaDataParser extends AbstractMetaDataParser implements MetaDat
     function getLayout ()
     {
     	$viewdefs = array () ;
-    	$fielddefs = $this->_fielddefs;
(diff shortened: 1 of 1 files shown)


dfstrauss:
After feedback on the related issue #2765 I was unable to reproduce the problem. So closing the pull request.

PR #2770, opened 2016-12-07
add missing method stream_metadata for streamWrapper class for upload…

ahildoer:
The streamWrapper interface prescribes the method stream_metadata per http://php.net/manual/en/streamwrapper.stream-metadata.php. We have added it here.

## Description
We simply added the missing method.

## Motivation and Context
We get errors on import and the status and errors csv files are not populating correctly, leaving the UI lacking proper feedback for the user.

## How To Test This
- Import a file larger than the maximum split. For default install this is 100 rows. So, 200+ should do it.
- Make sure the file has various errors in it, at different rows (some in the 1-100 range, some in the 200+ range)
- Watch the error log and look at the results at the end of the import. Errors will be missing.

## Types of changes

### Final checklist

### DIFF
--- a/include/upload_file.php
+++ b/include/upload_file.php
@@ -820,6 +820,32 @@ class UploadStream
         return @stat(self::path($path));
     }
 
+    public function stream_metadata( $path, $option, $value )
+    {
+
+        $path = self::path( $path );
+
+        switch ( $option )
+        {
+            case STREAM_META_TOUCH:
+                return @touch($path, $value[0], $value[1]);
+                break;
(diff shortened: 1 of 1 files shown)


samus-aran:
Hi there, 

However ideally the called functionality should be using existing classes/utils that are in the system so we don't introduce new calls that supress errors. 

For example the @ chown should be sugar_chown from the utils functions here.
https://github.com/salesagility/SuiteCRM/blob/7.8.x/include/utils/sugar_file_utils.php

Will close but if anyone wants to pick this up then do so. :+1:

PR #2782, opened 2016-12-09
m2m: fix SecurityGroup instance checking

majkel89:
Fix the way `M2MRelationship::add` method checks if given relation is added between `User` / `ACLRole` and `SecurityGroup` beans

## Description
When `User`, `ACLRole` or `SecurityGroup` is overloaded and the custom class is registered in `include/modules_override.php`  then user in no longer able to assign users/roles to security groups

## How To Test This
1. Overload any of the bean classes: `User`, `ACLRole`, `SecurityGroup`
2. Go to `Administration` -> `Security Groups` -> go to preview of any Security Group -> try to add User and/or Role to the security group.
3. Current behavior: 500 error when adding user or group. Expected behavior: user / role is added to security group

## Types of changes
- [x] Bug fix (Cannot add user/role to security group when any of the classes are overloaded)

### DIFF
--- a/data/Relationships/M2MRelationship.php
+++ b/data/Relationships/M2MRelationship.php
@@ -132,14 +132,18 @@ class M2MRelationship extends SugarRelationship
      */
     public function add($lhs, $rhs, $additionalFields = array())
     {
+        require_once("modules/Users/User.php");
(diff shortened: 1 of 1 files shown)


Dillon-Brown:
Closed due to inactivity

Issue #2760, opened 2016-12-05
Error in Reports page after upgrade: Missing argument 4 for AOR_Report::ACLAccess(),


jssldb:
#### Issue

We recently upgraded from Sugar Enterprise, to CE, then to Suite CRM. On opening the Reports tab for the first time, we get this error about the ACLAccess function in the AOR Reports module. We can't then access the page.

#### Expected Behavior

Expect to be able to access the Reports page.

#### Actual Behavior
The following error is displayed: "Warning: Missing argument 4 for AOR_Report::ACLAccess(), called in /../../../include/MVC/Controller/SugarController.php on line 356 and defined in /../../../modules/AOR_Reports/AOR_Report.php on line 120"

#### Possible Fix

I've looked through some files and seen that ACLAccess() is called in several different modules, but it's only in AOR Reports that it requires the fourth argument, even though ACLAccess is always called with only one argument, with the other two having default values. Does the fourth argument need to be present at all in AOR_Reports? 

#### Steps to Reproduce

1. In menu, go to All > Reports
2. ???
3. Error

#### Context

We won't be able to use the Reports functionality if we can't get past this error.

#### Your Environment

* SuiteCRM Version used: 7.7.8
* Google Chrome Version 54.0.2840.98 (64-bit):

dhayner:
jssldb - Why did you close this?  I have the same issue on a fresh install of 7.7.8 .  Thanks

jssldb:
I thought that it must just be an issue with our install, because I looked at their demo and didn't see anything like our error. But it looks like it's a bigger issue, so I'll reopen.

shogunpol:
@jssldb , is this mean you cannot open report module at all? , did you tried with fresh install?

jssldb:
What I ended up doing, until we got an answer, was simply open the AOR_Report.php and comment out the fourth argument. I couldn't find a single place that actually called for that argument, and all other modules with a similar function only had three arguments.

After doing that, we were able to use to use the Report module.

ijdavie:
@jssldb I would try replacing the line 

function ACLAccess($view,$is_owner='not_set',$in_group='not_set', $target_module){

with 

function ACLAccess($view,$is_owner='not_set',$in_group='not_set', $target_module=""){

from what I can see its complaining it is not being set, adding this may solve the problem. 

if you can confirm if this fixes it that would be great, Thanks.

jssldb:
[...]

PR #2804, opened 2016-12-20
Fixes #2803 - Restrict Export options based on user permissions for report's target module

### LINKED ISSUE #2803: AOR Reports - Restrict Export function on report target modules
#### Issue

A previous fix covered restricted users from creating and seeing the results of reports based on the target module of the report. However, this is not extended across to exporting the report, which will allow the restricted user to see data they should not be able to.
#### Expected Behavior

A user who has restricted access to a module should not be able to export reports targeting that module.
#### Actual Behavior

Any user can export a report despite restrictions on the target module of the report.
#### Possible Fix

Do another access check on the user's permissions for the target module when trying to export or when building the export button, similar to previous fixes regarding the target module restrictions
#### Steps to Reproduce

1. Go to report list view on a user with denied access to a module that has a report.
2. Select report and click export, it will create a csv with all the data.
3. Go into the report, click export and the same will occur.
#### Your Environment

* SuiteCRM Version used: 7.7.8
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Chromium
* Environment name and version (e.g. MySQL, PHP 7): PHP 5.6

MikeyJC:
## Description

Added more access checks on AOR_Reports and export.php to handle users not having access to the report's target module. It will only build the export buttons when the user has access to the target module. When users try to export from the list view, if they don't have access to the target module, an alert appears and they are given the standard export disabled screen.
References issue #2803
References issue #2760
## Motivation and Context

Restricts users from taking data on reports they should not be allowed to see.
## How To Test This

Go to the reports module as a user with restricted access to a module that is the target of a report and try to export.
## Types of changes

### Final checklist


custom/include/language/en_us.lang.php
```diff
@@ -0,0 +1,17 @@
...
+  'test5' => 'test ',
+  'test6' => 'tset6',
+  'test7' => 'test7',
+  'test8' => 'test 8 ',
+  'quote' => 'quote',
+  'quote2' => 'quote"2',
+  'quote3' => 'quote',
+);
+
+$GLOBALS['app_list_strings']['LBL_REPORTS_RESTRICTED'] = 'A report you have selected is targeting a module you do not have access to. Please select a report with a target module you have access to.';
```
horus68: Why create a custom file to include a language string and remove it from include/language/en_us.lang.php

MikeyJC: I was having issues installing these changes via a package when the string was originally in the core language file. This was more for my own work, and you're right that it shouldn't be there, I'll revert the commit. Thanks for bringing this to my attention!

Dillon-Brown:
One of the reference issues has been resolved. This would need to be remade for hotfix-7.8.x and the conflicts resolved.

Issue #2512, opened 2016-10-29
Upgrade PHPMailer to 5.2.16. Use XOAUTH2 classs for secure GMail.


chris001:
#### Issue

1. PHPMailer 5.2.13 is outdated.  Should be upgraded to 5.2.16.
2. PHPMailer comes with the XOAUTH2 class, which recommended by Google as the login method for applications to securely authenticate to GMail.  The purpose of XOAUTH2 login is to prevent theft of your Google account, much greater protection from email account hacking, etc.  Google disables by default normal password based login because allowing applications to perform password login to email is terribly insecure.  Getting XOAUTH2 into SuiteCRM would eliminate the constant problems users are experiencing in trying to connect suitecrm to one of the biggeset email providers, gmail.
#### Expected Behavior

Send email thru GMail with no errors.
#### Actual Behavior

Error LOGIN-PLAIN unsupported by default, use XOAUTH2, or enable "Less secure apps" which is easily hacked.
#### Possible Fix

Here is an example of using PHPMailer to login to gmail with XOAUTH2, and sending an email.

https://github.com/PHPMailer/PHPMailer/blob/master/examples/gmail_xoauth.phps
#### Steps to Reproduce

1.
2.
3.
4.
#### Your Environment

- SuiteCRM Version used: 7.7.6
- Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Firefox 50.0b10
- Environment name and version (e.g. MySQL, PHP 7): PHP  5.6, Mysql 5.6
- Operating System and version (e.g Ubuntu 16.04): Debian 8

chris001:
See #2831 
PHPMailer should be upgraded urgently to 5.2.19 due to security issue, remote code execution:
https://legalhackers.com/advisories/PHPMailer-Exploit-Remote-Code-Exec-CVE-2016-10033-Vuln.html

* This issue makes it more urgent for SuiteCRM to start to actively use PSR composer/packagist.org.  All user/customer installations of SuiteCRM could be easily configured so that composer would by default download and upgrade both the CRM application itself and the 3rd party libraries such as PHPmailer automatically when a new important bug fix or security fix version is released.

chris001:
PR #2840 should fix this by means of `composer update`, running on a daily `cron` job.

horus68:
@chris001 see related (PHP mailer update) https://github.com/salesagility/SuiteCRM/pull/3340

chris001:
[...]

PR #2840, opened 2016-12-28
Add PHPMailer, composer-custom-directory-installer

### LINKED ISSUE #2831: Security issue in PHPMailer
Maybe Suite needs to check for that vulnerability:

https://legalhackers.com/advisories/PHPMailer-Exploit-Remote-Code-Exec-CVE-2016-10033-Vuln.html

chris001:
This update to `composer.json` allows for effortless automatic security updates to third party libraries, such as `PHPMailer`.

## Description

This should fix #2831 and #2512 and greatly reduce issues related to bugs in outdated 3rd party libraries, because of the easy update, which should be set on a daily `cron` job on linux or `scheduled task` on windows.

Add `phpmailer` requirement to composer.
Add `composer-customer-directory-installer` requirement to composer.
Add correct path to update, the `phpmailer` folder within SuiteCRM `include` folder.

## Motivation and Context

From time to time, serious bugs are reported in third party libraries, and it's critical to get the fixed library version onto the user's production instance as quickly and automatically as possible.
Until now, when you ran `composer update`, this `composer.json` configuration file was effectively empty, so it was not updating any of the SuiteCRM application third party libraries.
Until now.
Now, updates will happen.  Composer will download and install the latest updated bug fixed and security fixed code libraries, directly from the author's online `packagist.org` repository, as soon as the user or the system executes the one line `composer update` command.

## How To Test This

Install `composer` on the system.  On linux, typically the command is either `yum install composer` or `apt-get install composer`
Install this file `composer.json` in the main folder of the SuiteCRM application.
From the main folder of the SuiteCRM application, run:
 `composer update`
The new version of PHPMailer will download and install into the correct folder.

Optional:
run `composer self-update` to update composer itself to the current version.

## Types of changes

### Final checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,6 +9,8 @@
     "vendor-dir": "vendor"
   },
   "require": {
+    "phpmailer/phpmailer": "^5.2",
+    "mnsami/composer-custom-directory-installer": "^1.1"
   },
   "require-dev": {
     "phpunit/phpunit": "^3.7",
@@ -21,5 +23,9 @@
       "": "src/"
     }
   },
-  "extra": {}
+  "extra":{
+    "installer-paths":{
+      "./include/phpmailer/": ["phpmailer/phpmailer"]
+    }
+  }
 }


mattlorimer:
that an interesting approach, but I think we should just start moving 3rd party components to the standard  vendor folder in a upcoming release.

chris001:
@mattlorimer 
Moving 3rd party components to the standard `vendor` folder would only work if there is no module out there which depends on finding the classic 3rd party components in the `include` folder and breaks when it fails to find them in `include` where they've always been.
Obviously, to get around this, we would have to add symlinks from `include/phpmailer` to `vendor/phpmailer/phpmailer` and so forth, one symlink for each of the 3rd party libraries.

PR #2867, opened 2017-01-05
Fixed #2866 removed links parameter from the build_report_html function signature

### LINKED ISSUE #2866: AOR Report pdf links cannot be created.
#### Issue
When checking the "Link" box in a report, it creates a Web Link in the online report, but not in a PDF

#### Expected Behavior
If user checks the Link check-box opposite item, this item should be created as a link in generated PDF. 

#### Actual Behavior
User checks the Link check-box opposite item, but the item only shows as a link inside of the generated report. The item is not displayed as a link in a PDF though.

#### Steps to Reproduce
1. All -> Reports -> Create Report
2. Report Module -> Opportunities (for example)
3. Fields -> add Opportunity Name -> tick Display, tick Link
4. Save
5. Generated report would display Opportunity Name as a link.
6. Click on Download PDF (from the dropdown menu).
7. Generated PDF will not display Opportunity Name as a link.

#### Context
this issue is caused by the following statement:
- modules/AOR_Reports/controller.php line 232
$printable = $this->bean->build_group_report(-1,false);

If I change it to 
$printable = $this->bean->build_group_report(-1,true); then the PDF will show Opportunity Name as a link. 
Which makes the following argument valid - there is no point to have a  $links = true parameter in the function signature 
function build_report_html($offset = -1, $links = true, $group_value = '', $tableIdentifier = '', $extra = array()){
as this parameter need to be always true.

vladbar:
## Description
references issue #2866
Removed links parameter from the build_report_html function signature.

There is no point to have a $links = true parameter in the function build_report_html signature as every time it is called, it needs to be true. If the parameter always needs to be true then there is no need to have this parameter.

## Motivation and Context
In one place links was set to false and users were complaining that whether they choose display as link or not PDF do not display links at all

## How To Test This
1. All -> Reports -> Create Report
2. Report Module -> Opportunities (for example)
3. Fields -> add Opportunity Name -> tick Display, tick Link
4. Save
5. Generated report would display Opportunity Name as a link.
6. Click on Download PDF (from the dropdown menu).
7. Generated PDF will display Opportunity Name as a link.

## Types of changes

### DIFF
--- a/modules/AOR_Reports/AOR_Report.php
+++ b/modules/AOR_Reports/AOR_Report.php
@@ -513,20 +513,20 @@ class AOR_Report extends Basic {
             while ($row = $this->db->fetchByAssoc($result)) {
                 if($html != '') $html .= '<br />';
 
-                $html .= $this->build_report_html($offset, $links, $row[$field_label], '', $extra);
+                $html .= $this->build_report_html($offset, $row[$field_label], '', $extra);
 
             }
         }
 
         if($html == ''){
-            $html = $this->build_report_html($offset, $links);
+            $html = $this->build_report_html($offset);
         }
         return $html;
 
     }
 
 
-    function build_report_html($offset = -1, $links = true, $group_value = '', $tableIdentifier = '', $extra = array()){
+    function build_report_html($offset = -1, $group_value = '', $tableIdentifier = '', $extra = array()){
 
         global $beanList, $sugar_config;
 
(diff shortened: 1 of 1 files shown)


Mac-Rae:
Please rebase this pull request onto hotfix-7.10.x (For 7.10.x Issues) or hotfix (For 7.11.x Issues) as the Master & Hotfix branches have changed purpose and no longer allow direct PRs since submission. 

Please take a look at our [contribution guidelines](https://github.com/salesagility/SuiteCRM/blob/master/.github/CONTRIBUTING.md) for more information.

Dillon-Brown:
While this parameter is unused, we wouldn't be able to remove it as that would break backwards compatibility.

PR #2880, opened 2017-01-09
Fixed #2183 - only check for duplicate field names if it is not an update

### LINKED ISSUE #2183: Studio: Can't modify fields - getting 'Invalid Value: Field Name already exists'
Modifying a newly created field is not possible. SuiteCRM shows 'Invalid Value: Field Name already exists'

#### Issue

Create a field in Studio. Modifying i.e. the Display Label or Max Size is not possible any more. SuiteCRM show error message.
#### Expected Behavior

Modifying several attributes should be possible.
#### Actual Behavior

See above.
#### Steps to Reproduce

1. Fresh 7.7.4 install
2. Create a new field in Studio
3. 'Repair JS Files'
4. Logout
5. Clear browser cache
6. Login
7. Modify the created field is not possible ('Invalid Value: Field Name already exists')
#### Your Environment

- Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): FF 48.0.2
- Environment name and version (e.g. MySQL, PHP 7): php 56, mysql 5.6
- Operating System and version (e.g Ubuntu 16.04): FreeBSD 10.2

vladbar:
## Description
references issue #2183
The following file - modules/ModuleBuilder/tpls/MBModule/field.tpl had this input field with a value always true - <input type='hidden' name='is_update' value='true'>

So the form could not differentiate whether it was creating a new field or just updating an existing field.

This caused a problem with saving already existing fields - it tried to check whether the field name already exists (this check is only required if the user creates a new field, if the user updates an already existing field, then there is no need to check for duplicate field name as the user cannot change the field name anyway it will always stay the same in case of an update).

The following changes were applied: 
1. modules/ModuleBuilder/views/view.modulefield.php
        if($field_name == '') {
            $is_update = false;
        } else {
            $is_update = true;
        }
        $fv->ss->assign('is_update', $is_update);

This change is required to tell the template modules/ModuleBuilder/tpls/MBModule/field.tpl whether it's an update or creation of a new field.
$field_name will be '' if the user clicks on 'Add field' (creation of a new field).
$field_name will be '... specific field name ...' if the user clicks on already existing field name (update of already existing field).

2. modules/ModuleBuilder/tpls/MBModule/field.tpl
{if $is_update}
	<input type='hidden' name='is_update' value='true'>
{else}
	<input type='hidden' name='is_update' value='false'>
{/if}

This change outputs is_update hidden field with a true or false value instead of just always true value.

3. jssource/src_files/include/javascript/sugar_3.js
    if(document.getElementsByName('is_update')[0].value == 'false') {
         ... check for duplicate name ...
    }
    Only check for duplicate name if the form creates a new field 

## How To Test This
1. Navigate to Admin -> Studio -> Accounts -> Fields -> click on description
2. Change the value of Display Label: to New Description 
3. Save
4. The field should successfully save without error - 'Invalid Value: Field Name already exists'
5. Display label for the field should now be New Description 

1. Navigate to Admin -> Studio -> Accounts -> Fields -> click on 'ADD FIELD'
2. Enter description into Field Name:	
3. Save
4. 'Invalid Value: Field Name already exists' error message should appear as description field already exists.

## Types of changes

### DIFF
--- a/jssource/src_files/include/javascript/sugar_3.js
+++ b/jssource/src_files/include/javascript/sugar_3.js
@@ -1208,15 +1208,17 @@ function validate_form(formname, startsWith){
 									}
 								});
 
-								for (k = 0; k < current_fields.length; k++) {
-									if(isError != true) {
-										val = current_fields[k].toUpperCase();
-										if ((operator == "==" && val == item1) || (operator == "!=" && val != item1)) {
-											isError = true;
-											add_error_style(formname, validate[formname][i][nameIndex], 'Invalid Value: Field Name already exists');
+								if(document.getElementsByName('is_update')[0].value == 'false') {
+									for (k = 0; k < current_fields.length; k++) {
+										if(isError != true) {
+											val = current_fields[k].toUpperCase();
+											if ((operator == "==" && val == item1) || (operator == "!=" && val != item1)) {
+												isError = true;
+												add_error_style(formname, validate[formname][i][nameIndex], 'Invalid Value: Field Name already exists');
+											}
 										}
 									}
-								}
+								}s
 							break;
 							case 'verified':
 							if(trim(form[validate[formname][i][nameIndex]].value) == 'false'){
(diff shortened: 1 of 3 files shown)


vladbar:
Reviewed it is still working. You probably forgot to Rebuild JavaScript.

PR #2897, opened 2017-01-12
Add User lockout and user unlock options 

JimMackin:
## Description

Provides admin settings for automatically locking users after a configurable amount of failed logins. Also provides admin users the ability to unlock locked users.
Also provides an admin setting for automatically unlocking users after a configurable amount of time.

## Motivation and Context

Currently there is no limit to the number of login attempts that a user can make, this could in theory allow brute force attacks on the login screen. This PR allows an automated way of preventing this.

## How To Test This

Setting the new maxfailedlogins config setting and deliberately failing to login, say, 3 times will prevent future logins until the user is unlocked on the user screen by an admin.
Setting the automaticunlocktime config setting however will unlock users x minutes after they are locked.

## Types of changes

### Final checklist

### DIFF
--- a/modules/Administration/PasswordManager.tpl
+++ b/modules/Administration/PasswordManager.tpl
@@ -250,7 +250,7 @@
 									{/if}
 									</table>
 
-									<table id="userResetPassId" name="userResetPassName" width="100%" border="0" cellspacing="1" cellpadding="0" class="edit view">
+									<table id="userLockId" name="userLockName" width="100%" border="0" cellspacing="1" cellpadding="0" class="edit view">
 										<tr>
 											<th align="left" scope="row" colspan="2"><h4>{$MOD.LBL_USER_LOCKOUT}</h4></th>
 										</tr>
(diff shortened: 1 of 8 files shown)


JimMackin:
One thing to raise is that currently config settings are only saved if they already exist due to how the configurator works. This means that new installs will have the default config settings but upgrades may not be able to save the settings. We may need to add to the upgrade script to add these missing config settings.

modules/Users/authentication/SugarAuthenticate/SugarAuthenticate.php
```diff
@@ -83,6 +83,21 @@ public function SugarAuthenticate(){
 		self::__construct();
 	}
 
+	private function isUserLockedOut(User $user){
```
mattlorimer: Docblocks Comment missing

modules/Administration/PasswordManager.tpl
```diff
@@ -250,8 +250,41 @@
 									{/if}
 									</table>
 
-
-
+									<table id="userResetPassId" name="userResetPassName" width="100%" border="0" cellspacing="1" cellpadding="0" class="edit view">
```
mattlorimer: shows even when LDAP or SAML are selected, should be hidden, unless this support them??

modules/Users/views/view.detail.php
```diff
@@ -122,6 +122,11 @@ function preDisplay() {
                     }
                 }
             }
+            if(is_admin($current_user)){
```
JimMackin: The two checks above also check for if the user is editing their own record, this shouldn't matter since if the user is locked they wont be viewing their own record but seems safer to keep this check separate?

mattlorimer: yes but your inside that elseif check, so that is right I am talking about the if statement on line 111

JimMackin: Ah, you're right. Fixed.

modules/Administration/language/en_us.lang.php
```diff
@@ -1270,5 +1270,9 @@
     'LBL_AUTO_UNLOCK_TIME_UNITS' => 'minutes',
     'ERR_MAX_FAILED_LOGINS' => 'Please specify a valid value for the number of max failed logins',
     'ERR_AUTOMATIC_UNLOCK_TIME' => 'Please specify a valid value for the automatic unlock time',
+    'LBL_ENABLE_MAX_FAILED_LOGINS' => 'Lock users after a number of failed logins',
+    'LBL_ENABLE_MAX_FAILED_LOGINS_HELP' => '',
```
horus68: Empty language string!

horus68:
I see this is merged in the feature branch. What SuiteCRM release is this planned for?

horus68:
@Dillon-Brown I see this is merged in the feature branch. What SuiteCRM release is this planned for?
Please consider also the PR: #2953

horus68:
@Dillon-Brown I see this is merged in the feature branch. But what SuiteCRM release is this planned for?
Please consider also the related PR: #2953

SinergiaCRM:
Hi @serhiisamko091184,

Can we bump this up?

Not sure if @JimMackin is still active on this. If not, with his permission, we could merge it with SuiteCRM latest version.

PR #2930, opened 2017-01-17
Fixed #707 - added conditional statement to check if action is not clone

### LINKED ISSUE #707: Studio - Clone a field doesn't work

vladbar:
## Description
references issue #707

Added a conditional statement inside submitForm function in order to prevent it from redirecting the user before the clone action has actually happened.
The statement checks if that the action is not a clone.

## Motivation and Context
This fix is required in order to fix Studio -> Clone field functionality. 

At present when Clone button is pressed inside Edit Field tab, it brings a floating div with the message:
"This operation is completed successfully"
But instead, it closes the Edit Field tab and then returns to Edit Fields tab
When you pressed the X on the floating div, it closes doing nothing.

## How To Test This
1. Navigate to Studio -> any module -> Fields -> any field.
2. click Clone
3. The following message - "This operation is completed successfully" will be presented.
4. Then  the Edit Field tab is closed and no cloning operation is performed

## Types of changes


pribeiro42:
Assessed. Working.

PR #2936, opened 2017-01-17
Hiding nav menus when the current module text is too large.

daniel-samson:
This prevents the toolbar on the right hand side from wrapping to the next line.

## Description

When a user access a module that has a name which is very long. The toolbar on the right hand side wraps to the next line. This is due to a change I made the the spacing of the home button.

So to ensure that this doesn't happen again I have implemented a script which is similar the one we used for non filter menus; where it hides the menu item. However this implementation doesn't put the element in the All menu.

## Motivation and Context

* prevents the nav bar style from breaking.

## How To Test This

* Compile style.css from sass
* Access a user that utilises the filter menus
* Access a module like MAP ADDRESS CACHE

## Types of changes

### Final checklist

### DIFF
--- a/themes/SuiteP/css/navbar.scss
+++ b/themes/SuiteP/css/navbar.scss
@@ -2125,6 +2125,26 @@
   }
 }
 
+.topnav #overflow-menu > li.with-actions .dropdown-menu {
+  display: none;
+  visibility: collapse;
+}
+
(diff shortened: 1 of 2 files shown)


samus-aran:
Remember to supply your compiled css when merging into develop @daniel-samson

PR #2937, opened 2017-01-17
Fixing detail view tabs (extra white space)

daniel-samson:
## Description

Due a fix I made for the mobile. When a user accesses the detail view the first tab has a lot of white space at the bottom of the tab. The reason is that I needed to remove the inline style so that the user could see the tabs in the mobile. This change adds the display:none property using the css instead of inline thus fixing the problem.

## Motivation and Context

* Bug fix

## How To Test This

* compile sass
* access detail view
* check that the tab doesn't have extra white space at the bottom.

## Types of changes

### Final checklist

### DIFF
--- a/themes/SuiteP/css/tabs.scss
+++ b/themes/SuiteP/css/tabs.scss
@@ -247,3 +247,8 @@
   border-bottom-left-radius: 4px;
   border-bottom-right-radius: 4px;
 }
+
+
+.tab-content .tab-pane-NOBOOTSTRAPTOGGLER:not(.active) {
+  display: none;
+}


themes/SuiteP/css/tabs.scss
```diff
@@ -246,4 +246,9 @@
   border-top-right-radius: 4px;
   border-bottom-left-radius: 4px;
   border-bottom-right-radius: 4px;
-}
\ No newline at end of file
+}
+
+
+.tab-content .tab-pane-NOBOOTSTRAPTOGGLER:not(.active) {
```
samus-aran: Explained that this white space is used for the mobile view, but the correct display:none was not in the desktop.

daniel-samson: It's a style sheet issue / Bootstrap JavaScript issue.

We can not employ the default behaviour in bootstrap due to the mobile first tab. As it needs to exist in the same place as the first tab, which the Bootstrap JavaScript cannot handle correctly.  Because the DOM is different that from what bootstrap expects. The reason it needs to be in the same location as the first tab is to ensure that the drop down displays correctly on a mobile. So in the detail there is a custom script tag which handles this tab behaviour correctly.

There used to be a selector for that applied across mobile, desktop, and tablet It employed the display: none across all tabs regardless of which tab was active. Then the .active is meant to override the property but it didn't. Which prevented the tab's from being in view on the mobile.  .fade css class only effects the opacity of each tab and not the display property. The .active class only defines the display: block. So in order to work around these issues and hide the non active tabs I have added a CSS selector to any non active tab content and set the display property to none.

PR #2966, opened 2017-01-23
Fix #2927 - Emails View Relationship popup collapses on second use

### LINKED ISSUE #2927: Suite P - Emails View Relationship popup collapses on second use
#### Issue
Using the Suite P theme in version 7.7.9. In the Emails component right click on an email and select the 'View Relationships' option. The 'Email Record' popup displays with sufficient height to view the contents (see attachment 1). 

If the popup is closed and then the action is repeated (right click on an email and select the 'View Relationships' option) the popup reappears but only the popup header is visible (none of the content is visible) rendering the popup useless (see attachment 2). 

Refreshing the page works around the issue but is very inconvenient to have to do so every time.

[This forum post](https://suitecrm.com/forum/suite-themes/12065-suitep-quick-create-from-email-multiselect-dropdown-box-size#40926) describes the same bad behaviour for the 'Quick Create' -> 'Contact' popup

#### Expected Behavior
The popup should open every time with at least adequate minimum height to make the content view-able.

#### Actual Behavior
The popup opens up the first time with adequate height to make the content view-able, but all subsequent openings result in the popup having far too little height to view any contents.

#### Possible Fix
Looking at the attachments you can see the height of the container is calculated the first time at 381px but the second time the height is calculated at only 10px!

I suggest that when dynamically calculating the height of a popup's content there should be a lower limit (say 150px) that the height will not be set lower than ever. This would at least allow the content to be viewed. 

#### Steps to Reproduce
1. Visit [Live Demo](http://demo.suiteondemand.com/index.php?module=Emails&action=index&parentTab=All)
2. Got to Emails module and select 'My Sent Emails'
3. Right click on an email and select the 'View Relationships' option.
4. Close popup
5. Right click on an email and select the 'View Relationships' option.

#### Context
Our users spend most of there time in the Emails module so this is a very annoying and time wasting bug.

#### Your Environment

* SuiteCRM Version used: Version 7.7.9
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Chrome Version 55.0.2883.87 (64-bit) 
* Environment name and version (e.g. MySQL, PHP 7): MySQL 5.6, PHP 5.4.36
* Operating System and version (e.g Ubuntu 16.04): Ubuntu 14.04.1 LTS

colinhunter:
Set contH to document.body["scrollHeight"] as default.

## Description
#2927 - On second click body.childNodes[i] isn't available so height calculation defaults to original (default) setting. Setting this to document.body["scrollHeight"] ensure the height calculation works on subsequent clicks.

## Motivation and Context
This changes ensures the the 'view relationships' pop up is viewable on multiple clicks.
Once implemented, requires rebuild JS grouping files.

## How To Test This
For the email list, right click an email, select 'view relationships'. After the pop up opens, close it and repeat. The subsequent pop up will remain viewable, where previously it was limited to a height of 10px.

## Types of changes

### Final checklist

### DIFF
--- a/modules/Emails/javascript/ajax.js
+++ b/modules/Emails/javascript/ajax.js
@@ -853,7 +853,7 @@ AjaxObject.detailView = {
 			});
 			SED.quickCreateDialog.renderEvent.subscribe(function() {
             	var viewHeight = YAHOO.util.Dom.getViewportHeight();
-            	var contH = 0;
+            	var contH = document.body["scrollHeight"];
             	for (var i in this.body.childNodes) {
             		if (this.body.childNodes[i].offsetHeight)
             			contH += this.body.childNodes[i].offsetHeight;


Dillon-Brown:
Assessed

PR #2995, opened 2017-01-27
Fixed #2266 - Email Template Editor

### LINKED ISSUE #2266: Access to core HTML for Emailtemplate existing and new
#### Issue

The latest version of SuiteCRM introduced a more plug and play approach to code when creating emails. It is really funky but for those of us long in the tooth with SugarCRM we have got used to building the HTML separately and pasting the whole lot into the template builder. Unfortunately the new area only allows you to put HTML inside the container so you don't have access to the core HTML. I've been speaking on a forum and a few others have come across this as a stumbling block. [https://suitecrm.com/forum/installation-upgrade-help/10008-email-templates-html-editor](url) 
#### Expected Behavior

The edit source function should in reality allow you to edit all the HTML not just a container.
#### Actual Behavior

You can only edit the inside container HTML so if you create your email as standalone HTML you can't copy and paste it in and it works.
#### Possible Fix

We are getting around it at the moment by re-enabling the TinyMCE editor which gives the option to edit the true source. A suggestion might be to just allow the new HTML source editor to give you access to the true source not just a container.
#### Steps to Reproduce

1.
2.
3.
4.
#### Context

We have had to add a hack to make it work, it isn't pretty but without it old templates couldn't be reused and to build new ones would require retraining and redesigning our standard design template.
#### Your Environment

- Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Chrome 53.0.2785.116 m (64-bit)
- Environment name and version (e.g. MySQL, PHP 7): PHP 5.9
- Operating System and version (e.g Ubuntu 16.04): Ubuntu 14.04

sgaved:
## Description

The email template editor has been modified to have a checkbox which brings up a textbox which allows users to enter the html directly without the editor modifying it.
## Motivation and Context

Address Issue #2266 - Access to core HTML for Emailtemplate existing and new 
The current editor modifies html in such a way that the formatting of many templates is lost when pasted in to the html source of it. 

## How To Test This

Create email template with html and see the html editor modify the template heavily. Test with fixes to see the html is intact(although it does go through the html purifier). After adding files please clear your browsers js cache. 
## Types of changes

### Final checklist


gody01:
I checked out this solution. It doesn't address real problem, just enables switching between raw HTML and broken WYSIWYG editor.
Out users wants to edit templates/Email with normal WYSIWYG editor, but without template which breaks their work.
If this switch could switch between standard tinyMCE and mozaik it would be perfect.

Issue #782, opened 2015-12-15
PSR Compliance


JimMackin:
It would be nice if SuiteCRM was [PSR-1](http://www.php-fig.org/psr/psr-1/) and [PSR-4](http://www.php-fig.org/psr/psr-4/) compliant. These standards provide a coding and autoloading standard which helps with creating and reusing reusable code. It also helps in using third party libraries which SuiteCRM could do better.

In the future other PSRs could also be considered (such as the logging interface and the caching interface).

daniel-samson:
I agree. Plus we could create an autoloader that handles the custom code better. Removing the need to use require_once

vjaro:
Why not use composer? This is basically what it is for.
I have been using composer together with suitecrm then packaging the codes that i continue to reuse for other project.

JimMackin:
Composer would be a good idea for third party libraries, PSR-4 would simplify loading and including code from within SuiteCRM.

jakubpas:
https://github.com/salesagility/SuiteCRM/pull/1114

- Introduce installation by composer - you can actually test it now:

```composer create-project jakubpas/suitecrm your_application_name```

- introduce PS-4 namespaces for SuiteCRM
- moving internal and abandoned classes to namespaced /src (currently SugarLogger and XTemplate)
- moving external  libraries to composer's /vendor (currenty Smarty as example)
- some minor refactoring

The" jakubpas/suitecrm" can be replaced by official SuiteCRM packagist location as soon as it will be set up.

mattlorimer:
PSR Compliance has been added to the roadmap

PR #3007, opened 2017-01-30
FIX #3005 - New cache system cache lite

### LINKED ISSUE #3005: New Cache system for SuiteCRM
#### Expected Behavior

A new caching method could be implemented using the CacheLite library[1] to prevent many queries being executed to the database, and as a new cache method for SuiteCRM.

[1] https://pear.php.net/package/Cache_Lite/

#### Actual Behavior

As we know SuiteCRM has many cache classes but use by default SugarCacheAbstract class. This cache method use RAM. This is good because RAM is fast, but the cache lasts for a limited time: the request, so in every request has to regenerate the cache. Besides the cache is not shared between users.

CacheLite stores the cache in files, which is slower than using RAM, but has many other benefices:

- The cache last much more time than the request, and you can configure that duration ($sugar_config['CacheLite']['LifeTime'] = 600; in config_override.php)
- The cache is shared between users.
- Reseting the cache is as simple as deleting the cache files.
- You can separate the cache in 'groups', so you can delete the cache of certain groups and the whole.

Abuelodelanada:
**Expected Behavior**

A new caching method could be implemented using the CacheLite library[1] to prevent many queries being executed to the database, and as a new cache method for SuiteCRM.

[1] https://pear.php.net/package/Cache_Lite/

**Actual Behavior**

As we know SuiteCRM has many cache classes but use by default SugarCacheAbstract class. This cache method use RAM. This is good because RAM is fast, but the cache lasts for a limited time: the request, so in every request has to regenerate the cache. Besides the cache is not shared between users.

CacheLite stores the cache in files, which is slower than using RAM, but has many other benefices:

    

- The cache last much more time than the request, and you can configure that duration ($sugar_config['CacheLite']['LifeTime'] = 600; in config_override.php)

- The cache is shared between users.

-  Reseting the cache is as simple as deleting the cache files.
    
- You can separate the cache in 'groups', so you can delete the cache of certain groups and the whole.

## Motivation and Context

Prevent many queries being executed to the database

## How To Test This

Go to Admin, and click in Repair:

suitecrm_admin1

Now we have a new option "Flush the cache". After click in that option, the cache is clean:

suitecrm_admin2

You can see the cache actions in suitecrm.log

suitecrm_log

The cache files are stored in cache/ folder

suitecrm_cache_files

**Some notes about this:**

- Now the default Cache method is the one we implement. This can be changed in the file: include/SugarCache/SugarCachePHPCacheLite.php
May be this could be seted up in config_override.php.

- We saw that the queries executed by the class OutboundEmail (include/OutboundEmail/OutboundEmail.php) were a lot and always the same, and changes rarely. This was the first implementation of the new cache method.
Example: In 53 minutes (the duration of a 200MB SuiteCRM log before it rotates) this class executed 100.000 queries(!) of a total of 413.000 queries.

- We made a test in a development instance of our SuiteCRM implementation, and saw that the queries executed were 44% less!.

This is a good performance improvement for SuiteCRM, and we have it in production.

## Types of changes

### Final checklist

### DIFF
--- a/include/CacheLite/Cache/Lite.php
+++ b/include/CacheLite/Cache/Lite.php
@@ -1,0 +1,896 @@
+<?php
+
+/**
+* Fast, light and safe Cache Class
+*
+* Cache_Lite is a fast, light and safe cache system. It's optimized
+* for file containers. It is fast and safe (because it uses file
+* locking and/or anti-corruption tests).
+*
+* There are some examples in the 'docs/examples' file
+* Technical choices are described in the 'docs/technical' file
+*
+* Memory Caching is from an original idea of
+* Mike BENOIT <ipso@snappymail.ca>
+*
+* Nota : A chinese documentation (thanks to RainX <china_1982@163.com>) is
+* available at :
+* http://rainx.phpmore.com/manual/cache_lite.html
+*
+* @package Cache_Lite
+* @category Caching
+* @author Fabien MARTY <fab@php.net>
+* @author Markus Tacker <tacker@php.net>
+*/
+
+define('CACHE_LITE_ERROR_RETURN', 1);
+define('CACHE_LITE_ERROR_DIE', 8);
+
+class Cache_Lite
+{
+
+    // --- Private properties ---
+
+    /**
+    * Directory where to put the cache files
+    * (make sure to add a trailing slash)
+    *
+    * @var string $_cacheDir
+    */
+    var $_cacheDir = '/tmp/';
+
+    /**
+    * Enable / disable caching
+    *
+    * (can be very usefull for the debug of cached scripts)
+    *
+    * @var boolean $_caching
+    */
+    var $_caching = true;
+
+    /**
+    * Cache lifetime (in seconds)
+    *
+    * If null, the cache is valid forever.
+    *
+    * @var int $_lifeTime
+    */
+    var $_lifeTime = 3600;
+
+    /**
+    * Enable / disable fileLocking
+    *
+    * (can avoid cache corruption under bad circumstances)
+    *
+    * @var boolean $_fileLocking
+    */
+    var $_fileLocking = true;
+
+    /**
+    * Timestamp of the last valid cache
+    *
+    * @var int $_refreshTime
+    */
+    var $_refreshTime;
+
+    /**
+    * File name (with path)
+    *
+    * @var string $_file
+    */
+    var $_file;
+    
+    /**
+    * File name (without path)
+    *
+    * @var string $_fileName
+    */
+    var $_fileName;
+
+    /**
+    * Enable / disable write control (the cache is read just after writing to detect corrupt entries)
+    *
+    * Enable write control will lightly slow the cache writing but not the cache reading
+    * Write control can detect some corrupt cache files but maybe it's not a perfect control
+    *
+    * @var boolean $_writeControl
+    */
+    var $_writeControl = true;
+
+    /**
+    * Enable / disable read control
+    *
+    * If enabled, a control key is embeded in cache file and this key is compared with the one
+    * calculated after the reading.
+    *
+    * @var boolean $_writeControl
+    */
+    var $_readControl = true;
+
+    /**
+    * Type of read control (only if read control is enabled)
+    *
+    * Available values are :
+    * 'md5' for a md5 hash control (best but slowest)
+    * 'crc32' for a crc32 hash control (lightly less safe but faster, better choice)
+    * 'strlen' for a length only test (fastest)
+    *
+    * @var boolean $_readControlType
+    */
+    var $_readControlType = 'crc32';
+
+    /**
+    * Pear error mode (when raiseError is called)
+    *
+    * (see PEAR doc)
+    *
+    * @see setToDebug()
+    * @var int $_pearErrorMode
+    */
+    var $_pearErrorMode = CACHE_LITE_ERROR_RETURN;
+    
+    /**
+    * Current cache id
+    *
+    * @var string $_id
+    */
+    var $_id;
+
+    /**
+    * Current cache group
+    *
+    * @var string $_group
+    */
+    var $_group;
+
+    /**
+    * Enable / Disable "Memory Caching"
+    *
+    * NB : There is no lifetime for memory caching ! 
+    *
+    * @var boolean $_memoryCaching
+    */
+    var $_memoryCaching = false;
+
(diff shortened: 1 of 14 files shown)


Abuelodelanada:
FIX https://github.com/salesagility/SuiteCRM/issues/3005

gunnicom:
Why using a disk cache if there are several caches (apc, memcache ... (That btw are used by suitecrm)) that could do the same in RAM?

Abuelodelanada:
Hi @gunnicom 

APC is an **unmantained an dead** opcode chache. 
Please take a look at: http://php.net/manual/en/intro.apc.php

This feature with Cache_Lite library does not cache opcode. This method cache queries results, arrays, etc.

You can use Memcached and would be better in performance terms than this implementation with Cache_Lite.

But for this you have to install, configure and maintain another software in the server and there a a lot of situations where you cannot do this because may be you are not the sysadmin, or because the sysadmin do not want to install, configure and maintain another software, or because...

SuiteCRM inherits from SugarCRM 9 cache systems:

1. SugarCacheAPC.php 
2. SugarCacheFile.php
3. SugarCacheMemcached.php
4. SugarCacheMemcache.php 
5. SugarCacheMemory.php
6. SugarCacheRedis.php
7. SugarCachesMash.php
8. SugarCacheWincache.php
9. SugarCacheZend.php

Where:

1. Is an opcode and dead cache method and we should delete from SuiteCRM
2. This cache system replicates the Cache_Lite library functionality in a primitive way, and without having the ability to store cache in groups
3. This cache system is a good choise but you have to install, configure and mantaind Memcached
4. This is the same as 3... 
5. This cache system is one SuiteCRM uses by default, but the cache lasts for a limited time: the request and the cache is not shared between users.
6. This one is a good choise, but you have to install, configure and maintain Redis.
7. This one is useless because "zfunctions" does not exists. I assume these functions are used to work with compressed cache files on disk, but I could never make it work. Did you?
8. This one is only for Windows Servers.
9. This is an outdated implementation for ZendCache monster. We should update it.!

Well, in fact there are 5 usable cache system: 2, 3, 5, 6, 8 and 9

This implementation with Cache_Lite library could replace in a better way the one implemented in 2 (SugarCacheFile.php)

We are using this implementation in a 4 nodes SugarCRM CE 6.5.18 implementation with 2000 unique users every day and works great. Saved us thousands of queries to the database.

gunnicom:
I should have said apcu. Did now know you dont know that. apcu is not an opcode cache.

Abuelodelanada:
Ok, apcu.
Yes I know apcu. This one is not implemented in SuiteCRM and would be a good choise too, but in our implementation we need to provide 2 extra features:

- The ability to store the cache in different groups (and delete the entire group of cache).
- Provide a way to remove the cache just deleting cache files.

That's why we use Cache_Lite library.

chris001:
The library should be added with `composer` so that it'll get updates for bug fixes from the author.
`$ composer require pear/cache_lite`

Abuelodelanada:
@chris001 using composer is a good idea, but there are some points to check:

- Using composer supposes SuiteCRM package will not have the libraries, because will be installed during the installation of SuiteCRM. This adds requirements to SuiteCRM: 
  - composer itself
  - Internet connection in the server (As far as I remember, the installer do not download anything)
- PHP version: SuiteCRM supports 5.3x and composer needs 5.3.2+

chris001:
@Abuelodelanada 
* Suite is adding support for PSR-4 Composer Autoloading for this exact reason, security updates to 3rd party libraries, see #782.  Probably the current 3rd party libraries would continue be shipped, for those without internet connection initially, and the system would check daily for updates from the author, and install.  Case in point, the recent high priority security vulnerability issue with PHPMailer #2831 #2840   It's not too much to expect most CRM servers to be connected to the internet, otherwise, how else will users be able to send email campaigns, or receive support requests from customers online? How else will the server operating system update itself?
* Support for PHP 5.3, from PHP organization itself, ended in middle 2014, 2.5 yrs ago.  The current supported versions of PHP, by PHP.net, are 5.6, 7.0, and 7.1.   http://php.net/supported-versions.php  It's fair to require CRM servers to run the currently supported version of PHP: 5.6 minimum (on security patch status only), preferably 7.0+ (for which new features are being developed).

Abuelodelanada:
@Abuelodelanada

- Great to hear about PSR-4! I did not see #782 before.
- About: _"It's not too much to expect most CRM servers to be connected to the internet, otherwise, how else will users be able to send email campaigns, or receive support requests from customers online? How else will the server operating system update itself?"_
  - If you implement a CRM in a Bank you will not have internet connection in the CRM.... that's my case ;-) Anyway, you only install the CRM the first time, so it is not so important.

- About PHP 5.3 I completely agree with you, just something to consider.

samus-aran:
[New Cache system for SuiteCRM](https://trello.com/c/kSHWxf6v/36-new-cache-system-for-suitecrm)

chris001:
Can we have comments from @daniel-samson @samus-aran @salesagility @gregsoper @mattlorimer about this PR.  It improves performance by reducing load on the database, and is more easily manageable in the file system than the existing file cache.  Why isn't this PR merged yet.  What else do you need. Thank you.

horus68:
@Dillon-Brown will this PR to be considered in the future?

Mac-Rae:

Abuelodelanada:
> Please resolve the conflicts on this pull request bringing it up to date.

Hi @Mac-Rae 

I think many things have been changed since 2017. I'll try to do a fresh new PR, I think it is easier to resolve the conflicts

Mac-Rae:
Hi @Abuelodelanada,

We would love to see a new PR or an update to this PR whenever you're able. If you would like assistance from the core team please feel free to reach out and we will see what we can do to assist.

In the meantime, for the cleanliness of the repo, I will close this PR over.

Thanks :+1:

Abuelodelanada:
> Hi @Abuelodelanada,
> 
> We would love to see a new PR or an update to this PR whenever you're able. If you would like assistance from the core team please feel free to reach out and we will see what we can do to assist.
> 
> In the meantime, for the cleanliness of the repo, I will close this PR over.
> 
> Thanks +1

Hi @Mac-Rae 

In the following weeks I will return with a new PR for this feature.... We have it in production and save us a lot of time!

Mac-Rae:
That's great to hear @Abuelodelanada,

Let us know if you run into any issues!

PR #3008, opened 2017-01-30
Fix #3006 Now we can have module vardefs definitions or custom data in the DOM

### LINKED ISSUE #3006: Vardefs definition and custom data in DOM
When developing some complex JS functionality is very useful to have module vardefs definitions or custom data in the DOM so you can read it from JS.

Abuelodelanada:
## Description

When developing some complex JS functionality is very useful to have module vardefs definitions or custom data in the DOM so you can read it from JS.

## Motivation and Context

getvardefsdata

getvardefsdata_dom

adddomjs

adddomjs_dom

## Types of changes

### Final checklist


Abuelodelanada:
FIX https://github.com/salesagility/SuiteCRM/issues/3006

samus-aran:
[Vardefs definition and custom data in DOM](https://trello.com/c/ikthNYNG/35-vardefs-definition-and-custom-data-in-dom)

Abuelodelanada:
Hi @samus-aran @Dillon-Brown

Like with PR #3078 with @marin-h we are very interested in trying to get this PR into SuiteCRM.... It is a feature we have in a SugarCRM CE instance that we are going to migrate to SuiteCRM 7.10.x shortly.

Do you consider it suitable for including in 7.10.x ?

if so, we can re-adapt the code of this PR in order to avoid the conflicts that has appeared.

clemente-raposo:
Hi @Abuelodelanada, @samus-aran, @Dillon-Brown,

I think having some vardef info can be usefull, as it would make some developments easier / cleaner. 
Things like checking if its required, checking the type, etc. (currently this needs to be done by looking up in the DOM or in the validations array/object). 

Anyway, two points that I would have to dig a little deeper: security and performance.

Also, I wouldn't keep this info in the data of some DOM element but it the SUGAR global. I think thats what was already proposed by @Abuelodelanada.

Usage scenarios:

I've used this kind of info for building js helpers/lib, that provides a wrapper/view around fields and some more things that sometimes are usefull.

samus-aran:
Good points @clemente-raposo !
@Abuelodelanada Have you had any thoughts on the security & performance? Do you have experience of using this for clients and it was an adverse affect. Or do we only want to include particular vardefs rather than all or as Clemente stated as just required? Is that consideration?

Jason-Dang:
Hi all,

Just to chime in with some thoughts...

May be worth having a default set of (safe) vardef keys that are output to the SUGAR globals. This would cover useful things like `required` and `type` etc. as @clemente-raposo mentions.

Additionally we could add the ability in the vardefs themselves to define extra keys (as and array) to output such as custom keys etc. as needed.

Thoughts?

Abuelodelanada:
> Hi all,
> 
> Just to chime in with some thoughts...
> 
> May be worth having a default set of (safe) vardef keys that are output to the SUGAR globals. This would cover useful things like `required` and `type` etc. as @clemente-raposo mentions.
> 
> Additionally we could add the ability in the vardefs themselves to define extra keys (as and array) to output such as custom keys etc. as needed.
> 
> Thoughts?

Hi @Jason-Dang I think all this can be implemented with this PR... We can also add some lines in the SugarView __construct() in order to have vardefs definition in SUGAR global by default.

clemente-raposo:
Hi @samus-aran,  @Abuelodelanada 

Thanks @Abuelodelanada. Agreed.

@samus-aran. I think it would be good to have this.

samus-aran:
Awaiting Travis to complete

PR #3009, opened 2017-01-30
Fixed #508 - Now we can add multiple JS files to a view in only one minified JS file

### LINKED ISSUE #508: Add multiple JS files to a view in only one minifyed JS file.

Abuelodelanada:

editviewdefs_accounts

- Each file is not minifyed.

view edit php_accounts

## Types of changes

### Final checklist


Abuelodelanada:
FIX https://github.com/salesagility/SuiteCRM/issues/508

chris001:
@Abuelodelanada Could this be adapted to override the existing default function for adding javascript to a page.

samus-aran:
[Add multiple JS files to a view in only one minifyed JS file.](https://trello.com/c/4Fnb2OhF/146-add-multiple-js-files-to-a-view-in-only-one-minifyed-js-file)

Abuelodelanada:
Hi @samus-aran 

We have made this feature for [our big implementation](https://www.gcoop.coop/gestion-de-contacto-con-el-asociado-banco-credicoop).
For us was **very** useful in performance terms. 
@chris001 suggested to override the existing default function for adding javascript to a page.

We did it this way so that both are available and to avoid generating conflicts with what was already there.

Do you think it's worth working on resolving the conflicts that appeared after we did the PR, or do you think we should follow Chris's suggestion?

There is also the possibility that this is not a feature that you want in SuiteCRM and it's fine ;-)

Abuelodelanada:
Hi @clemente-raposo 
I'll take a look at your suggestion

Abuelodelanada:
Closed because PR #6672 is a better solution.

PR #3016, opened 2017-01-31
Change the PHP version that install checks to 5.3.0 from 5.2.0.

pribeiro42:
Bumps the PHP version checked by the installer to 5.3.0

## Description
Changes the PHP version checked by the installer to 5.3.0

## Motivation and Context
SuiteCRM 7.7.x requires, at least, PHP version 5.3. 
While tinkering with the installer, I've noticed that the php_version check is still validating 5.2.x, but, further down the same file there is code that uses Namespaces (introduced in 5.3).
This change prevents this mismatch, and allows SuiteCRM to refuse anything below 5.3.0.

## Types of changes

### Final checklist

### DIFF
--- a/install.php
+++ b/install.php
@@ -52,8 +52,8 @@ if(isset($_POST['smtp_tab_selected'])) {
 }
 
 //session_destroy();
-if (version_compare(phpversion(),'5.2.0') < 0) {
-	$msg = 'Minimum PHP version required is 5.2.0.  You are using PHP version  '. phpversion();
+if (version_compare(phpversion(),'5.3.0') < 0) {
+	$msg = 'Minimum PHP version required is 5.3.0.  You are using PHP version  '. phpversion();
     die($msg);
 }
 $session_id = session_id();


horus68:
@pribeiro42 I believe you should use the Hotfix branch instead

pribeiro42:
Done!
Thanks @horus68

chris001:
Support for PHP 5.3 ended in middle 2014, 2.5 yrs ago. The current supported versions of PHP, by PHP.net, are 5.6, 7.0, and 7.1. http://php.net/supported-versions.php It's fair to require CRM servers to run the currently supported version of PHP: 5.6 minimum (and this version is on security patch status only), preferably 7.0+ (actively developed).

horus68:
Many servers has 5.3 and users can't changed it. If you do not allow a software to run 5.3 they will not work or not install.
I understand the concerns on PHP version but this is a minimum required to work.
There must be users education on the need to raise the level ... I would prefer an alert then a ban. If we have an alert on a email not setup on the admin screen why not a security alert also?

On the 7.x many software are not ready for this and I assume SuiteCRm is not ready yet (or just some of the plugins available)

chris001:
This is today's stats for the version of PHP reported on all sites running wordpress, a common php application.
```
$ curl http://api.wordpress.org/stats/php/1.0/ | jq '.'
{
  "5.2": 5.822,
  "5.3": 16.116,
  "5.4": 25.214,
  "5.5": 15.848,
  "5.6": 31.199,
  "7.0": 5.398,
  "7.1": 0.402
}
```
If not a hard requirement for PHP 5.6 minimum, at least there needs to be a big red warning alert, "PHP 5.6 Apache + FPM, runs much faster, with far less bugs, and is recommended over PHP 5.3 Apache + fastcgi which runs slow, contains bugs, and is about 2.5+ years past its expiration date."

chris001:
This warning appears when running a PHP application on the command line using the version shipped with CentOS 7, the current stable version of that popular operating system.
`Your version of PHP, 5.4.16, is affected by CVE-2013-6420 and cannot safely perform certificate validation, we strongly suggest you upgrade.`

Between when PHP 5.4.16 was released, June 2013, and now, as you can see here, http://php.net/ChangeLog-5.php#5.4.16 , many hundreds of bugs have been fixed in PHP 5.

Anyone still running PHP 5.3 is certainly having far worse computing experience with more bugs and lost time, than someone running the latest updated version of 5.6.

pribeiro42:
@chris001 my intent with this tiny fix is simply to avoid an akward situation that it can happen, according to yous message about the php versions in use there are at least 5k sites that allow the installer to run (because the check is made against php 5.2) and, at the same time it crashes with a syntax error because around line 281 there's a line with:
`catch (\Exception $e) `

Also, if you go to https://suitecrm.com/wiki/index.php/Compatibility_Matrix the minimum PHP version for the 7.7.x version is 5.3

So, while i'm totally in favour of supporting only 5.6 as absolute minimum, I've proposed the fix in line with the available documentation. Maybe we should fix documentation also....

chris001:
@pribeiro42 
Your intention is good, the installer must check the minimum version the same as Compatibility Matrix in the wiki.
At the same time, there needs to be a stronger warning added for versions of PHP less than or equal to 5.5, telling the user a few very good reasons to upgrade to the recommended version, either 5.6 or 7.0.  The best reasons are *the list of hundreds of bugs in their current version* - serious vulnerabilities, including the one which makes it impossible to safely performing certificate validation.  In today's day and age of widespread hacking, information security is a very high priority, anyone in their right mind who has valuable information to secure, would make the right choice and either upgrade their PHP, or get off the obsolete server, onto a host which is running a regularly updated current supported stable version of PHP.

pgorod:
I just noticed that the [Compatibility Matrix](https://suitecrm.com/wiki/index.php/Compatibility_Matrix) had removed PHP 5.3 from the supported versions in SuiteCRM 7.5, but reintroduced it in SuiteCRM 7.7 (this is the latest, there isn't anything in the Matrix about 7.8 yet).

I wonder if that was a typo, or maybe it was deliberate...

pribeiro42:
@pgorod, I've also noticed the change but assumed it was deliberate... If it was a typo, then this PR is moot and it should be better to check php_version against 5.5

pribeiro42:
@chris001, @horus68, @pgorod i'm closing this PR. I've "discovered" half-a-dozen more places where PHP version is checked, not counting the included libs (phpmailer, Zend stuff, etc...). 

I think i'll rewrite this into a more sane approach... thanks for all the input.

PR #3025, opened 2017-02-01
Fixed Desktop Toolbar visual overflow issue

daniel-samson:
## Description

This is a second attempt to fix #2936. Using the height of the desktop toolbar to detect if the navigation menus are wrapping to the next line, instead of the width of the desktop toolbar.

screenshot_20170201_151308

## Motivation and Context

* Fixes bug with the navigation bar in the large desktop view

## How To Test This

* Add many menu filters
* Resize browser width to around 1605px
* Access a module who name is longer the 17 characters

The navigation elements should not wrap. However this hides some navigation menus:
screenshot_20170201_152327

Menu items should not wrap to the next line
## Types of changes

### Final checklist

### DIFF
--- a/themes/SuiteP/tpls/_headerModuleList.tpl
+++ b/themes/SuiteP/tpls/_headerModuleList.tpl
@@ -287,7 +287,7 @@
                         </li>
                     {/foreach}
                 </ul>
-                {* 7.8 Hide filter menu items when the windows is too small to display them *}
+                {* 7.8 Hide filter menu items when the window is too small to display them *}
             {literal}
                 <script>
                   var windowResize = function() {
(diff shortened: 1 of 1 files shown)


chris001:
This is much improved over the bug of the wrapping menus.  However, when the screen is too narrow to show them all, if you drop some of the navigation menu filters, then you might never be able to access certain nav menu filters.  Why not detect there's not enough space and collapse all the menu filters into one hamburger menu button - the three horizontal lines - then show the one menu filter for the page you're on right now.

You'd have, in order left to right: 1. Home icon, 2. Hamburger menu icon (which stands for ALL modules), and 3. One Navigation menu filter for the page you're on right now, it can be up to 3 words - which is the current module page you're on. Some are long: Target Lists, Security Groups Management, Administration, PDF Templates, Project Templates, Opportunities, Map Address Cache.  One filter at a time should fit.

daniel-samson:
@chris001 Thank you for your peer review. Due to feedback we received in 7.7.x. Some users complained that SuiteP is moving to the mobile view too early. So I can't display the ham burger menu to early.

Even though the filter is hidden. You can still see the modules in the All menu. 

Unfortunately, this is a limitation of the theme. If there isn't space to display something in a way that complies to the design, then I am limited to what I can display. Adding an extra menu will make it more confusing for new users. It will most likely frustrate experienced users. 

Also, administrators should prioritise the filter so that the most important things are on the left. As that is the established design convention.

chris001:
Thanks @daniel-samson 
After more tests, the issue is still failing the menu in the majority of cases, wrapping still occurs. It seems the calculation for the navigation menu width, and the decision about which menu module filters to drop from the navigation menu, takes place too early, before the submenu filter is added to the navigation menu, presumably by a postponed javascript function. The result is the navigation ends up being too wide for the screen, and wraps, as you've shown in your screenshot above.

daniel-samson:
@chris001 Could you give me more details about your set up eg php version, browser, resolution, which module, etc?

I am having trouble replicating this issue after my fix.

chris001:
@daniel-samson 
Sure, the screenshot showing resolution, browser version, php version, modules, are on #3169 
In the navigation menu, there are 11 menu filters, in this instance: 
1. Sales, 
2. Mrktng, 
3. Spprt, 
4. Ops, 
5. Actvts, 
6. Collab, 
7. Map, 
8. Evnt, 
9. Rprt, 
10. Mgmt, 
11. and the system generated "All", 
12. Home icon, 
13. and the dynamically added module submenu for the module you're currently in e.g. "Opportunities".

samus-aran:
@chris001 - We have replicated the issue between the sizes of 1200px and 1250px.  However we can only seem to replicate it when we load the page less than 1200 and then manually (and slowly) resize to the css blind spot.  Once it reaches past the 1250 mark the top right menu does not wrap anymore even if you resize it down to hamburger menu and manually (and slowly) resize back up.

Can you confirm this similar action?

chris001:
@samus-aran - The bug happens every time when:
1. Set browser client width to around 1200 pixels, 
2. Might depend on the having this set of 12 module filters which happens to require a certain width to all fit onto the area of the nav menu.
3. No need to manually resize the browser window. Just login, and browse from Home into any module, ie Opportunities, 

What's causing the bug.  The app apparently is failing to call the javascript which removes module filter menus until it fits the navigation menu into the available width.  Or the javascript miscalculates and thinks it doesn't need to remove any menu filters.

Some screenshots of the bug in action:
suitecrm 1204px opp menu wrap bug

suitecrm 1207px project resource calndar menu wrap bug

suitecrm 1218px opp menu wrap bug

suitecrm 1219px proj templates menu wrap bug

suitecrm 1236px project resource calendar working menu wrap bug

daniel-samson:
@chris001 can you please refer to #3184 ?

Does that fix your issue?

chris001:
@daniel-samson Yes, this fixes the wrapping bug, when the browser viewport is less than about 1200 px, it drops down to smart phone mode - hamburger menu, plus the one menu filter for the current module.  At least it's not breaking the app now!  Thanks!

chris001:
Spoke too soon, there's still one module where this bug happens: the Create Person Form page in Campaigns module.
suitecrm wrap bug create person form

PR #3036, opened 2017-02-02
Fixed #2971 - A blank screen is displayed instead of an message.

### LINKED ISSUE #2971: A blank screen is displayed instead of an message.
#### Issue
If the URL of a deleted CRM Record is entered into the browser a blank screen is displayed instead of an message. In the previous versions a message was displayed, something like
that the user either has no access or the record is deleted

#### Expected Behavior
Displaying of an message

#### Actual Behavior
A blank screen is displayed instead of an message.

#### Steps to Reproduce

1. Create an Test-Call
2. Copy URL of Test-Call
test_call

3. Delete Test-Call record
delete_call

4. Insert the copied URL of Test-Call in browser
5. It will display a blank page instead of error message as in earlier versions
blank_site

#### Your Environment
  SuiteCRM Version used: 7.7.8
  Browser name and version: Firefox 50.1.0 (64-bit)
  Environment name and version: MySQL, PHP 7.0
  Operating System and version: Ubuntu 16.04 (64-bit)

ChangezKhan:
## Description
I have replaced 
`sugar_die($GLOBALS['app_strings']['ERROR_NO_RECORD']);` 
with 
`print('<h1 class="error">' . $GLOBALS['app_strings']['ERROR_NO_RECORD'] . '</h1>');return;`

## Motivation and Context
Well, whenever user tries to access some record which is deleted or is not available in CRM, it should show some message on screen. But it wasn't. #2971 

## How To Test This

1. Open `Call` module.
2. Create an Test-Call
3. Copy URL of Test-Call
4. Delete Test-Call record
5. Insert the copied URL of Test-Call in browser

## Types of changes

### Final checklist


pgorod:
Hhmm... this doesn't look right to me... `sugar_die` is not the same as simply printing an error and returning from whatever function you're in. This might be changing a lot of behaviors in the app.

Maybe what you need is an improved `sugar_die` function, which would let you do everything in one place and save you from all these edits. Just my 2 cents, I'm not an expert or an authority.

gunnerman1:
I am not where I can look at the `sugar_die` function but I concur with pgorod. Functions are usually called for a reason. I did find [this GitHub page](https://gist.github.com/KyleGawryluk/7418600)  which hypothetically suggests that `sugar_die` may already handle the functionality you seek and may simply require your html be passed to it.

ChangezKhan:
Hello all, kindly refer #2971, so you would get idea on why I replaced `sugar_die` function.

samus-aran:
Hi @ChangezKhan - Alotofforcing the developers to consider how to handled that particular user case per view (per module).  thinking that should be a better way at a higher level to make it constant.  Until that happens should we not just echo out the error_message to resolve the issue because the sugar_die is used at least 200+ times.

ChangezKhan:
Hello @samus-aran, yeah,  I was also thinking the same. I was also feeling uncomfortable with echoing out the error like this. So, kindly merge new solution and close this one.

PR #3060, opened 2017-02-07
Fix #450 - Modifying Address with Inline Editing

### LINKED ISSUE #450: Inline editing of address buggy
Inline editing of the address only works for the actual "address" field. It is not possible to edit the city, post code, or country this way. 

Furthermore, when inline editing the "address" field, the other fields related to the address dissapear until the page is refreshed.  

SuiteCRM 7.3.1. Tried on Chrome and Firefox on Windows.

kdjmonaghan:
## Description
Fix for Issue #450 - Previously, multiple fields (particularly, address) would not be editable with inline editing as inline editing would return a single field to edit and would also return a single field to the page without a page refresh.

This fix also fixes a parsing issue where Telephone Numbers containing a formatting similar to as follows: (area code) long number, e.g: (00000) 000000 would return with an exception error in the console and be unable to be changed with inline editing due to a failed html parse.

Additionally when further working on this change - it was discovered that the key event for pressing enter key to submit an Inline Editable field was using keypress (for press and release) and not keydown like it should have been and also did not belong in the block of code that it was in, but at the top of the document as it would stack a history if you'd open and close multiple inline editable fields.

## Motivation and Context
Frustrating user experience if they're editing information on one page and discover they cannot update information to do with addresses on the same page due to it being a multiple record request in the database. With this fix the user is able to edit addresses and telephone numbers with inline editing.

## How To Test This
In a new installation with populated data proceed with the following steps:

* Contacts -> View Contacts
* Click on a contact's name to go into detailed view.
* Begin using inline editing to update address by double clicking the address field.
* Make changes to the information.
* Save.

## Types of changes

### Final checklist

### DIFF
--- a/include/InlineEditing/InlineEditing.php
+++ b/include/InlineEditing/InlineEditing.php
@@ -56,104 +56,94 @@ function getEditFieldHTML($module, $fieldname, $aow_field, $view = 'EditView', $
     // set the filename for this control
     $file = create_cache_directory('include/InlineEditing/') . $module . $view . $alt_type . $fieldname . '.tpl';
 
+    if (!isset($vardef)) {
+        require_once($beanFiles[$beanList[$module]]);
+        $focus = new $beanList[$module];
+        $vardef = $focus->getFieldDefinition($fieldname);
+    }
+
+    $displayParams = array();
+    //$displayParams['formName'] = 'EditView';
+
+    // if this is the id relation field, then don't have a pop-up selector.
+    if ($vardef['type'] == 'relate' && $vardef['id_name'] == $vardef['name']) {
+        $vardef['type'] = 'varchar';
+    }
+
+    if (isset($vardef['precision'])) unset($vardef['precision']);
+
+    //$vardef['precision'] = $locale->getPrecedentPreference('default_currency_significant_digits', $current_user);
+
+    //TODO Fix datetimecomebo
+    //temp work around
+    if ($vardef['type'] == 'datetime') {
+        $vardef['type'] = 'datetimecombo';
+    }
+
+    // trim down textbox display
+    if ($vardef['type'] == 'text') {
+        $vardef['rows'] = 2;
+        $vardef['cols'] = 32;
+    }
+
+    // create the dropdowns for the parent type fields
+    if ($vardef['type'] == 'parent_type') {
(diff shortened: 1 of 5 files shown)


include/InlineEditing/InlineEditing.php
```diff
@@ -128,6 +128,7 @@ function getEditFieldHTML($module, $fieldname, $aow_field, $view = 'EditView', $
     $validMultipleElementNames = array(
```
mattlorimer: seems very specific, anyone could add any address

kdjmonaghan: An alternative approach to doing a type check for the input field you're turning into an inline edit that I thought would be adding a attribute to the field in the template to detect if it should be handled as an array or not. The problem with this is function inputs such as email currently fall over when they're treated as a multiple editable section.

Primary is used in the detail view while "billing_..." is used elsewhere, where there is inline Editing. Without some approach to know which fields should become inline editable, with the current layout structure, fields such as "Country" under default sections like "Contacts" turn into multi-edit field as they retrieve across multiple columns, even if they're displaying the one column to the user.

Currently looking into the attribute approach instead as I'd prefer to apply an attribute to trigger multiple elements rather than appending to an array.

kdjmonaghan: Going to look into removing "group" from email1 to remove the issue we have with emails when not using a comparison array to check for multiple fields as the "group" in email1 appears to be legacy. If so, I'll remove the comparison array and the "group" key from the email1 array in vardefs. :)

kdjmonaghan:
This needs a way to associate a  difference between a parent (which be a field we would want to load multiple inline editable fields when clicked into, example, detail view's billing address field) against ones we do not (list view, city/country etc, we wouldn't want these to return the group's inline editable fields, only that one field...)...

Checking array for the existing field name being present against current vardef's name is a possibly solution, but would also be problematic as custom fields would need to be appended to this. Adjusting the vardef to have a parent key would also be a problem to custom fields.

Currently thinking over an approach which could work universally...

kdjmonaghan:
Moved to https://github.com/salesagility/SuiteCRM/pull/3106 due to using hotfix as a base for this when I should have made another branch.

PR #3073, opened 2017-02-09
Fixed - Panels in the edit view (SuiteP)

daniel-samson:
## Description

When the user edits a record and then clicks on the head of a panel. It results in every panel being collapsed instead of the one the users selected. 

The issue is due to the JavaScript selecting every panel. So I have corrected how the JavaScript adds/removes the classes.

## Motivation and Context

* This fixes the issue

## How To Test This

* Edit a record
* Check that the panel collapse only collapses the correct panel
* Test the different configurations of tabs and panels.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [x] My code follows the code style of this project found [here](https://suitecrm.com/wiki/index.php/Coding_Standards).
- [ ] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://suitecrm.com/wiki/index.php/Contributing_to_SuiteCRM) guidelines.

PR #3076, opened 2017-02-09
Fixed - Edit View Tabs for mobile (SuiteP)

daniel-samson:
## Description

Same is as #2988 but in the edit view.

When a administrator/developer changes the layout of a detail view in studio, so that the panel types are mixed between "tab" and "panel" eg Tab, Tab, Panel, Tab, Panel. This causes and issue with the javascript for the first tab in the mobile edit view. Where as a User: after selecting a tab in the dropdown menu in the first tab, the contents of the tab displays as none / blank.

## Motivation and Context

* fixes the bug
* includes the fix in #3073 (in order to prevent merge conflicts)

## How To Test This

* In studio add panels and set the type (Tab/Panel) in a random order.
* Check that you can see those tabs in the mobile view.

## Types of changes

### Final checklist

### DIFF
--- a/themes/SuiteP/include/EditView/EditView.tpl
+++ b/themes/SuiteP/include/EditView/EditView.tpl
@@ -143,7 +143,7 @@
 
             <div class="panel panel-default">
                 <div class="panel-heading {{$panelHeadingCollapse}}">
-                    <a class="{{$collapsed}}" role="button" data-toggle="collapse" aria-expanded="false">
+                    <a class="{{$collapsed}}" role="button" data-toggle="collapse-edit" aria-expanded="false">
                         <div class="col-xs-10 col-sm-11 col-md-11">
                             {sugar_translate label='{{$label}}' module='{{$module}}'}
                         </div>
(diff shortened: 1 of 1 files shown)


samus-aran:
Works :+1: 

However I just have a separate query....

Should the top dropdwon title change to the current tab?  As on the Mobile it isn't clear where you are.

pr3076

themes/SuiteP/include/EditView/EditView.tpl
```diff
@@ -215,33 +235,34 @@ $(document).ready(function() {ldelim}
...
+                console.log(tab)
             }
         });
 
-        $('a[data-toggle="collapse"]').click(function(e){
-            var content;
+        $('a[data-toggle="collapse-edit"]').click(function(e){
+          // this == <a> link in the header of the panel:
+          //
+          // .panel
```
samus-aran: Do you need to have this explanation?

daniel-samson:
@samus-aran  We should look into this. Perhaps we need an Icon instead or just a single label called "tabs". Maybe we could add a check box next to the selected tab in the menu. There isn't much room in the mobile view. If the label is too big, it will force the action menu to overflow onto the next line.

I have also amended my PR to match your review

Issue #3131, opened 2017-02-15
Standardise Format Changes (Noticing changes for no reason/benefit)


eggsurplus:
Noticing a trend as I check changes from version to version. Code styles/preferences are overriding code that has been there for many years and is dependent on the whim of whomever is editing a specific file. Take for example /data/SugarBean.php with the recent changes in the 7.8.0 release:

https://www.dropbox.com/s/ywqm69t5u6om8xc/Screenshot%202017-02-15%2016.11.57.png?dl=0

Here the valid entry point check is moved to be low the header comment and the parenthesis are removed from the require_once.

Another case: 

https://www.dropbox.com/s/moftokujzvyq2ur/Screenshot%202017-02-15%2016.17.35.png?dl=0

The == true is removed which makes it harder to read.

There are many more like these.

While these changes are not syntactically incorrect, they add no value and in many cases goes against best practices. While it is recommended to have a styling guide, it shouldn't be enforced randomly. I suggest erring to the side of history within the project and only changing toward one's personal preferences in new code moving forward, if at all. Adopting a chameleon method to coding to blend in with past practices may be best until all code is under a unified coding standard.

eggsurplus:
The biggest issue is that there is no consistency to the changes. While it may change one way in a file, a completely differently file will make a change in the exact opposite direction. A styleguide for the SuiteCRM project could help unify the code base. There are many benefits to this.

chris001:
I agree @eggsurplus , using `php-cs-fixer` mentioned in #942 automatically formats the inconsistently formatted code.  Travis-CI should be setup to call it right here on github every time a PHP file has changes merged.

pribeiro42:
@chris001 Please take notice that PSR-2 style conflicts with [SuiteCRM Coding Standards](https://suitecrm.com/wiki/index.php/Coding_Standards) on:
  - space vs. tab question (PSR-2 mandates use of 4 spaces, SuiteCRM mandates one tab set at 4 spaces);
  - opening braces for classes (PSR-2 mandates opening brace on new line, SuiteCRM mandates opening brace on same line);

It would be nice if either SuiteCRM Coding Standars aligned with PSR-2 or the php-cs-fixer could apply the SuiteCRM Coding Standards...

samus-aran:
Hey guys.

Just would like to address your concerns with a little clarify from our side.  

[...]

Issue #942, opened 2016-02-10
Remove all PHP warning and use PHP CS


gagarine:
Hello,

I tried SuiteCRM but it was not possible to install it without having trillion of PHP warning....

"These are just warnings and can be ignored, you can turn off warnings in the php.ini settings." is not an acceptable answer.  Because I wanted a CRM quickly up and running I was using the "one click install" of my hosting. SuiteCRM was simply not running with all those warning so I switch to vtiger for the moment.

I think it make normal peoples worried and developer thinking your code is garbage. If you can't fix simple PHP warning can I trust you for other things?

For my point of view the project should use http://cs.sensiolabs.org/ (perhaps with permissive settings) and use it has a first test to reject bad pull request.

EDIT:  Warnings  was showing up with PHP 5.6 on the installer...

gagarine:
Anyway SuiteCRM should handle error using http://php.net/manual/en/function.set-error-handler.php so error can be logged in database or files instead to be displayed to the end user. Or at least error can be displayed in a nice way without breaking the website.

The other things is http://php.net/manual/en/function.error-reporting.php to be able to set the error we do care about in the backend. Without having to change php.ini.

So user will have the choice to hide and log error or to display error but in a nice "error box". And user can chose witch level of error he want to log or display.

chris001:
:+1:

pgorod:
Although most of these errors are just annoying, some of them seem to actually break the application (I don't fully understand why).

I just tried installing SuiteCRM over XAMPP on Windows and I got this exact behavior as described here:
https://suitecrm.com/forum/installation-upgrade-help/12442-warning-declaration-of-sugaremailaddress-should-be-compatible-with-sugarbean

**Note that I had E_STRICT errors turned off.**

I removed the offending line (not 1162, which is where the class ends, but near the beginning, where the Save function was declared) and SuiteCRM installed correctly.

It's a pity to lose first-time customers installing SuiteCRM because of this kind of thing. If somebody could explain to me how this kind of issue should be fixed I'd be willing to try. 

Maybe this is not just a suggestion, it seems to be a bug (a fatal one) on some circumstances.

pgorod:
There is useful information in #2697 provided by @gunnicom - that bug was closed for being a duplicate of this one, but let's keep it under our sights, there's better detailed technical information there.

Maybe @gunnicom knows enough PHP (I don't) to help me get around the error that breaks the installs on XAMPP:

[...]

PR #3133, opened 2017-02-16
Enable Travis-CS to call PHP-CS-Fixer. #942 and #3131

chris001:
Addresses #3131 and #942.
When code is merged, Travis-CI calls `PHP-CS-Fixer` which helps locate and fix warnings, and make the code adhere to PSR-1 and PSR-2 coding standards.
* http://www.php-fig.org/psr/psr-1/
* http://www.php-fig.org/psr/psr-2/

## Description

Add steps to `.travis.yml` to download and `php-cs-fixer`.

## Motivation and Context

Inconsistently formatted code make is more difficult to read and see what's happening.   PSR-1 and PSR-2 standards enforce consistently formatted code.

## How To Test This

Commit a php file to the repo.  Travis-CI will run with the settings in this commit and show code to fix.

You can remove the parameter `--dry-run`, to auto fix the code format, instead of just showing where the code needs to be fixed, to adhere to the PSR-1 and 2 standards.

## Types of changes

### Final checklist


pgorod:
Just a question, would these PHP-CS-Fixer block the Travis checks, causing PR's to fail? Or can they be configured to act just as warnings (at least initially)?

Or in other words, could we enable this without having to correct all the hundreds of problems it's going to find? And then we could work on that bit by bit later?

chris001:
The way it works is, Travis runs php-cs-fixer, and if any tests fail, it puts up warnings at the bottom of the PR page, as on this page, it found some things to take a closer look at.  So, you really should not merge until you find and fix whatever is causing that check error.

PR #3155, opened 2017-02-17
Reports fail for non-admins with related fields

eggsurplus:
If a report is made by a normal user within a group it will fail if there are related fields in the report. This is due to the table alias not being passed along to the logic to include the SecurityGroups queries.


pgorod:
Jason, if I'm not mistaken, [you're supposed to](https://suitecrm.com/wiki/index.php/Contributing_to_SuiteCRM#Make_a_Bug_Fix_Pull_Request) target the `hotfix` branch for fixes, and the `develop` branch for new features, but never directly `master`.

I hope you can find an option to just change the base branch of this PR, without having to re-do it...

modules/AOR_Reports/AOR_Report.php
```diff
@@ -1208,7 +1208,7 @@ function build_report_query_join($name, $alias, $parentAlias, SugarBean $module,
                         $join = $module->$name->getJoin($params, true);
                         $query['join'][$alias] = $join['join'];
                         if($rel_module != null) {
-                            $query['join'][$alias] .= $this->build_report_access_query($rel_module, $name);
+                            $query['join'][$alias] .= $this->build_report_access_query($rel_module, "`$alias`");
```
mattlorimer: I think backticks are only supported in MySQL

eggsurplus: I followed the same practice already used in that file elsewhere.

eggsurplus:
@pgorod Sorry, I'll remember that next time. Anything that I need to do for this specific PR?

pgorod:
@eggsurplus I see that mattlorimer has already changed the base branch on this PR, so now it's correct.

modules/AOR_Reports/AOR_Report.php
chris001: Matt, Jason, See #3165 for correct way to let the db manager handle adding backticks depending on if it's using to a database which supports them or not.

chris001: See also PR #1086 very similar to this one.

samus-aran:
@eggsurplus Is it possible to update your PR to reflect removing the backtick (didn't want to take your commit/fix away from you) :)  We have updated areas outside DBManager to remove backticks directly into dynamic queries.

eggsurplus:
Yes. Are you saying that the DBManager logic will no longer format with ticks going forward?

samus-aran:
Apologies @eggsurplus I should've been more clearer.

The DBManager deals with the backticks by a function.

`$this->db->quoteIdentifier($table_alias)`

https://github.com/salesagility/SuiteCRM/pull/3165/files#diff-aede7afff7f65db47e05554c029a1ceeR1381

eggsurplus:
I take some responsibility as well. Too much context switching! I made the change as requested.

chris001:
Right.  Microsoft SQL Server and PostgreSQL don't support ticks. MySQL allows ticks yet doesn't require them.  Standard SQL says use double quotes!
http://stackoverflow.com/questions/10573922/what-does-the-sql-standard-say-about-usage-of-backtick

samus-aran:
Thanks for that @chris001.

So at the moment outside of the MySQL DBManager there shouldn't be any backticks implemented into queries.

The function:

`$this->db->quoteIdentifier()`

provides the appropriate encapsulation.  MSSQL = [ ]  and MySQl is the backticks.  Consideration to standardise this to double quotes is noted.

However just looking it up there could be a consideration on the SQL Server for QUOTED_IDENTIFIER = OFF.

chris001:
This should be delved into deeper and determine what other database abstraction layers handle quoting ticks etc, e.g. Doctrine DB Abstraction Layer, it's sort of the state of the art.
https://github.com/doctrine/dbal
`# composer require doctrine/dbal`

PR #3210, opened 2017-03-02
tpl editor bgcolor set to #fafafa (off-white)

gymad:
## Description

Change the current background colour of the Email Templater Editor's blocks to be 'off-white' #fafafa.
(ref: SCRM-330)
-	tpl editor bgcolor set to #fafafa (off-white)

## Types of changes

### Final checklist

### DIFF
--- a/include/javascript/mozaik/jquery.mozaik.js
+++ b/include/javascript/mozaik/jquery.mozaik.js
@@ -211,7 +211,6 @@ var plgBackground = {
             namespace: false,
             ace: true,
             width: '600px',
-            bgColor: '#fafafa',
             toolPlugins: [plgBackground],
             uploadPathField: null
         }, options);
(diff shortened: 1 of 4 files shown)


include/javascript/mozaik/jquery.mozaik.js
```diff
@@ -46,7 +46,7 @@ var mozaik = {
     getMozaikInnerHTML: function(name, innertext, style) {
         //style = ($(innertext).attr('style') ? $(innertext).attr('style') + ';' : '') + style;
         //style = style.replace(/;\s*;/, ';');
-        var html = '<div class="mozaik-inner"' + (name ? ' data-name="' + name + '"' : '') + ' style="' + style + '">' + innertext + '</div>';
+        var html = '<div class="mozaik-inner"' + (name ? ' data-name="' + name + '"' : '') + ' style="background-color: #fafafa; ' + style + '">' + innertext + '</div>';
```
daniel-samson: It would be better if you implemented this as css class and then added a variable to variables.css. That way it can be configured.

gymad: Definitely not. If you add and apply the background-color only in the css it's looks like the color on the template block in the editor area because the browser render the colour on it (as normally) but it's never get inline on the element and it doesn't saved to the database so later when you will send the emails without the background colour information then it never send out via email contents and the recipients doesn't see the changed default background colour.

PR #3220, opened 2017-03-04
Add Visual Studio exclusions to .gitignore

pgorod:
Microsoft Visual Studio is a free IDE that can be used for PHP projects, including SuiteCRM.

This makes the generated project files be excluded from Git, similarly to what is already there for other IDE's.


pgorod:
Hmm there's a rogue commit in there, I'll try to remove it.

PR #3256, opened 2017-03-10
Fix #3088 - Send correct password reset email

### LINKED ISSUE #3088: Password reset sends new password only for admin account
If I try to send new password via "user detail" -> "Password reset" button, system sends "new account email" with new password for **admin** account.  

#### Issue

There is no other way to reset user password then do it manually in database

#### Expected Behavior

System should send new password for specific user account

#### Actual Behavior

Systems sends new password for admin account

#### Steps to Reproduce

1. Create new employee - user and use some your email address
2. In user detail view click on "Reset password button"
3. You receive new email in your admin mailbox with new temporary email
4.

#### Your Environment

* SuiteCRM Version used: 7.8.1.
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): All browsers
* Environment name and version (e.g. MySQL, PHP 7): PHP 5.6.29,  Ver 15.1 Distrib 10.0.29-MariaDB 
* Operating System and version (e.g Ubuntu 16.04): Debian 8 Jessie

kdjmonaghan:
## Description
Fix for #3088  - When logged in as the Admin user and you go to reset a user's password, the password sent is always the admin user's password.

## Motivation and Context
Bug raised by a user. Broken functionality due to caching wrong information, potential for locking yourself out of the admin account if the e-mail ended up in junk, etc... Had to be resolved.

## How To Test This
* Set up the email settings for mailcatcher
* Log into the admin account
* Go to the Admin Panel
* Pick User Management
* Click on the name of a user account who isn't the Admin user.
* Click Actions -> Reset Password
* Check mailcatcher to make sure it was sending out to the user's email address not the admin's email address.

Please note: When you apply the code changes, please assure the cache directory has been cleared (or the template relating to the change has been deleted) as the template will have a cached id to it and will not be corrected, even with the code fix, until the file is deleted or rebuilt.

## Types of changes

### Final checklist


markghayden:
Is there a work-around for this problem?  Kind of hard to set up users with this bug.

thanks, Mark

chris001:
You can edit the 2 files as shown here, if you want to try out the fix to this bug before it gets merged into the next version and released as an upgrade zip file.

kdjmonaghan:
Hi @markghayden,

As @chris001 - if you need an urgent fix you can take the two files in this PR and apply them to your instance. This will be merged in and rolled into an upgrade pack at some point in the future - but for an immediate fix, please take the changes and put them in your instance. :+1:

markghayden:
Thanks for the information.   I did try applying the changes and they didn’t seem to fix the problem.  We found a work-around by using the login “Forgot Password” setting.  It is good to hear that this will be put into a release at some point in the future.

It seems a little strange to run into a blocker bug in basic functionality that is 100% reproducible.  Is 7.8.2 considered a stable release?  Is there a release that is considered stable?

—Mark

kdjmonaghan:
@markghayden In the instance I was using, this fix resolved the issue with always sending the admin reset out, please explain the step by step you're doing that this doesn't resolve your issue.

Also, please try to repair and rebuild and delete your cache files (as remember, this was a cached file and will need the cache removed for the change to actually work) - hopefully these steps resolve it for you.

7.8.2 is considered a stable release.

markghayden:
Thanks for the extra information.  Just confirmed that clearing the cache made the changes work. —Mark

chris001:
@kdjmonaghan  Would be brilliant to add a method to detect a change in source code, and programatically clear the cache and trigger the quick repair/rebuild in the background.

kdjmonaghan:
I could be mistaken, but I think you're able to clear the cache with the repair option in the admin panel... detecting a code change could be quite cumbersome with the way that Sugar's underlying code works, especially as it modifies permissions quite a lot which can come across as file changes also, which would basically make any form of detection trigger, immediately, when really nothing changes other than the permissions on the file.

I realised I didn't mention about clearing the cache in the steps of resolving the issue in the opening paragraph, so I'll add it in now.

modules/Users/metadata/detailviewdefs.php
```diff
@@ -49,7 +49,9 @@
                      * To customise in an upgrade safe way, You need to create custom view instead.
                      * Then override UsersViewDetail::preDisplay().
                      */
-                    'buttons' => array (),
+                    'buttons' => array (
+                        array('customCode' => '<input title="'.translate('LBL_GENERATE_PASSWORD_BUTTON_TITLE','Users').'" class="button" LANGUAGE=javascript onclick="generatepwd(\'{$fields.id.value}\');" type="button" name="password" value="'.translate('LBL_GENERATE_PASSWORD_BUTTON_LABEL','Users').'">"')
```
samus-aran: Can not use this here as stated by the above comments:

/**
                     * Actions for users are configured in modules/Users/views/view.detail.php
                     * This is to control security access to the actions based on the user and system preferences.
                     * To customise in an upgrade safe way, You need to create custom view instead.
                     * Then override UsersViewDetail::preDisplay().
                     */

modules/Users/views/view.detail.php
```diff
@@ -125,7 +125,7 @@ function preDisplay() {
 
                     if (!$this->bean->portal_only && !$this->bean->is_group && !$this->bean->external_auth_only
                         && isset($sugar_config['passwordsetting']['SystemGeneratedPasswordON']) && $sugar_config['passwordsetting']['SystemGeneratedPasswordON']){
-                        $this->dv->defs['templateMeta']['form']['buttons'][] = array('customCode' => '<input title="'.translate('LBL_GENERATE_PASSWORD_BUTTON_TITLE','Users').'" class="button" LANGUAGE=javascript onclick="generatepwd(\''.$this->bean->id.'\');" type="button" name="password" value="'.translate('LBL_GENERATE_PASSWORD_BUTTON_LABEL','Users').'">"');
+//                        $this->dv->defs['templateMeta']['form']['buttons'][] = array('customCode' => '<input title="'.translate('LBL_GENERATE_PASSWORD_BUTTON_TITLE','Users').'" class="button" LANGUAGE=javascript onclick="generatepwd(\''.$this->bean->id.'\');" type="button" name="password" value="'.translate('LBL_GENERATE_PASSWORD_BUTTON_LABEL','Users').'">"');
```
samus-aran: Instead of this (though would be interested as to when this became  problem) update the customCode to become 'cache' friendly like so:


`$this->dv->defs['templateMeta']['form']['buttons'][] = array('customCode' => '<input title="'.translate('LBL_GENERATE_PASSWORD_BUTTON_TITLE','Users').'" class="button" LANGUAGE=javascript onclick="generatepwd(\'{$fields.id.value}\');" type="button" name="password" value="'.translate('LBL_GENERATE_PASSWORD_BUTTON_LABEL','Users').'">"');
                    }`

samus-aran:
Added a commit to address a couple of the other areas in that Detail tpl, but @kdjmonaghan to complete the review by understanding the changes :)

modules/Users/metadata/detailviewdefs.php
kdjmonaghan: In it's previous point it was caching by following the above comments - following the above comments results in the system caching the id into the template, which is why it was moved here so it was a constant check and not saved into the tpl... So we'd be back to square one.

Suggestions for a resolve without the above solution or the commented recommendation?

cumanacr:
@samus-aran Hi I just want to comment that, this is an important fix it should be on master already I was having the same problem on Version 7.8.3, I copied the files and the password reset is now working as expected.

Regards

PR #3260, opened 2017-03-10
Fix #3202 - Repair functionality of "reply-to" checkbox on profile

### LINKED ISSUE #3202: Reply-to address not saved in Profile > Email Settings
#### Issue
When you go to Profile > Email Settings, there is a *Reply-to* checkbox after each email address. However, if you check the checkbox and press *Save*, the changes are not saved.

#### Expected Behavior
When you tick the checkbox and press *Save*, the checked checkboxes should remain checked. 

#### Actual Behavior
After saving, the Reply-to checkboxes are empty again. 

#### Steps to Reproduce
1. Go to Profile
2. Scroll down to Email Settings
3. Tick one or more Reply-to checkboxes
4. Press Save

#### Context
We want our employees to be able to choose from a set of Reply-to email addresses when sending emails to clients. Right now, you can only use your primary email address to send email (your other email addresses are not shown in the from dropdown in the compose email screen). 

#### Your Environment
* SuiteCRM Version used: 7.8.2
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Chrome Version 55.0.2883.87 (64-bit)
* Environment name and version (e.g. MySQL, PHP 7): Ubuntu 14.04, PHP 5.5.9, MySQL 5.5.54

kdjmonaghan:
## Description
Fix for Issue #3202 - When selecting the reply all check inside of the Profile tab and pressing save, the checkbox wasn't being saved as checked. This also applied for any record that could had the checkbox checked, but then the user wanted to uncheck the box.

Inclusion of double quote in line 302, appending the name attribute, was breaking the record to not actually handle the form submission properly (due to the association key being wrong, most likely due to the double quote ending the attribute (escaping the attribute, entirely) early) and append or update the record to the system as the result selected by the user.

## Motivation and Context
Issue raised by community member. Broken functionality that should be corrected.

## How To Test This
* Login to SuiteCRM
* Select Profile from the top-right dropdown
* Scroll to the bottom of the profile page.
* Switch the radio (or apply to the radios) for the reply to e-mail.
* Save the changes to the profile
* Go back to the profile to see if the e-mail has been set as the reply e-mail.
* Check the database tables to see if the flag has been set.

Table Structure for checking:

The email_addresses table where email_address is a match to the e-mail you're trying to set
taking it's id from the id column and then using the id to search against the email_addr_bean_rel where email_address_id is a match to the id you took from email_addresses

## Types of changes

### Final checklist

### DIFF
--- a/jssource/src_files/include/SugarEmailAddress/SugarEmailAddress.js
+++ b/jssource/src_files/include/SugarEmailAddress/SugarEmailAddress.js
@@ -299,7 +299,7 @@
       // Reply to checkbox
       var replyToCheckbox = lineContainer.find('input#email-address-reply-to-flag');
       if (replyToCheckbox.length == 1) {
-        replyToCheckbox.attr('name', this.module + _eaw.id + '"emailAddressReplyToFlag');
+        replyToCheckbox.attr('name', this.module + _eaw.id + 'emailAddressReplyToFlag');
         replyToCheckbox.attr('id', this.module + _eaw.id + 'emailAddressReplyToFlag' + _eaw.totalEmailAddresses);
         replyToCheckbox.attr('value', this.module + _eaw.id + 'emailAddress' + _eaw.totalEmailAddresses);
         replyToCheckbox.attr('tabindex', tabIndexCount);
--- a/jssource/src_files/include/SugarEmailAddress/SugarEmailAddress.js
+++ b/jssource/src_files/include/SugarEmailAddress/SugarEmailAddress.js
@@ -315,48 +315,6 @@
         removeButton.prop('disabled', true);
       }
 
-      // Reply to checkbox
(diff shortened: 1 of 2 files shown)


jssource/src_files/include/SugarEmailAddress/SugarEmailAddress.js
```diff
@@ -298,15 +298,34 @@
...
+        replyToCheckbox.prop("checked", true);
+      }
+
+
+      // Prevent users from removing their primary email address
+      if (this.module == 'Users' &&  replyToCheckbox.attr("checked")) {
+        removeButton.prop('disabled', true);
+      }
+
+      // Reply to checkbox
```
samus-aran: Remove commented out code - drop the habit of being unsure to commit to your code - if it needs reverting the history will provide that :)

PR #3268, opened 2017-03-13
Fixes #2787 by checking allowed cron users

### LINKED ISSUE #2787: Robustness Big Issue 1: cron shouldn't be allowed to run as root
This could also be titled "Solve permissions problems once and for all" :-)

After recent tests reliably reproducing permissions degradation problems in a controlled testing environment, and finally finding a proper explanation for them, I venture saying that **we are POSSIBLY looking at something that will solve ALL permissions problems**, or at least we're **SURELY looking at something that will solve a LOT of permissions problems** that lots of people experience.

We're talking about 5 or 6 cases of this problem every week on the forums, not counting all the other instances where people don't ask, because they already know how to do it. I know I used to fix permissions before and after every upgrade, and some people run cron scripts to fix them every hour... let's put an end to that, shall we?

You can read about the tests here: https://github.com/salesagility/SuiteCRM/issues/1688

### What is permissions degradation in SuiteCRM? 

It's when file permissions are set correctly, but at a later moment are changed in a way that causes problems. To cause problems, normally it's because files get harder permissions (either ownership by a more secured user like `root`, or losing some mode of access like `writeable`) and then when SuiteCRM needs to read or write those same files, it gets access denied errors. These will break some functionality, probably at an unexpected point, resulting in poor logging/messages to the user and admins.

### Why does it happen?

It certainly doesn't happen just "because", for no reason at all. It happens because some process running on the server does the changes; it could be a third party process messing with the files, but that's really strange. Why would any process just randomly mess with another app's files? 

It could be (and it is) the SuiteCRM application itself. If it's not the web application, then it would have to be the Scheduler jobs (and it is).

### How should Scheduler jobs be configured?

They should run under the same user that is used for the Web Server. For example, on a typical Ubuntu install, this would be `www-data`. On Bitnami installs it would be `bitnami`, etc.

### Does anybody configure Schedulers to run as root? Why would they?

I did : - )

I didn't know I could edit the crontab of a user that does not allow logins, like `www-data` (with  `crontab -e -u www-data`). Initially I didn't even consider that there was a separate cron for each user. So I just edited crontab as root and made cron.php always run as root. And the jobs worked, so I just forgot about it... Not good practice, I understand now. But Linux actually _induces_ ignorant admins to do just that.

Also, the fact is that there is nothing in the wiki documentation about which user to use. The advice commonly given on the forums doesn't mention it. 'SuiteCRM for Developers' doesn't mention it. Finally, the line suggested by the installer doesn't mention it. 

From asking in the forums in these past weeks, I know other people did it (just one example: https://suitecrm.com/forum/installation-upgrade-help/12155-import-data-from-a-sugarcrm-6-5-20-db-to-suitecrm-7-7-8).

I suspect LOTS of other people did it.

### So, how exactly do Schedulers break permissions when running as root?

They write files, and these will get ownership by root. Then other processes like the SuiteCRM app running from the web server won't be able to access them. **The most relevant case has to do with caching**: just by accessing certain common data structures, vardefs and other elements, the system will generate cached files. The odds of this happening increase during installations, upgrades, and Quick Repair and Rebuilds, when caches are flushed and get rebuilt from scratch, so whatever the cron jobs touch first, before the rest of the web app, will be owned by root.

### How to fix it?

**For individual installs**, configure cron correctly.

**For the project**, to really increase the real-world stability and robustness of the app, and to remove the bad fame of being an app that breaks inexplicably, a combination of these things:

1. Specify in the **wiki installation instructions** that the crontab must be configured to run under the same user as the web server, recommending usage of crontab -e -u TheWebServerUserName.

2. Inside the app, make the **same message appear in the installer scripts** (install/confirmSettings.php, install/ready.php). Tell the user specifically the user name to use.

3. Make the **same message appear in the Admin-->Schedulers page** (modules/Schedulers/Scheduler.php) 

4. **Make cron.php check under which user it is running**. If running as root, log a FATAL error `cron.php is running as root. This will cause serious errors and must be corrected.`. Then `die();`.

5. Have the SuiteCRM installer **save both user and group in config.php** after install. These might be valuable clues for cron.php to use to compare to its current user.

### Other users?

As a final note, strictly speaking, these problems could arise with users different from root, basically with any user that is not the web server user. But I expect root is the culprit 99% of the time.

I've thought about this in the past weeks and it's not easy to get a more robust solution. When running from within `cron.php` (from CLI) there's no easy way to find the web server user. Maybe solution number 5 above would be a good place to start. Possibly in the future we should be moving to a situation where config.php includes _a list_ of possible users under which SuiteCRM and cron will run. All others would cause the app to exit. This would bring the issue of consistent user configuration under a wiser attention from admins.

But for now, let's start with just checking for `root`, to keep things simple.

### High Priority?

I know everybody thinks 'their' bugs are High Priority. But this on definitely is (and it's not 'my' bug, my system is fixed now and _stays_ fixed!). 

**This is High Priority on all counts**: number of people affected, long-standing bug, critically breaks affected systems, very hard to diagnose, affects SuiteCRM public fame, cripples adoption by new users, consumes forum resources greatly and needlessly, etc.

pgorod:
This PR (finally) fixes #2787.

### Why we should care
I see this in the forums almost every week. When releases come out and people start upgrading, this problem shows several times a week. I would say all the years we've been suffering from this, and the hours of work consumed by admins and support, all add up to a huge and senseless pain.

### Aren't educational efforts enough?
Sure, people should know how to configure things correctly. However, we're not providing instructions for this in the wiki, there are tons of wrong instructions out there on the Internet, and our installer misleads people.
No, educational efforts are not enough - they are _what we have been failing at_ for all these years. No matter how often I explain all this in the forums, it comes up again and again.

### Why it's not that simple
Because configurations can be very different from system to system, and we need to accomodate them all. Because it's hard to know from within cron, which user your other scripts (the web app) run under. So this information needs to be recorded in one place, then read in another.

### What it does, and how it impacts people
- It does not affect Windows at all. This is a problem of permissions degradation caused by cron misconfiguration, it was diagnosed in Linux, and the fix affects Linux only (in case you're wondering, `exec('whoami');` works on Windows)
- new installations on Linux will get proper advice as to how to setup cron correctly for their specific case.
- On upgrades, it does not affect people with standardly configured Linux systems (web server user = cron user). These should just go on working.
- On upgrades, people with non-standard Linux configurations will have to review their config.php to keep cron working. They will get explanative log entries and receive instructions about this in the Release notes.

### Technical details
- **on install time**, add an entry to an array in `config.php` called `allowed_cron_users` containing the current web server user. 
- **on install time**, provide correct instructions for cron configuration with that specific user's crontab.
- **on upgrade time**, add the same entry to the same array, if it's not there yet (for people coming from versions of SuiteCRM before this mechanism was introduced)
- for these automatic additions of users to the array, **`root` is treated as a special case**, to discourage its use and to force manual intervention from an admin
- **on Schedulers screen**, show specific correct configuration instructions also
- **when `cron.php` runs**, check if the current user is in the array, and exit if not, leaving appropriate log message 

### Security
Any attack that compromises `cron.php` _or any of the Scheduler scripts that it runs_ will be much more severe if cron is running as `root`, something that happens in a lot of installations out there. We're doing people a favor by nudging them towards a safer configuration.

### Educational efforts

As I said, these are not enough, but they _are_ needed.

#### Suggested text snippet to include in Release Notes:

> New mechanism to check which user cron runs under, to avoid permissions degradation issues. IMPORTANT: after upgrade, check your logs and confirm that your Schedulers are running.

#### Suggested addition to the [wiki installation instructions](https://suitecrm.com/wiki/index.php/Installation):

Insert the following section just before the Title "Upgrading SuiteCRM". This will make it the final section of the previous title ("Installing SuiteCRM").

> "Post-Installation steps"
> 
> To complete installation, the Scheduler jobs need to be enabled. This is done by asking your operating system to run a script called cron.php every minute. This, in turn, manages all SuiteCRM jobs that are supposed to run at scheduled times, supporting functionalities like search indexing, workflows, email notifications, database maintenance, etc.
> 
> In Windows this is setup on the system's "Task Scheduler"; in Linux and iOS, it is setup in crontab. During the installation you will receive detailed instructions with the specific commands you need to run to setup these jobs. After installation, you can still access these instructions in Admin area, under option "Schedulers", at the bottom of the screen.
> 
> To make sure your Scheduler jobs don't create permissions problems, the standard practice is to run them under the same user that your Web server runs under. For different configurations, the allowed_cron_users array in config.php must be adjusted.
> 

#### In the forums

Of course some people will get into trouble with this when upgrading and show up in the forums saying "my cron jobs aren't working", and we'll simply answer: complete your upgrade by following instructions, here is the link to configure cron correctly. We have an admin's attention, he scheduled downtime for this, he's probably even testing on a clone of his production system.

Compare this with what we have now: people come up weeks or months after creating the problem, with complains from end-users and managers, with random symptoms and sometimes varied and extensive system rot in production. After several diagnosis attempts, they are told to chmod their permissions, and the cycle of permissions' degradation starts over, to need fixing again later on...

## How To Test This
I tested these situations:
1. Installation (check `config.php` gets a correct `cron_allowed_users` entry).
2. Upgrade (to check the same as above. I had to create a tweaked upgrade package with the new files).
3. Running cron, both with an allowed user and a non-allowed user.
4. You can also try some variations on the user (root or non-root), etc

## Types of changes

- [ X ] Bug fix (non-breaking change which fixes an issue)
- [ X ] New feature (non-breaking change which adds functionality)
- [ X ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [ X ] **My change requires a change to the documentation (text suggestions provided).**
- [ X ] I have read the [**How to Contribute**](https://suitecrm.com/wiki/index.php
/Contributing_to_SuiteCRM) guidelines.


horus68:
Language strings added to translations packs for versions 7.8.2.3
Just ping me if there are any language changes in this PR

pgorod:
@horus68 you add the strings _before_ the PR's get merged? Doesn't that make things complicated for you when the merges are delayed for a long time?

horus68:
No... if I just include the new strings and let the old ones stay around.
This alow to test the PR with translation but to work on previous versions.
This also gives translators the time to translate. Then when a PR is merged its ready to be used by all!

chris001:
I have a similar fix for this issue, testing it now.

anothermouse:
Just looking at some of the shared hosting environments that don't directly allow shell commands, or run in a chrooted environment. Perhaps:
     posix_getpwuid(posix_geteuid())['name'];
is a better alternative to exec('whoami'); ?

pgorod:
People on those systems could still configure the user manually in `config.php`, but I agree it would be better if we could handle things automatically everywhere. I'd like to avoid diverging too much from Windows, though. Anything in the code related to permissions already diverges _a lot_. We're basically always assuming Windows is entirely permissive which, of course, is not true everywhere, and I believe will tend to be less so in the future.

Anyway, thanks @anothermouse because your comment is obviously something I need to look at. Do you know of any place where I can test those cases of shared hostings with those limitations? I'd like to see exactly how `exec` fails and how I could test for that.

pgorod:
Ok, I got it - if I add `exec` to `disabled_functions` in my `php.ini` I can emulate does scenarios. Testing now.

pgorod:
I added some more logic, it's probably overkill, but it might come in useful in some weird hosting out there. Anyway it now works on systems with `exec` disabled.

samus-aran:
[Robustness Big Issue 1: cron shouldn't be allowed to run as root](https://trello.com/c/wTto2LbF/48-robustness-big-issue-1-cron-shouldn-t-be-allowed-to-run-as-root)

install/install_utils.php
```diff
@@ -2169,3 +2169,66 @@ function enableInsideViewConnector()
     // $mapping is brought in from the mapping.php file above
     $source->saveMappingHook($mapping);
 }
+
+/**
+ * Gets the username of the user under which the PHP script is currently running
+ * Notes:
+ * - works on Windows and Linux, tries a variety of methods to accomodate different systems and hosting restrictions
```
mattlorimer: typo accommodate

install/install_utils.php
```diff
@@ -2169,3 +2169,66 @@ function enableInsideViewConnector()
...
+ * - on Windows, return full username in form DOMAIN\USER
+ * - returns empty string if failed
+*/
+
+function getRunningUser()
+{
+    // works on Windows and Linux, but might return null on systems that include exec in disabled_functions in php.ini (typical in shared hosting)
```
mattlorimer: exceeds 120 characters, break into 2 lines

install/install_utils.php
```diff
@@ -2169,3 +2169,66 @@ function enableInsideViewConnector()
...
+    return ($runningUser == null) ? "" : $runningUser;
+}
+
+/**
+ * Adds a username to the allowed_cron_users array in config.php
+ * @param string $addUser the name of the user to add [usually obtained with getRunningUser()]
+ * Notes:
+ * - this is Linux only, does nothing on Windows
+ * - does not repeat the user if he is already there
+ * - creates the sub-array if previoulsy unexisting
```
mattlorimer: typo previously

install/install_utils.php
```diff
@@ -2169,3 +2169,66 @@ function enableInsideViewConnector()
...
+    }
+    if (!array_key_exists('allowed_cron_users', $sugar_config['cron'])) {
+        $sugar_config['cron']['allowed_cron_users'] = array();
+    }
+    if (!in_array($addUser, $sugar_config['cron']['allowed_cron_users'])) {
+        if ($addUser == 'root') {
+            $addUser = 'root_REMOVE_THIS_NOTICE_IF_YOU_REALLY_WANT_TO_ALLOW_ROOT';
+            if (!in_array($addUser, $sugar_config['cron']['allowed_cron_users'])) {
+                $sugar_config['cron']['allowed_cron_users'][] = $addUser;
+                $GLOBALS['log']->error("You're using 'root' as the web-server user. This should be avoided for security reasons. Review allowed_cron_users configuration in config.php.");
```
mattlorimer: exceeds 120 characters, break into 2 lines

install/install_utils.php
```diff
@@ -2169,3 +2169,66 @@ function enableInsideViewConnector()
...
+    // works on Windows and Linux, but might return null on systems that include exec in disabled_functions in php.ini (typical in shared hosting)
+    $runningUser = exec('whoami');
+
+    if ($runningUser == null) {  // matches null, false and ""
+        if (is_windows())
+            $runningUser = getenv('USERDOMAIN').'\\'.getenv('USERNAME');
+        else
+            $runningUser = posix_getpwuid(posix_geteuid())['name'];
+    }
+    return ($runningUser == null) ? "" : $runningUser;
```
mattlorimer: use single quotes

install/install_utils.php
```diff
@@ -2169,3 +2169,66 @@ function enableInsideViewConnector()
...
+ * - does not repeat the user if he is already there
+ * - creates the sub-array if previoulsy unexisting
+ * - special treatment for user 'root' to require manual intervention from an admin to allow
+ */
+
+function addCronAllowedUser(string $addUser)
+{
+    global $sugar_config;
+
+    if (is_windows() || !isset($sugar_config)|| !isset($addUser) || ($addUser == "")) {
```
mattlorimer: use single quotes

install/install_utils.php
```diff
@@ -2169,3 +2169,66 @@ function enableInsideViewConnector()
...
+ * Adds a username to the allowed_cron_users array in config.php
+ * @param string $addUser the name of the user to add [usually obtained with getRunningUser()]
+ * Notes:
+ * - this is Linux only, does nothing on Windows
+ * - does not repeat the user if he is already there
+ * - creates the sub-array if previoulsy unexisting
+ * - special treatment for user 'root' to require manual intervention from an admin to allow
+ */
+
+function addCronAllowedUser(string $addUser)
```
mattlorimer: scalar type hints e.g string, are not supported in php < 7

install/install_utils.php
```diff
@@ -2169,3 +2169,66 @@ function enableInsideViewConnector()
...
+ * - returns empty string if failed
+*/
+
+function getRunningUser()
+{
+    // works on Windows and Linux, but might return null on systems that include exec in disabled_functions in php.ini (typical in shared hosting)
+    $runningUser = exec('whoami');
+
+    if ($runningUser == null) {  // matches null, false and ""
+        if (is_windows())
```
anothermouse: Can we change line 2184 to use the getRunningUser() function for consistency?

pgorod: @anothermouse this is _inside_ getRunningUser, so that would cause an endless recursion. This function uses both methods, whoami and the one you suggested, it's probably overkill but my idea was to cover as many different scenarios as possible (where one function works, but not the other, and vice-versa).

pgorod:
@mattlorimer please confirm if I understood you correctly: I moved both functions, `getRunningUser` and `addCronAllowedUser` to `include/utils.php`.

install/install_utils.php
anothermouse: Sorry - my mistake....I didn't examine it closely enough as I was only looking at the diffs. I was thinking about the uniformity of using:
posix_getpwuid(posix_geteuid())['name'];
and avoiding the use of exec('whoami')

samus-aran:
@pgorod - Yes that is what he meant so we can use it else where when required.

Couple of things on my side and its regarding testing which I have only really got round to doing. Anyway - when installing.  The amendments to actually add the allowed_cron_users to the config does not execute because complete_install.php never gets called.  The last step within the SuiteCRM install process is 'performSetup.php' so I would suggest putting your functionality in there.  We have two places but I recommend the latter - I'm trying to find the best place myself.

Now I was going to look at this today and push up to your branch but I wanted to make sure I wasn't missing something extremely obvious as to why my allowed_cron_users were not appearing in my config.php

Reason why I noticed it was when I was running the cron, I expected an error as there was not anything in config.php but it didn't as the check expects it to be there... in which case we should put an error log to notify that there is not allowed_cron_users set up. https://github.com/Accendere/SuiteCRM/blob/manage-web-server-users/cron.php#L56

Another minor thing is to update to where the file to be pointing. https://github.com/Accendere/SuiteCRM/blob/manage-web-server-users/cron.php#L53

pgorod:
Hi @samus-aran thanks for the tips.

I fixed the `require` in `cron.php`.

I added a warning in `cron.php` when no `allowed_cron_users` entry is found in `config.php`.

Regarding the code in `install.php`, I did test that (quite a while ago - maybe I need to try it again) and I believe the code ran. Maybe `complete_setup.php` is not getting required as a file, but it seems the ` case 'complete_install.php':` section in `install.php` is running. 

It just occurred to me - you don't happen to be using a silent install method? I didn't test that (and probably should). Anyway the place we choose to put that code in, should work for both kinds of installs...

samus-aran:
Thanks @pgorod - I would recommend again to test it as I've just cloned your repo, checkout manage-web-server-users and ran install.php (via the browser, not silent install) and allowed_cron_users are not being inserted into config.php.  My only thinking is that this has changed since you tested (due to our delay on picking this up) but I think putting it in performSetup would actually be more beneficial as complete_install is pretty redundant (it only cleared the session and redirected you to login page which performSetup does).

My steps were:
- clone https://github.com/Accendere/SuiteCRM
- checkout manage-web-server-users
- chown www-data:www-data -R .
- chmod 775 -R . (yeah abit lazy, but for quickness)
- navigated to url/install.php and ran through installation wizard, completed and directed to login page.
- open newly created config.php - search for allowed_cron_users (not found).

pgorod:
Ok I did the changes, I agree it looks better this way, and probably gets this working also for silent installs.

I can't test right now (no access to my installation tests VM on a Sunday) but I hope to do so soon. Anyway I pushed the changes so that if you get a chance, you can test it. Thanks.

samus-aran:
Thanks @pgorod - Tested that and it works - Superduper.  I'll do a couple of other tests but overall very happy with that.

pgorod:
Great!  :-)

The test that gave me more work was the upgrade. I had to install an earlier version (7.8.1), then get an upgrade package for 7.8.2 and tweak it manually, inserting my files into the zip.

I was afraid that my new file `end.php` would not be in place at the end _of that same upgrade_, which was required. But thankfully it was copied somewhere in the middle of the upgrade, and was in place when the `end.php` was called.

pgorod:
Thanks @samus-aran!

Don't forget to copy-paste the texts above for the Release Notes and the Wiki (changing whatever you feel needs changing).

PR #3267, opened 2017-03-13
Optimistic Locking not working in some cases

gunnicom:
In some cases sugar retrieves more object of the edited type. This leads to that the optimistic locking id is not the one thats in the edit view, but some other (maybe from relation).
So if someone saves the same record in a different session, the one that saves later simply overwrites the first changes, and does not show the dialog for conflicting saves.
This change makes sure the id from the request is stored in the session, not the id from the last fetched record that maybe from a related bean.

## Motivation and Context
On conflicting saves under specific circumstances the conflict dialog is not shown.

## How To Test This
Not quite sure about the exact requirements for this bug. I assume it is some custom relationships, that relate to the same module, or to an other module and then to the same module again.

Expected:
When saving a record and someone else changed the record within the same time a dialog about that conflicting change should show.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist
- [x] My code follows the code style of this project found [here](https://suitecrm.com/wiki/index.php/Coding_Standards).
- [ ] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://suitecrm.com/wiki/index.php/Contributing_to_SuiteCRM) guidelines.

PR #3276, opened 2017-03-14
Optimistic Locking not working in some cases

gunnicom:
In some cases sugar retrieves more object of the edited type. This leads to that the optimistic locking id is not the one thats in the edit view, but some other (maybe from relation).
So if someone saves the same record in a different session, the one that saves later simply overwrites the first changes, and does not show the dialog for conflicting saves.
This change makes sure the id from the request is stored in the session, not the id from the last fetched record that maybe from a related bean.

## Motivation and Context
On conflicting saves under specific circumstances the conflict dialog is not shown.

## How To Test This
Not quite sure about the exact requirements for this bug. I assume it is some custom relationships, that relate to the same module, or to an other module and then to the same module again.

Expected:
When saving a record and someone else changed the record within the same time a dialog about that conflicting change should show.

## Types of changes

### Final checklist

### DIFF
--- a/data/SugarBean.php
+++ b/data/SugarBean.php
@@ -3863,7 +3863,7 @@ class SugarBean
 
         global $module, $action;
         //Just to get optimistic locking working for this release
-        if ($this->optimistic_lock && $module == $this->module_dir && $action == 'EditView' && $id == $_REQUEST['record']) {
+        if ($this->optimistic_lock && $module == $this->module_dir && $action == 'EditView' && isset($_REQUEST["record"]) && $id == $_REQUEST['record']) {
             $_SESSION['o_lock_id'] = $id;
             $_SESSION['o_lock_dm'] = $this->date_modified;
             $_SESSION['o_lock_on'] = $this->object_name;
--- a/data/SugarBean.php
+++ b/data/SugarBean.php
@@ -3863,7 +3863,7 @@ class SugarBean
 
         global $module, $action;
         //Just to get optimistic locking working for this release
-        if ($this->optimistic_lock && $module == $this->module_dir && $action == 'EditView') {
+        if ($this->optimistic_lock && $module == $this->module_dir && $action == 'EditView' && $id == $_REQUEST['record']) {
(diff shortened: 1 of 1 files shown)


gunnicom:
Replaces https://github.com/salesagility/SuiteCRM/pull/3267 (wrong branch)

chris001:
@gunnicom FYI you can edit the branch, it's a lot faster, no need to make a new PR.

gunnicom:
@chris001 I did that in the other PR, but that pulled in some completely unrelated commits.

chris001:
@gunnicom   In case you're interested in exploring some good phpunit test examples for locking, with step by step how to, here you go:
https://github.com/doctrine/doctrine2/tree/master/tests

gunnicom:
@chris001 Thx, But at the moment i dont want to put in too much work, and maybe then in a year or two the PR gets deprecated. If salesagility decreases the queue of issues and PR maybe i will do more.

chris001:
@gunnicom Understood.  PR's need to get reviewed and added faster than they currently are being added, which requires more time dedicated to review.  It's a circular situation, which can be solved by automatic tests with phpunit, because when the phpunit tests pass, you can be sure there is no new bug introduced due by the new PR. The greatest part is the phpunit tests you write them once at the same time you code the bugfix or new feature, they work forever, and they only take a few seconds for the system to run, much faster than a human tester.

data/SugarBean.php
```diff
@@ -3863,7 +3863,7 @@ public function retrieve($id = -1, $encode = true, $deleted = true)
 
         global $module, $action;
         //Just to get optimistic locking working for this release
-        if ($this->optimistic_lock && $module == $this->module_dir && $action == 'EditView') {
+        if ($this->optimistic_lock && $module == $this->module_dir && $action == 'EditView' && $id == $_REQUEST['record']) {
```
gymad: It's ok if request has a 'record' but can you please also check for the requested key exists? (for e.g isset($_REQUEST['record']) && $id && ....)

PR #3284, opened 2017-03-16
Fix #3238 - Hide save and continue button on contacts subpanel

### LINKED ISSUE #3238: Save and continue, adding customers
When adding contacts on the account page, there is a button save and continue.
This button does nothing, so only saving the customer does something.

#### Expected Behavior
I expect it to add the contact and continue to a new field, so I can add more contacts.

#### Actual Behavior
Nothing happens.

#### Context
Low priority

#### Your Environment
* SuiteCRM Version used: Versie 7.8.2 Sugar / Versie 6.5.24 (Maken 509)
* Browser name and version: Chrome Version 56.0.2924.87 (64-bit)
* Environment name and version: MariaDB 10.1.21-r0, php5-5.6.30-r0
* Operating System and version: Alpine Linux 3.5.2

kdjmonaghan:
## Description
A fix provided by @imtg-suitecrm in relation to Issue #3238 - Sub panel when you are in quick create shows Save and Continue button at the bottom when it should not as this functionality does nothing currently.

## Motivation and Context

UI indicates functionality that does not belong/does not work, removing this from the view would enhance user experience and not frustrate users when it does not work as they think it would.

## How To Test This
* Open Contacts module
* Go into the Edit View on a contact
* Scroll down to the sub panel
* Create a new record
* Save and Continue button is removed from the bottom section of the subpanel

## Types of changes

### Final checklist

### DIFF
--- a/themes/SuiteP/include/EditView/actions_buttons.tpl
+++ b/themes/SuiteP/include/EditView/actions_buttons.tpl
@@ -7,7 +7,7 @@
         {{sugar_button module="$module" id="SAVE" view="$view" form_id="$form_id"}}
         {{sugar_button module="$module" id="CANCEL" view="$view" form_id="$form_id" }}
     {{/if}}
-    {if $showVCRControl}
+    {if $showVCRControl && $view != 'QuickCreate'}
             <button type="button" id="save_and_continue" class="button saveAndContinue" title="{$app_strings.LBL_SAVE_AND_CONTINUE}" onClick="SUGAR.saveAndContinue(this);">
                 {$APP.LBL_SAVE_AND_CONTINUE}
             </button>


themes/SuiteP/include/EditView/actions_buttons.tpl
```diff
@@ -7,7 +7,7 @@
         {{sugar_button module="$module" id="SAVE" view="$view" form_id="$form_id"}}
         {{sugar_button module="$module" id="CANCEL" view="$view" form_id="$form_id" }}
     {{/if}}
-    {if $showVCRControl}
```
samus-aran: We shouldn't be putting in template fixes that is very specific (don't want to use the term 'hack').  Need to have this resolved at a higher level i.e. before the template is parsed.

themes/SuiteP/include/EditView/actions_buttons.tpl
```diff
@@ -7,7 +7,7 @@
...
+    {if $showVCRControl && $view != 'QuickCreate'}
```
samus-aran: (original comment on wrong side)
We shouldn't be fixing template issues with very specific use cases (don't want to use the term 'hack' here).  This should be resolved at a higher level i.e. before the template is parsed.

samus-aran:
For some reason I'm unable to push to your branch -  but essentially in inclue/EditView/SubpanelQuickCreate.php

`$this->ev = $this->getEditView();
		$this->ev->view = $this->viewType;
		$this->ev->showVCRControl = false;
		$this->ev->ss = new Sugar_Smarty();`

At line 86 set the view's showVCRControl to false here.

kdjmonaghan:
@samus-aran check #3323 as this PR was opened using the community member's branch.

Issue #3375, opened 2017-04-08
Report. The errors caused by the conditions of type "field = ''"


halterpc:
#### Issue

At AOR_Report, when exist a condition "Equal_To" and empty, the system add extra condition " OR XXX is null", but this condition has no parenthesis.

#### Expected Behavior

When you live with other conditions, the results of the report are unexpected

#### Actual Behavior

Correct results

#### Possible Fix

Ubicación: .../modules/AOR_Reports/AOR_Report.php
Clase, funcion: line 1550
Code:
Actual cde:
```
                    if($condition->value_type == 'Value' && !$condition->value && $condition->operator == 'Equal_To') {
                        $value = "{$value} OR {$field} IS NULL";
                    }

```

New proposed code:
```
                    if($condition->value_type == 'Value' && !$condition->value && $condition->operator == 'Equal_To') {
                        /* PPW - CODe - Modifiación CORE */
                        // $value = "{$value} OR {$field} IS NULL";
                        $value = "{$value} OR {$field} IS NULL)";
                        $field = "(".$field;
                    }
```
#### Steps to Reproduce

1. Create report and apply condition "equal" and empty
2. Create others conditions

#### Context

It´s a bug

#### Your Environment

* SuiteCRM Version used: 7.8.2
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): All
* Environment name and version (e.g. MySQL, PHP 7): Mysql/MariaDB, PHP7
* Operating System and version (e.g Ubuntu 16.04): Ubuntu 16.04

Dillon-Brown:
Hi @halterpc can you create a PR for this issue to salesagility/suitecrm:hotfix the referenced issue is a commit to your own fork.

halterpc:
I will resolve problem to PR, monday 17

Dillon-Brown:
Issue should now be resolved.

PR #3421, opened 2017-04-19
Report. The errors caused by the conditions of type "field = ''"

halterpc:
At AOR_Report, when exist a condition "Equal_To" and empty, the system
add extra condition " OR XXX is null", but this condition has no
parenthesis.


Dillon-Brown:
Hi @halterpc try and stick to this format for your commit messages so that the PR and the issues can be linked together and then closed once merged. Thanks for the contribution!

Fixed #issue - Issue title

So in this case you would want: Fixed #3375 - Report. The errors caused by the conditions of type "field = ''"

PR #3465, opened 2017-04-27
fix: validation of the Email action on Workflows

mrbarletta:
When the Workflow action is 'email' and it is related to the Bean processed, the field related might be empty. 

When this happens the workflows fails with a fatal error due to $linkedBean->emailAddress->getPrimaryAddress  call.   All the scheduled workflows won't run after a failure like this.

 $linkedBeans is present but with an empty array so this PR will check if the value is not empty before trying to access its properties.

## Motivation and Context

 When the Workflow action is 'email' and it is related to the Bean processed, the field related might be empty. 

When this happens the workflows fails with a fatal error due to $linkedBean->emailAddress->getPrimaryAddress  call.   All the scheduled workflows won't run after a failure like this.
## How To Test This

1. Create a workflow tied to Opportunities
2. Select any condition you like. 
3. Set an Email action to be sent to a related field. For example a non mandatory user field.
 
When the related field is empty, the job will fail. 

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOW_Actions/actions/actionSendEmail.php
+++ b/modules/AOW_Actions/actions/actionSendEmail.php
@@ -245,10 +245,12 @@ class actionSendEmail extends actionBase {
                         }
                         if($linkedBeans){
                             foreach($linkedBeans as $linkedBean) {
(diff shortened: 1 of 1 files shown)


Dillon-Brown:
Assessed

PR #3496, opened 2017-05-02
Adding importing emails

daniel-samson:
## Description

This change adds a the email feature from the detail and list views. 
There is also an indicator in the list view to show which emails are imported using a tick.

There now is a new view called DetailedNonImported. This as its own metadata file nonimporteddetailviewdefs.php. It is responsible for display emails which haven't been imported. The reason for its own definitions is simply to remove SuiteCRM fields and only have an import action in the action menu. I have swapped the name field of a subject function field. So that I can pass the Imap information to "action_DisplayDetailView" controller action which decides which view to use. Imported emails will use the standard detail view.

You can override the bulk actions via the listviewdefs.php:
```
$viewdefs['Emails']['ListView'] = array(
    'templateMeta' => array(
        'form' => array(
            'buttons' =>
                array(
          //...
                ),
            'actions' => array(
                array(
                    'customCode' => '<a href="javascript:void(0)" class="parent-dropdown-handler" id="delete_listview_top" onclick="return false;"><label class="selected-actions-label hidden-mobile">{$APP.LBL_BULK_ACTION_BUTTON_LABEL_MOBILE}</label><label class="selected-actions-label hidden-desktop">{$APP.LBL_BULK_ACTION_BUTTON_LABEL}</label></a>',
                ),
                array(
                    'customCode' => '<a data-action="emails-import-multiple" title="{$MOD.LBL_IMPORT}">{$MOD.LBL_IMPORT}</a>'
            //....

```

Example:
importlistview

## How To Test This

* Rebuild everything (css and js)
* Clear cache.
* Select emails
* Select folder
* Select emails
* Select Bulk Actions
* Select Import

## Types of changes

### Final checklist

### DIFF
--- a/include/EditView/EditView2.php
+++ b/include/EditView/EditView2.php
@@ -1,9 +1,11 @@
 <?php
-/*********************************************************************************
+/**
+ *
  * SugarCRM Community Edition is a customer relationship management program developed by
  * SugarCRM, Inc. Copyright (C) 2004-2013 SugarCRM Inc.
- * SuiteCRM is an extension to SugarCRM Community Edition developed by Salesagility Ltd.
- * Copyright (C) 2011 - 2014 Salesagility Ltd.
+ *
+ * SuiteCRM is an extension to SugarCRM Community Edition developed by SalesAgility Ltd.
+ * Copyright (C) 2011 - 2017 SalesAgility Ltd.
  *
  * This program is free software; you can redistribute it and/or modify it under
  * the terms of the GNU Affero General Public License version 3 as published by the
@@ -14,7 +16,7 @@
  *
  * This program is distributed in the hope that it will be useful, but WITHOUT
  * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
- * FOR A PARTICULAR PURPOSE.  See the GNU Affero General Public License for more
+ * FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more
  * details.
  *
  * You should have received a copy of the GNU Affero General Public License along with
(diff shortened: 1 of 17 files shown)


include/EditView/EditView2.php
```diff
@@ -2,7 +2,6 @@
 /*********************************************************************************
  * SugarCRM Community Edition is a customer relationship management program developed by
  * SugarCRM, Inc. Copyright (C) 2004-2013 SugarCRM Inc.
-
  * SuiteCRM is an extension to SugarCRM Community Edition developed by Salesagility Ltd.
```
gymad: can you please change the licence headers too

modules/Emails/controller.php
```diff
@@ -324,23 +333,34 @@ public function action_GetFolders()
         $this->view = 'ajax';
     }
 
+
+    public function action_DisplayDetailView()
+    {
+        $emails = BeanFactory::getBean("Emails");
+        $result = $emails->get_full_list("", "uid = {$_REQUEST['uid']}");
```
gymad: can you use the db->quote method on the request variable before use it in the sql queries, also make sure this place the "uid" could be only number or strings also acceptable because if it could be a string you should use quotes

samus-aran: "uid = '{$_REQUEST['uid']}'"

modules/InboundEmail/InboundEmail.php
```diff
@@ -4164,7 +4164,7 @@ public function handleTranserEncoding($str, $enc = 0)
      */
     public function getMessageId($header)
     {
-        $message_id = md5(print_r($header, true));
+        $message_id = md5( );
```
gymad: I don't really understand why we removed the unique header from this function, it was an intended direction or just accidentally?

modules/InboundEmail/InboundEmail.php
```diff
@@ -5000,6 +5008,115 @@ public function returnImportedEmail($msgNo, $uid, $forDisplay = false, $clean_em
         return true;
     }
 
+    public function returnNonImportedEmail($msgNo, $uid)
```
gymad: could you add some valid PHPDoc to this function too please? I'm asking it three different reason because I'm not really sure what the arguments mean, I don't know what the possible return values mean because this function too long

daniel-samson: done

modules/InboundEmail/InboundEmail.php
daniel-samson: done

modules/Emails/controller.php
daniel-samson: done

include/EditView/EditView2.php
daniel-samson: updated

modules/Emails/controller.php
daniel-samson: updated

modules/InboundEmail/InboundEmail.php
```diff
@@ -4166,7 +4166,7 @@ public function getMessageId($header)
     {
         $message_id = md5(print_r($header, true));
 
-        return $message_id;
+        return $message_id;#
```
gymad: hash?

daniel-samson: fixed it :D

PR #3503, opened 2017-05-03
Fixed #3499 - SugarBean - array_seach function doesn't use strict parameter

### LINKED ISSUE #3499: SugarBean - array_seach function doesn't use strict parameter
#### Issue

Storyline is as follows:
I tried to create WorkFlow with SendEmail action on custom module. For the recipient TO field I wanted to set it to Related field (that actual field is custom field and it's checked as required). However this field doesn't show in list, so I dig into the code to see where is the issue...

There should be field named 'Budget Manager' in dropdown list:
emal_relate

The definition of field:
field_def_relate

The snippet of field def as php array during execution:
fielddef

It truns out that **get_related_fields()** function from **SugarBean** class gets called and in this function there is a code snippet:
```php
foreach ($fieldDefs as $name => $properties) {
                if (array_search('relate', $properties) === 'type') {
                    $related_fields[$name] = $properties;
                }
            }
```
Since there is no $strict parameter there is no type comparison.
This results that `array_search('relate', $properties)` will return `'required'` in my example above probably because `'relate' == true` results in `true`. This then results in if condition `'required' == 'type'` which is then false and field doesn't get populated in $related_fields array 

 
#### Expected Behavior

There should be strict type checking parameter in call to array_search() function. With that the field won't be skipped because it's required.

#### Actual Behavior

There is no strict type checking parameter which results that the field is skipped.

#### Possible Fix

```php
foreach ($fieldDefs as $name => $properties) {
                if (array_search('relate', $properties, true) === 'type') {
                    $related_fields[$name] = $properties;
                }
            }
```
#### Steps to Reproduce

1. Create Relate field to Users module and set check the Required box
2. Create new Workflow with Send Email action
3. In recipient list select 'To' - 'Related Field'
4. The newly created relate field will be missing from list.

#### Context

I cannot create Workflow to send email because the user is missing from dropdown list.

#### Your Environment

* SuiteCRM Version used: 7.7.8 
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Firefox 50
* Environment name and version (e.g. MySQL, PHP 7): MySQL 5.5.50, PHP 5.5.9-1ubuntu4.19
* Operating System and version (e.g Ubuntu 16.04): Mint 17.1

ajedvaj:
Added strict parameter to array_search function call to include strict type comparison of field def value

Issue raised when trying to create WorkFlow with Send Email action where TO field was a Related Field and there wasn't expected related field in the dropdown list. The field was skipped because it's checked as Required - since array_search was used without the $strict parameter it would compare string value and boolean true value.

## How To Test This

1. In studio mark Relate field to Users module as Required
2. Create WorkFlow with Send Email action.  Set TO as Related Field and in dropdown list of fields there should now be this field.

## Types of changes

### Final checklist

### DIFF
--- a/data/SugarBean.php
+++ b/data/SugarBean.php
@@ -4330,7 +4330,7 @@ class SugarBean
         //find all definitions of type link.
         if (!empty($fieldDefs)) {
             foreach ($fieldDefs as $name => $properties) {
-                if (array_search('relate', $properties) === 'type') {
+                if (array_search('relate', $properties, true) === 'type') {
                     $related_fields[$name] = $properties;
                 }
             }


chris001:
Good job.

PR #3527, opened 2017-05-09
Fixing Compose View Related To

daniel-samson:
Task: scrm-8

## Description

When the user types in the related to field they are not presented with the auto-completion box.

The issue is due to the id of the compose view being randomised. SQS expects the id of the form to be the name of the view like "EditView", as the sqs_objects array stores each filed by the form id + the name of the field.

Also in this change:
I have added an if statement to the minifier so that when you are in development mode the minifier simply returns the javascript unminified. This helps developers debug client side issues faster, and also speeds up the JS rebuild processes. This will only effect developers and not users.

## Motivation and Context

This change corrects the issue.

## How To Test This

* Repair and Rebuild
* Repair JS Grouping
* Compose an email
* Type in the relate to field

You should be able to see the auto-completion box. Unless there are no records with a name which matches your input.

## Types of changes

### Final checklist

### DIFF
--- a/jssource/Minifier.php
+++ b/jssource/Minifier.php
@@ -121,6 +121,12 @@ class Minifier
      */
     static public function minify($js, $options = array())
     {
+        global $sugar_config;
+
+        if(isset($sugar_config['developerMode']) && $sugar_config['developerMode'] === true) {
+            return $js;
+        }
+
         try{
             ob_start();
             $currentOptions = array_merge(self::$defaultOptions, $options);
--- a/jssource/Minifier.php
+++ b/jssource/Minifier.php
@@ -123,7 +123,7 @@ class Minifier
     {
         global $sugar_config;
 
-        if(isset($sugar_config['developerMode']) && $sugar_config['developerMode'] === true) {
+        if(isset($sugar_config['developer_mode_disable_minifier']) && $sugar_config['developer_mode_disable_minifier'] === true) {
             return $js;
         }
 
--- a/modules/Emails/include/ComposeView/ComposeView.tpl
+++ b/modules/Emails/include/ComposeView/ComposeView.tpl
@@ -42,7 +42,7 @@
 {{sugar_include type="smarty" file=$headerTpl}}
 {sugar_include include=$includes}
 {* Compose view has a TEMP ID in case you want to display multi instance of the ComposeView *}
-<form class="compose-view" id="{$TEMP_ID}" name="ComposeView" method="POST" action="index.php?module=Emails&action=send">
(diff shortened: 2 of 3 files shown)


jssource/Minifier.php
```diff
@@ -121,6 +121,12 @@ class Minifier
      */
     static public function minify($js, $options = array())
     {
+        global $sugar_config;
+
+        if(isset($sugar_config['developerMode']) && $sugar_config['developerMode'] === true) {
```
gymad: Can we change the config variable name for something else, please? - for e.g usually I don't want to switch to developer mode only for testing because I would to see exactly same thing as the end-user but I want to remove the js-compressing to speed-up the slow cache rebuilding process.
Thanks

daniel-samson: I think we should make a developer_mode a prefix.

In the system settings, the developer mode check box should reveal other options including this one. Perhaps your stack trace feature #3486 could also be an option.

I have changed this to developer_mode_disable_minifier.

pgorod:
Hi @daniel-samson, I've been testing the Beta and all these email improvements look pretty cool.

I have a question that is related to the "related to" field ;-)

Do you think it would be a good idea to let users easily "relate to" whatever Contact, Lead, or Target they selected in the "To:" field? I mean: when you email someone, you expect that email to be saved as related to that Contact, right? Having to select the Contact twice (once in the "to" field, once in the "related to" field, is repetitive, and involves quite some work, searching and selecting twice for the same person.

However, I see possible difficulties because you can be emailing any combination of several people in the "to", the "cc" and the "bcc" fields, and SuiteCRM can't just guess which one(s) you want to relate the field to. Perhaps a nice solution could be buttons to "copy" the Contact from "to"/"cc"/"bcc" to "related to" and another button in the other direction (copy from "related to" to "to", "cc" or "bcc").

What do you think?

daniel-samson:
@pgorod 

At present, when you include a contact, account etc in the To, Cc, Bcc fields. It will show up in history subpanel those respected records. The history subpanel provides a custom filter so you can search emails. Create direct relationships could create duplicates and would make it harder to filter. The current query based method is the way to go for now.

The flex relate to is a means to say that this email is directly related to a module record. Relating an email to multiple accounts doesn't make much sense. If you have multiple contacts, then you should have an account which relates those contacts.  The typical use case here is that you would have a company with employees. 

Copying to the relate field doesn't make much sense. As the relate field doesn't work by searching the email address. However, you could customise the relate field to copy the primary email address to the to/bcc field provided that this field is empty. Given that the browser already has cut, copy and paste. It would be possible for the user to then decide on which field they would prefer. This would then show up in the history panel as well as relating the email directly. Perhaps you could post this as a suggestion or provide a pull request after the initial production release. 

Thanks for testing. :D <3

PR #3529, opened 2017-05-09
Allow SubPanel Rows to add Row Buttons

mrbarletta:
Currently, only the "edit_button", "close_button" and "remove_button" can be shown in each subpanel row. 

If you were to create a new button inside the subpanel it will not be rendered

With this little change, it is possible to add new buttons with its own SugarWidget class.

## Description

Added the 

## Motivation and Context

Row buttons in subpanels used to be customizable, adding custom logic to this buttons is a nice feature to have. I remember it used to work, maybe before the ListViewSubPanel creation.
## How To Test This

Add a new button to a sub panel - Maybe something like this:
``` 
 'create_button' => 
  array (
    'widget_class' => 'SubPanelCreateContract',
    'module' => 'AOS_Quotes',
    'width' => '5%',
    'default' => true,
  ),
````
Go the the Module which the subpanel row should be shown and check if the button is rendered. You will see that it will not, as the current code only renders for said buttons (edit/remove/close)
## Types of changes

- [ x] New feature (non-breaking change which adds functionality)

### Final checklist

- [x ] My code follows the code style of this project found [here](https://suitecrm.com/wiki/index.php/Coding_Standards).
- [x ] I have read the [**How to Contribute**](https://suitecrm.com/wiki/index.php/Contributing_to_SuiteCRM) guidelines.

### DIFF
--- a/include/ListView/ListViewSubPanel.php
+++ b/include/ListView/ListViewSubPanel.php
@@ -504,7 +504,13 @@
                                         unset($_content);
                                     }
                                     else {
-                                        $button_contents[$aVal][] = '';
+                                        $dontProcessThisFields = array("edit_button", "close_button","remove_button");
+                                        if(!in_array($list_field['name'],$dontProcessThisFields) && '' != ($_content = $layout_manager->widgetDisplay($list_field))){
+                                            $button_contents[$aVal][] = $_content;
+                                            unset($_content);
+                                        }else{
+                                            $button_contents[$aVal][] = '';
+                                        }
                                     }
                                 }
                                 else {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: include/ListView/ListViewSubPanel.php [hunk 1]
```diff
@@ -504,7 +504,13 @@ public function process_dynamic_listview_rows($data, $parent_data, $smartyTempla
                                         unset($_content);
                                     }
                                     else {
-                                        $button_contents[$aVal][] = '';
+                                        $dontProcessThisFields = array("edit_button", "close_button","remove_button");
```
daniel-samson: The word "fields" in SuiteCRM means something you can store in database. The word "actions" in SuiteCRM means something you can do. Since the array stores actions, can you please this name to something like "$doNotProcessTheseActions"?

mrbarletta:
Thanks @daniel-samson good advice :)

PR #3567, opened 2017-05-17
Check php version

pribeiro42:
## Description
This changes the way that php version comparisons are made, at installation time and at upgrade time.
It changes the way the minimum PHP version is defined and also the number of places where it is defined. I think is a saner approach than to have the same check / code repeated in half-dozen files.

Creates a new file: php_version.php where one can define a constant with the minimum supported PHP version and changes the version_compare stuff to use a modified check_php_version.

## Types of changes

### Final checklist

### DIFF
--- a/include/utils.php
+++ b/include/utils.php
@@ -44,6 +44,7 @@
  * All Rights Reserved.
  * Contributor(s): ______________________________________..
  ********************************************************************************/
+require_once 'php_version.php';
 require_once 'include/SugarObjects/SugarConfig.php';
 require_once 'include/utils/security_utils.php';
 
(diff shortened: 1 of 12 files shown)


include/utils.php
```diff
@@ -4501,6 +4481,9 @@ function str_split_php4($string, $length = 1)
     return $return;
 }
 
+/*
+ * WONTFIX: Not updating this.. This should be removed !!!
+ */
 if (version_compare(phpversion(), '5.0.0', '<')) {
```
pribeiro42: Hello @daniel-samson.
Maybe we could really delete this code, as its only purpose is to define the str_split function under PHP4. If you agree i'll take care of that.

install.php
```diff
@@ -51,11 +53,23 @@
...
-//session_destroy();
-if (version_compare(phpversion(),'5.2.0') < 0) {
-	$msg = 'Minimum PHP version required is 5.2.0.  You are using PHP version  '. phpversion();
+/**
+ * Check php version
+ *
+ * If less than minimum we refuse to install.
+ */
+if (check_php_version() === -1) {
+	$msg = 'The recommended PHP version to install SuiteCRM is ';
```
pribeiro42: I really don't know. At this point there is no definition of "Installation Language". This gets executed really early in the installation process, before loading the language definition files, the Localization support and all that...

install.php
```diff
@@ -1,42 +1,42 @@
...
+ * The interactive user interfaces in modified source and object code versions 
+ * of this program must display Appropriate Legal Notices, as required under 
+ * Section 5 of the GNU Affero General Public License version 3. 
+ * 
+ * In accordance with Section 7(b) of the GNU Affero General Public License version 3, 
+ * these Appropriate Legal Notices must retain the display of the "Powered by 
+ * SugarCRM" logo and "Supercharged by SuiteCRM" logo. If the display of the logos is not 
+ * reasonably feasible for technical reasons, the Appropriate Legal Notices must 
+ * display the words "Powered by SugarCRM" and "Supercharged by SuiteCRM". 
+ */
```
daniel-samson: You have removed sugarEntry definition when you placed the license header.

`if(!defined('sugarEntry'))define('sugarEntry', true);`

It is breaking the test suite. Can you please add it back in and fix it to match the coding standard?

```
if(!defined('sugarEntry')) {
    define('sugarEntry', true);
}
```

include/utils.php
daniel-samson: I agree. I think your right. We should remove these statements. We don't support anything less than version 5.3. I ran a quick search across the project. I can't see any benefit for it being kept in. Can you please remove these statements?

install.php
daniel-samson: Lets leave this for now.

Dillon-Brown:
[Install check version](https://trello.com/c/LJ5yEhZZ/27-install-check-version)

Issue #3577, opened 2017-05-21
 PHP Notice:  A non well formed numeric value encountered in SUITECRM/include/ListView/ListViewSmarty.php on line [102|108]


jobst:
PHP Notice:  A non well formed numeric value encountered in SUITECRM/include/ListView/ListViewSmarty.php on line [102|108]

#### Issue

The problem is related to the WIDTH parameter passed in the function process of class  ListViewSmarty.php
If the width parameter is passed in pixel (i.e. 10) all is fine an the calculation continues without exception and correctly. If the width parameter is passed as a percentage the error happens and the calculation does not produce the correct values.

#### Expected Behavior

Well this is tricky. In real terms the programmer of ANY module should provide a pixel width and not a percentage ... that WAS fine for the OLD way of creating web displays. These days everything is responsive so you should use a percentage!
So thinking about this the class function "process" would need to be re-written to make it possible to specify width parms in %, ie 10%.

#### Actual Behavior

a PHP notice is raised but ALSO incorrect values are calculated!

#### Possible Fix

Tricky again ... needs to be decided as part of the direction SuiteCRM is taking and the code needs to be changed accordingly ....

#### Steps to Reproduce

1. click on contacts
2. wait for the list to come up
3. check the PHP log, its full of "A non well formed numeric value encountered ...."

#### Context

Well the display of the columns is skewed as the values in the function "process" are not evaluated correctly.

#### Your Environment

* SuiteCRM Version used: 7.8.2
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): N/A
* Environment name and version (e.g. MySQL, PHP 7): N/A
* Operating System and version (e.g Ubuntu 16.04): N/A

chris001:
You could update the function to detect incoming percentage parameters and handle them.

Note, this is an old way of generating HTML templates in PHP on the server, and should be upgraded, sooner or later, to responsive CSS3 templates handled by the browser.

jobst:
@chris001 done updating, fix proposed. Not sure whether I messed up the branches - still getting used to this - sorry.

chris001:
@jobst You merged your changes into your own fork, that's the first step, now, you need to make a PR from your fork to this repo.
1. Go to your repo.  
2. Click Code tab.
3. Click Compare (right side of page).
4. Select `SalesAgility/SuiteCRM:hotfix` and `jobst/SuiteCRM:hotfix` (assuming your used the hotfix branch of your form).
[...]

PR #3580, opened 2017-05-23
FIXED - PHP Notice: A non well formed numeric value encountered #3577

jobst:
This is a fix for an issue: PHP Notice: A non well formed numeric value encountered in SUITECRM/include/ListView/ListViewSmarty.php on line [102|108] #3577

The problem here a different themes/templates using PIXEL or PERCENT as width.
This fix checks which one is used and decides what to do. 
In case PIXELS are used some adjustments are required
in case PERCENT is used nothing needs to be adjusted.

## Description

The issue leads to constant exception calls which cost time.
In the original code in this foreach loop

  foreach($this->displayColumns as $name => $params) {
         $totalWidth += $params['width'];
   }

when it encounters 10% there is a problem leading to loads of exceptions (which cost time).
The fixed proposed checks this and makes sure that no adjustment is required.

## Motivation and Context

Ugly code, the issue leads to constant exception calls which cost time, log file clogging
It also leads to incorrect values that change the layout.

## How To Test This

Use a responsive theme, then open any subpanel that has any lists, check the log file, its full of the error.

## Types of changes

### Final checklist

### DIFF
--- a/include/ListView/ListViewSmarty.php
+++ b/include/ListView/ListViewSmarty.php
@@ -102,10 +102,10 @@ class ListViewSmarty extends ListViewDisplay
         try {
           foreach($this->displayColumns as $name => $params) {
             // error_log("CHECKING COLUMNS of $name ->".print_r($params,1));
-            if(strpos($params['width'],"%"))
-              $widthInPercent=true;
-            else
+            if(is_numeric($params['width']))
               $widthInPixels=true;
+            else
+              $widthInPercent=true;
           }
           // error_log("FINISHED CHECKING COLUMNS, RESULT PIX ->".$widthInPixels."<- PER ->".$widthInPercent."<-");
           // debug, trigger the error
--- a/include/ListView/ListViewSmarty.php
+++ b/include/ListView/ListViewSmarty.php
@@ -95,21 +95,56 @@ class ListViewSmarty extends ListViewDisplay
         $this->tpl = $file;
         $this->data = $data;
 
-        $totalWidth = 0;
-        foreach($this->displayColumns as $name => $params) {
-            $totalWidth += $params['width'];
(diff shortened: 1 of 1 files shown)


jobst:
Fully agreed "SuiteCRM should be theme agnostic", but there are still themes AND modules that use hard-coded pixel values. Also there are many spelling mistake in the (width) definitions of lists within the subpanel, I have seen "20s" and "20p" for example. I just put the code there to get rid of the constant exceptions and also added a DEBUG message in PHP.

daniel-samson:
If there are still themes and modules that use hard coded pixel values. We need to refactor them. I propose that to need to fix the issues instead of fixing the symptoms. We should at least fix the core and the modules. We can live with templates in the old themes which have hard coded values. However in SuiteP, we must use the SCSS to handle this correctly. Since the SuiteP will be the only theme.

We could still keep the columns and slots in the metadata. This could act as our universal unit.

Since this issue is related to the subpanels. Would you be willing to refactor just the hard coded parts for the subpanels?

jobst:
@daniel-samson: It's not only subpanels, I only specified subpanels in "How to test this" as it is a sure way to find these. It's everywhere where there are list displayed, even on the Dashboard.

I added a debug statement to show how often this occurs and where it needs to be fixed.

Adding the code will 
  - get rid of the exceptions for now till it's all clean speeding up processing
  - help to find all of them, once the code clean you can scrap the added code.

judgej:
A fresh install on 7.9.4 was wriggled with these errors (ListViewSubPanel.php, line 103). Applying this change removes all these errors for me.

hoenie-ams:
thanks @jobst, this fixed it for me too (PHP 7.1).

connorshea:
seems like this was fixed by #5162?

PR #3583, opened 2017-05-23
Upgrade fix

gymad:
## Description

Upgrade fix (ref: scrm-519)

Upgrade process have to upgrade the UpgradeWizard files first.
This PR also contains a change to able to add extra files to pre-upgrade process manually in upgrade pack. If the upgrade contains an extra php file named files_to_add_post.php and it's contains an array definition like this: `$filesToAddPost = array('filename1', 'filename2', );` the upgrade process copy these files first.

## Motivation and Context

Impossible to upgrade the crm when the release change the upgrade process itself

## How To Test This

see: description

## Types of changes

### Final checklist

### DIFF
--- a/modules/UpgradeWizard/uw_utils.php
+++ b/modules/UpgradeWizard/uw_utils.php
@@ -761,6 +761,9 @@ function upgradeUWFiles($file) {
 	if(file_exists("$from_dir/include/UploadFile.php")) {
 		$allFiles[] = "$from_dir/include/UploadFile.php";
 	}
+	if(file_exists(realpath("$from_dir/../scripts/files_to_add_post"))) {
+		$allFiles[] = findAllFiles(realpath("$from_dir/../scripts/files_to_add_post"), $allFiles);
+	}
 
     upgradeUWFilesCopy($allFiles, $from_dir);
 }
--- a/modules/UpgradeWizard/uw_utils.php
+++ b/modules/UpgradeWizard/uw_utils.php
@@ -758,8 +758,12 @@ function upgradeUWFiles($file) {
     if(file_exists("$from_dir/include/utils/autoloader.php")) {
         $allFiles[] = "$from_dir/include/utils/autoloader.php";
     }
+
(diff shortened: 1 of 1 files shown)


modules/UpgradeWizard/uw_utils.php
```diff
@@ -759,6 +759,28 @@ function upgradeUWFiles($file) {
...
+	}
+
+	// add extra files to post upgrade process
+	if(file_exists(realpath("$from_dir/../scripts/files_to_add_post.php"))) {
+		include(realpath("$from_dir/../scripts/files_to_add_post.php"));
+		if(isset($filesToAddPost) && is_array($filesToAddPost) && $filesToAddPost) {
+			foreach($filesToAddPost as $file) {
+				if(file_exists("$from_dir/$file")) {
+					$allFiles[] = "$from_dir/$file";
+					$GLOBALS['log']->info("File added to post upgrade: $from_dir/$file");
```
daniel-samson: Should this be an error instead of info?

gymad: It's the expected behavior, the log just inform that everything passed normaly

chris001: This will fail on Windows.  Try using `DIRECTORY_SEPARATOR` instead of `/` as separator. Windows uses a different separator for file system paths (backslash) than Linux (slash).

gymad: Everyone calm down, it's works on Windows.
@chris001: as far as I known both path separator works properly on Windows systems.
Thanks for your notice.

chris001: OK `/` works on Windows and Linux today, but I prefer to keep it 100% unbreakable on every OS, and future proof. This function does the job:
```
function file_build_path() {
    return join(DIRECTORY_SEPARATOR, func_get_args($segments));
}
```
Usage:   `$pathname = file_build_path($from_dir, $file);`

gymad: Yes @chris001 , you right, mee too

PR #3584, opened 2017-05-23
Feature email from field

daniel-samson:
Task: SCRM-554

## Description

This change adds the from field to the compose view and to the detail views. This will also work with the new draft feature.

For Developers:
I have also refactored the populateBean method so that we can make it more robust. You can now pass the request variables into the method call. This allows developers to use this method from other contexts and not just via http. In the future we can then write better unit tests for this method.

## Motivation and Context

* Feature enables users to select which email account the wish to use

## How To Test This

* Repair and Rebuild.
* Repair JS Groupings

Test 1:
* Compose an email
* Save as draft
* Edit Draft

Test2:
* View Email

## Types of changes

### Final checklist


modules/Emails/controller.php
```diff
@@ -78,6 +78,11 @@
     $GLOBALS['sugar_config']['http_referer']['actions'][] = 'ImportFromListView';
```
gymad: this file need some refactoring

daniel-samson:
@gymad I have updated my changes to match your code review. Seems that when I merged branches. This caused some issues. I haves fixed the changes.

gymad:
some console.log still there in EmailsComposeView.js

PR #3644, opened 2017-06-02
Feature ImportView Modal: Assign to and Related To

daniel-samson:
## Description

This change adds a new view called ImportView. This view has it's own metadata, allowing developers to easily add / remove fields. When the user selects import from the actions and bulk actions menu, they will be presented with a modal dialog with the assigned to and related to fields. 

This change also add a new feature into the SugarController which will check if the action exists in the controller and then temporary adds it to the referring http global, allowing developers to add custom actions without having to change the XSS protection check.

I have moved the Emails Controller into a new file EmailsController. As this is more PSR compliant. I have added a require_once into the controller.php for backwards compatibility.

I have updated the message box plugin so that it is easier to find the OK and cancel button.

I have updated the emails module so that it uses the basic sugar object template. As this fixes the assign to field.

I have updated Inbound email so that the description fields displays plain test, and the description_html displays html or a html version of the description field when html is not available.

I have updated the added new plugins for the import process, and update the ImportEmailAction to support the new plugins.

## Motivation and Context

* Add missing functionality
* Fixes the description field
* Cleans up code duplication
* Cleans up emails controller

## How To Test This

* repair and rebuild
* repair minified js
* repair js grouping

* Test Importing on list view
* Test importing on detail view

## Types of changes

### Final checklist


modules/Emails/include/ImportView/ImportView.tpl
```diff
@@ -0,0 +1,185 @@
+{*
+/**
+ *
+ * SugarCRM Community Edition is a customer relationship management program developed by
```
gymad: licence header is not up to date

daniel-samson: fixed

PR #3647, opened 2017-06-03
Fixed #3645 - User defined smtp settings not used when sending email

### LINKED ISSUE #3645: User defined smtp settings not used when sending email (ver 7.9)
#### Issue
I am referring to the Mail Account settings that can be found in the User profile (Profile - scroll all the way down - Settings - Mail Accounts - Add or Edit)
Here the user can add new SMTP settings; then the user can Add or Edit a Mail Account and specify these SMTP settings to be used with that mail account. This step works fine and settings are saved.

#### Expected Behavior
After defining SMTP settings and attaching them to a Mail Account as per above, one would expect emails sent from that Mail account to use the SMTP settings as defined.

#### Actual Behavior
This is not happening. The system smtp settings are used instead.

#### Possible Fix
First I confirmed that the smtp selection is correctly stored in the database along with the other settings in the inbound_email table. No problem here.
I could see that an id in the form of random key var_char is there to identify each record -- we shall call it inbound_email_id.
I then checked the email composer form (when you go to email - compose) and I saw that it properly lets you select your email address and sets the correct inbound_email_id in a hidden form element which is then submitted. All good so far.
Then I think I found an issue in modules/Emails/vardefs.php: there is no definition for the inbound_email_id that I could find, so that inbound_email_id form submission has nowhere to go...
So I created one by adding:
```
        'inbound_email_id' => array(
            'name' => 'inbound_email_id',
            'type' => 'varchar',
            'source' => 'non-db',
            'inline_edit' => false,
```
So now the bean engine does its magic and sets the bean inbound_email_id from the $_REQUEST data
Then next issue in line is the send() function in Emails.php
The send() function calls the setMailer() function which is the function that sets the smtp settings (among other things).
Here is setMailer definition
`function setMailer($mail, $mailer_id='', $ieId='')`
I could see that its third parameter $ieId is there to provide the inbound_email_id as one way to identify the from account. Indeed when I feed it the inbound_email_id manually it works great and sends via the correct smtp settings.
So in send() function of Email.php I replace 
```
$mail = $this->setMailer($mail);
```
with
```
$ieId = $this->inbound_email_id;
$mail = $this->setMailer($mail,'',$ieId);
```
And that does the trick. Emails are now sent through the smtp settings specified with the corresponding inbound email settings.

#### Steps to Reproduce

1. You need a way to know which SMTP server is used when sending. An easy way is to use gmail smtp because it will always set the from_address to the smtp user; make sure you use one gmail or google apps account for the system smtp 
2. Under a test user, create a new entry under Outgoing SMTP Mail Servers under Profile - scroll all the way down - Settings - Mail Accounts; here make sure to use another gmail or google apps account so that it will be clear whether that smtp user is used or not when receiving email.
2. Create or edit the corresponding incoming account under Mail account, select SMTP settings created previously. Save all and close.
3. Compose new email and send to an email address you can access. Check email and notice from info is from system smtp userid.

#### Your Environment
* SuiteCRM Version used: 7.9
* Browser: Chrome Version Version 58.0.3029.110 (64-bit)
* Environment name and version : MySQL 5.7.18, PHP 5.6

dawansv:
Added inbound_email_id when sending mail in order to get correct smtp settings

## Description
fix #3645 

## Motivation and Context
Changes needed to send from user specified smtp settings

## How To Test This
1. You need a way to know which SMTP server is used when sending. An easy way is to use gmail smtp because it will always set the from_address to the smtp user; make sure you use one gmail or google apps account for the system smtp 
2. Under a test user, create a new entry under Outgoing SMTP Mail Servers under Profile - scroll all the way down - Settings - Mail Accounts; here make sure to use another gmail or google apps account so that it will be clear whether that smtp user is used or not when receiving email.
2. Create or edit the corresponding incoming account under Mail account, select SMTP settings created previously. Save all and close.
3. Compose new email and send to an email address you can access. Check email and confirm it is from user smtp userid, not from system smtp userid.

## Types of changes
- [ x] Bug fix (non-breaking change which fixes an issue)

### Final checklist
- [x ] My code follows the code style of this project found [here](https://suitecrm.com/wiki/index.php/Coding_Standards).
- [x ] I have read the [**How to Contribute**](https://suitecrm.com/wiki/index.php/Contributing_to_SuiteCRM) guidelines.


modules/Emails/vardefs.php
```diff
@@ -483,6 +483,16 @@
             'inline_edit' => false,
         ),
 
+        'inbound_email_id' => array(
```
daniel-samson: This field is not needed. The mailbox_id field already stores the id of the inbound email record.

modules/Emails/Email.php
```diff
@@ -2697,7 +2697,8 @@ public function send()
             }
         }
 
-        $mail = $this->setMailer($mail);
+        $ieId = $this->inbound_email_id;
```
daniel-samson: use mailbox_id instead.

Dillon-Brown:
Works as expected

pgorod:
@Dillon-Brown you know you can make your assessments as code reviews instead of simple comments?

You can find that button `Review changes`  on the `Files changed` tab of the PR. Even unprivileged users like me can do it.

Unless this part of your flow to do it like this...

dawansv:
@daniel-samson 

Thanks. Missed that one. Will update.

On a side not, I wanted to test my changes with the latest hotfix so I first did a resync of my local hotfix branch with the upstream branch. But then I get an "underfined" error when sending an email or trying to update smtp settings in the admin interface.  Went back to my origin fork from a few days ago and works fine. To be clear I am just talking hotfix branches here without any of my changes.

I am doing something stupid or are other experiencing same problem on current hotfix branch? Easier way to test is just to try to edit smtp settings from the admin - Outbound Email option.

Here is the compare between my original hotfix branch and the curent one:
https://github.com/dawansv/SuiteCRM/compare/hotfix...salesagility:hotfix

Will have to test more and submit a new bug if I can confirm this.

dawansv:
@daniel-samson 

Never mind the undefined error. It's a problem on my side. Sorry about the noise.

dawansv:
Add to modify controller.php to set mailbox_id in action_send() because it is not populated automatically since it's different from the REQUEST variable name.

Small conflict in controller.php because of some end of line difference, but I can't figure out how to get rid of it... Grrrr.

modules/Emails/EmailsController.php
```diff
@@ -135,6 +135,7 @@ public function action_ComposeView()
     public function action_send()
     {
         $this->bean = $this->bean->populateBeanFromRequest($this->bean, $_REQUEST);
+        $this->bean->mailbox_id = $_REQUEST['inbound_email_id'];
```
daniel-samson: Can you please move this into the populateBeanFromRequest? So that we can reuse this code as much as possible.

I have just noticed that this method isn't using the $request variable. We will need to update this.

dawansv:
@daniel-samson 

OK I moved the mailbox_id assignment to populateBeanFromRequest.  I also removed it from function actionSaveDraft() in EmailController.php since it is now superfluous there as well. I did a grep -r * -e "mailbox_id" on the whole tree and as far as I can tell, there is no similar assignment anywhere else.

As for the assignment in populateBeanFromRequest, I used $_REQUEST instead of $request because the rest of the code does the same.  It looks like this function is pending cleanup as relating to the use of $request (as per comment in header), but I am not sure which direction you want it to go...

samus-aran:
@dawansv From what I can review it looks fine the execution.  Are you complete and we can get this re-tested internally?

PR #3677, opened 2017-06-08
Feature: Make sugar as search Plugin for Firefox

gunnicom:
Firefox allows new search plugins with opensearch ( http://www.opensearch.org/Home ).
This small patch allows sugar as search plugin.

### DIFF
--- a/sugarsearch.php
+++ b/sugarsearch.php
@@ -5,7 +5,7 @@ print <<<EOF
 <?xml version="1.0" encoding="UTF-8" ?> 
 <OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/"> 
     <ShortName>SuS</ShortName>  
-    <Description>Sugar Suche</Description> 
+    <Description>Sugar Search</Description> 
     <InputEncoding>UTF-8</InputEncoding> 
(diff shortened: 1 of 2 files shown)


gunnicom:
Changed "Suche" to "Search" (Maybe if anyone know how to use translation framework, help is welcome)
For the hotfix branch, i think as Suite7 theme is removed in 7.9 i did not want to make this pull request against hotfix and chose hotfix-7.8.

chris001:
@gunnicom 
http://support.sugarcrm.com/Documentation/Sugar_Developer/Sugar_Developer_Guide_6.5/Module_Framework/Language/

Retrieving Labels

There are two ways to retrieve a label. The first is to use the `translate()` function found in `include/utils.php`. In using this function, the label will be retrieved for the current users language. This function can also be used to retrieve labels from `mod_strings`, `app_strings` or `app_list_strings`.

An example of this is:

```
require_once('include/utils.php');
$label = translate('LBL_KEY', 'Accounts');
```
Alternatively, you can also use the global variable `$mod_strings` as follows:

```
global $mod_strings;

$label = '';
if (isset($mod_strings['LBL_KEY']))
{
    $label = $mod_strings['LBL_KEY'];
}
```

pgorod:
I was testing this old PR and I was able to get it to work. Nice!

I am on Firefox 64 and SuiteCRM 7.10.11. I made the changes in `themes/SuiteP/tpls/_head.tpl`. I can see the new `link` element is in my HTML after I refresh.

For other people who come in here testing this: make sure you get around any caching issues until you see the new `link rel="search" ...` element in the `<head>` of your page.

When you do, your Firefox will show something like this on the search bar: 

image

See that green `+` sign? That means you can now add the new search engine:

image

The PR needs a little more work, though:

1. For HTML4 correctness, leave out the closing `/` in the `<link>` element.
1. Consider what could be a better location for `sugarsearch.php`, and call it by a better name ;-), I suggest `include/opensearch.php`
1. Make the strings translatable. This includes fixing one more `SugarCRM Suche` in the `_head.tpl` file.
1. Add this to the other themes.
1. Add something about this to Documentation. Note that this is not just for Firefox, OpenSearch is supported by (at least) Firefox, Edge, Internet Explorer, Safari, and Chrome.

@gunnicom are you willing to update the PR with any of this? Do you want me to take care of any of the above? I'd be happy to.

gunnicom:
@pgorod Maybe i will have a look at this tomorrow. Totally forgot about that one :D

gunnicom:
Did a new PR with the comments from pgorod included. I think it looks nicer now:
https://github.com/salesagility/SuiteCRM/pull/6642

PR #3770, opened 2017-06-22
Fix #3753 no available themes

### LINKED ISSUE #3753: 7.9 Upgrade fails if SuiteP is disabled, and breaks instance
This is from a couple of issues in the forums, like [this one](https://suitecrm.com/forum/suitecrm-7-0-discussion/14657-7-9-1-upgrade#49090). Thanks to user **LISpeedyG**.

### Issue
Some people have disabled SuiteP theme in their instances, so that their users couldn't select the new theme. Sometimes this was a necessity because of theme-specific customizations, other times it was a matter of taste.

This seems to be possible with one of these ways (or both?):

In `config.php` something like this:

      'disabled_themes' => 'SuiteP,Suite7',

In `config_override.php` something like this: 

     $sugar_config['disabled_themes'] = 'SuiteP,Suite7';

### Expected Behavior
SuiteCRM should re-enable the SuiteP theme and complete the upgrade successfully; or at the very least, give a proper error message asking the admin to go and enable the theme.

### Actual Behavior
The upgrade is interrupted at the last stage of the Wizard, no error shown.

**SuiteCRM is now broken**, doesn't work at all, and it's hard to diagnose. Nothing relevant in the logs.

### Steps to Reproduce
1. Get a working 7.8 instance and disable the SuiteP theme in `config.php` or `config_override.php`
2. Try the upgrade to 7.9

### Context
This is not easy to diagnose. It is not very serious for people who have a way to go back (backups, VM snapshots) but is highly critical for people who don't. On the other hand, it shouldn't be _that_ common.

### Your Environment
* SuiteCRM Version used: 7.8 upgrading to 7.9

pgorod:
## Description
This is the minimal fix, just a better message. It would be nice to handle this better in upgrades to 7.9 where SuiteP is disabled.

## Motivation and Context
See issue #3753 

## How To Test This
Disable all themes through config_override.php, for example.
Or disable SuiteP on a 7.8 instance, then upgrade to 7.9

## Types of changes

- [ X ] Bug fix (non-breaking change which fixes an issue)

### Final checklist

- [ X ] I have read the [**How to Contribute**](https://suitecrm.com/wiki/index.php/Contributing_to_SuiteCRM) guidelines.

### DIFF
--- a/include/SugarTheme/SugarTheme.php
+++ b/include/SugarTheme/SugarTheme.php
@@ -1424,7 +1424,7 @@ class SugarThemeRegistry
         }
         asort($themelist, SORT_STRING);
         if (count($themelist)==0) {
-		$GLOBALS['log']->fatal("availableThemes() is returning empty array! Check disabled_themes in config.php and config_override.php");
+		$GLOBALS['log']->fatal('availableThemes() is returning empty array! Check disabled_themes in config.php and config_override.php');
         }
         return $themelist;
     }
--- a/include/SugarTheme/SugarTheme.php
+++ b/include/SugarTheme/SugarTheme.php
@@ -1424,8 +1424,7 @@ class SugarThemeRegistry
         }
         asort($themelist, SORT_STRING);
         if (count($themelist)==0) {
-		$themelist= array('SuiteP' => 'Suite P');
-		$GLOBALS['log']->error('availableThemes() was returning empty array, SuiteP added! Check disabled_themes in config.php and config_override.php');
+     		$GLOBALS['log']->fatal('availableThemes() is returning an empty array! Check disabled_themes in config.php and config_override.php');
         }
         return $themelist;
     }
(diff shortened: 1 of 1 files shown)


pgorod:
This last commit changes the approach. It is questionable whether this is better. 

It's better for the current situation, but leaves a hard-coded theme name in the code which is not very good for the future. 

So let's hear some opinions? 
@samus-aran @daniel-samson and whoever else wants to have their say... thanks.

Dillon-Brown:
@pgorod I agree that having a hard-coded theme is not the best solution for the future. I definitely think that the logging message is good idea.

I had actually already resolved this issue in the upgrade script internally, its currently still being tested. It seems I forgot to assign the original issue to myself, my apologies. Do you think having the check in the upgrade script is a better option?

pgorod:
Of course avoiding the hard-coded mention to `SuiteP` is the code is preferable if the upgrade script can handle it, so go ahead with your fix.

Anyway, I would still leave the FATAL error when `availableThemes` is returning an empty array, it's a useful safety net for the future.

So I'll try and delete (revert) today's commit. Thanks!

mengbo123456:
image

Issue #3766, opened 2017-06-22
7.9.1 Bounce email handling broken; importOneEmail only half-deprecated


pgorod:
#### Issue
From the forums. A user [complains](https://suitecrm.com/forum/installation-upgrade-help/14730-bounce-handling-account-what-does-it-do?limitstart=0) bounce handling is not working in Campaigns.

#### Expected Behavior
Bounced emails should be recognized as such in the Campaign.

#### Actual Behavior
The Campaign data ignores bounced emails.

#### Possible Fix
I have a somewhat speculative possibility for why this is happening. Feel free to ignore it if you discover that it's not the real cause, or if you can't even reproduce this issue.

It seems in 7.9 a function called importOneEmail was deprecated. It says so in a comment in the code. See [here](https://github.com/salesagility/SuiteCRM/blob/master/modules/InboundEmail/InboundEmail.php#L4584).

However, quite a few [calls to that function](https://github.com/salesagility/SuiteCRM/search?utf8=%E2%9C%93&q=importOneEmail&type=) are still present in the code.

The function replacing this seems to be called [returnImportedEmail](https://github.com/salesagility/SuiteCRM/blob/master/modules/InboundEmail/InboundEmail.php#L4883). It includes a call to `HandleMailboxType` which handles bounced emails. 

But the current flow from the Scheduler job doesn't seem to use the new function, this seems to be the sequence (I hope I'm correct, I just followed this "manually", not while debugging):

1. The scheduler [job](https://github.com/salesagility/SuiteCRM/blob/032cc000331e1d9a411e468f292e8f9d14d553fd/modules/Schedulers/_AddJobsHere.php#L396-L413) runs
2. It Calls ImportMessages, which calls ImportMailboxMessages
3. That calls importOneEmail (instead of calling the new function)
4. Maybe when it then reaches the call to `HandleMailboxType` there is something not quite right, and it doesn't work.

Should we just change the call to `importOneEmail` to cal `returnImportedEmail` instead?

#### Steps to Reproduce
1. Define bounce account
2. Run a campaign and make sure some address bounces
3. check the campaign status and see the bounce isn't there

#### Your Environment

* SuiteCRM Version used: 7.9.1

sergio91pt:
Some public "bounce emails", may be useful for testing:

- bouncetest@tribulant.com (immediately rejected with SMTP error 550)
[...]

PR #3777, opened 2017-06-23
Hotfix scrm 569 bounce email handling broken

gymad:
## Description

- This PR contains some fox for PHP errors (Email Template variables did not handle the variables correctly after html entity cleanup)
- Also contains a fix for Inbound email duplication check: related to bounced messages in campaign stat (related issue #3766)

## Motivation and Context

Campaign stat doesn't shows the bounced messages after the scheduled jobs run

## How To Test This

- create a new campaign with some bounced email address in target list
- check campaign stat after scheduled jobs

## Types of changes

### Final checklist

### DIFF
--- a/modules/EmailTemplates/EmailTemplate.php
+++ b/modules/EmailTemplates/EmailTemplate.php
@@ -300,7 +300,10 @@ class EmailTemplate extends SugarBean
     {
         if (empty($this->body) && !empty($this->body_html)) {
             global $sugar_config;
-            $this->body = strip_tags(html_entity_decode($this->body_html, ENT_COMPAT, $sugar_config['default_charset']));
+            $bodyCleanup = html_entity_decode($this->body_html, ENT_COMPAT, $sugar_config['default_charset']);
+            $bodyCleanup = preg_replace('/(\$\w+\b)([^\s])/', '$1 $2', $bodyCleanup);
+            $bodyCleanup = strip_tags($bodyCleanup);
+            $this->body = $bodyCleanup;
         }
         $this->created_by_name = get_assigned_user_name($this->created_by);
         $this->modified_by_name = get_assigned_user_name($this->modified_user_id);
--- a/modules/EmailTemplates/EmailTemplate.php
+++ b/modules/EmailTemplates/EmailTemplate.php
@@ -300,9 +300,17 @@ class EmailTemplate extends SugarBean
     {
         if (empty($this->body) && !empty($this->body_html)) {
             global $sugar_config;
+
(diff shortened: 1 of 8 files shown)


modules/EmailTemplates/Save.php
```diff
@@ -53,7 +53,7 @@
 
 require_once('modules/EmailTemplates/EmailTemplateFormBase.php');
 $form = new EmailTemplateFormBase();
-sugar_cache_clear('select_array:'.$focus->object_name.'namebase_module=\''.$focus->base_module.'\'name');
+sugar_cache_clear('select_array:'.$focus->object_name.'namebase_module=\''. (isset($focus->base_module) ? $focus->base_module : null).'\'name');
```
sergio91pt: You should use `''` instead of `null` to make the intent clear.

modules/EmailTemplates/EmailTemplate.php
```diff
@@ -300,7 +300,10 @@ function fill_in_additional_detail_fields()
     {
         if (empty($this->body) && !empty($this->body_html)) {
             global $sugar_config;
-            $this->body = strip_tags(html_entity_decode($this->body_html, ENT_COMPAT, $sugar_config['default_charset']));
+            $bodyCleanup = html_entity_decode($this->body_html, ENT_COMPAT, $sugar_config['default_charset']);
+            $bodyCleanup = preg_replace('/(\$\w+\b)([^\s])/', '$1 $2', $bodyCleanup);
```
sergio91pt: What is the purpose of this hack? Please add it as a comment to the code.

modules/EmailTemplates/Save.php
gymad: It doesn't matter by functional behavior.
In PHP at string concatenation the `(string)null` value cast to string evaluated as an empty string `''` .
Also in PHP the undefined variables and object properties evaluated as `null` and indicate a php notice. 
so if it's `null` that clearly means it is an undefined variable which evaluated as an empty string without php notice.

modules/InboundEmail/InboundEmail.php
```diff
@@ -4407,7 +4407,7 @@ public function importDupeCheck($message_id, $header, $textHeader)
         $r = $this->db->query($query, true);
         $a = $this->db->fetchByAssoc($r);
```
daniel-samson: Can we change these variable names to something that is more meaningful?

Could we also change the $a['c']  to something that is a little more descriptive?

gymad: It was not in the PR but thanks for your notice :)

samus-aran:
From testing default emails bounces but specific unique test cases were discovered.  But this was established also for the older versions 7.8.x so separating this issue appropriately.

PR #3801, opened 2017-06-27
Fix. Demo data was not installed for Travis tests!

chris001:
Install sample SugarCRM demo data so that Travis PHPunit / Codeception tests will have some data to run their tests upon.

## Motivation and Context

This allows the Travis automated tests to run with sample data.  
Previously, no sample data was installed (!!!!).

## How To Test This

Push a commit to an existing PR, or create a new PR.
The Travis build should install the SugarCRM sample data.
The Travis test results should show fewer F (FAIL) when those fails were due to lack of test data.

## Types of changes

### Final checklist

### DIFF
--- a/tests/travis_config_si.php
+++ b/tests/travis_config_si.php
@@ -1,5 +1,6 @@
 <?php
 $sugar_config_si  = array (
+    'demoData' => 'yes',
     'dbUSRData' => 'create',
     'default_date_format' => 'Y-m-d',
     'default_decimal_seperator' => '.',


daniel-samson:
The demo data is not suitable for automated testing. We need to use data mocking for tests so that they can be run in isolation and as part of a larger sub system. The demo data is fine for manual testing but its purpose is really to demonstrate the features to prospecting users.

chris001:
@daniel-samson 
Thank you.  Yes I'm aware and agree, adding demo data is not the ultra perfect solution from God himself, or herself, depending on your religious orientation.  However, installing demo data for use by the tests, does have a value.  My intention in adding the demo data is so that some test will no longer halt before running at all, because they're running on absolutely empty database, which is an unusual environmental condition, and they cannot run on no data.  Functional tests need data in the database in order to run, and unfortunately most of these so-called unit tests are in reality functional tests misclassified as unit tests.  They are still tests however, and we still want them to run to completion, as much as they possibly can, and they need data to do so.  Hence the need for the installer to install that good ol' sample data in the database.

The next iteration of adding data, to please the testing gods, would be to use the glorious `dbunit` and restore the demo data to its original state, in between each test, to eliminate test-to-test changes in the environment.  And then, to use multiple different sets of demo data, to create many combinations and permutations of data inputs to run the tests on.

daniel-samson:
Since we are intending to correct this in 7.10, would it not be better to hold off for now? That way, we do not waste your efforts on something that will be soon changed. 

In regards, to your tooling suggestions, why not make a branch from the feature codeception branch and have a look into the rules for PHP cs fixer and Travis CI? 

We need to create custom rules for areas that will not be psr2 such as the controller actions. So that PHP cs is a little more forgiving.

Dillon-Brown:
Closing this due to it being an inappropriate fix. We'll be addressing this properly in 7.10

PR #3806, opened 2017-06-28
Workflow Enhancements 7.9.1 - Process Once Per Day

stevejenkin:
We needed to find a solution of stopping a repeatable running workflow from triggering every time the Workflow schedule runs,

## Description

Conditions when this is an issue:

Time-based workflow condition which triggers on a particular day. On that day the conditions are met all day and the workflow will trigger every time the schedule runs (e.g. every minute). Repeated runs is required as the date may change to another value in the future and the record should be considered to be processed in the workflow again.

Proposed Solution:

Add checkbox next to the Repeated Runs checkbox called 'Trigger Only Once Per Day'.
When this checkbox is ticked, once it has run once for a particular record it will not run again for 24 hours for that particular record. There is also an Hour & Minute field to specify at what time each day this workflow should run.

## Motivation and Context

We needed to find a solution of stopping a repeatable running workflow from triggering every time the Workflow schedule runs. 

## How To Test This

Create a workflow using the module Accounts.

The condition is: When a date field in the Accounts module (e.g Last Contacted Date) > 60 days ago.

The action is: Create a new call.

What will happen in this instance is that each time the workflow scheduler runs a new call will be created, we cannot un-select the 'Repeated Runs' check box because the date in the field will change again and we want the record to be processed again in the future.

With our development we have implemented a check box which specifies that the workflow should only 'Trigger Only Once Per Day'. We have also added an hours and minutes field which allows the user to specify when that will be.

## Types of changes

### Final checklist

- [X ] My code follows the code style of this project found [here](https://suitecrm.com/wiki/index.php/Coding_Standards).
- [X ] My change requires a change to the documentation.
- [X ] I have read the [**How to Contribute**](https://suitecrm.com/wiki/index.php/Contributing_to_SuiteCRM) guidelines.

### DIFF
--- a/modules/AOW_WorkFlow/AOW_WorkFlow.php
+++ b/modules/AOW_WorkFlow/AOW_WorkFlow.php
@@ -122,7 +122,44 @@ class AOW_WorkFlow extends Basic {
 		$flows = AOW_WorkFlow::get_full_list(''," aow_workflow.status = 'Active'  AND (aow_workflow.run_when = 'Always' OR aow_workflow.run_when = 'In_Scheduler' OR aow_workflow.run_when = 'Create') ");
 
         foreach($flows as $flow){
-            $flow->run_flow();
+            ## START Trigger Only Once Per Day
+            $temp_trigger = true;
+            if($flow->trigger_only_once_per_day)
+            {
+                $query = "SELECT * FROM aow_workflow_execute WHERE aow_workflow_id = '".$flow->id."' AND 
+                        DATE_FORMAT(execute_datetime, '%Y-%m-%d') = DATE_FORMAT(now(), '%Y-%m-%d')";
+                    
+                $results = $GLOBALS['db']->query($query, true);
+                $num_row = $GLOBALS['db']->getRowCount($results);
+                
+                if($num_row > 0)
+                {
+                    $temp_trigger = false;
+                }else{
+                    $query = "SELECT now() FROM dual WHERE TIME_TO_SEC(CONCAT(DATE_FORMAT(CurDate(), '%Y-%m-%d'), ' ".$flow->once_per_hour.":".$flow->once_per_min."'))  > TIME_TO_SEC(now())";
+                    
+                    $results = $GLOBALS['db']->query($query, true);
+                    $num_row = $GLOBALS['db']->getRowCount($results);
+                    
+                    if($num_row > 0)
+                    {
+                        $temp_trigger = false;
+                    }
+                }
+            }
+            
+            if($temp_trigger) 
+            {
+                $flow->run_flow();
+            
+                if($flow->trigger_only_once_per_day)
+                {
+                    $qry_ins = "INSERT INTO aow_workflow_execute(aow_workflow_id, execute_datetime)
+                                VALUES('".$flow->id."', now())";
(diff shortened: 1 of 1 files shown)


modules/AOW_WorkFlow/AOW_WorkFlow.php
```diff
@@ -210,7 +210,13 @@ function build_flow_query_join($name, SugarBean $module, $type, $query = array()
                 case 'custom':
                     $query['join'][$name] = 'LEFT JOIN '.$module->get_custom_table_name().' '.$name.' ON '.$module->table_name.'.id = '. $name.'.id_c ';
                     break;
-
+				
+                case 'custom_relationship':
+                    $main_name = str_replace("_cstm", "", $name);
+                    $query['join'][$name] = 'LEFT JOIN '.$module->get_custom_table_name().' '.$name.' ON '.$main_name.'.id = '. $name.'.id_c ';
+                    #echo $query['join'][$name];
```
daniel-samson: debug statement? please remove.

modules/AOW_WorkFlow/AOW_WorkFlow.php
```diff
@@ -277,11 +283,17 @@ function build_query_where(AOW_Condition $condition, $module, $query = array()){
...
+		
+        $oepl_rel = '';
+        
         if(isset($path[0]) && $path[0] != $module->module_dir){
             foreach($path as $rel){
                 $query = $this->build_flow_query_join($rel, $condition_module, 'relationship', $query);
                 $condition_module = new $beanList[getRelatedModule($condition_module->module_dir,$rel)];
                 $table_alias = $rel;
+                
+                $oepl_rel = $rel;
```
daniel-samson: What is $oepl_rel short for? Can you use the full name here?

stevejenkin:
@daniel-samson 

I have uploaded all the files related to this development, does it make things clearer now?

stevejenkin:
Any update on this please?

daniel-samson:
@stevejenkin git hub is showing unknown repository. I am unable to clone your branch or see your commits. Have you re based or removed the branch?

stevejenkin:
@daniel-samson I have removed the branch unfortunately, how would you like to proceed?

pgorod:
I have a suggestion for a design to solve this same problem, but with a more generic solution, and I believe it isn't too much work to develop:

The problem is: sometimes it's useful to have a middle ground between `Repeated runs`and `No Repeated Runs`. It would be nice to reset `Repeated Runs` daily... or monthly, or any other period.

So I suggest instead of looking at this like a problem of resetting `Repeated Runs` periodically, we could do it like this:

- in Workflows, add option `Run` as `Scheduler, specific job`

- in Schedulers, add an option in `Add Scheduler` called `Specific Workflow` with parameter letting you choose existing `Specific job` Workflows from Dropdown

- in general workflow execution, ignore these Workflows when launched from the generic Scheduler job, and run them when run from the specific one

This way the full possibilities of `crontab` settings could be applied to any specific Workflow.

Mac-Rae:

pgorod:
Agree, this is better to close. But the discussion is interesting, and I followed up with a suggestion here #8292.

Issue #3741, opened 2017-06-18
7.9.1 Can't compose any email. No outbound accounts shown in "From" dropdown


pgorod:
This is on a 7.9.1 test instance, upgraded from 7.9 (and from many earlier versions).

#### Steps to Reproduce
1. Configure system outbound email account, in my case Gmail. Click the checkbox to `allow users to use this account`.
2. Send test email (works well)
3. Go to `Emails` module (this is the Admin user, and has no configured Inbound email)
4. Click `Compose`. Note that there is nothing in the `From` dropdown box.
5. Fill in a valid `To` address, fill in `Subject` and `Body`.
6. If you click `Send`, you get an error `Error Sending Email. Please contact your administrator for assistance.`

#### In the logs

suitecrm.log shows this at the moment the `Send` button is pressed:

```
Sun Jun 18 15:12:53 2017 [1470][1][FATAL] SugarPHPMailer encountered an error: Invalid address:  (addAnAddress Reply-To): null
Sun Jun 18 15:12:53 2017 [1470][1][FATAL] SugarPHPMailer encountered an error: Invalid address:  (punyEncode) null
```
#### In the database

This is a small CSV with the only row in my `outbound_email` table:

```
"id","name","type","user_id","mail_sendtype","mail_smtptype","mail_smtpserver","mail_smtpport","mail_smtpuser","mail_smtppass","mail_smtpauth_req","mail_smtpssl","date_entered","date_modified","modified_user_id","created_by","deleted","assigned_user_id"
"b841a42e-7e21-d813-f871-594684657b17","system","system","1","smtp","other","smtp.gmail.com","587","obfuscated@gmail.com","TLOBFUSCATEDsjkdskjfdskjfdsjafcYPA=","1","2","2017-06-18 13:46:24","2017-06-18 13:46:24","1","1","0",NULL
```

#### Environment
* SuiteCRM Version used: 7.9.1
* Browser name and version: Firefox latest, and Edge. Same behavior.
* Environment name and version: MySQL 5.7, PHP 7
* Operating System and version: Ubuntu 16.04

pgorod:
@Dillon-Brown were you able to reproduce this problem? I'm trying to understand if I have a problem specific to my system (and a few other people out there) or if this is happening to everybody. Thanks.

Dillon-Brown:
@pgorod I'm able to replicate this. With no inbound email account configured; the "From" field won't populate.

samus-aran:
[...]

PR #3812, opened 2017-06-28
Fix scrm 638

daniel-samson:
## Description

Ensures that a user can only send an email via their own inbound email account.

## Motivation and Context

* Adds a check to the send action in the email controller.
* It verifies that they user is subscribed to the inbound email in question
* It verifies the the from address is in the users inbound email account or in the system settings. 

## How To Test This

* Repair and rebuild
* Set up a user with a personal email account
* Take a note of the inbound email id and the from address
* Set up a second user
* Login in to the second user account 
* set the url  to http://path-to-suitecrm-instance/index.php?module=Emails&action=Send&inbound_email_record=[noted inbound email record id]&from_addr=[noted from email address]

You should see and error message, and when you refresh or access the suitecrm again it should ask you to login again.

## Types of changes

### Final checklist


daniel-samson:
@gymad  @sergio91pt  @pgorod @samus-aran can you please review and test this.

modules/Emails/EmailsController.php
```diff
@@ -139,19 +139,43 @@ public function action_ComposeView()
      */
     public function action_send()
     {
-        $this->bean = $this->bean->populateBeanFromRequest($this->bean, $_REQUEST);
-        $this->bean->save();
+        global $current_user;
+        global $app_strings;
 
-        $this->bean->handleMultipleFileAttachments();
+        $this->bean = $this->bean->populateBeanFromRequest($this->bean);
```
Dillon-Brown: Missing parameter: $_REQUEST

modules/Emails/EmailsController.php
```diff
@@ -139,19 +139,43 @@ public function action_ComposeView()
...
-        if ($this->bean->send()) {
-            $this->bean->status = 'sent';
+        if($this->userIsAllowedToSendEmail($current_user, $inboundEmailAccount, $this->bean)) {
             $this->bean->save();
+
+            $this->bean->handleMultipleFileAttachments();
+
+            if ($this->bean->send()) {
+                $this->bean->status = 'sent';
+                $this->bean->save();
```
sergio91pt: Shouldn't the status also be saved when it failed to send?
I know you didn't touch this part, but it is interesting nonetheless.

modules/Emails/EmailsController.php
```diff
@@ -139,19 +139,43 @@ public function action_ComposeView()
...
+            } else {
+                $this->bean->status = 'sent_error';
```
sergio91pt: phpdoc of $status says this is `dom_email_status` but that dropdown has `send_error` and not `sent_error`.
Will you take a look, or should we open a new issue to investigate this?

screenshot from 2017-06-29 12 40 34

```bash
$ grep -Frni sent_error modules/*Email*
modules/Emails/views/view.sendemail.php:83:            case 'sent_error':
modules/Emails/EmailsController.php:146:            $this->bean->status = 'sent_error';

$ grep -Frni send_error modules/*Email*
modules/Emails/Save.php:189:            $focus->status = 'send_error';
modules/Emails/Email.php:781:            $this->status = 'send_error';
modules/Emails/Email.php:1153:                $this->status = 'send_error';
modules/Emails/Email.php:3076:            if ($this->type == 'out' && $this->status == 'send_error') {
modules/Emails/Email.php:3082:            if (($this->type == 'out' && $this->status == 'send_error') || $this->type == 'draft') {
```

modules/Emails/EmailsController.php
```diff
@@ -627,4 +651,49 @@ protected function setAfterImport($importedEmailId, $request)
...
         return $emails;
     }
+
+    /**
+     * @param User $requestedUser
+     * @param InboundEmail $requestedInboundEmail
+     * @param Email $requestedEmail
+     * @return bool false if user doesn't have access
+     */
+    protected function userIsAllowedToSendEmail($requestedUser,  $requestedInboundEmail, $requestedEmail)
```
sergio91pt: 2 spaces after `$requestedUser,` :wink:

modules/Emails/EmailsController.php
```diff
@@ -627,4 +651,49 @@ protected function setAfterImport($importedEmailId, $request)
...
+    {
+        $hasAccess = false;
+        $hasAccessToInboundEmailAccount = false;
+
+        // Check that user is allowed to use inbound email account
+        $usersInboundEmailAccounts = $requestedInboundEmail->retrieveByGroupId($requestedUser->id);
```
sergio91pt: Note about the called function: Field group_id in the table inbound_email is not indexed. New issue?

modules/Emails/EmailsController.php
daniel-samson: Added

modules/Emails/EmailsController.php
daniel-samson: Removed space

modules/Emails/EmailsController.php
daniel-samson: Renamed to sent_error.

modules/Emails/EmailsController.php
```diff
@@ -627,4 +651,49 @@ protected function setAfterImport($importedEmailId, $request)
...
+        foreach ($usersInboundEmailAccounts as $inboundEmailId => $userInboundEmail) {
+            if($userInboundEmail->id === $requestedInboundEmail->id) {
+                $hasAccessToInboundEmailAccount = true;
+                break;
+            }
+        }
+
```
sergio91pt: You can return early if `$hasAccessToInboundEmailAccount` is false, to avoid extra work, since this code forces the existence of an inbound account (ref #3741)

modules/Emails/EmailsController.php
```diff
@@ -627,4 +651,49 @@ protected function setAfterImport($importedEmailId, $request)
...
+        // Check that the from address is the same as the inbound email account
+        $isFromAddressTheSame = false;
+        if($inboundEmailStoredOptions['from_addr'] === $requestedEmail->from_addr) {
+            $isFromAddressTheSame = true;
+        }
+
+        // check if user is using the system account, as the email address for the system account
+        // is likely to be different
+        $outboundEmailAccount = new OutboundEmail();
```
sergio91pt: if `$isFromAddressTheSame === true` avoid doing extra work

sergio91pt:
I just took a look at the code. Give me a few days to test properly.

modules/Emails/EmailsController.php
daniel-samson: Adding save call. However we need to ensure that this will still show up in the list view when the status was draft.

modules/InboundEmail/InboundEmail.php
```diff
@@ -6046,6 +6056,11 @@ public function retrieveByGroupFolderId($groupFolderId)
     /**
      * Retrieves an array of I-E beans that the user has team access to
      */
+    /**
```
sergio91pt: Delete this line and the previous one

daniel-samson: Fixed

daniel-samson:
@sergio91pt Just wanted to thank you for offering to test this pull request. Let us know how you get on. We are about to release 7.9.2 soon. So we will merge this pull request in for 7.9.3.

modules/Emails/EmailsController.php
```diff
@@ -139,19 +139,51 @@ public function action_ComposeView()
...
     {
-        $this->bean = $this->bean->populateBeanFromRequest($this->bean, $_REQUEST);
-        $this->bean->save();
+        global $current_user;
+        global $app_strings;
 
-        $this->bean->handleMultipleFileAttachments();
+        $this->bean = $this->bean->populateBeanFromRequest($this->bean, $_REQUEST);
+        $inboundEmailAccount = new InboundEmail();
+        $inboundEmailAccount->retrieve($_REQUEST['inbound_email_id']);
```
Dillon-Brown: There seems to be multiple uses of the inbound_email_id and inbound_email_record parameter. inbound_email_id is not returning email id and is instead returning null.

sergio91pt: It's used by Detail/ListView and when importing and marking emails, but not when sending emails.

It is present in the "compose view" url when you forward or reply to an email. The url param value is blank with the Mozaik editor, but populated with the other 2 editors.

Either way, it doesn't seem to cause any problems. I tested without this PR (but with #3645) and the `inbound_email_record` id was ignored and emails would still be sent through the system account.

daniel-samson: :+1:

modules/Emails/EmailsController.php
```diff
@@ -627,4 +659,49 @@ protected function setAfterImport($importedEmailId, $request)
...
+        // Check that user is allowed to use inbound email account
+        $usersInboundEmailAccounts = $requestedInboundEmail->retrieveByGroupId($requestedUser->id);
+        foreach ($usersInboundEmailAccounts as $inboundEmailId => $userInboundEmail) {
+            if($userInboundEmail->id === $requestedInboundEmail->id) {
+                $hasAccessToInboundEmailAccount = true;
+                break;
+            }
+        }
+
+        $inboundEmailStoredOptions = $requestedInboundEmail->getStoredOptions();
```
daniel-samson: Added additional check.

modules/Emails/EmailsController.php
```diff
@@ -639,4 +673,78 @@ protected function setAfterImport($importedEmailId, $request)
...
+        // Check that the from address is the same as the inbound email account
+        $isFromAddressTheSame = false;
+        if ($inboundEmailStoredOptions['from_addr'] === $requestedEmail->from_addr) {
+            $isFromAddressTheSame = true;
+        }
+        // check if user is using the system account, as the email address for the system account
+        // is likely to be different
+        $outboundEmailAccount = new OutboundEmail();
+        if(empty($inboundEmailStoredOptions['outbound_email'])) {
+            $outboundEmailAccount->getUserMailerSettings();
```
sergio91pt: Email#setMailer calls OutboundEmail#getInboundMailerSettings, which has different logic for this case than OutboundEmail#getUserMailerSettings. Both methods also require a User as the first argument.

You can use `$outboundEmailAccount->getInboundMailerSettings($requestedUser, $requestedInboundEmail->id);` as it applies the same logic when the `outbound_email` is not set, or `$outboundEmailAccount->getInboundMailerSettings($requestedUser, '', $requestedInboundEmail->id);` to be clearer at the expense of an extra SELECT query.

modules/Emails/EmailsController.php
```diff
@@ -639,4 +673,78 @@ protected function setAfterImport($importedEmailId, $request)
...
+                break;
+            }
+        }
+
+        $inboundEmailStoredOptions = $requestedInboundEmail->getStoredOptions();
+
+        // if group email account check that user is allowed to use email account
+        if ($requestedInboundEmail->isGroupEmailAccount()) {
+            if ($inboundEmailStoredOptions['allow_outbound_group_usage'] === true) {
+                $hasAccessToInboundEmailAccount = true;
```
daniel-samson: I do not think that it is a good idea these methods, as they may be used by custom code.

modules/Emails/EmailsController.php
daniel-samson: The I was using this to get the system mailer settings. I have updated my code to use the system mail settings .

modules/Emails/EmailsController.php
```diff
@@ -639,4 +673,78 @@ protected function setAfterImport($importedEmailId, $request)
...
+        // is likely to be different
+        $outboundEmailAccount = new OutboundEmail();
+        if(empty($inboundEmailStoredOptions['outbound_email'])) {
+            $outboundEmailAccount->getUserMailerSettings();
+        } else {
+            $outboundEmailAccount->retrieve($inboundEmailStoredOptions['outbound_email']);
+        }
+
+        $isAllowedToUseOutboundEmail = false;
+        if ($outboundEmailAccount->type === 'system') {
```
daniel-samson: if the outbound email is null then we will not be able to send an email. As I have previous mentioned in comment above, I meant to fallback to the the system email and then check to see if the user is allowed to use the system account. I have updated my code to better reflect this. Please note that the $isAllowedToUseOutboundEmail is assigned to false. If the id is null it will not match the if conditions which assignthe variable to true. So this will deliberately fail the $hasAccess check.

sergio91pt: I think you misunderstood what I was trying to say here

$outboundEmailAccount->type may be null if the $outboundEmailAccount->retrieve is called with an id (that is not null) but doesn't exist. In this case getUserMailerSettings when called by Email#setMailer, will return the system account. That's why we need to do the system check when `$outboundEmailAccount->type === null`.

This is actually quite easy to get, by removing an outbound account that is used by an inbound account.

modules/Emails/EmailsController.php
sergio91pt: @daniel-samson I don't understand what you're trying to say. Can you elaborate please?

When I say that `retrieveAllByGroupId`/`retrieveAllByGroupIdWithGroupAccounts` are broken, is because they clearly do not do what the documentation says it does.

It says:
> Retrieves an array of I-E beans that the user has team access to including group

But it does:

Retrieves an array of I-E beans with all the group email accounts and (optionally) the personal accounts the user has access to.

modules/Emails/EmailsController.php
sergio91pt: @daniel-samson The options I listed are correct in this case and not `getSystemMailerSettings`, because OutboundEmail#getInboundMailerSettings will look for an outbound account with the same id as the inbound account, when `!isset($opts['outbound_email'])`.

You need to remember `userIsAllowedToSendEmail` does not choose the outbound account to use, so it needs to use the same logic the module uses when really sending email.

modules/Emails/EmailsController.php
gymad: Hi @sergio91pt 
$outboundEmailAccount->type could be 'system' or 'user', if it's anything else by default (for e.g if it's null) then the $isAllowedToUseOutboundEmail = false, and it change the return of function to false correctly.

sergio91pt: I see. Then we're changing the current behavior of defaulting to the system `outbound_email` account, when the wanted `outbound_email` does not exist. I actually have an account in this situation on my dev instance (it wasn't done by hand..).

In this case it may not make sense to blindly destroy the user's session, as the user may tried to use a non-existing outbound account associated with his inbound account, as opposed to trying to use an account he does not have access to.

modules/Emails/EmailsController.php
```diff
@@ -722,7 +722,7 @@ protected function userIsAllowedToSendEmail($requestedUser, $requestedInboundEma
         }
 
         $isAllowedToUseOutboundEmail = false;
-        if ($outboundEmailAccount->type === 'system') {
+        if ($outboundEmailAccount->type === 'system' || $outboundEmailAccount->type === 'system-override') {
```
sergio91pt: @daniel-samson nice find, but it should be in the else if clause  (`user`), since when using system-override accounts `isAllowUserAccessToSystemDefaultOutbound()` will probaly return false. They're used to specify your own username and password, when the CRM does not allow users to use the system account after all.

daniel-samson: Feature is removed in https://github.com/salesagility/SuiteCRM/pull/3873/ so we don't need it

modules/Emails/EmailsController.php
samus-aran: @sergio91pt Thanks for your scenario. We'll tackle that as a parallel issue.  Internal Reference is SCRMBT-20.  Believe there is additional 'type' not documented ;)

PR #3862, opened 2017-07-10
Fixed #3848 - get email by mailbox_id

### LINKED ISSUE #3848: SuiteCRM 7.9.2 Imported Email can load incorrect email from another user account
SuiteCRM 7.9.2 Imported Email can load incorrect email from another user account.  Seems to be only on emails that share same uid after an import.  This makes email module unusable with multiple users importing emails.

#### Issue
 Importing an email with uid=1 as USER1 will show any email with uid=1 as imported on every users's email list view.  Every user USER2, USER3, ect. will see the correct message for them in list view even if sharing a uid, but clicking will direct all users to the first imported email.

#### Expected Behavior
Multiple users can use email module

#### Actual Behavior
 Importing an email with uid=1 as USER1 will show any email with uid=1 as imported on every users's email list view.  Every user USER2, USER3 ect. will see the correct message for them in list view even if sharing a uid, but clicking will direct all users to the first imported email.

#### Possible Fix
Possible Fix could be *select using message_id or mailbox_id with uid.  Below is an example of my sql log
----------------------------
SELECT emails.*, users.user_name as assigned_user_name
 FROM emails
 LEFT JOIN users ON emails.assigned_user_id=users.id 
WHERE uid = '17' AND  emails.deleted=0 
 ORDER BY date_sent DESC
---------------------------
This pulls every email with the uid of 17 and forwards every user to the first value returned.  ***update*** uid and message number are generated on the imap server, so it will not be unique across different imap accounts.

#### Steps to Reproduce
SuiteCRM 7.9.2 Upgraded from 7.9.1.  Bug was present in 7.91 as well.

Create more than one user account.  Set up different IMAP boxes for each user.  Find a message on any user account that has the same uid as a message from another user account  (easier having only a dozen or so messages in each mailbox).  Import email on one account.  Then go to another user account and click on the message with the same uid, and it will direct you to the wrong email.

#### Context
I am trying to implement SuiteCRM as a solution for my company.  I need multiple users to import emails.

#### Your Environment
SuiteCRM 7.9.2 upgrade from 7.9.1 (bug was present in 7.9.1 as well)
MSSQL 2005
PHP 5.6.3
IIS Server

gymad:
## Description

get email by mailbox_id - show emails only for owner users
## Motivation and Context

please see related issue: #3848
## How To Test This

please see related issue: #3848
## Types of changes

### Final checklist

### DIFF
--- a/modules/Emails/EmailsController.php
+++ b/modules/Emails/EmailsController.php
@@ -317,7 +317,7 @@ class EmailsController extends SugarController
     {
         global $db;
         $emails = BeanFactory::getBean("Emails");
-        $result = $emails->get_full_list('', "uid = '{$db->quote($_REQUEST['uid'])}'");
+        $result = $emails->get_full_list('', "mailbox_id = '" . $db->quote($_REQUEST['inbound_email_record']) . "' AND uid = '{$db->quote($_REQUEST['uid'])}'");
         if (empty($result)) {
             $this->view = 'detailnonimported';
         } else {


cprider:
This worked, thank you!  A lingering issue with this bug, messages with the same uid are showing as imported across all user accounts.  This is my local fix:  
Change line 522 SuiteCRM/modules/Emails/include/ListView/ListViewDataEmails.php

  "emails.uid LIKE '" . $uid . "' and emails.mailbox_id LIKE '". $inboundEmail->id . "'");  //cprmod mssql fix and show only current mailbox import

PR #3869, opened 2017-07-10
Default/Inactive Email Account Issue

gymad:
## Description

If an Inbound Mail Account is marked as both Inactive, and is not selected as the Default, it still appears as an option on the "From" dropdown when Composing an Email address. (Despite being both Inactive and not chosen)
## Motivation and Context

ref(scrm 625)
## How To Test This

1) Create 2 inbound mail accounts via the "Settings" option on a User Profile
2) Ensure both can be seen when Composing under the "From" dropdown
3) Navigate back to "Settings" and set one as Inactive, (and ensure it is not set as default)
4) Navigate to Email Composing, and view the options under "From" dropdown.
## Types of changes

### Final checklist

### DIFF
--- a/modules/Emails/javascript/EmailUI.js
+++ b/modules/Emails/javascript/EmailUI.js
@@ -2796,6 +2796,7 @@ SE.folders = {
     /**
      * Updates user's group folder subscriptsion (Sugar only)
      * @param ieID The group folder to add to the tree view
+     * @see SE.folders.setFolderSelection()
      */
     retrieveGroupFolderSubscriptions : function() {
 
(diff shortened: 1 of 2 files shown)


modules/Emails/javascript/EmailUI.js
```diff
@@ -2804,6 +2805,8 @@ SE.folders = {
     	for(i=0;i<a_rs.length;i++)
     	{
     		var t_record = a_rs[i];
+          // if it's not working try something like this:
```
daniel-samson: Something about this comment smells bad. Has this been tested? It doesn't emit confidence.

Perhaps consider using a js doc in front of the code instead of this comment? 
```
  /**
   * @see SE.folders.setFolderSelection()
   */
```

PR #3875, opened 2017-07-12
disable user to edit other employees data

gymad:
## Description

disable users to edit other employees data
and some php error fixes
## Motivation and Context

users can modify other employees data
## How To Test This

try modify an other one employees data with a non-admin user
## Types of changes

### Final checklist


include/InlineEditing/InlineEditing.php
```diff
@@ -392,7 +397,7 @@ function formatDisplayValue($bean, $value, $vardef, $method = "save")
     }
 
     //If field is of type link and name.
-    if ($vardef['link'] && $vardef['type'] == "name" && $_REQUEST['view'] != "DetailView") {
+    if (isset($vardef['link']) && $vardef['link'] && $vardef['type'] == "name" && $_REQUEST['view'] != "DetailView") {
```
daniel-samson: It is possible to use === and !==. Consider changing them?

modules/Employees/Employee.php
```diff
@@ -295,6 +295,29 @@ function hasCustomFields()
...
+     * @param bool $check_notify
+     * @return bool|string
+     */
+    public function save($check_notify = false)
+    {
+        global $current_user;
+        if ($current_user->id) {
+            if (!is_admin($current_user)) {
+                if ($this->id && $current_user->id != $this->id) {
+                    $GLOBALS['log']->fatal("You can change only your own employee data.");
```
daniel-samson: Should this be a warn or a security instead of a fatal error?

As nothing is has broken down, instead it is a security violation.

Also It might be worth changing the message to include the name of the user, so that a administrator can see who is causing this issue.

modules/Home/controller.php
```diff
@@ -89,19 +89,26 @@ public function action_getValidationRules(){
 
                 $fielddef = $bean->field_defs[$_REQUEST['field']];
 
-                if(!$fielddef['required']){
+                if(!isset($fielddef['required']) || !$fielddef['required']){
                     $fielddef['required'] = false;
                 }
 
-                if($fielddef['name'] == "email1" || $fielddef['email2']){
+                if ($fielddef['name'] == "email1" || (isset($fielddef['email2']) && $fielddef['email2'])) {
```
daniel-samson: what about email3...email{n}?

Each email address which is posted to the controller will be set to email{n} in the bean. Should we use a regex match instead?

gymad: It's an other issue because its change the behavior of home controller. How should it works?

PR #3891, opened 2017-07-15
Fix #3890 make AOR and AOW modules more code-customizable

### LINKED ISSUE #3890: AOR and AOW modules not very code-customization friendly
Probably for historic reasons (these modules used to reside in `custom` dir themselves), some files that could be customizable are not using the standard mechanisms. This came up in the forums when a user was trying to customize.

So I just went and searched for all references and introduced the mechanism. This is a quick Saturday morning job and I must say _I didn't test_ at all. But since the change is simple I am confident that it works.

In case anybody wants to check that `get_custom_file_if_exists` function I call, [here it is](https://github.com/salesagility/SuiteCRM/blob/084f2f3f1f805cea2603323b1924f998ebffe0e9/include/utils.php#L5212-L5215).

A generic test of the AOW module and the AOR module should suffice to validate this. The includes and requires have to be right for it to work at all.

pgorod:
## Description
See Issue #3890 for explanation.

Basically, it makes some includes and requires in these two modules use the standard mechanism to pick up custom code from the custom directory.

## How To Test This
- reading the code with a good eye for missing parenthesis or typos
- generic test of AOR and AOW modules, just see if they're not hopelessly broken

## Types of changes

- [ X ] New feature (non-breaking change which adds functionality)

### Final checklist

- [ X ] I have read the [**How to Contribute**](https://suitecrm.com/wiki/index.php/Contributing_to_SuiteCRM) guidelines.


pgorod:
A [user in the forum](https://www.suitecrm.com/forum/suitecrm-7-0-discussion/15061-custom-aor-reports#50710)s tested this correctly for AOR (but didn't try AOW).

pgorod:
@Dillon-Brown I just resolved conflicts here.

gunnicom:
Good idea. Is the change for custom AOW_Actions, too?
 I tried adding custom Actions in custom/Extensions/ folder, but seems not to be rebuild, and in custom/modules/ folder they will be overwritten on Quick/Repair-Rebuild.

modules/AOW_Actions/actions/actionCreateRecord.php
```diff
@@ -282,19 +281,21 @@ public function set_record(SugarBean $record, SugarBean $bean, $params = array()
                                 break;
                         }
                         break;
-                    case 'Round_Robin':
```
samus-aran: @gymad - we happy with the deference of 'Case' versus 'case'? :)

gymad: You right @samus-aran, I didn't realized this one, fortunatelly it does not breaks anything cause both are a valid part of the php lang. Also I think we will having much more issue about this handreds line long messy switch..case..else..if labyrinth here.

pgorod: I fixed this in a couple of new commits, also fixing other tiny objections from Codacy.

pgorod:
@mattlorimer my PR is a bit superficial, I just edited the requires to look in `custom`, without much consideration about why it's requiring in the first place. To be honest I'd need to spend a lot more time understanding the code sufficiently to feel comfortable editing it more.

Do you think we can just merge this one as it is, and open a new issue with your comment to remind us to improve the things you mention later?

mattlorimer:
@pgorod defiantly not, it would be simple to do it correctly and if we merged it, it would put a bad practice / example into core

PR #3907, opened 2017-07-18
Fixed #1394 - Bug in Quote to Invoice Conversion

### LINKED ISSUE #1394: Bug in Quote to Invoice Conversion
I found a bug in Quote to Invoice conversion. When you convert to invoice it still links the quote group to the invoice. When you change something in quote, you will see that change in the Invoice. In code it is creating new group for invoice so those groups should relate with the new line items.

I hope you understand the problem and will fix this in your next release.

thanks

regards:
Ahmed

sgaved:
## Description

Previously, when creating an invoice, it would create new line items for each item in the quote but use the existing group. This meant that if you modified a quote then the totals on the invoice would be wrong. To reproduce bug:
1. Create a quote
2. Convert to invoice
3. Modify quote so that the total is different
4. Look at invoice and see the totals are now wrong
Relates to issue #1394 
## Motivation and Context

Fixes a bug
## How To Test This

Create a quote and then convert to invoice. Modify the original quote and then ensure that the totals are still correct on the invoice.
## Types of changes

### Final checklist

### DIFF
--- a/modules/AOS_Quotes/converToInvoice.php
+++ b/modules/AOS_Quotes/converToInvoice.php
@@ -83,7 +83,9 @@
 	//Setting Group Line Items
 	$sql = "SELECT * FROM aos_line_item_groups WHERE parent_type = 'AOS_Quotes' AND parent_id = '".$quote->id."' AND deleted = 0";
   	$result = $this->bean->db->query($sql);
+	$quoteToInvoiceGroupIds = array();
 	while ($row = $this->bean->db->fetchByAssoc($result)) {
+		$quoteGroupId = $row['id'];
 		$row['id'] = '';
 		$row['parent_id'] = $invoice->id;
 		$row['parent_type'] = 'AOS_Invoices';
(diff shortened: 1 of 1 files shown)


pgorod:
Link to the [Forum discussion](https://suitecrm.com/forum/suitecrm-7-0-discussion/15094-invoices-line-item-groups-point-on-quotes-line-item-groups-in-database-instead-of-actual-invoices-line-item-groups#50735) that originated this.

ajedvaj:
Yup, this seems to resolve this issue. 

Just one thing...
Have you tested it with line item groups disabled? 
Because than the $quoteToInvoiceGroupIds array is just an empty array and there is no group:
$quoteToInvoiceGroupIds[$row['group_id']] will be uninitialized or something like that. Not sure if it will give Notice..

It will probably result that the $row['group_id'] becomes empty string or null and system will work, but maybe it should be explicitly covered not to be left to wonder...

Maybe something like this:
$row['group_id'] = empty($quoteToInvoiceGroupIds) ? null : $quoteToInvoiceGroupIds[$row['group_id']]

Anyway, definitely thumbs up!

daniel-samson:
@Dillon-Brown Can you please assess?

PR #3914, opened 2017-07-19
Fix scrmbt 21 email check mssql

gymad:
## Description

This PR contains a fix for Compose Email Popup.
'From' field failed when trying to retrieve options for drop-drown select.
ref: SCRMBT-21

## Motivation and Context

Error log shows up a message: [FATAL] Exception in Controller: Cannot tell if the inbound email account is a personal account or a group account.

@daniel-samson: Can you please look at your two new function in https://github.com/salesagility/SuiteCRM/commit/73d21ca5ffde14256dcda033afdff4b269d3637a#diff-2aadc7f50e2480defbfecef7a0295a66
Problem was the database doesn't store string values in a bool field or mssql driver doesn't retrieve strings when the value is bool in this field: 'inbound_email.is_personal' but the IF condition with '===' check for strings and it's failed in each possible case.
In same time, these function returns are bool typed values (in PHPDoc above) so I assume we have two type of Email Accounts: personal/group. Is that correct?

## Types of changes

### Final checklist

### DIFF
--- a/include/SugarFolders/SugarFolders.php
+++ b/include/SugarFolders/SugarFolders.php
@@ -536,6 +536,7 @@ class SugarFolder
      * Builds up a metacollection of user/group folders to be passed to processor methods
      * @param object User object, defaults to $current_user
      * @return array Array of abstract folder objects
+     * @throws \SugarFolderEmptyException
      */
     public function retrieveFoldersForProcessing($user, $subscribed = true)
     {
--- a/modules/InboundEmail/InboundEmail.php
+++ b/modules/InboundEmail/InboundEmail.php
@@ -308,38 +308,16 @@ class InboundEmail extends SugarBean
 
     /**
      * @return bool
-     * @throws Exception
      */
     public function isPersonalEmailAccount() {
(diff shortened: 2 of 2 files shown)


pgorod:
@gymad will this, by any chance, also fix #3741?

gymad:
@pgorod I don't think so, as I see: my commit changes the InboundEmail::isGroupEmailAccount() and InboundEmail::isPersonalEmailAccount() functions only. These function called from EmailsController::action_GetFromFields() and it's called via ajax in modules/Emails/include/ComposeView/EmailsComposeView.js at $.fn.EmailsComposeView / self.construct(). So when the compose view generates options for "From" drop-down field and call the getFromFields action which is goes trough on non-personal 'bounce' mailbox typed elements of inbound email accounts (see in InboundEmail::retrieveAllByGroupIdWithGroupAccounts()) and pass the results in the ajax response. In this case in the drop-down should contains Inbound Email Accounts so I don't know how the "From" field relates to the outbound accounts as you mention in your issue #3741. Probably it is a separated issue.

PR #3918, opened 2017-07-19
Fixing getFileName and layouts in module builder and studio.

daniel-samson:
## Description

This change fixes some issue with module builder and studio.  I have also added PHP doc to the files I have changed and updated  the license headers in each class. I have only updated the  access modifiers, where I believe it is safe to do so.

Issue 1:
When you create a new module in module builder, you are unable to select any of the layouts or edit fields. 

Issue 2:
When you create a new module in module builder, the filter icons displays the incorrect icon and the text is blank.

Issue 3:
I have updated the classes which call getFileName method call so that they handle the new packageName variable.

## Motivation and Context

* Fixes some issues

## How To Test This

* Repair and Rebuild
* Test each feature for
    * Module Builder
    * Studio

## Types of changes

### Final checklist

### DIFF
--- a/modules/ModuleBuilder/MB/MBLanguage.php
+++ b/modules/ModuleBuilder/MB/MBLanguage.php
@@ -181,7 +181,7 @@ class MBLanguage{
                 'LNK_NEW_RECORD'=>$mod_strings['LBL_CREATE'] ." ". $this->label,
                 'LNK_LIST'=>$mod_strings['LBL_VIEW'] ." ". $this->label,
                 'LNK_IMPORT_'.strtoupper($this->key_name)=>translate('LBL_IMPORT') ." ". $this->label,
-                'LBL_SEARCH_FORM_TITLE'=>$mod_strings['LBL_SEARCH'] ." ". $this->label,
+                'LBL_SEARCH_FORM_TITLE'=>$mod_strings['LBL_SEARCH_BUTTON'] ." ". $this->label,
                 'LBL_HISTORY_SUBPANEL_TITLE'=>$mod_strings['LBL_HISTORY'],
                 'LBL_ACTIVITIES_SUBPANEL_TITLE'=>$mod_strings['LBL_ACTIVITIES'],
                 'LBL_'.strtoupper($this->key_name).'_SUBPANEL_TITLE'=>$this->label,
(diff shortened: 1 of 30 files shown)


pgorod:
That an interesting question that @gymad raises - these PR's are harder to read and understand when so much code cleanup is in there too. It is difficult to see exactly what the PR is intending to change, and what are just formatting changes.

But we need to be careful with code cleanups because they also require testing. We've been bitten by this more than once - a tiny mistake in a change that was not intended to change the way the code works, but ends up breaking things (like that query that lost a space character in 7.9.0...).

So we would probably want to be able to do code reviews with separated fixes/code cleanups, but then test them together (if they affect the same files). I don't know if this is practical, or exactly how it be achieved...

horus68:
@gymad On code cleanup I made the same request for language files: all SuiteCRM code files should have a code cleanup version, as in full files edit without code changes and a new version should be released for that. 
This is needed now because of the Coding standards approved and once every year for copyright statements.
This will avoid having to change each files when there is a need to change the code on that file. But then I found that the option was to only a file change when it was needed. This do not allow quality and time speed.

Code cleanup required to follow the Coding Standards https://suitecrm.com/wiki/index.php/Coding_Standards:
- License header for every file (languages are now done)
- Entry point if - die (languages are now done)
- Copyright version - this needs to be done every January (languages was updated for 2017 on version 7.9.3)
- Code Indentation normalization (not done!)
- Use single quotes in language files + Arrays formatting   (not done!)

I believe this would take some days for the same person but will avoid any further time lost to everyone else.
Can you please discuss this internally with SalesAgility?

fernengel:
Hi!

Sorry I'm a bit late on commenting...
I originally added a comment to issue #4032 (layout editing does not work on Module Builder).

Unfortunately I don't know yet how to get the pull request to test your fix, but I corrected the DeployedMetaDataImplementation.php-file manually just to see if it works (added argument $packageName to getFileName()).

Therefore I might have missed other fixes of your pull request...but my question is:
Is the fix tested sufficiently so that both Module Builder AND Studio work?
Maybe I broke my SuiteCRM, but I'm having problems with editing the Layouts in "Studio".
Also, even though the Module Builder now loads the layout-editing-window again, it does not seem to work reliably. It replaces fields in the layout always with the field "Modified By Name".

So, don't panic :) Maybe I just broke my instance.
But anybody who tested the fix...have you made sure, Module Builder AND Studio are working and are actually usable?

Thanks a lot!

Best regards
David

JBrace1990:
This has also broken all of our tabs and turns them into panels whenever the views are updated.

daniel-samson:
see #4097

Issue #3904, opened 2017-07-18
7.9.3 Cant send Emails Out


petedyer:
7.9.3

Since upgrade cant send any messages out. The following appears in the log file User xxx attempted to send an email using incorrect email account settings in which they do not have access to. This didn't happen in 7.9.2

pgorod:
I think (not 100% sure, though), that there is a change requiring individual users to send email via their own accounts, instead of allowing a system account like before. So maybe you can check if that user's Email configurations.

sergio91pt:
Are you trying to send email through a personal account? Try to re-enter the outbound settings, as if your account was linked to a non-existent outbound account (or removed), you'll get that message (#3812  - SCRMBT-20).

petedyer:
Hi no its not the personal account its a set of three different outbound accounts. Cant send any emails with them. Gone back in and edited and saved them but not getting any further. Only occurring since last nights update

pgorod:
@petedyer where exactly did you define the settings for those accounts?

Mausino:
@petedyer  has your issue something with #3901 ???

petedyer:
I've attached a screenshot. Group settings are done via group settings. Outbound via the outbound. Then from profile(admin user in this case) allow access to the three group emails.
![email-screenshot-issue](https://user-images.githubusercontent.com/17161452/28321906-76568636-6bcc-11e7-890c-ffaab207be63.jpg)
 Note this is a system thats not a clean install but one weve been upgrading every since early days of suite(in case this makes a difference)

petedyer:
Hi @Mausino  we are using mysql and no we do get the group emails appearing the from dropdown so this doesn't happen to us

petedyer:
Is it something to do with the security in  terms of the fact these three are group accounts and appear in the from compose field BUT these three are not set in the outbound settings(well one is)

petedyer:
Hi #pgorod are we not meant to be using outbound accounts but using individual accounts to send out out email - bit confised here so can someone kindly clear this up.

halest:
FYI, I had to comment out the whole `$this->userIsAllowedToSendEmail` function on EmailController.php Line 149 (changed it to `if (1===1)` ) to get it so send mail for the moment, otherwise it would just time out on the ajax-dialogue

petedyer:
[...]

PR #3921, opened 2017-07-19
Hotfix 3904 email sending issue

gymad:
## Description

This PR is a possible fix for issue #3904, contains:
 - a fix in user access check for sending email
 - show more informative validation message in log

## Types of changes

### Final checklist

### DIFF
--- a/modules/Emails/EmailsController.php
+++ b/modules/Emails/EmailsController.php
@@ -813,12 +813,14 @@ class EmailsController extends SugarController
 
         $inboundEmailStoredOptions = $requestedInboundEmail->getStoredOptions();
 
-        // if group email account, check that user is allowed to use group email account
-        if ($requestedInboundEmail->isGroupEmailAccount()) {
-            if ($inboundEmailStoredOptions['allow_outbound_group_usage'] === true) {
-                $hasAccessToInboundEmailAccount = true;
-            } else {
-                $hasAccessToInboundEmailAccount = false;
+        if (!$hasAccessToInboundEmailAccount) {
+            // if group email account, check that user is allowed to use group email account
+            if ($requestedInboundEmail->isGroupEmailAccount()) {
+                if ($inboundEmailStoredOptions['allow_outbound_group_usage'] === true) {
+                    $hasAccessToInboundEmailAccount = true;
+                } else {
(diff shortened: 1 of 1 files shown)


modules/Emails/EmailsController.php
```diff
@@ -813,12 +814,14 @@ protected function userIsAllowedToSendEmail($requestedUser, $requestedInboundEma
 
         $inboundEmailStoredOptions = $requestedInboundEmail->getStoredOptions();
 
-        // if group email account, check that user is allowed to use group email account
-        if ($requestedInboundEmail->isGroupEmailAccount()) {
-            if ($inboundEmailStoredOptions['allow_outbound_group_usage'] === true) {
-                $hasAccessToInboundEmailAccount = true;
-            } else {
-                $hasAccessToInboundEmailAccount = false;
+        if (!$hasAccessToInboundEmailAccount) {
```
sergio91pt: Logically this doesn't make sense. retrieveAllByGroupIdWithGroupAccounts is supposed to retrieve all group (and personal) accounts the user has access but right now it's broken and retrieves all group accounts regardless if the user has access or not.

So, with the current broken behavior, $hasAccessToInboundEmailAccount is always true for group inbound accounts, so the next if is equivalient to `if (false)`.

If retrieveAllByGroupIdWithGroupAccounts performed as documented, `if (!$hasAccessToInboundEmailAccount)` means the user really doesn't have access to the inbound group account, but we allow access to the outbound account if the inbound account allows outbound access to their group members (although the user doesn't belong to this group).

gymad: The $requestedInboundEmail could be anything and the retrieveAllByGroupIdWithGroupAccounts should return only the non-personal accounts. Btw you'r right, this part logically removable if isGroupEmailAccount() will works properly.

edwinlee2000: I applied this to 7.9.7 and still doesn't work.
Configured both system and personal outgoing email, tested ok. But got error when trying to compose, then send.
UI showed: "Error Sending Email. Please contact your administrator for assistance."
Log showed: [SECURITY] User is not allowed to send email
Reason was:
User is not allowed to use outbound email.
[SECURITY] User XXX attempted to send an email using incorrect email account settings in which they do not have access to.
Please help help !

PR #4187, opened 2017-08-30
Password security formats

gymad:
## Description

Adding more password formatting to password settings page on admin.
ref: scrm-744
## Types of changes

### Final checklist


modules/Users/password_utils.php
```diff
@@ -1,9 +1,10 @@
 <?php
-if(!defined('sugarEntry') || !sugarEntry) die('Not A Valid Entry Point');
+if (!defined('sugarEntry') || !sugarEntry) {
+    die('Not A Valid Entry Point');
+}
 /*********************************************************************************
```
daniel-samson: You need to update the license header so that the license is at the top

modules/Users/password_utils.php
```diff
@@ -39,98 +40,106 @@
...
+    $current_user->retrieve($usr_id);
+    $type = 'syst';
+
+    if ($current_user->portal_only == '0') {
+        global $mod_strings;
+        $res = $GLOBALS['sugar_config']['passwordsetting'];
+        if ($type != '') {
+            switch ($res[$type . 'expiration']) {
+
+                case '1':
```
daniel-samson: It would be nice to use a class with const members to better describe what the numbers mean. Like an enum

themes/SuiteP/css/editview.scss
```diff
@@ -1191,6 +1191,8 @@ p.error {
 
 .error {
   border: none;
+  color: #f08377;
+  background: transparent;
```
daniel-samson: This will change all error messages. Consider creating a specialisation class for this purpose?

modules/Users/password_utils.php
gymad: Yes, I aggree but I don't know what does it mean, as you see it already there. This part of PR contains a PSR code cleanup only. May we have to take the clean up commits in separated PR?

PR #4196, opened 2017-09-03
Fixed #4146 - Outbound Emails SMTP Connect

### LINKED ISSUE #4146: Outbound Emails SMTP Connect ()
#### Issue
Cannot send outbound emails.

#### Expected Behavior
The outbound emails should be working.

#### Actual Behavior
I used to use a web hosting company for my email. I recently switched to gmail and now have SSL/TLS auth. But it is not working.

#### Steps to Reproduce
Sending any email from anywhere gives an error message.
This includes sending out emails from the admin section in setting up outbound emails with trying to test. As well as just saving the settings there assuming they are correct and trying to send an email from the quote section by selecting "Email Quotation"

#### Version Info
SuiteCRM 7.9.4

Error Message:
Wed Aug 23 04:47:31 2017 [5628][5a6f3c04-6cdc-72e2-d455-58ca04892a20][FATAL] SugarPHPMailer encountered an error: The outbound mail server selected for the mail account you are using is invalid. Check the settings or select a different mail server for the mail account.
Wed Aug 23 04:47:31 2017 [5628][5a6f3c04-6cdc-72e2-d455-58ca04892a20][FATAL] SugarPHPMailer encountered an error: SMTP connect() failed. https://github.com/PHPMailer/PHPMailer/wiki/Troubleshooting

image

stancel:
This pull request fixes the ability to properly connect to an SMTP Server (ex. Gmail/GSuite) where the hosted SuiteCRM site does not have a valid cert setup. The issue was seen and reported by @mruch2 and after debugging realized that the only way to solve the issue was by passing an SMTP SSL option of verify_peer = false. Hopefully this fix will help others including those setting up a local dev/testing environment to help move SuiteCRM along.

## Description

The fix merges in an SMTP options array into the options array parameter and sets the SSL 'verify peer' option to false as the default. It then passes that SMTP options array to the parent function, which is then properly handled by the PHPMailer library.

## Motivation and Context

Please see the referenced issue #4146 for the motivation and context along with @pgorod. 

## How To Test This

1) Login to SuiteCRM as an Admin
2) Navigate to Admin --> Outbound Email 
3) Add a new or Edit the existing 'system' outbound email account, using SMTP (Gmail/GSuite was used in testing)
4) Save the account settings (username, password, etc.) and validate that everything save correctly
5) Edit the account and click the "Send Test Email" button 
6) The email should now properly go through with no errors thrown.

## Types of changes

### Final checklist

### DIFF
--- a/include/SugarPHPMailer.php
+++ b/include/SugarPHPMailer.php
@@ -359,7 +359,15 @@ eoq;
      */
     public function smtpConnect($options = array())
     {
-        $connection = parent::smtpConnect();
+        // Proposed Fix for #4146 - Outbound Emails SMTP Connect () 
+        // https://github.com/salesagility/SuiteCRM/issues/4146
+        $option2 = array(
+                'ssl' => array(
+                    'verify_peer' => false
+                )
+            );
+        $mergedArray = array_merge($options, $option2);       
+        $connection = parent::smtpConnect($mergedArray);
         if (!$connection) {
             global $app_strings;
             if (isset($this->oe) && $this->oe->type === 'system') {


pgorod:
What exactly are the security implications of this?

stancel:
@pgorod - This essentially disable checking of the certificates, ***which I would not recommend in a PROD environment but only in a DEV environment***. I don't think putting this code here is a bad idea if it is commented out and only uncommented by someone in their development environment like the reporting user @mruch2 has done (notice allow edits from maintainers is enabled). 

Ideally everyone would use a cert that was signed by a recognized CA or create a self signed one and add it as trusted in their OS. Teaching someone how to do these steps is outside of the scope of helping them get their SMTP connection working inside SuiteCRM though. By adding this code here (it should be commented out though) and telling people using local/development environments to only try uncommenting this code to see if it resolves their connection issues could be very useful though. 

Take the pull request and comment out the code to help troubleshooting if you'd like or leave it. Either way, I was just trying to help.

pgorod:
You _have_ been helpful and I believe we made progress on the original issue. But I'm still scratching my head at some aspects of this problem, and of this solution.

I don't think the current PR can go into the code like this, because of the way it affects security decisions, but let's let a few days pass and see if the arrival of Monday brings some help from other people to reach a better diagnosis and solution.

Thanks for all your contribution.

isleshocky77:
Just going through some email related issues and ran into this. I would throw my hat in to say this should NOT go into the codebase.

@stancel I run into this issue all the time as well, and you will in most software packages. You can and are better off creating your own signed cert and trusting it on your local machine for development. To have something like this in the code base is extremely bad for security. If you need more help, google "Creating your own Certificate Authority".  You'll then be able to install that root certificate in development browsers, machines, curl, etc. OR only enable it while developing.

pgorod:
I agree and I suggest to @Dillon-Brown that this could be closed. I can't see this option going into our code. We already have these online discussions that can be found on Google, and help anyone who faces this problem and is willing to take a risky approach to workaround it.

Thanks to all involved!

PR #4228, opened 2017-09-08
Added return instruction to One2MRelationship::add() function

ajedvaj:
Set return value of function as the value returned from call to parent::add() function

## Description

One2MRelationship::add() functions last command is to call parent::add() function but it does not return value which is returned from parent::add(). 
Documentation states that it should return boolean value:
`@return boolean true if successful`

## Motivation and Context

Function needs to return value as documented and intended to do so.

## How To Test This

Make a call to create relationship between beans.

## Types of changes

### Final checklist

### DIFF
--- a/data/Relationships/One2MRelationship.php
+++ b/data/Relationships/One2MRelationship.php
@@ -160,7 +160,7 @@ class One2MRelationship extends M2MRelationship
             }
             
             // Add relationship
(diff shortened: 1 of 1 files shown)


daniel-samson:
@Dillon-Brown can you please label ready to merge?

PR #4266, opened 2017-09-16
Allow 'function' fields in email and pdf templates

HVStechnik:
#### Issue
E-mails send by the Workflow module based on an Email template do not contain function fields (source = 'function' in vardef definition). Those fields are not automatically calculated, when the bean is initiated and the db row retrieved. For standard views the function fields are calculated during displaying. For email templates they however stayed blank.

#### Solution
Update the templateParser.php file within the AOS_PDF_Templates module in order to calculate function fields while creating the replacement array for the template markers.
The solution is similar to the calculation of those fields in /include/EditView/EditView2.php lines 1000 to 1040

## How To Test This
Create an email template including a function field from a vardef and try to send it automatically.

## Types of changes

### Final checklist

Thanks to pgorod for guiding me through my first git pull request!


HVStechnik:
Sorry, I do not get the codecov report here. What is the problem? What do I have to do in order to have this pull request merged? 
Best regards

Mac-Rae:

HVStechnik:
How do I check the clahub agreement? I have signed the agreement myself already. How can the "commit's parents" not have signed the CLA if I base on the SA SuiteCRM develop branch?

Dillon-Brown:
Hi @HVStechnik, the issue with the CLAHub is that your commit is coming from what Git sees as two different authors, "HVStechnik" and "Max Wilckens". I would suggest either editing the author of the last commit and force pushing back up or doing a rebase (This may be of some assistance: https://www.git-tower.com/learn/git/faq/change-author-name-email)

Let me know if you run into any difficulties, I can rebase this for you with just the "HVStechnik" account as the author if that makes it easier. Thanks!

HVStechnik:
Thanks @Dillon-Brown, I managed to change the author, but now travis fails. I think that the php5.6 build currently fails for all new pull requests....

pgorod:
@Dillon-Brown it would be nice to merge this one, I think it's low risk and useful.

You don't have to label it "Enhancement", in my opinion, since it's technically a bug to have fields that conform to a spec, and render well in some views but fail when sent via email. This doesn't require the extra "Enhancement" review, I believe...

HVStechnik:
@pgorod I reviewed the custom function options once again and got a bit confused:

Within the vardefs there are three different versions of how to include a custom function to calculate a field value:
1. ```$vardefs[$field]['function']```
The ```function``` vardef key may be an array that stores a file to be included (key: ```include```), the function to be called (key: ```name```), and an optional return mode (key: ```returns```: if set to 'html' the function's return value replaces the field's value, otherwise the function's return value is set as the ```options``` key of the vardefs array and defines e.g., the options in an enum field).
The code is implemented in [EditView2](https://github.com/salesagility/SuiteCRM/blob/3591f7c09190a4a5c17c37d60e4c598d1157ad57/include/EditView/EditView2.php#L651-L655), [InlineEditing](https://github.com/salesagility/SuiteCRM/blob/c9d2132df8db5955a0809301dacb9fb5db872b6c/include/InlineEditing/InlineEditing.php#L216-L228), [MassUpdate](https://github.com/salesagility/SuiteCRM/blob/c9d2132df8db5955a0809301dacb9fb5db872b6c/include/MassUpdate.php#L711-L728), [SearchForm2](https://github.com/salesagility/SuiteCRM/blob/c9d2132df8db5955a0809301dacb9fb5db872b6c/include/SearchForm/SearchForm2.php#L670-L699), and in [SugarFieldBase](https://github.com/salesagility/SuiteCRM/blob/ef1b41965c056937894dd3018d3b149c9c25a9a9/include/SugarFields/Fields/Base/SugarFieldBase.php)
It seems to be called in EditViews, but I have no idea whether it is also called in Detail or List views (?) With ```'returns' => 'html'``` it should be similar to the other two options:

2. ```$vardefs[$field]['source'] == 'function'```
This is, what I was adressing in this PR. If ```['source'] == 'function'```, SugarBean looks for ```function_class```, ```function_name```, and ```function_params``` in the vardef array.
This is implemented in [SugarBean](https://github.com/salesagility/SuiteCRM/blob/c9d2132df8db5955a0809301dacb9fb5db872b6c/data/SugarBean.php#L1365-L1411) and [EditView2](https://github.com/salesagility/SuiteCRM/blob/1391bd278c0aa101ae297d27afda4668f7cdbc5b/include/EditView/EditView2.php#L1004-L1046). In SugarBean it is implemented in the process_union_list_query function, which should call the function in ListViews. DetailViews2 extends the EditView class (file: EditView2.php) and thus also implements this code.

3. ```['type'] == 'function'```
Last, but not least... This has the same parameter as ```['source'] == 'function'``` (```function_class```, ```function_name```, and ```function_params```)
This is implemented in [SearchForm2](https://github.com/salesagility/SuiteCRM/blob/c9d2132df8db5955a0809301dacb9fb5db872b6c/include/SearchForm/SearchForm2.php#L703). But I think this version has the disadvantage, that the approapriate SugarField should not be rendered any more (because not specific type is given). I assume that it should read ```['source'] == 'function'``` in [SearchForm2](https://github.com/salesagility/SuiteCRM/blob/c9d2132df8db5955a0809301dacb9fb5db872b6c/include/SearchForm/SearchForm2.php#L703)

I would very much recommend to simplify this a little bit:
I think the ```'returns' => 'html'``` cases should better implement option (2). The other implementations of (1) could be simplified by a utils function that implements the redundant code. And (3) potentially just is a bug...

After having done this review, I think we might want to leverage the [callFunction function in EditView](https://github.com/salesagility/SuiteCRM/blob/3591f7c09190a4a5c17c37d60e4c598d1157ad57/include/EditView/EditView2.php#L1004-L1046) that does exactly the same as implemented by this PR. Would it be an option to move that function to SugarBean.php and call it whenever a custom function field needs to be evaluated?
What do you think?

pgorod:
@HVStechnik thanks for taking the time to study this. I guess some reorganization would be beneficial. We'd have to be careful not to make the changes break any existing customizations that people might have out there, relying on one of these mechanisms.

Another thing that would be quite useful would be to improve (or start?) some Documentation about this... I don't think there's any place really teaching how to use this.

Issue #4276, opened 2017-09-18
v7.8.6 LTS - User Management dropdown menu edit button doesn't work


JBrace1990:
#### Issue

When in the user management screen (http://atlas.c-4analytics.com/index.php?module=Users&action=index) and viewing a user (http://atlas.c-4analytics.com/index.php?module=Users&offset=1&stamp=1505740554063185800&return_module=Users&action=DetailView&record=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx), the Actions -> Edit button throws a JS error.

~~As a note, this potentially also affects the quick edit for users, which is not working as intended. This could be duplicated in #4270, but I'm unsure as to whether this is one root issue or 2.~~ (Edit: Fix has nothing to do with this)

#### Expected Behavior

The edit button should take you to the edit screen

#### Actual Behavior

Throws a JS error

#### Possible Fix

The ID of the record is not quoted, so JS throws a "Uncaught SyntaxError: Invalid or unexpected token" error

#### Steps to Reproduce

1. Go to the User Management screen under Admin
2. View a users Record
3. Click "Actions"
4. Click "Edit"

#### Context

Makes it impossible for normal users without JS knowledge to edit a user account. 

#### Your Environment

* SuiteCRM Version used: 7.8.6 (Latest LTS)
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): N/a
* Environment name and version (e.g. MySQL, PHP 7): N/a
* Operating System and version (e.g Ubuntu 16.04): N/a

Dillon-Brown:
Looks like this is also present in 7.9.5

JBrace1990:
The possibility is there, since they contain a lot of the same edits. 

I found the commit that this happened in: https://github.com/salesagility/SuiteCRM/commit/831b56fa70b677060d280ebcaa56dd2f31498cda line 117. The single quotes are just on the wrong side.

This looks like it's been fixed in Master now, since I just went to edit it on a branch and it's fine.

JBrace1990:
Create a PR for this that fixes the issue #4277

Dillon-Brown:
Duplicate of #4232

Issue #4232, opened 2017-09-10
User Detail - Edit action stopped working after 7.8.6 upgrade


kaasig-nv:
#### Issue
When I navigate to the User detail page (Admin > User Management > User Detail > Action > Edit ), the edit action is not working anymore, the action is broken with a javascript error.

#### Expected Behavior
On click on edit, the application should navigate to the edit user page.

#### Actual Behavior
The page stays as it is

#### Possible Fix
suitecrm/modules/Users/views/view.detail.php
Line no 117 and 121, the id field has to be in quotes

this.form.return_id.value='".'{$fields.id.value}'."';
![usereditaction](https://user-images.githubusercontent.com/25651530/30253633-3d422184-96dd-11e7-828b-b47cc5bf9bf8.png)

#### Steps to Reproduce
1.Go to Admin Options
2.Click on User Management
3.Click on a user to view detail record
4.Under Actions, click on Edit action.

#### Context
Trying to edit a user detail record

#### Your Environment

* SuiteCRM Version used: 7.8.6
* Browser name and version : Google Chrome 60.0.3112.90
* Environment name and version : MariaDb 10.1, PHP 7.0
* Operating System and version : Debian Jessie

PR #4277, opened 2017-09-18
SuiteCRM #4276 - Fix for edit button not working to edit users.

JBrace1990:
This is a potential fix for #4276.


Dillon-Brown:
Looks like this has already been resolved by #4232

PR #4289, opened 2017-09-20
Fixed #4287 Use the correct module for module name translations.

### LINKED ISSUE #4287: Module Name Incorrect in Assigned Emails
#### Issue

It's sometimes possible for the module name to be incorrect in assignment notifications or not filled in at all leading to an email with the subject "Assigned Module Name" rather than, for example "Assigned Case". This is caused in cases where the current module is not the same as the one that is being assigned (as may be the case in Workflow or logic hooks).
#### Expected Behavior

The module name in emails should always match the module for which the user is assigned.
#### Actual Behavior

Module name in the email instead refers to the current module, which may differ from the module that the email is sent out for.
#### Possible Fix

The create_notification_email in SugarBean has the following line:
`$xtpl->assign("OBJECT", translate('LBL_MODULE_NAME');`
This will use the current mod_strings rather than the mod strings of the bean. Instead it should be:
`$xtpl->assign("OBJECT", translate('LBL_MODULE_NAME',$this->module_name));`

#### Steps to Reproduce

1. Create a workflow that triggers on save of a contact
2. Have this workflow set the assigned user of the contact's account.
3. Save a new contact with a known account.
4. Observe the received email which will have incorrect copy.

#### Your Environment

* SuiteCRM Version used: 7.8.5
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Firefox
* Environment name and version (e.g. MySQL, PHP 7): MySQL, PHP 5.6

JimMackin:
## Description

Use the appropriate module when translating the module name in assignment notications. Previously this would rely on the `mod_strings` matching the current bean which is not always the case.

## Motivation and Context

Fixed #4287 - module name is incorrect in assignment emails in some conditions.
## How To Test This

Follow the steps in #4287 - It should now show the correct module name in the email.
## Types of changes

### Final checklist

### DIFF
--- a/data/SugarBean.php
+++ b/data/SugarBean.php
@@ -2741,7 +2741,7 @@ class SugarBean
         if (in_array('set_notification_body', get_class_methods($this))) {
             $xtpl = $this->set_notification_body($xtpl, $this);
         } else {
-            $xtpl->assign("OBJECT", translate('LBL_MODULE_NAME'));
+            $xtpl->assign("OBJECT", translate('LBL_MODULE_NAME',$this->module_name));
             $template_name = "Default";
         }
         if (!empty($_SESSION["special_notification"]) && $_SESSION["special_notification"]) {


horus68:
@Dillon-Brown up!

PR #4344, opened 2017-09-28
Make Business Hours less mysterious

pgorod:
## Description
There is no Issue for this PR, I'm assuming that isn't strictly necessary.

## Motivation and Context
Nobody knows what "Business Hours" does. Yet after discussing this in the forums with a user that actually needed this feature, I tried a few tests to understand the feature, and here I basically improve the messages to the user to be more helpful.

## How To Test This
Check `Admin` screen, new message there.
Click `Business Hours`, new message there also.

I'm pushing this `hotfix` because it's such a simple fix I don't think it needs to wait until a development release, we can look at this as a bugfix in a message :-)

P.S. - I also took the chance to remove a few full-stops on the Admin screen descriptions, for consistency with the rest of the messages


pgorod:
@Dillon-Brown can I leave this for `hotfix` branch, or should I change it back to `hotfix-7.8.x`? I only changed it because git was giving me trouble, I didn't have a clean branch when I started the PR, and I couldn't solve it. I will basically re-create this from an up-to-date `hotfix-7.8.x` if that's necessary, is it?

Dillon-Brown:
@pgorod It's not strictly necessary as I can duplicate this PR for the hotfix-7.8.x branch and then merge them both at the same time. It would make things easier / get it merged faster if you re-created it though.

horus68:
> P.S. - I also took the chance to remove a few full-stops on the Admin screen descriptions, for consistency with the rest of the messages

When a string is a description its better to have a full stop. Full stops are used to indicate the end of sentences which are not questions or exclamations.
As for consistency there are so many sentences (see include language file) with those full stops that I don't know what is the rule in SuiteCRM!

pgorod:
I think I'll have to disagree with you on this one, @horus68... these aren't really sentences, just labels, so the full-stops seem too much. At least when I look at that Admin screen I feel it looks better without the full stops. 

Anyway for me this is just a matter of taste and I don't have strong opinions on the issue, so if you or anyone wants to go ahead and add full stops everywhere, that's ok with me.

horus68:
No problem for me. I only commented as they are in strings with a `_DESC` text in the key, so I assumed they are descriptive, not labels.
And labels should not have full stops!

likhobory:
@pgorod , I tried to test this feature, but wokrflow works at any case. (Version 7.9.5)


Conditions & actions:


-----------------------
Result:

pgorod:
@likhobory I tested with a Workflow set like this:

Run: `Only In The Scheduler`
Run On: `New Records `

Which settings do you have here? Can you try with these, to see if Business hours are respected? Thanks.

likhobory:
Thanks, you're right, it works as expected with 'Only In The Scheduler'  option.

pgorod:
I added an even better string, this needs to be 100% clear, why not?

pgorod:
I added an even better string, this needs to be 100% clear, why not?

I didn't test - I believe those escaped apostrophes won't break anything, but somebody please try it live...

mattlorimer:
so you mean there is a bug that business hours doesn't work on workflows running "on save"?

pgorod:
@mattlorimer yes I just tested a "On Save", new records, and it was executed immediately, despite being off work hours.

pgorod:
@mattlorimer is it really a bug if Business Hours doesn't restrict `On save` workflows? Isn't it possibly intentional? 

Anyway, that needs to be clarified before @Dillon-Brown copies this PR to the 7.8 branch, to determine if the current message (which mentions specifically only scheduler workflows) is correct or misleading...

mattlorimer:
it wasn't intentional, just not implemented yet

modules/Administration/language/en_us.lang.php
```diff
@@ -839,7 +839,7 @@
     'LBL_AOP_JOOMLA_ACCOUNT_CREATION_EMAIL_TEMPLATE' => 'Joomla Support Portal Account Creation Template',
 
     'LBL_AOP_BUSINESS_HOURS_SETTINGS' => 'Business Hours',
-    'LBL_AOP_BUSINESS_HOURS_DESC' => 'Change the business hours',
+    'LBL_AOP_BUSINESS_HOURS_DESC' => 'Restrict \'Only In The Scheduler\' Workflows to launch only in certain days and times',
```
Dillon-Brown: Use single and double quotes when appropriate. You shouldn't need to escape quotes within a string.

pgorod: Ok, I fixed this one and another similar case in the same file. Did I do it right?

modules/Administration/language/en_us.lang.php
```diff
@@ -839,7 +839,7 @@
     'LBL_AOP_JOOMLA_ACCOUNT_CREATION_EMAIL_TEMPLATE' => 'Joomla Support Portal Account Creation Template',
 
     'LBL_AOP_BUSINESS_HOURS_SETTINGS' => 'Business Hours',
-    'LBL_AOP_BUSINESS_HOURS_DESC' => 'Restrict \'Only In The Scheduler\' Workflows to launch only in certain days and times',
+    'LBL_AOP_BUSINESS_HOURS_DESC' => "Restrict 'Only In The Scheduler' Workflows to launch only in certain days and times",
```
horus68: Isn't the value test to be enclosed into single quotes, not double?
I think the most fashioned way is to use
'LBL_LABEL' => 'text "special" more text'

pgorod: I'll have to test it, but if I want to show single quotes (apostrophes) on the screen, it should be possible...

pgorod: Tested, works well. I see a lot of other strings in that file with double-quotes. According to the project coding standards, this should be avoided when variable substitution is not occurring inside the string. So you would use:

pgorod: 'Text'
"Text with $var"
But inthis case we have a reason to use double-quotes, which is: we want an apostrophe as part of the string.

Dillon-Brown:
Assessed

Issue #4346, opened 2017-09-29
Workflow not working if set date format for user is d/m/Y


anhnhatuit:
**Issue:**
I've create a workflow to modify record update i change any value in any module and i using date format is: d/m/Y. When i modify record then workflow not working, i found the issue bellow:

    $beanDateEnteredTimestamp = strtotime($timedate->asUser(new DateTime($timedate->fromDb($bean->date_entered))));
            $beanDateModifiedTimestamp = strtotime($timedate->asUser(new DateTime($timedate->fromDb($bean->date_modified))));
            $thisDateEnteredTimestamp = strtotime($this->date_entered);
            switch($this->flow_run_on){
                case'New_Records':
                    // it is an invalid bean if the user modify it now because the affection need on new records only!
                    if(!empty($bean->fetched_row) ||
                        $beanDateEnteredTimestamp < $thisDateEnteredTimestamp) {
                        return false;
                    }
                    Break;

if day of date created record > 12 example (13/01/2017) then $beanDateEnteredTimestamp will return false
and day date created of workflow <12 example (10/01/2017) then return valid value

=>  $beanDateEnteredTimestamp < $thisDateEnteredTimestamp is true then workflow will ignore.

**Possible Fix**

Because strtotime not support d/m/Y format then need change to a format strtotime support example: Y-m-d
Note: You can find the code in modules/AOW_Workflow/AOW_Workflow.php, check_valid_bean function
**Steps to Reproduce**
Config workflow:
 Create a workflow that triggers on save of a Cases
 Have this workflow set the assigned user is any user..
Test:
  Create new case.
 Go to detail case then case need assigned to a user.

**Context**

**Your Environment**

SuiteCRM Version used: 7.8.9
Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Firefox
Environment name and version (e.g. MySQL, PHP 7): MySQL, PHP 5.6
Operating System and version (e.g Ubuntu 16.04): Ubuntu 16.04

pgorod:
A couple of questions...

1. How _exactly_ did you configure your cron jobs to run?

2. You should have two `php.ini` in your system, one for the web server, one for the CLI. What setting do you have for `date.timezone` in each of them?

On my Ubuntu 16.04, the locations of these files is this
Web server `/etc/php/7.0/apache2/php.ini`
Command-line `/etc/php/7.0/cli/php.ini`

anhnhatuit:
Hi Pgorod,
[...]

PR #4357, opened 2017-10-03
Fixed issue #4346 workflow not working when user using date format is d/m/Y

anhnhatuit:
Fixed issue strtotime not working with d/m/Y format. I keep format as db format of bean date created, date modified and convert workflow date created to db format to make sure all workflow not ignore while condition ok.

## Description

Some workflow is not working when user using date format is d/m/Y. I investigated the cause. Because strtotime not working with date format d/m/Y so if day of bean created > 12 (Example: 30/10/2017) then strtotime will return false. And if the day of date created of workflow < 12 (Example: 10/01/2017) then it will return a valid value because it understand 10 is month (sttotime working with m/d/Y format) then if compare 2 value then bean created < workflow date is true => workflow will be ignore while actually bean created > workflow created.
See issue: #4346
## Motivation and Context

-  Make sure workflow always working if condtion ok.

## How To Test This
1.  Go to Profile > Advanced tab > Locale settings > Date format and change to d/m/Y format.
2.  Change date your computer to 10/01/2017 and create new workflow on any module with Run On: New Records and execute any action (Ex: create new record or modify record,...)
3.   Change date your computer to day > 12 (Ex: 20/01/2017) and create new record on module you have configured workflow to check.
4.   Workflow will not working because strtotime (20/01/2017) = false < strtotime(10/01/2017) is valid value.

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOW_WorkFlow/AOW_WorkFlow.php
+++ b/modules/AOW_WorkFlow/AOW_WorkFlow.php
@@ -461,11 +461,14 @@ class AOW_WorkFlow extends Basic {
 
 
         if($this->flow_run_on){
-
-            // database time correction with the user's time-zoneqq
-            $beanDateEnteredTimestamp = strtotime($timedate->asUser(new DateTime($timedate->fromDb($bean->date_entered))));
-            $beanDateModifiedTimestamp = strtotime($timedate->asUser(new DateTime($timedate->fromDb($bean->date_modified))));
-            $thisDateEnteredTimestamp = strtotime($this->date_entered);
+		
+	    // strtotime not working with d/m/Y format, need convert to db format to make sure feature working expected.
+            // Keep db format to stable with strtotime
+            $beanDateEnteredTimestamp = strtotime($bean->date_entered);
+            $beanDateModifiedTimestamp = strtotime($bean->date_modified);
+		
+            // Convert user date format to db format to stable with strtotime
+            $thisDateEnteredTimestamp = strtotime($timedate->fromUser($this->date_entered)->asDb());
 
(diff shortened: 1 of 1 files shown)


pgorod:
Your code seems to be throwing a fatal error in the tests:

`Fatal error: Uncaught Error: Call to a member function asDb() on null in /home/travis/build/salesagility/SuiteCRM/modules/AOW_WorkFlow/AOW_WorkFlow.php on line 471`

Maybe there some extra check that needs to be made for some edge case.

anhnhatuit:
Hi @pgorod ,

How to check $timedate is global variable loaded or workflo date created not null when run test?

Thanks!

pgorod:
I don't really know what this code does, and I am no expert in PHP, but your changes to these two variables

`$beanDateEnteredTimestamp`
`$beanDateModifiedTimestamp`

seems to me you solve the `strtotime` problem you were experiencing, but at the cost of changing how all this works - people want Workflow logic operating in their own timezones, not in db timezone (GMT). So you can't just remove those `AsUser` calls, unless I'm missing something.

anhnhatuit:
Hi @pgorod ,
Ok. I got it, I'll change to one user format working with strtotime to make sure it' work with timezone
Thanks!

samus-aran:
Hi @anhnhatuit Looks like this has been resolved in another commit here.https://github.com/salesagility/SuiteCRM/commit/0f2f884f88cd488b5d763de79cb10ef13457820c
Feel free to reopen the original issue if still a problem. But thank you again for your contribution and providing a potential solution.

PR #4385, opened 2017-10-09
Typo: Triple quotes instead of only two

gunnicom:
Typo: Triple quotes instead of only two


Dillon-Brown:
Well spotted.

gunnicom:
Wasn't me, Was HTMLTidy in my Browser ;)

PR #4396, opened 2017-10-10
 Missing space, put deprecated width and align in css

gunnicom:
Missing space, put deprecated width and align in css.
align and width are deprecated, and css should be used here.

HTML generated before (missing space between ' and width ):
`<div style='white-space: normal;'width='100%' align='left'>`
After:
`<div style='white-space: normal; width:100%; text-align:left;'>`


Dillon-Brown:
Assessed

PR #4398, opened 2017-10-10
Tabcount should count only Tabs and not Panels

gunnicom:
Tabcount should count only Tabs and not Panels 
In Detailview it is correct:
https://github.com/salesagility/SuiteCRM/blob/e0a0a9cd8ad2f01049ef2d412f024fb435151652/include/DetailView/DetailView.tpl#L52-L59


Dillon-Brown:
@gunnicom Should be to the "hotfix-7.8.x" branch as well as a few of your other PR's. If you want you can change them to "hotfix" and when it's time to merge I can make new duplicate PR's to the "hotfix-7.8.x branch to save you from remaking them.

PR #4536, opened 2017-11-17
Feature/pwdsecure 2 - fix password expires issue

daniel-samson:
## Description

This change corrects the issue that a users password expires immediately, which cause by the changes to password management due to the defaults on the config have not been set correctly.
 
## Motivation and Context

- Bug fix

## How To Test This

- Install and log into a user
- You password should only expire when it has expired

## Types of changes

### Final checklist


modules/Users/password_utils.php
```diff
@@ -83,8 +83,7 @@ function  hasPasswordExpired($username){
     $current_user= new user();
     $usr_id=$current_user->retrieve_user_id($username);
 	$current_user->retrieve($usr_id);
-	$type = '';
-        $type='syst';
+    $type='syst';
```
daniel-samson: Should this be blank or syst?

daniel-samson: it should be syst.

PR #4569, opened 2017-11-23
Fixed #4496 - bug of wrong SugarDateTime minutes method used

### LINKED ISSUE #4496: Bug on Scheduler.php . Wrong minutes getter used.
#### Issue
I notice my php error logs were filling up with this warning on multiple line numbers on the same range:
A non-numeric value encountered in /crm/modules/Schedulers/Scheduler.php on line 423

Basically the issue show up when the variables `$currentMin` and `$startMin' are being used.

#### Expected Behavior
No warnings. Scheduler properly executed.

#### Actual Behavior
I have not done full trace of what problems this could be causing, but I guess that the scheduler is not using the current "now" minutes for the calculation of the scheduler.

#### Possible Fix
I traced the reason of the problem and it's because it's using the wrong getter for minutes. From the class 'SugarDateTime' the correct way to get the minutes is using the getter 'min'.

However these two lines are found on the Scheduler.php
```
402  $currentMin = $timedate->getNow()->minute;
417  $startMin = $timedate->fromDb($focus->date_time_start)->minute;
```

Instead they should say:
```
402  $currentMin = $timedate->getNow()->min*1;
417  $startMin = $timedate->fromDb($focus->date_time_start)->min*1;
```

The multiplication by 1 is optional, just  to make the result numeric instead of the by default string.

#### Steps to Reproduce
1. Activate the scheduler
2. Send all the PHP warnings to be saved on a file.
3. Review the warnings file after the scheduler is executed.

#### Context
We are trying to certify the SuiteCRM to run on PHP 7.1, but we have found already a lot deprecated warnings and a lot of unchecked variables. So far we are ignoring the ones that do not affect the behavior, but this one could cost somebody head, if the schedule is not properly executed on time.

I will suggest to be medium priority.

#### Your Environment
* SuiteCRM Version used: 7.8.5 (because we need themes).
* Browser name and version: Irrelevant
* Environment name and version:  PHP 7.1.10
* Operating System and version : Windows Server 2016

cherrador:
Additional of the wrong method was being used, also by multiplying the result with 1, it makes sure that that the result is numeric. This was causing warnings on PHP 7.1.  Code to reduce leading zeros is not needed anymore.

## Types of changes

### Final checklist

### DIFF
--- a/modules/Schedulers/Scheduler.php
+++ b/modules/Schedulers/Scheduler.php
@@ -399,10 +399,7 @@ class Scheduler extends SugarBean {
 		// derive minutes
 		//$currentMin = date('i');
         $minName = array();
-		$currentMin = $timedate->getNow()->minute;
-		if(substr($currentMin, 0, 1) == '0') {
-			$currentMin = substr($currentMin, 1, 1);
-		}
+		$currentMin = $timedate->getNow()->min*1;
 		if($mins == '*') {
 			$GLOBALS['log']->debug('----->got * mins');
 			for($i=0; $i<60; $i++) {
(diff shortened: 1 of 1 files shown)


Dillon-Brown:
Hi @cherrador, thanks for the contribution! would you be able to remake this to the hotfix-7.8.x branch? that way we can merge this into both 7.8 and 7.9.

PR #4632, opened 2017-12-06
Fix detail panels on upgrade

daniel-samson:
## Description

When a User upgrades from SugarCRM, any JavaScript customisation break on the panels. This is due the change after SuiteCRM 7.7.x, where the panels in the editview and the detail view were changed from tables to divs.

## Motivation and Context

This change adds the old class and a data-id attribute to allow these scripts either continue to work or at least be easy to update.

## How To Test This

* Repair and Rebuild

## Types of changes

### Final checklist

### DIFF
--- a/themes/SuiteP/include/EditView/EditView.tpl
+++ b/themes/SuiteP/include/EditView/EditView.tpl
@@ -170,7 +170,7 @@
                     </a>
 
                 </div>
-                <div class="panel-body {{$collapse} panelContainer}" id="detailpanel_{{$panelCount}}" data-id="{{$label_upper}}">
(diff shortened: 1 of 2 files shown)


themes/SuiteP/include/EditView/EditView.tpl
```diff
@@ -170,7 +170,7 @@
                     </a>
 
                 </div>
-                <div class="panel-body {{$collapse}}" id="detailpanel_{{$panelCount}}">
+                <div class="panel-body {{$collapse} panelContainer}" id="detailpanel_{{$panelCount}}" data-id="{{$label_upper}}">
```
gymad: panelContainer in between the smarty markers '{', '}'. Is it works properly?

daniel-samson: It actually does...surprisingly :D. Just fixed it now.

PR #4640, opened 2017-12-07
API - Removed 'type' from the list of forbidden attributes

PedroErnst:
Removed the 'type' field from the list of reserved API keywords because there are some models that have a 'type' field that needs to be editable.
As far as I can see there is no real reason to forbid this keyword.

## Motivation and Context
It was impossible to set the 'type' attribute on any record that contained such a field (Alerts, Cases, Bugs, Meetings, etc) because the API had it as a reserved keyword. Passing type in the list of attributes resulted in an error being returned by the API.

## Types of changes

### Final checklist


daniel-samson:
This word is a reserved word in the JSON API spec. We need to find an alternative solution to this problem.

PR #4642, opened 2017-12-08
Fix/api avoid type keyword

PedroErnst:
Added an alias for the 'type' field to be able to send it through the API. 

## Description

When posting a type attribute the API will return an error and inform about the alias.
When posting the alias the API will convert it to the type attribute on the bean, enabling the CRM to correctly save the record.

## Motivation and Context

It was impossible to set the 'type' attribute on any record that contained such a field (Alerts, Cases, Bugs, Meetings, etc) because the API had it as a reserved keyword. Passing type in the list of attributes resulted in an error being returned by the API.

## Types of changes

### Final checklist


PedroErnst:
@Dillon-Brown merge conflicts are solved

PR #4673, opened 2017-12-13
Fixed #4671 - bug - bulk action, filter and column chooser missing attributes

### LINKED ISSUE #4671: bug - bulk action, filter and column chooser missing attributes
Ok I understand that this might not sound **missing critical to you but it is for us.**

The issue:
If I do NOT use the legacy search "methods" then those buttons are not labelled correctly for screen readers to be usable by a blind user (we do have about 10 blind users working for us, it is our business).

so simply put we need to make the buttons visible to screen readers using "alt" or "title" ...  
Currently they see "void(0)" for them... and can not really know what will be the consequence clicking on them...

Why it is so important to us...
I need to remove the legacy search in order to get column chooser to work.
and the column chooser is a feature that we need so the user can customize their listview.

I understand that you might not consider this as urgent but please point me in the right direction so I can update the file myself.

So I tested and if I add " title="Filter" to the element it will show correctly  
`<a href="javascript:void(0)" class="glyphicon glyphicon-filter parent-dropdown-handler" onclick="listViewSearchIcon.toggleSearchDialog('latest'); $('#searchDialog .nav-tabs .active').removeClass('active'); $('#searchDialog .nav-tabs li').first().addClass('active'); $('#searchDialog').modal('toggle');" title="Filter"></a>`

and the same for column chooser 
`<a href="#" class="glyphicon glyphicon-th-list" data-toggle="modal" data-target=".modal-columns-filter"  title="Column Choser"></a>`

So where can i find those element declarations ?

btw I am running 7.8.8
Thank you

Dillon-Brown:
## Description

Add missing attribute to make the buttons visible to screen readers using "alt" or "title" ...
Currently they see "void(0)".

## Motivation and Context

Issue reference: #4671 

## Types of changes

### Final checklist

### DIFF
--- a/include/ListView/ListViewColumnsFilterLink.tpl
+++ b/include/ListView/ListViewColumnsFilterLink.tpl
@@ -1,5 +1,5 @@
 <ul class="clickMenu selectmenu SugarActionMenu columnsFilterLink listViewLinkButton listViewLinkButton_{$action_menu_location}" onclick="columnsFilter.onOpen();">
     <li class="sugar_action_button">
-        <a href="#" class="glyphicon glyphicon-th-list" data-toggle="modal" data-target=".modal-columns-filter"></a>
+        <a href="#" class="glyphicon glyphicon-th-list" data-toggle="modal" data-target=".modal-columns-filter" title="Column Choser"></a>
     </li>
 </ul>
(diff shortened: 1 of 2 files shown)


horus68:
They are hard coded in code files. 
But all users should have a translated interface!
@Dillon-Brown can we have translated strings?

PR #4701, opened 2017-12-14
Improve non filtered email performance

daniel-samson:
## Description

When a user accesses the emails module. If they have Inbox which have a large number of emails (or greater than 2000 emails), the have to wait in excess of 60 seconds. 

## Motivation and Context

This change helps to improve the performance by up 12x, when the user is not using a filter. It still takes on average around up to 10 seconds to load, hover this is an improvement on the 60 seconds I mentioned earlier.

I tested these loading times against my inbox which has around 6500 emails in it. I was running this on my local computer.

Due to the limitations that we cannot cache the emails, we can only improve the performance on the sort by date field.

## How To Test This

- Set up emails
- Access emails module
- Measure loading times.

## Types of changes

### Final checklist

### DIFF
--- a/modules/InboundEmail/InboundEmail.php
+++ b/modules/InboundEmail/InboundEmail.php
@@ -368,38 +368,89 @@ class InboundEmail extends SugarBean
         $filterCriteria = NULL;
 
 
-        foreach($filter as $filterField => $filterFieldValue) {
-            if(empty($filterFieldValue))
-            {
-                continue;
+        if(!empty($filter)) {
+            foreach($filter as $filterField => $filterFieldValue) {
+                if(empty($filterFieldValue))
+                {
+                    continue;
+                }
+
+                // Convert to a blank string as NULL will break the IMAP request
+                if($filterCriteria == NULL) {
+                    $filterCriteria = '';
+                }
+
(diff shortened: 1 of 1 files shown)


pgorod:
@daniel-samson there is [a thread on the forums](https://suitecrm.com/forum/installation-upgrade-help/17034-after-upgrade-to-7-9-8-no-emails-are-shown-on-my-inbox-module) with some users that are getting a problem with their inboxes - nothing shows.

One of them recently discovered that with a filter, he sees emails, only with no filters he gets the problem, and he suggested that this PR could be the cause.

If you can have a look there and suggest any further steps for diagnosis that would be great. If we can't reproduce, maybe we could give them a special `InboundEmail.php` file with extra logging code... Thanks!

pgorod:
A user on that thread reports that #4731 fixes this problem.

fcorluka:
Hi @pgorod ,

yes there is issue with showing mails.
I have clean 7.10.7 installation, created few mails but when go inside Email can not see any.
Same problem is on SuiteCRM offical Demo.
Also, solution in #4731 seems to be implemented in 7.10.7 but not working.

Issue #4718, opened 2017-12-16
SuiteCRM 7.9.8 Inbound email with a hyphen in the name will not display emails


Cheezecake245:
#### Issue

Need to setup an inbound gmail account with a hyphen in the domain name (example@my-domain.com) that's been giving me trouble.

#### Expected Behavior

Show emails from hyphenated gmail domain in the imap view

#### Actual Behavior

 I tried everything I could find. Outbound works, inbound shows it works, cron jobs are running, then shows no emails. I can see the emails in the sql database, they just aren't showing in the list. Then I tried a different gmail account that ends in @gmail.com and it works fine. I can see all the emails and interact with the imap server, mark things as read and unread and whatnot. Tried another @gmail address and it worked fine too. Then setup another instance of SuiteCrm on another fresh machine, and got the same result, inbound works for @gmail accounts, but not for the hyphened email. Only thing in the SuiteCrm.log is [12748][1][FATAL] *** FOLDERS: addBean() is trying to save to a non-saved or non-existent folder.

#### Steps to Reproduce

1. add hyphenated domain name email to the inbound email.
2.run cron job
3.no emails ;(

#### Context

This is keeping my client with hyphenated domain name from receiving emails within SuiteCrm

#### Your Environment

* SuiteCRM Version used: 7.9.8
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Chrome Version 63.0.3239.84 (Official Build) (64-bit)
* Environment name and version (e.g. MySQL, PHP 7): 10.1.29-MariaDB,Apache/2.4.29 (Win32) OpenSSL/1.0.2n PHP/7.1.12
* Operating System and version (e.g Ubuntu 16.04):  tested on Windows 10 Pro 64bit, Windows Server 2013

PedroErnst:
Investigating this issue I came across a small bug in the InboundEmail class, which a priori didn't seem to have any connection to this, but after fixing it I was able to retrieve the emails from a hyphenated domain address where before I wasn't getting any return.

After fixing this apparently unrelated bug I have not been able to reproduce this issue anymore.

Cheezecake245:
This has solved my issue as well, thank you!

horus68:
@Cheezecake245 please reopen this. Even if your issue is solved it is not implemented in code yet (PR merged). Usually issues are only closed if created by mistake or when a PR is merged (and it can be done automatically or by hand)

PR #4731, opened 2017-12-18
Fix/invalid sequence for imap function

PedroErnst:
Make sure the sequence passed to imap_fetch_overview is not invalid.

## Motivation and Context

Sometimes the sequence would contain negative numbers or be otherwise invalid.
Issue #4718 
Issue #4746

## Types of changes

### Final checklist

### DIFF
--- a/modules/InboundEmail/InboundEmail.php
+++ b/modules/InboundEmail/InboundEmail.php
@@ -415,12 +415,8 @@ class InboundEmail extends SugarBean
                     $lastMsg = $firstMsg + (int)$pageSize;
                 }
             }
-            if ($firstMsg < 1){
-                $firstMsg = 1;
-            }
-            if ($lastMsg < $firstMsg){
-                $lastMsg = $firstMsg;
-            }
+            $firstMsg = $firstMsg < 1 ? 1 : $firstMsg;
+            $lastMsg = $lastMsg < $firstMsg ? $firstMsg : $lastMsg;
 
             $sequence  = $firstMsg . ':' . $lastMsg;
             $emailSortedHeaders = imap_fetch_overview(
--- a/modules/InboundEmail/InboundEmail.php
+++ b/modules/InboundEmail/InboundEmail.php
@@ -415,6 +415,12 @@ class InboundEmail extends SugarBean
                     $lastMsg = $firstMsg + (int)$pageSize;
                 }
             }
+            if ($firstMsg < 1){
+                $firstMsg = 1;
+            }
+            if ($lastMsg < $firstMsg){
+                $lastMsg = $firstMsg;
+            }
 
             $sequence  = $firstMsg . ':' . $lastMsg;
             $emailSortedHeaders = imap_fetch_overview(


fcorluka:
Do anyone knows when this will be on master branch?

I have a fresh installation of 7.10.7  and have the same issue!
I thought the is a problem in my code (error in .zip) so I went into SuiteCRM DEMO, and there is the same issue. I will try this fix, but it I believe this needs to be fixed ASAP because this way seems that Email module is broken.

Dillon-Brown:
Hi @fcorluka, so far I haven't been able to replicate this on 7.10.9 but I do encounter an issue with the SMTP connection in demo, i'll investigate further and get back to you (Just to note, I wouldn't suggest setting up personal email on demo as this is publicly visible to everyone using the instance).

fcorluka:
Hi @Dillon-Brown,
these are screeshots taken from Offical Demo.
If I am doing something wrong please correct me.
Under emails there are none shown:
demo-email1

Under accounts there are mails (under History subpanel):
demo-email2

When I apply filter in Emails, againg none are shown:
demo-filter

On my local instance (v. 7.10.7) when I apply filter I can see mails regarding to the filter.
Best regards,
Filip

Dillon-Brown:
@fcorluka The emails under "History" have been archived to that account which is why you can see them. The emails that should appear in the "Emails" module come from the personal account you set up in user profile here: 
email

fcorluka:
Hi @Dillon-Brown,
but how is than a situation where I can on my local instance set filter and see emails (that were not in INBOX), but can not see them when a search is cleared?
Can I have multiple INBOX accounts and be able to show all emails to all users?

Best regards,
Filip

Dillon-Brown:
@fcorluka Is it possible you are searching in a different email folder (And thus having results returned) then when you clear the filter it's taking you back to inbox?

emails

fcorluka:
Hi,
I made some progress in understanding how EMAIL module works. As for your advice, I have set up mail accounts (inbound), and really can see all mail from that accounts (in separate folders); and that is OK for me.

Another problem is that I can not see attachments and creating case from emails is not working.
I will try now fresh 7.10.9 to see is the problem presists.

Dillon-Brown:
@fcorluka Did you make sure to select an "Account Name" on the "Create Case" popup? it seems to be working for me on 7.10.9. I do however see the issue with the attachments, if you import an email you can see the attachments in the "attachments" sub-panel but the the "Has attachments" column doesn't display anything, this looks like a bug.

fcorluka:
@Dillon-Brown I am not sure what popup do you mean?
So far I had numerous problems with EMAIL module, that I don't know where to start and does it even has the sense to write it here in comments, or to open Issue.

I had set up this:
inbound-mail

I had two inbound emails and could see both in Emails.
What I could not is:
- see subpanel in the mail (including attachments)
- get CASE to be created

I did some modification, deleted and returned Inbound mail account, and now:
- can not see other folders (other mail accounts)

Deleted all and recreated only 1:
- can not see any mail

Dillon-Brown:
@fcorluka For the attachment issue, did you import the email? you won't see any subpanels until you do, this is what I see on 7.10.9:

attachment


When you were unable to see any mail after recreating inbound email did you make sure you set it to active? this can be done in user profile -> settings:

email_settings

If you can reliably replicate any of these issues then yes I would suggest raising a separate issue for each one instead of a comment, that way we can prioritise it and investigate further.

fcorluka:
Hi @Dillon-Brown ,

I haven't imported email, I have just set up email inbound account and all were there, but not attachments.

Also, 'Active' status was checked. 
I must recreate each of these scenarios (screenshots) and create issues, as You suggested.

Best regards,
Filip

PR #4738, opened 2017-12-18
Disable php timeout while report export

isleshocky77:
## Description

Exports of Reports can be very large and often take more than the default php timeout of 30 seconds. 30 seconds make sense as a sane timeout when trying to display a page or submit data; however, when building an export it does not.

A better solution would be to background this and let the user know when it's completed; however, for now just disabling the php timeout while running the export action makes much more sense.

## Motivation and Context

Receiving a gray screen while attempting to export a report with an error message

```
PHP Fatal error:  Maximum execution time of 30 seconds exceeded in /opt/suitecrm/modules/AOW_WorkFlow/aow_utils.php
```

## How To Test This

Attempt to run a large export which would normally take more than 30 seconds without this patch and with this patch. See that one throws a gray screen with a 500 message while with the patch it succeeds.

## Types of changes

### Final checklist


Dillon-Brown:
Assessed

PR #4769, opened 2017-12-21
If there is no google maps API key don't throw php notice

gunnicom:
If there is no google maps API key don't throw php notice


Dillon-Brown:
Assessed

Dillon-Brown:
Already merged into hotfix. Thanks!

PR #4785, opened 2017-12-27
Enable sending emails from system outgoing email account

PedroErnst:
Added the system email account as a selectable option to the dropdown list of From Addresses when composing a new email.

Removes the need to create an Incoming Email Account in order to send email, as long as there is a system account email address existant and enabled for users to send emails from it.

## Motivation and Context

Issue #3741
On upgrade from 7.8 to 7.9, there is no from address available to send from when composing an email,

it should allow me to send from the system email address, if "Allow users to use this account for outgoing email" is checked

Secondly to use a personal outbound email, I had to add a inbound account, where i would expect it to just pick up my outbound accounts.

strange error:-

Wed Dec 13 17:42:49 2017 [27469][1][SECURITY] User Administrator attempted to send an email using incorrect email account settings in which they do not have access to.

## Types of changes

### Final checklist

### DIFF
--- a/modules/Emails/EmailsController.php
+++ b/modules/Emails/EmailsController.php
@@ -451,7 +451,7 @@ class EmailsController extends SugarController
                 'id' => $system->id,
                 'attributes' => array(
                     'from' => $system->mail_smtpuser,
-                    'name' => $current_user->name,
+                    'name' => $system->name,
                     'oe' => $system->mail_smtpuser,
                 ),
                 'prepend' => false,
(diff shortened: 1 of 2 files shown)


pgorod:
This fixes #3741, right?

PedroErnst:
Oh thanks for pointing that out, I didn't know there was an issue for it.

Yes, this should fix that precise issue.

The only thing I don't know is if this solution is the most elegant one, I think it can be done much better. Maybe when @daniel-samson and @gymad come back they can have a look too.

Issue #4291, opened 2017-09-20
Dynamicly created labels, potentially removed by mistake


pgorod:
Ok, @horus68 and @Dillon-Brown, I made a search over the entire codebase with a really weird grep to find lines that:
- include the string `LBL`
- either include the string `_'` or the string `'_`, to look for places where strings are being composed from parts

That gave me 135 lines. I cleaned that list up manually (hard work that strained my poor eyes) to the list below which has 67 lines.

```php
modules/ModuleBuilder/javascript/ModuleBuilder.js:288:				ModuleBuilder.history.popup_window.setHeader( module_str + ' : ' + SUGAR.language.get('ModuleBuilder', 'LBL_' + layout.toUpperCase()) + SUGAR.language.get('ModuleBuilder', 'LBL_HISTORY_TITLE'));
modules/ModuleBuilder/javascript/studio2.js:388:		var panelLabel = 'lbl_' + view +  '_panel' + panelNumber;
modules/ModuleBuilder/javascript/studio2PanelDD.js:124:					var panelLabelNoID = 'lbl_' + view +  '_panel';
modules/ModuleBuilder/views/view.searchview.php:109: 				$searchLabel = 'LBL_' . strtoupper ( $this->editLayout) ;
modules/ModuleBuilder/views/view.module.php:78:            $translated_type[$type]=translate('LBL_TYPE_'.strtoupper($type),'ModuleBuilder');
modules/ModuleBuilder/MB/MBLanguage.php:187:                'LBL_'.strtoupper($this->key_name).'_SUBPANEL_TITLE'=>$this->label,
modules/ModuleBuilder/MB/MBField.php:64:		if(empty($this->vname))$this->vname = 'LBL_' . strtoupper($this->name);
modules/ModuleBuilder/parsers/relationships/ActivitiesRelationship.php:128:                'system_label' => 'LBL_' . strtoupper($this->relationship_name . '_FROM_' . $this->getRightModuleSystemLabel()) . '_TITLE',
modules/ModuleBuilder/parsers/relationships/ActivitiesRelationship.php:133:                'system_label' => 'LBL_' . strtoupper($this->relationship_name . '_FROM_' . $this->getLeftModuleSystemLabel()) . '_TITLE',
[...]

PR #4795, opened 2017-12-28
Remove unused and inexistant mod string assignment

PedroErnst:
An assignment for a template variable that was not being used was causing a PHP notice.

## Motivation and Context

Removes a notice.

## Types of changes

### Final checklist

### DIFF
--- a/modules/Import/views/view.step2.php
+++ b/modules/Import/views/view.step2.php
@@ -129,7 +129,7 @@ class ImportViewStep2 extends ImportView
                 "INSTRUCTION_STEP" => $mod_strings["LBL_{$lang_key}_NUM_$i"],
             );
         }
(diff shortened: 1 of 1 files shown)


pgorod:
Hi @PedroErnst. Are you _sure_ it's not being used? 

We currently have made a big mess out of unused language strings, especially in cases like this where the label name is built from variables. We removed a lot of strings that should be there, and we'll have to reinstate them. I just want to make sure this is not one of them...

Please see #4291 
Thanks!

PedroErnst:
I am _fairly_ sure.

The label appears to be built from variables, but in fact that variable is hard coded just a few lines further up in the function, and will always be LBL_IMPORT_CUSTOM_TITLE.

Which does not appear in any of the language files.

And the key, INSTRUCTIONS_TITLE does not appear anywhere either.

So, I'm not a 100% but really close.

pgorod:
Ok... I think you might be right. :-)

Dillon-Brown:
Assessed

Issue #3902, opened 2017-07-18
Accessing calls from home page dashlet - cannot reschedule the call - greyed out. 


jonnie00:
#### Issue

If we select a call from the home page dashlet and then try and reschedule it the reschedule pop up appears but it appears to have a grey overlay and cannot be edited. The page itself can be scrolled up and down. (Please see the attached screen shot)

To clear it, you have to access an alternative page.

If we access the same call directly from the calls page, rescheduling works fine.

We are now running 7.9.3 on a linux server but the error also occurs in 7.9.1. We have checked the file permissions and run rebuild and config, repair roles and dashlets etc..

We have also replicated this error on the suitecrm demo site @ softaculous.com/demos/suitecrm

#### Expected Behavior

As above.  The call should be able to be rescheduled
#### Actual Behavior

The rescheduled call pop up box is greyed out and no action can clear it.  Navigating away to a different page is the only way to clear the error
#### Steps to Reproduce

1. Create a call.
2. Access the call via the 'my calls' dashlet from the home page and attempt to reschedule the call.
3. The reschedule call box is greyed out and no further action can be taken.
4. Accessing the same call directly from the 'calls' page works correctly

#### Context

high priority

#### Your Environment

* SuiteCRM Version used: 7.9.3
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Firefox and Chrome
* Environment name and version (e.g. MySQL, PHP 7):
* Operating System and version (e.g Ubuntu 16.04):
![call_reschedule_issue](https://user-images.githubusercontent.com/29843802/28315693-7d0ffb48-6bb7-11e7-81b2-2ce588b978ea.png)

pgorod:
I've also been able to reproduce this in my test system. When the call is opened from the Dahslet (and only when this is the case) then the Reschedule button gives that greyed-out effect.

jonnie00:
Any update on this issue please?

jonnie00:
This is now an issue in 7.8.8 on ALL call rescheduling not just when accessed from the home page.  Any update / help on this is greatly appreciated.

pgorod:
@jonnie00 I think there was a fix for a z-index issue like this one, I'm not sure if it went into the 7.10 Beta3.

If it is easy for you to try a Beta3 upgrade on a clone of your system, then maybe you could try it. But don't do it in your production system, this is really an early Beta!

[...]

PR #4797, opened 2017-12-28
Fix #4176 date picker

### LINKED ISSUE #4176: Dashlet's calendar popup windows open in background partially hidden/unaccessible
#### Issue

When you have a dashlet which in edit mode has a date field with a calendar button next to it, when launched the calendar popup window appears underneath the previous popup hence disallowing to navigate the calendar. 

#### Expected Behavior

The popup should appear in the foreground
#### Actual Behavior

obraz

#### Steps to Reproduce

It's 100% repeatable on 7.9.4 as well as demo version.

#### Your Environment

   *  SuiteCRM Version used: 7.9.4
    * Firefox 55.0.3. 32-bit or Chrome 60.0.3112.113 (64-bit)
    * MariaDB 10.2.6. , PHP 7.1.8
    * CentOS Linux release 7.3.1611 (Core)

e-reeley:
Date picker was showing behind the popup display.

## Description

The z-index was changed to allow the date picker to be accessed and in full display.

## How To Test This

1) Have the 'Outcome By Month' dashlet on display
2) Click the edit icon
3) Click on the date icon to select a date

The css will need to be rebuilt for this change.

## Types of changes

### Final checklist

### DIFF
--- a/themes/SuiteP/css/suitep-base/editview.scss
+++ b/themes/SuiteP/css/suitep-base/editview.scss
@@ -5422,7 +5422,6 @@ table tr#delegates_search {
 /* popup in studio */
 .masked .yui-simple-dialog {
   // Overridding inline style
-  z-index: 100000!important;
 }
 
 /* popup dialog on home page */


pgorod:
@e-reeley this fix you made got me wondering - a few months ago we started having lots of different z-index issues around the app. It's hard to know how things will work out with z-indexes since what counts is their relative position to other screen elements, and in generic YUI functions we don't even know in which screens they will appear...

An example is this one #3902

e-reeley:
@pgorod Yeah, this fix was a bit strange to me. I did seem to notice that there was a variety of z-indexes and altering only the values did not always make the expected difference.

It may be worth over time having a clean up of these values to allow us to alter them as we require.

pgorod:
This is all a bit risky - in the same way that somebody changed something a few months ago, and we got unexpected side-effects, we can be changing things now and only find out later what we broke in some other screen. And these issues are simple, but they often break functionality for end-users.

So definitely a more coherent approach would be necessary... but I fear it's too big an effort.

Is #3902 something you can fix easily? We have users in the forums complaining about that one, and it seems it recently became worse. Before, it just affected calls records in dashlets, now it seems to affect all call screens.

e-reeley:
Yeah, I agree, the simple removal of the z-index was something that was not the most well rounded fix, but it can often be rather difficult to test all areas which would be potentially impacted.

I agree, hopefully it is something that we can look into over time, but it looks like a big fix. This does not make it out of the question though, especially since this is hopefully the start of fixing this.

I will take a look into the linked issue and see what I can do. Thank you for adding this one in, it seems that not only have a lot of people experienced this issue, but there seems to be a lot of closely linked issues.

Dillon-Brown:
Closing in favour of #4901

Issue #4802, opened 2017-12-28
Email DetailView shows white screen when email contains un-handled date header format


isleshocky77:
#### Issue

According to the RFC for the email date header ([RFC 822](https://tools.ietf.org/html/rfc822#section-5.1
) or newer [RFC2822](https://tools.ietf.org/html/rfc2822#section-3.3)) a date header can have an optional "Day of the week in front" (e.g. "[ day "," ] date time").

However, code committed which is currently in v7.9.8 always expects it to be there

```php
# modules/InboundEmail/InboundEmail.php:5270
                // find if string has (UTC) in timezone
                $pattern = "/\([\w-+\/]+\)/i";
                $timezoneTextFound = preg_match($pattern,  $header[0]->date);

                $dateTimeFormat = 'D, d M Y H:i:s O *';
                if($timezoneTextFound === false || $timezoneTextFound === 0) {
                    $dateTimeFormat = 'D, d M Y H:i:s O';
                }

                $dateTime = DateTime::createFromFormat(
                    $dateTimeFormat,
                    $header[0]->date
                );
```

This means that if an email comes in with a date without the Day of the week (e.g. Date: 28 Dec 2017 10:58:42 -0500) that this code causes an exception to be thrown which is not handled.

#### Expected Behavior

You should be able to open up and view any RFC compliant email sent

#### Actual Behavior

Attempting to open a non-imported email which has a Date header without a day of the week shows an empty white screen

#### Possible Fix

Correct the pattern matching on the Date header in _modules/InboundEmail/InboundEmail.php_

#### Steps to Reproduce

1. Attempt to open an email with a Date header matching "28 Dec 2017 10:58:42 -0500" which has not yet been imported
2. Notice White screen and email not shown

#### Context

Unable to open all emails in order to import them into the CRM.

It appears this code was introduced fairly recently #3607 and is currently affecting all releases which would make it fairly high priority.

#### Your Environment

* SuiteCRM Version used: 7.9.8
* Environment name and version (e.g. MySQL, PHP 7): PHP 5.6

pgorod:
I don't think the current regular expression is doing anything at the moment - it won't match anything with a space in it, for example. If I'm not mistaken, all dates will contain spaces in them... you can play with it [here](https://regex101.com/r/G2DAZm/1).

So that also probably needs fixing; and then the issue with the day-of-week.

Note that PHP's DateTime object already has constants for this:
http://php.net/manual/en/class.datetime.php
includes 
[...]

PR #4804, opened 2017-12-28
Checking for RFC822 and RFC2822 dates including optional parts in Unimported Inbound Emails

isleshocky77:
## Description
See #4802
This deals with an exception when an inbound email does not have a day of the week in the date.

## Motivation and Context
Inbound emails are showing a white screen with an unhandled exception.

## How To Test This
Look at an unimported inbound email with different valid RFC dates

## Types of changes

### Final checklist

### DIFF
--- a/modules/InboundEmail/InboundEmail.php
+++ b/modules/InboundEmail/InboundEmail.php
@@ -5268,39 +5268,42 @@ class InboundEmail extends SugarBean
             $email->type = 'inbound';
 
 
-            if (empty($email->date_entered)) {
+            if(empty($email->date_entered)) {
 
-                // find if string has (UTC) in timezone
-                $pattern = "/\([\w-+\/]+\)/i";
-                $timezoneTextFound = preg_match($pattern, $header[0]->date);
+                $possibleFormats = [
+                    \DateTime::RFC2822,
+                    str_replace(['D, '], '', \DateTime::RFC2822), // day-of-week is optional
+                    str_replace([':s'], '', \DateTime::RFC2822), // seconds are optional
+                    str_replace(['D, ', ':s'], '', \DateTime::RFC2822), // day-of-week is optional, seconds are optional
+                    \DateTime::RFC822,
+                    str_replace(['D, '], '', \DateTime::RFC822), // day is optional
+                    str_replace([':s'], '', \DateTime::RFC822), // seconds are optional
+                    str_replace(['D, ', ':s'], '', \DateTime::RFC822), // day is optional, seconds are optional
+                ];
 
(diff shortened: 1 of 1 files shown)


modules/InboundEmail/InboundEmail.php
```diff
@@ -5268,39 +5268,42 @@ public function returnNonImportedEmail($msgNo, $uid)
...
-                    $dateTimeFormat = 'D, d M Y H:i:s O';
+            if(empty($email->date_entered)) {
+
+                $possibleFormats = [
+                    \DateTime::RFC2822,
+                    str_replace(['D, '], '', \DateTime::RFC2822), // day-of-week is optional
+                    str_replace([':s'], '', \DateTime::RFC2822), // seconds are optional
+                    str_replace(['D, ', ':s'], '', \DateTime::RFC2822), // day-of-week is optional, seconds are optional
+                    \DateTime::RFC822,
+                    str_replace(['D, '], '', \DateTime::RFC822), // day is optional
```
PedroErnst: Shouldn't day be 'd' ?

isleshocky77: @PedroErnst No, this is referring to the leading "day of week" variable [PHP DateTime](http://php.net/manual/en/class.datetime.php). The comment refers to what the RFC calls it. In 2822 they refer to it as "day-of-week"; in 822 they refer to it as "day". 

Both actually refer to what PHP calls 
> D | A textual representation of a day, three letters | Mon through Sun

[Source](http://php.net/manual/en/function.date.php#refsect1-function.date-parameters)

modules/InboundEmail/InboundEmail.php
```diff
@@ -5268,39 +5268,42 @@ public function returnNonImportedEmail($msgNo, $uid)
...
+                        sprintf('Expected header Date to comply with RFC882 or RFC2882, but actual is "%s"', $header[0]->date)
                     );
                 }
+
+                $email->date_entered = $timedate->asUser($dateTime, $current_user);
+                $email->date_modified = $timedate->asUser($dateTime, $current_user);
+                $email->date_start = $timedate->asUserDate($dateTime);
+                $email->time_start = $timedate->asUserTime($dateTime);
+
+                $systemUser =  BeanFactory::getBean('Users', 1);
```
mattlorimer: if getting the system user is really necessary, it would be better to use the user function getSystemUser, as it not guaranteed the user with id 1 exists

isleshocky77: I don't know enough about that to change it.  I would argue it shouldn't be a part of this PR.

Dillon-Brown:
Assessed

PR #4835, opened 2018-01-03
Feature/gdpr show "opt in" tick in email fields.

daniel-samson:
## Description

This change adds:

Provided that the confirm opt in feature is enabled. This will display a tick next emails. The color of the tick will depend on it's status.

## Motivation and Context

- GDPR Compliance

## How To Test This

- Enable confirm opt in
- look at email address fields

## Types of changes

### Final checklist


Dillon-Brown:
Only issue seems to be with updating the "opt_in_email_created" record in the db with a value. The detailview correctly updates with "Pending Confrim opt in, Confirm opt in sent" but the listview remains unchanged "Pending Confirm opt in, Confirm opt in not sent".

daniel-samson:
@Dillon-Brown 

I have made fixes issue 1 and update the colors for issue 2.

daniel-samson:
@mattlorimer Fixed the issues raised.

mattlorimer:
it still showing on employees, it shouldn't  show for users

daniel-samson:
@mattlorimer I have changed the "getEmailAddressOptInStatus"  to not show on users and employees

include/SugarObjects/templates/basic/Basic.php
```diff
@@ -113,17 +159,16 @@ public function getEmailAddressConfirmOptIn($emailField)
 
         global $sugar_config;
 
-        if (!$sugar_config['email_enable_confirm_opt_in']) {
+        if (!$sugar_config['email_enable_confirm_opt_in']) {#
             global $log;
             $log->warn('Confirm Opt In is not enabled.');
-            return true;
+            return false;
```
gymad: Use `true` as default when `email_enable_confirm_opt_in` config value is not set or `false` as disabled, so it won't block or change any email behaviour when this option disabled by configuration.

include/SugarObjects/templates/basic/Basic.php
```diff
@@ -134,24 +179,23 @@ public function getEmailAddressConfirmOptIn($emailField)
      * @throws RuntimeException
      * @throws InvalidArgumentException
      */
-    private function getEmailAddressId($emailField)
+    protected function getEmailAddressId($emailField)
     {
-        $this->getEmailAddressValidateArguments($emailField);
```
gymad: Why should remove the validation?

include/SugarObjects/templates/basic/Basic.php
```diff
@@ -134,24 +179,23 @@ public function getEmailAddressConfirmOptIn($emailField)
...
+        global $log;
 
         $emailAddress = $this->cleanUpEmailAddress($this->{$emailField});
 
         if (!$emailAddress) {
-            global $log;
```
gymad: Load and use the `$log` only if it's needed.

include/SugarObjects/templates/basic/Basic.php
```diff
@@ -113,17 +159,16 @@ public function getEmailAddressConfirmOptIn($emailField)
...
         }
 
         $emailAddressId = $this->getEmailAddressId($emailField);
-        $emailAddress = BeanFactory::getBean('EmailAddresses', $emailAddressId);
-        $confirmOptIn = $emailAddress->confirm_opt_in;
-
-        return $confirmOptIn;
+        /** @var EmailAddress $emailAddressBean */
+        $emailAddressBean = BeanFactory::getBean('EmailAddresses');
+        return  $emailAddressBean->retrieve($emailAddressId);
```
gymad: Illogical return value: The function called `getEmailAddressConfirmOptIn` return an `EmailAddress` object instead the status of Opt-in confirmation as it was initially. Rename the function to `getEmailAddress` or revert the return value to a `boolean`.

include/SugarObjects/templates/basic/Basic.php
```diff
@@ -160,7 +204,9 @@ private function getEmailAddressId($emailField)
         }
 
         if (!$found) {
-            throw new RuntimeException('A Basic bean has not selected email address. (' . $emailAddress . ')');
+            // Changed exception to error as demo data is never selected.
```
gymad: If you remove the exception, perhaps you also wanted to change the phpdoc above.

include/SugarObjects/templates/basic/Basic.php
```diff
@@ -134,24 +179,23 @@ public function getEmailAddressConfirmOptIn($emailField)
...
-            global $log;
+
             $log->warn('Trying to get an empty email address.');
             return null;
         }
 
-        if (empty($this->emailAddress->addresses)) {
-            $this->retrieve();
-        }
+        $this->retrieve();
```
gymad: In case, if it has no any email address, unnecessary to retrieve hole bean for an email address ID in return value at the end.

metadata/email_addressesMetaData.php
```diff
@@ -80,8 +80,12 @@
             'default'        => 0,
             'vname'         => 'LBL_OPT_OUT',
         ),
-            
-                'confirm_opt_in' => array(
+        'opt_in_email_created' => array(
```
gymad: Where and why did you use this `opt_in_email_created` for? I don't find any reference to this in your PR?

modules/EmailAddresses/EmailAddress.php
```diff
@@ -52,20 +52,56 @@
...
+
+	/**
+     * @var int|bool $invalid_email
+     */
+	public $invalid_email = 0;
+
+	/**
+     * @var TimeDate $opt_in_email_created
+     */
+    public $opt_in_email_created;
```
gymad: Where does it get any value?

modules/EmailAddresses/EmailAddress.php
```diff
@@ -52,20 +52,56 @@
...
+
+    /**
+     * EmailAddress constructor.
+     */
+    function __construct()
```
gymad: `function __construct()` visibility should be `public`

modules/EmailMan/EmailMan.php
```diff
@@ -628,7 +628,7 @@ public function sendEmail($mail, $save_emails = 1, $testmode = false)
         if (
                 $module instanceof Basic &&
                 $sugar_config['email_enable_confirm_opt_in'] &&
-                !$module->getEmailAddressConfirmOptIn('email1')
+                empty($module->getEmailAddressConfirmOptIn('email1')->confirm_opt_in)
```
gymad: Non-logical using of `getEmailAddressConfirmOptIn()` method. It return an Email Address instead Confirm Opt-In status which has a property named confirm_opt_in?

include/SugarObjects/templates/basic/FunctionField/displayEmailAddressOptInField.php
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * @param \Basic $focus
+ * @param $field
+ * @param $value
+ * @param $view
+ * @return string
+ */
+function displayEmailAddressOptInField($focus, $field, $value, $view)
+{
+    global $app_strings, $app_list_strings, $mod_strings;
```
gymad: Unused global variables?

modules/Emails/include/displayEmailAddressOptInField.php
```diff
@@ -0,0 +1,59 @@
...
+
+
+/**
+ * @param \Email $focus
+ * @param $field
+ * @param $value
+ * @param $view
+ * @return string
+ */
+function displayEmailAddressOptInField($focus, $field, $value, $view)
```
gymad: Code duplication: exactly same functionality also implemented in file `include/SugarObjects/templates/basic/FunctionField/displayEmailAddressOptInField.php`.

themes/SuiteP/css/Dusk/style.css
```diff
@@ -12373,7 +12373,7 @@ div.qtip.qtip-default {
 
 .modal-content .search_form .view select,
 .modal-content .search_form .view select#saved_search_select {
-  border: 2px solid #A5E8D6; }
+  border: 2px solid #c4e3f5; }
```
gymad: Using a hard-coded colour.

themes/SuiteP/css/Dusk/style.css
```diff
@@ -14109,6 +14109,26 @@ p.msg {
...
+
+.confirm-opt-in-tick {
+  float: right;
+  padding: 0 10px; }
+
+.email-opt-in {
+  font-size: 14px; }
+
+.email-opt-in-not-sent {
+  color: #dadbdc; }
```
gymad: Using a hard-coded colour.

themes/SuiteP/css/Dusk/style.css
```diff
@@ -14109,6 +14109,26 @@ p.msg {
...
+
+.email-opt-in-sent {
+  color: #dadbdc; }
```
gymad: Using a hard-coded colour.

themes/SuiteP/css/Dusk/style.css
```diff
@@ -14109,6 +14109,26 @@ p.msg {
...
+
+.email-opt-in-confirmed {
+  color: #6EA674; }
```
gymad: Using a hard-coded colour.

include/SugarObjects/templates/basic/Basic.php
```diff
@@ -84,26 +85,71 @@ public function get_summary_text()
...
+     * @global array $sugar_config
+     * @global \LoggerManager $log
+     * @param string $emailField
+     * @return string
+     */
+    public function getEmailAddressOptInStatus($emailField)
+    {
+        $emailAddress = $this->getEmailAddressConfirmOptIn($emailField);
+
+        $doNotDisplayOptInTickForModule = array(
```
gymad: It seems like an external argument or parameter, separate it and/or make it configurable.

include/SugarObjects/templates/basic/Basic.php
```diff
@@ -84,26 +85,71 @@ public function get_summary_text()
...
         $tickHtml = '';
-        $confirmOptIn = $this->getEmailAddressConfirmOptIn($emailField);
-        if ($confirmOptIn && $sugar_config['email_enable_confirm_opt_in']) {
-            $tickTitle = $app_strings['LBL_CONFIRM_OPT_IN_TITLE'];
-            $tickHtml = '<span class="confirm-opt-in-tick" title="' . $tickTitle . '">&#10004;</span>';
+
+        if ($sugar_config['email_enable_confirm_opt_in']) {
+            $template = new Sugar_Smarty();
+            $template->assign('APP', $app_strings);
+            $template->assign('APP_LIST_STRINGS', $app_list_strings);
```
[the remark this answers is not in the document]
daniel-samson: It's true that it isn't used in this pull request. However, we will want to make this available to developers who wish to customise this template. APP, APT_LIST_STRINGS, and MOD are variables which are expected by suitecrm developers for smarty templates. We should keep the pattern.

include/SugarObjects/templates/basic/Basic.php
```diff
@@ -84,26 +85,71 @@ public function get_summary_text()
...
+            $template->assign('MOD', $mod_strings);
```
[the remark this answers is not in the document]
daniel-samson: Same reason as above.

include/SugarObjects/templates/basic/Basic.php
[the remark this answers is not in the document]
gymad: If you want to make it customizable then you should use the `get_custom_file_if_exists()` function to check and load the custom file before smarty `fetch()`.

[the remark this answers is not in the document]
daniel-samson: I will add get_custom_file_if_exists inside the fetch function. That way all tpl's can be over ridden. I thought we already had this in  v7.7 - when we implemented SuiteP.

include/SugarObjects/templates/basic/Basic.php
daniel-samson: So $log is used twice in this method, it two separate scopes. So we need to move the global out of this if statement (or scope) to guarantee that it can be accessed in both cases. see line 208.

gymad: ok

include/SugarObjects/templates/basic/Basic.php
daniel-samson: I removed this method because I was having trouble with the emails module. getEmailAddressValidateArguments assumes that the filed will end in a number like email1, email2, ... etc. Which works great for Accounts, Contacts and Leads.

However, the Emails module stores the email address in email_address field. Which will not pass this validation.

So ... we need to come up with something which better handles these constraints. The real issue is where this code is situated. The responsibility for the email1, should have been implemented as a separate class, via an interface, or perhaps it we should employ a trait. Ideally this would be in the EmailAddresses module instead of the SugarBean. Since we cannot time travel back and change the original problem. We will have to accept the existing code in SugarBean.

We could use the inheritance tree. What If we changed the method to be protected instead. Then override the method. Should these methods existing in the basic type? Should this in company and person instead?

modules/EmailAddresses/EmailAddress.php
daniel-samson: Please explain?

opt_in_email_created is defined in the var_defs or rather in the metadata/email_addressesMetaData.php file. So I have added the property to make it easier to find with and IDE.

modules/EmailMan/EmailMan.php
daniel-samson: I see what you mean. What if change this to getEmailAddressFromSugarField instead?

modules/EmailAddresses/EmailAddress.php
daniel-samson: done.

metadata/email_addressesMetaData.php
daniel-samson: It isn't used in this PR. In my next pr we will use it.

modules/Emails/Email.php
```diff
@@ -4225,4 +4225,50 @@ protected function handleBodyInPlainTextFormat($mail)
         $mail->Body = $this->decodeDuringSend($mail->Body);
         $this->description = $mail->Body;
     }
+
+
+    public function getEmailAddressConfirmOptInTick($emailField)
```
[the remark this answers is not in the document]
daniel-samson: Similar to the message earlier. The validate fails due to the check for the number at the end. Lets move the duplicated part into an other method.

modules/Emails/Email.php
```diff
@@ -4225,4 +4225,50 @@ protected function handleBodyInPlainTextFormat($mail)
...
+        global $mod_strings;
+
+        $tickHtml = '';
+
+        if ($sugar_config['email_enable_confirm_opt_in']) {
+            $template = new Sugar_Smarty();
+            $template->assign('APP', $app_strings);
+            $template->assign('APP_LIST_STRINGS', $app_list_strings);
+            $template->assign('MOD', $mod_strings);
+            $template->assign('OPT_IN', $this->getEmailAddressOptInStatus($emailField));
```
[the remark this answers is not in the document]
daniel-samson: This has changes to OPT_IN_STATUS. You may be looking at a older commit ?

modules/Emails/Email.php
```diff
@@ -4225,4 +4225,50 @@ protected function handleBodyInPlainTextFormat($mail)
...
+    /**
+     *
+     * @global array $sugar_config
+     * @global \LoggerManager $log
+     * @param string $emailField
+     * @return \EmailAddress
+     * @throws RuntimeException
+     * @throws InvalidArgumentException
+     */
+    public function getEmailAddressConfirmOptIn($emailField)
```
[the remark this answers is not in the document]
daniel-samson: Moved to parent, and split out the validation

modules/Emails/include/displayEmailAddressOptInField.php
daniel-samson: So the Basic object does:
```return $focus->getConfirmOptInTickFromSugarEmailAddressField($field);```

the emails modules does:
```return $focus->getConfirmOptInTickFromSugarEmailAddressField('from_name');```

the first example (basic) is for when the email{n} is passed in to the function field. the emails module passes the "from name". This needs to be force as there doesn't seem to be a way to pass the field argument in via the vardefs.

File: themes/SuiteP/css/Dusk/style.css (hunk 13 above)
Comment (daniel-samson) [reply themes/SuiteP/css/Dusk/style.css]: it's not hard coded. I will minify these css files.

File: include/SugarObjects/templates/basic/Basic.php (hunk 17 above)
Comment (daniel-samson) [reply include/SugarObjects/templates/basic/Basic.php]: I moved into a static variable which can be overridden if needed. I am not sure we want administrators changing this value.

File: include/SugarObjects/templates/basic/Basic.php (hunk 5 above)
Comment (daniel-samson) [reply include/SugarObjects/templates/basic/Basic.php]: Php Doc Updated

File: include/SugarObjects/templates/basic/Basic.php (hunk 6 above)
Comment (daniel-samson) [reply include/SugarObjects/templates/basic/Basic.php]: When you retrieve the bean it will get the email addresses. I noticed that emails module requires the retrieve.

File: include/SugarObjects/templates/basic/Basic.php (hunk 2 above)
Comment (gymad) [reply include/SugarObjects/templates/basic/Basic.php]: if the `email_address` field also valid, why don't you just add it to the validation instead removing the validation at all?

File: modules/EmailAddresses/EmailAddress.php (hunk 8 above)
Comment (gymad) [reply modules/EmailAddresses/EmailAddress.php]: Thats ok, my problem is I don't find anywhere in code base or in this PR where we using this variable. Where do you set the value of it?

File: modules/EmailMan/EmailMan.php (hunk 10 above)
Comment (gymad) [reply modules/EmailMan/EmailMan.php]: I meant, if $module instance of Basic and the Basic class implemented a getEmailAddressConfirmOptIn() properly as it was then it should not changed.

File: metadata/email_addressesMetaData.php (hunk 7 above)
Comment (gymad) [reply metadata/email_addressesMetaData.php]: "It isn't used in this PR. In my next pr we will use it." - Why not implementing this in that PR then? Or finished the task here? - Sorry but I disagree, I think totally pointless and confusing to implementing a half task here an other one half of same task in an other PR...

File: modules/Emails/Email.php (hunk 20 above)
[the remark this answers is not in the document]
Comment (gymad) [reply modules/Emails/Email.php]: If a validation fails with a valid data, then the validation is wrong or incomplete. It means we should fix the validation instead just remove.

File: modules/Emails/Email.php (hunk 21 above)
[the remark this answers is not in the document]
Comment (gymad) [reply modules/Emails/Email.php]: My comment was 6 hours ago:
gymad 6 hours ago  Member
This variable OPT_IN never used in template displayEmailAddressOptInField.tpl

Your commit was 2 hours ago:
commit 204f9631f93debf1bbf7b94877a3a70fa36d2681 @daniel-samson daniel-samson committed 2 hours ago.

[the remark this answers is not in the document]
Comment (gymad) [reply modules/Emails/Email.php]: so it's fine.

File: modules/Emails/Email.php [hunk 23]
```diff
@@ -4225,4 +4234,47 @@ protected function handleBodyInPlainTextFormat($mail)
...
+     *
+     * @global array $sugar_config
+     * @global \LoggerManager $log
+     * @param string $emailField
+     * @return \EmailAddress
+     * @throws RuntimeException
+     * @throws InvalidArgumentException
+     */
+    public function fromSugarEmailAddressField($emailField)
+    {
```
gymad: This function body still duplicated code. Exactly same method name and body lines in `include/SugarObjects/templates/basic/Basic.php` at line 142 and line 144 to 156.

modules/Emails/Email.php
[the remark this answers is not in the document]
daniel-samson: I have fixed it. I made one validation method for sugar email address field and other one for the email module. The fromSugarEmailAddressField method will call the validation method. fromSugarEmailAddressField is used by most of these the methods


I have also added this into the getConfirmOptInTickFromSugarEmailAddressField just in case the fromSugarEmailAddressField is not called.

modules/Emails/Email.php
daniel-samson: Sorry I didn't see that in the email.php. I have removed it

include/SugarObjects/templates/basic/Basic.php
gymad: But the developers may want. Anyway, It's hard coded at the moment but it should be a function parameter.

include/SugarObjects/templates/basic/Basic.php
```diff
@@ -82,76 +88,100 @@ public function get_summary_text()
...
+            if($emailAddress->invalid_email) {
+                return 'INVALID_EMAIL';
+            }
+
+            if ($emailAddress->opt_out == "1") {
+                return 'OPT_OUT';
+            }
+
+            if ($emailAddress->confirm_opt_in == "1") {
+                return "OPT_IN_PENDING_EMAIL_CONFIRMED";
```
gymad: php CS standars: use single quote `'` char at string constants

include/SugarObjects/templates/basic/Basic.php
```diff
@@ -82,76 +88,100 @@ public function get_summary_text()
      * @throws RuntimeException
      * @throws InvalidArgumentException
      */
-    public function getEmailAddressConfirmOptInTick($emailField)
+    public function getConfirmOptInTickFromSugarEmailAddressField($emailField)
```
gymad: Update for php doc: This method doesn't throws `RuntimeException` anymore.

include/SugarObjects/templates/basic/Basic.php
```diff
@@ -82,76 +88,100 @@ public function get_summary_text()
...
-            $tickTitle = $app_strings['LBL_CONFIRM_OPT_IN_TITLE'];
-            $tickHtml = '<span class="confirm-opt-in-tick" title="' . $tickTitle . '">&#10004;</span>';
+
+    /**
+     * @global array $sugar_config
+     * @global \LoggerManager $log
+     * @param string $emailField
+     * @return string
+     */
+    public function getOptInStatusFromSugarField($emailField)
```
gymad: Update php doc: This method doesn't use globals (`$sugar_config` and `$log`) anymore.

include/SugarObjects/templates/basic/Basic.php
```diff
@@ -82,76 +88,100 @@ public function get_summary_text()
...
+        return 'UNKNOWN_OPT_IN_STATUS';
     }
 
     /**
      *
      * @global array $sugar_config
      * @global \LoggerManager $log
      * @param string $emailField
-     * @return boolean
+     * @return \EmailAddress|\SugarBean
```
gymad: Update php doc: It doesn't return `SugarBean` object anymore.

include/SugarObjects/templates/basic/Basic.php
```diff
@@ -82,76 +88,100 @@ public function get_summary_text()
...
      * @throws RuntimeException
      * @throws InvalidArgumentException
      */
-    public function getEmailAddressConfirmOptIn($emailField)
+    public function fromSugarEmailAddressField($emailField)
```
gymad: Update php doc: It doesn't throw `RuntimeException` anymore.

include/SugarObjects/templates/basic/Basic.php
```diff
@@ -82,76 +88,100 @@ public function get_summary_text()
...
+            }
+
+            if ($emailAddress->opt_out == "1") {
+                return 'OPT_OUT';
+            }
+
+            if ($emailAddress->confirm_opt_in == "1") {
+                return "OPT_IN_PENDING_EMAIL_CONFIRMED";
+            } elseif(!empty($emailAddress->opt_in_email_created)) {
+                return "OPT_IN_PENDING_EMAIL_SENT";
```
gymad: php CS standars: use single quote `'` char at string constants

include/SugarObjects/templates/basic/Basic.php
```diff
@@ -82,76 +88,100 @@ public function get_summary_text()
...
+            } elseif(empty($emailAddress->opt_in_email_created)) {
+                return "OPT_IN_PENDING_EMAIL_NOT_SENT";
```
gymad: php CS standars: use single quote `'` char at string constants

include/SugarObjects/templates/basic/Basic.php
```diff
@@ -82,76 +88,100 @@ public function get_summary_text()
...
+        return $emailAddressBean->retrieve($emailAddressId);
     }
 
     /**
      *
      * @global \LoggerManager $log
      * @param string $emailField
      * @return string|null EmailAddress ID or null on error
-     * @throws RuntimeException
-     * @throws InvalidArgumentException
```
gymad: Update php doc: This method still can throws `InvalidArgumentException`.

include/SugarObjects/templates/basic/Basic.php
gymad: Still don't need to retrieve the all of the Basic bean even if it's already retrieved. Check the ID first if it's more make sense for you.

include/SugarObjects/templates/basic/Basic.php
```diff
@@ -160,21 +190,50 @@ private function getEmailAddressId($emailField)
...
-            throw new RuntimeException('A Basic bean has not selected email address. (' . $emailAddress . ')');
+            // Changed exception to error as demo data is never selected.
+            $log->fatal('A Basic bean has not selected email address. (' . $emailAddress . ')');
+            return null;
         }
 
         return $emailAddressId;
     }
 
+    protected function displayOptInFromSugarEmailAddressField($emailField)
```
gymad: Add php doc

include/SugarObjects/templates/basic/FunctionField/displayEmailAddressOptInField.php
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+if (!defined('sugarEntry') || !sugarEntry) {
+    die('Not A Valid Entry Point');
+}
+
+
+/**
+ * @param \Basic $focus
+ * @param $field
```
gymad: Update php doc: param type is not defined.

include/SugarObjects/templates/basic/FunctionField/displayEmailAddressOptInField.php
```diff
@@ -0,0 +1,61 @@
...
+ * @param $value
```
gymad: Update php doc: param type is not defined.

include/SugarObjects/templates/basic/FunctionField/displayEmailAddressOptInField.php
```diff
@@ -0,0 +1,61 @@
...
+ * @param $view
```
gymad: Update php doc: param type is not defined.

include/SugarObjects/templates/basic/Basic.php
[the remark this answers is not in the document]
gymad: ok

include/SugarObjects/templates/basic/Basic.php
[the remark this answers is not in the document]
gymad: ok

modules/EmailAddresses/EmailAddress.php
```diff
@@ -52,20 +52,52 @@
  */
 class EmailAddress extends SugarEmailAddress
 {
-	var $disable_row_level_security = true;
+    /**
```
gymad: Can you please fix the indents in this file.

modules/Emails/include/displayEmailAddressOptInField.php
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+if (!defined('sugarEntry') || !sugarEntry) {
+    die('Not A Valid Entry Point');
+}
+
+
+/**
+ * @param \Email $focus
+ * @param $field
```
gymad: Update php doc: param type is not defined.

modules/Emails/include/displayEmailAddressOptInField.php
```diff
@@ -0,0 +1,59 @@
...
+ * @param $value
```
gymad: Update php doc: param type is not defined.

modules/Emails/include/displayEmailAddressOptInField.php
```diff
@@ -0,0 +1,59 @@
...
+ * @param $view
```
gymad: Update php doc: param type is not defined.

modules/Emails/include/displayEmailAddressOptInField.php
gymad: ok

include/SugarObjects/templates/basic/Basic.php
daniel-samson: i have updated updated php doc.

include/SugarObjects/templates/basic/Basic.php
```diff
@@ -82,76 +88,100 @@ public function get_summary_text()
      * @throws RuntimeException
      * @throws InvalidArgumentException
      */
-    public function getEmailAddressConfirmOptInTick($emailField)
+    public function getConfirmOptInTickFromSugarEmailAddressField($emailField)
     {
-        $this->getEmailAddressValidateArguments($emailField);
+        $this->validateSugarEmailAddressField($emailField);
 
-        global $sugar_config, $app_strings;
```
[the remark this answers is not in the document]
daniel-samson: I have removed them from php doc.

include/SugarObjects/templates/basic/Basic.php
daniel-samson: I have removed them.

include/SugarObjects/templates/basic/Basic.php
daniel-samson: Whoops. I have fixed it.

include/SugarObjects/templates/basic/Basic.php
daniel-samson: I have fixed it.

include/SugarObjects/templates/basic/Basic.php
daniel-samson: I have fixed it.

include/SugarObjects/templates/basic/Basic.php
daniel-samson: Updated PHP Doc. There is problem with the retrieve method. The code inspection flags this up, because the PHP Doc in the SugarBean::retrieve() returns a SugarBean.

I suppose it's more truthful to say this returns an EmailAddress.

include/SugarObjects/templates/basic/Basic.php
daniel-samson: Updated php doc

include/SugarObjects/templates/basic/Basic.php
daniel-samson: updated php doc

include/SugarObjects/templates/basic/Basic.php
daniel-samson: Added

modules/Emails/include/displayEmailAddressOptInField.php
daniel-samson: updated

modules/Emails/include/displayEmailAddressOptInField.php
daniel-samson: updated

include/SugarObjects/templates/basic/Basic.php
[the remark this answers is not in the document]
gymad: ok

Issue #4789, opened 2017-12-27
403 Error Upon Survey Submission 


Glen206:
After selecting the submit button on a test survey I was sent to the thank you page at /modules/Surveys/Entry/Thanks.php

![image](https://user-images.githubusercontent.com/2359340/34389401-648cdbc0-eb07-11e7-825d-acd52b9d96b2.png)

Double  
#### Expected Behavior
Should be shown a thank you page

#### Actual Behavior
Get a 403 Forbidden message

#### Possible Fix
Not Sure

#### Steps to Reproduce
1. Create a survey
2. Take the Survey
3. Hit The submit button

#### Your Environment
SuiteCRM  7.10.0. beta-3
Chrome (64-bit)):
MySQL, PHP 7
Centos

Glen206:
I have the folders below set to 775 and files in them set to 755
SurveyQuestionOptions
SurveyQuestionsResponses
SurveyQuestions
SurveyResponses
Surveys

on another note but related, I do get the proper message when I close the survey and then try to take it again. I get the message telling me the survey is now closed

pgorod:
Permissions don't mean anything by themselves. You have to check also the ownership, and match both with your web server user.

I try accessing this on my system directly and it worked fine. Can you try directly accessing the file by typing in your browser

`http://www.yourserver.com/modules/Surveys/Entry/Thanks.php`

Lastly, check if you have any .htaccess file denying access (you could have some long-forgotten rule like `entry*` that happens to block this).

Glen206:
Thanks for the clue, it was in the .htaccess file with this line in it:

RedirectMatch 403 /+(soap|cache|xtemplate|data|examples|include|log4php|metadata|modules)/+.*\.(php|tpl)

once I removed modules, it works fine.

pgorod:
Well, then that needs attention because that's the default rule that the SuiteCRM installer writes...

So this _is_ a bug that needs fixing.

pgorod:
Well, then that needs attention because that's the default rule that the SuiteCRM installer writes...

https://github.com/salesagility/SuiteCRM/search?utf8=%E2%9C%93&q=RedirectMatch+modules&type=

So this _is_ a bug that needs fixing.

PedroErnst:
This should fix the issue without having to modify the htaccess file.

PR #4848, opened 2018-01-04
Fix/survey thanks page

PedroErnst:
Make the survey redirect to an entry point instead of pointing to the php file directly to avoid htaccess issues.
Pass the survey name as GET parameter to be able to display it.

## Motivation and Context

Issue #4789 

## How To Test This

Create a survey.
Answer the question.
Verify the thanks page opens and displays the correct survey name.

## Types of changes

### Final checklist

### DIFF
--- a/modules/Surveys/Entry/Thanks.php
+++ b/modules/Surveys/Entry/Thanks.php
@@ -20,11 +20,6 @@ $surveyName = !empty($_REQUEST['name']) ? $_REQUEST['name'] : 'Survey';
 </head>
 <body>
 <div class="container">
-    <div class="row">
-        <div class="col-md-offset-3 col-md-6">
-            <img height=100 src="survey_logo.jpg"/>
-        </div>
-    </div>
     <div class="row well">
         <div class="col-md-offset-2 col-md-8">
             <h1><?php echo $surveyName; ?></h1>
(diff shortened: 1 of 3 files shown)


modules/Surveys/Entry/Thanks.php
```diff
@@ -14,14 +20,9 @@
...
 <div class="container">
-    <div class="row">
-        <div class="col-md-offset-3 col-md-6">
-            <img height=100 src="survey_logo.jpg"/>
-        </div>
-    </div>
     <div class="row well">
         <div class="col-md-offset-2 col-md-8">
-            <h1><?= $survey->name ?></h1>
+            <h1><?php echo $surveyName; ?></h1>
```
daniel-samson: Question: The short hand tags "<?=" are available on php 5.4 and above. SuiteCRM 7.10 has a minimum requirement of 5.5.9. So why do we need to change these?

PedroErnst: PSR-1

PedroErnst: Yes my bad, that change is unnecessary

daniel-samson: Ah OK. I wanted to make sure that the community member who raised it, was using the correct php version.

Issue #3631, opened 2017-06-01
Attachments in Emails 7.9 aren't showing - In List View


Mausino:
#### Issue

In Email version 7.9 are not showing any email attachments 

I have set up the column Has Attachment
![attachment1](https://cloud.githubusercontent.com/assets/5773777/26670418/e12587f4-46b1-11e7-8b58-5e7086d8c97f.png)

And there are not there (in email client like mozilla, every second mail has the attachment)
![attachment2](https://cloud.githubusercontent.com/assets/5773777/26670449/046557a8-46b2-11e7-97b2-4c095f4a4322.png)

 #### Expected Behavior

Will indicate if email has or doesn't have attachments

#### Steps to Reproduce

1. Go to Email molule :)
2. Look on column Has Attachment

#### Your Environment

* SuiteCRM Version used: SuiteCRM 7.9
* Browser name and version: Chrome 59.x 64bit
* Environment name and version: MySQL 5.6, PHP 7.0.16
* Operating System and version: Win 10 Home

Dillon-Brown:
@Mausino Are the attachments missing only on the list-view? e.g. are the actual attachments themselves present on the email record?

Mausino:
@Dillon-Brown  no in detaiview i see the issue :(

Thunderbird
![at2](https://cloud.githubusercontent.com/assets/5773777/26673825/8a9c063a-46be-11e7-9dea-e1e06a1f9a60.png)

SuiteCRM 7.9
![at1](https://cloud.githubusercontent.com/assets/5773777/26673837/96df4f10-46be-11e7-9216-99b35cf9f1a6.png)

Dillon-Brown:
@Mausino This is due to the email not being imported into the CRM. The attachment icon won't appear until an email has been successfully been imported.

Mausino:
@Dillon-Brown after import of emial and  doesn't see it on listview... in detailview i see the attachment field and there the name of the file....

But i didn't understand you in your sentence: "This is due to the email not being imported into the CRM. The attachment icon won't appear until an email has been successfully been imported."

You mean that will no possible see the email icon in listview and the name of file in detailview until system or user don't import email ???  If yes, how we will know if email has the attachment before importing... Without this BASIC function will the emails totally usefull !!!  

@pgorod @horus68 @samus-aran @daniel-samson please guys what are you thinking about it.. ???

Dillon-Brown:
[...]

PR #4866, opened 2018-01-08
Fix/email show attachment

PedroErnst:
Removed a bug when checking unimported emails for attachments.

## Motivation and Context

Issue #3631 

## How To Test This

Any emails with attachments should now display the paperclip in listview.

## Types of changes

### Final checklist

### DIFF
--- a/modules/InboundEmail/InboundEmail.php
+++ b/modules/InboundEmail/InboundEmail.php
@@ -2660,7 +2660,7 @@ class InboundEmail extends SugarBean
 
             $focusUser = new User();
             $focusUser->retrieve($groupId);
-            $mailerId = (isset($_REQUEST['outbound_emaihttp://localhost:9001/SuiteCRM-github-develop/index.php?module=Users&action=EditView&record=1#l'])) ? $_REQUEST['outbound_email'] : "";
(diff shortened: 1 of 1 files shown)


modules/InboundEmail/InboundEmail.php
```diff
@@ -504,6 +495,27 @@ function($a, $b) use($sortCRM) {  // defaults to DESC order
...
+                    return true;
+                }
+            }
+        } else {
+            if (isset($imapStructure->disposition) && $imapStructure->disposition == 'attachment') {
+                return true;
+            }
+        }
+        return false;
+    }
```
zoltankocsardi: You can split the if condition, like:

`        if (!isset($imapStructure->parts)
            && isset($imapStructure->disposition)
            && $imapStructure->disposition == 'attachment') {
            return true;
        }

        if (isset($imapStructure->parts)) {
            foreach ($imapStructure->parts as $part) {
                if ($this->mesageStructureHasAttachment($part)) {
                    return true;
                }
            }
        }

        return false;
`

I believe it's more readable and you return a case immediately

PedroErnst: Yes I agree, makes it easier to read.

Issue #3370, opened 2017-04-07
Links to the website urls without http://


matteoraggi:
#### Issue
Hi, often by hand is easy to write a url like: www.eforte.it instead of www.eforte.it , and I discovered that if I write the urls without http or https in front ot the url, then the link is goign to www.SUITECRM/www.eforte.it instead of www.eforte.it . I think that it is a bug, and vtigercrm was having the same (and bigger) bug too, till few years ago.

Here a video should explain better of my words:
www.eforte.it/suitecrm-links-web-url.mp4

When inline changing a URL on a account field, then confirm to save.
If you click the new URL it gets this prefix
https://demo.suiteondemand.com
as in
https://demo.suiteondemand.com/www.beansthe.com
But if you refresh the page its ok!
Also if you do not use the inline edit (use the full edit view) then no problem too.

Apache php 5.6 suitecrm 7.8.2
If needed you can find more info here:
https://suitecrm.com/forum/suitecrm-7-0-discussion/13777-links-to-the-website-urls-without-http#46185

PR #4873, opened 2018-01-08
Fix/missing http from links

PedroErnst:
Making sure that the behaviour when inline editing a url is the same as when editing throught the Edit View, and that links without http:// or https:// prefix don't create problems.

## Motivation and Context

Issue #3370

## How To Test This

Go to Detail View of an Account that has a website
Inline edit the URL

## Types of changes

### Final checklist

### DIFF
--- a/include/InlineEditing/InlineEditing.php
+++ b/include/InlineEditing/InlineEditing.php
@@ -520,10 +520,11 @@ function formatDisplayValue($bean, $value, $vardef, $method = "save")
             $value .= "</a>";
         }
     }
-	if($vardef['type'] == "url")
-	{
-		$value = '<a href='.$value.' target="_blank">'.$value.'</a>';
-	}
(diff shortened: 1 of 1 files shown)


gunnicom:
Did not look exactly into this, but what if for some reason you want a relative link.
Say you have a field where you want to link within your suite and want it to work if it is on http or https no matter what? Like "index.php?action=...."
Won't your fix break such links?

PedroErnst:
My fix only changes the value that is returned in the ajax response after inline editing, making it equal to the value that would be displayed if the field was edited in the edit view, or if the page was simply reloaded after doing the inline edit.

So it does not affect any behavior or any data being saved in the db, it only makes the way it is displayed consistent with other views.

PedroErnst:
@Dillon-Brown Yes that's right. Changed now.

PR #4906, opened 2018-01-10
Improved tooltip regarding DB configuration on install

pgorod:
This is a consequence of the discoveries in this discussion: #3168 

Just an improved label.


Dillon-Brown:
Assessed

Issue #4907, opened 2018-01-11
Adding a Contact/Lead through Calls or Meetings entry bypasses mandatory field requirements


jsamelko:
#### Issue
The 'Create an invitee' sub-panel available when creating a new Call or Meeting allows a user to create new Contacts in the system that don't meet mandatory field requirements, and cannot be disabled/hidden. Additionally, duplicate checks are not performed against existing contacts and multiple entries are created

#### Expected Behavior
The Create as Contact / Create as Lead forms should be customizable as with Layouts for EditView/QuickCreate, or be able to be hidden. The form requests only very basic information, and where the EditView and QuickCreate layouts may have mandatory fields for 'Account' or 'Phone' this enables those fields to be skipped. 

This can result in a large number of incorrect/incomplete Contact records with very minimal information that require rework/additional time to correct and update

#### Actual Behavior
A user can enter a contact with only First Name, Last Name and Email address, where requirements might additionally mandate Phone, Assigned User and Account. If edit rights are set to owner then the entering user is not able to later update/correct the record, and an Admin user or user with All edit access must correct each entry manually or follow-up

#### Possible Fix
- Allow the 'Create an invitee' sub-panel to be customized through Studio, or allow for it to be disabled/hidden per user/role/system-wide. 
- Assign the created Contact the the creating user
- Perform a duplicate check on existing Contact records

#### Steps to Reproduce
1. Ensure current user Edit access set to Owner for Contacts
2. In Studio, select the contacts module and set Field 'title' as a required field
3. Test entry of a new Contact and confirm Title required before save
4. Log a new Call. Create an invitee > As Contact > enter Name & Email identical to earlier entry
5. Confirm in Contacts a new entry has been created without the required Title field, as a duplicate of existing Contact
6. Attempt to edit to correct/update. As Contact does not have Assigned User, the SuiteCRM user is not able to edit

#### Context
Users in our company are adding new Contacts to Call and Meeting records without sufficient information to update or correct later (e.g. What division of the company is responsible for the contact, the contact's account and contact details, call type, etc)

Because of the layout in the Log Call/Schedule Meeting forms users easily confuse the 'Create an Invitee' button to mean they are creating a record for an existing contact against the Call being entered, or don't understand the repercussions of incomplete records created through this method. This regularly results in duplicate contacts and contacts with insufficient details for later follow-up or correction

#### Your Environment
* SuiteCRM Version used: 7.9.7
[...]

PR #4914, opened 2018-01-11
Fix/Add attendees through popup

PedroErnst:
Replaces the procedure for attaching (and creating) invitees and resources to the scheduler in these modules:
Calls
Meetings
Projects
Project Templates

## Description

Fixes issue #4907 

Instead of an incomplete and bug prone form the system now opens the same popup as from the subpanel.

screenshot from 2018-01-11 15-14-07

screenshot from 2018-01-11 15-14-33

Requires rebuilding scss for correct button visuals.

## Motivation and Context

Fixes potential issues with the way records were being created in these modules.
Creating records happens now from the standard view in a popup, with all the proper checks.

## How To Test This

See the mentioned issue.

## Types of changes

### Final checklist


daniel-samson:
Assessed

samus-aran:
My issue here is that, yes, it is consistent, but we are removing a very easy and quick way to add invitees. Imagine you work in a call center and you are having to click the button, have a popup window open, find the search for that blah blah. Thats alot of extra clicks.

I think we can certainly use this for when we want to create an invitee, but moving the search area into the popup is a step backwards IHO.

Are we able to have the quick create form appear within the current form when you click the Create Invitee? Or would it be easier to keep the popup for the quick create, but ensure that when they select from the popup search that it too adds to the invitee list (which I'm assuming it does in your solution).

PR #4950, opened 2018-01-15
Fix #533

### LINKED ISSUE #533: Workflow - Setting Date field based on Another Date field. Always 1 Save behind.
https://suitecrm.com/forum/advanced-opensales-workflow-portal-support/6566-workflow-for-date-not-working

A rough Description would be if you:
Created Workflow to set Datefield2 to equal Datefield1 +6 Months, Does not change on record save.
Edit>Save record again, Datefield2 is set to the correct date.

However, if you make Datefield1 Equal to vastly different dates for each save, you can clearly see that Datefield2 is set to the previous value of Datefield1 +6 months

While not exactly a fix, Scheduled tasks does help with this as the next run will set it to the correct date.

Jason-Dang:
## Description

Updated date assignments to not use fetched_row as when the date/datetime values are pulled from a newly created record the values do not yet exist in the database and so fetched_row returns false. This causes the date/datetime value to be blank which results in the value being set to today/now. New assignment gets the values straight from the bean rather than the database which ensures the values are set correctly.
## Motivation and Context

When a Workflow is setup to populate a date/datetime value from a newly created record the value to be used is not retrieved as it does not yet exist in the database causing it to default to today/now.

## How To Test This

1. Create a Workflow that runs on new records for a module such as Invoices. 
2. Create an action on the Workflow to create a Task record. 
3. Add a field line to populate the required 'Subject' field on the Task record. 
4. Add another line to populate the 'Due Date' field of the Task record with the value of a Date/Datetime field such as 'Date Created' + 1 week. 
5. Save the Workflow.
6. Create an Invoice record and populate the date field that was set on the Workflow. 

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOW_Actions/actions/actionCreateRecord.php
+++ b/modules/AOW_Actions/actions/actionCreateRecord.php
@@ -248,11 +248,13 @@ class actionCreateRecord extends actionBase {
                                 if($params['value'][$key][0] === 'now'){
                                     $date = gmdate($dformat);
                                 } else if($params['value'][$key][0] === 'field'){
-                                    $date = $record->fetched_row[$params['field'][$key]];
+                                    $dateToUse = $params['field'][$key];
+                                    $date = $record->$dateToUse;
                                 } else if ($params['value'][$key][0] === 'today') {
                                     $date = $params['value'][$key][0];
                                 } else {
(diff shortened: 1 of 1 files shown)


Dillon-Brown:
Assessed

PR #4952, opened 2018-01-15
Feature/gdpr bulk optin

gymad:
## Description

- bulk action - confirm opt in

If Confirm opt in enabled:-

Bulk Action send confirm opt in Email from person/Company list view
Send email via email queue / background task
log confirm opt in date sent on email
Person refers to all module of type Person excluding Users
company refers to all module of type Company

## Motivation and Context

GDPR

## How To Test This

- QRR and Rebuild Minified JS files needed.
- Use BULK ACTION on list view to send confirm opt in emails via email queue

## Types of changes

### Final checklist


modules/EmailMan/EmailManDelivery.php
```diff
@@ -248,11 +252,35 @@
...
-            $GLOBALS['log']->fatal("Email delivery FAILURE:" . print_r($row, true));
+        if((empty($row['related_confirm_opt_in']) || $row['related_confirm_opt_in'] == '0')) {
+            if (!$emailman->sendEmail($mail, $massemailer_email_copy, $test)) {
+                $GLOBALS['log']->fatal("Email delivery FAILURE:" . print_r($row, true));
+            } else {
+                $GLOBALS['log']->debug("Email delivery SUCCESS:" . print_r($row, true));
+            }
         } else {
-            $GLOBALS['log']->debug("Email delivery SUCCESS:" . print_r($row, true));
+            if($confirmOptInEnabled) {
```
daniel-samson: is the if a automated formatting issue? should it be elseif?

include/EntryPointConfirmOptIn.php
```diff
@@ -43,40 +43,104 @@
 /**
  * Class confirm_opt_in
  */
-class ConfirmOptIn
+class EntryPointConfirmOptIn
```
daniel-samson: So this class is a view. The changes in this class is starting to look like a controller. I created ConfirmedOptIn to differentiate the display logic from the entry point logic. Perhaps I should have named it ConfirmedOptInView.

These changes will prevent someone from customising the process in the custom folder. The entry point is the file which could be overridden here. Which would let a third party customise how the opt in process works.

Also, if the email address logic is in the basic template, why do we need to move the confirmed opt in to a locate outside the basic template?

include/EntryPointConfirmOptIn.php
```diff
@@ -43,40 +43,104 @@
...
+            case 'confirmOptInSelected':
+                $output = $this->methodConfirmOptInSelected($post);
+                break;
+            default:
+                $output = $this->methodConfirmOptInUser($request);
+                break;
+        }
+        
+        echo $output;
+        sugar_cleanup();
```
daniel-samson: sugar_cleanup in the constructor makes it harder to customise.  It also unusual to control the process using a constructor. Typically a constructor sets up the state of a class. clean up code would normally be in a destructor.

However, I would be better to call this logic form the entrypoint file. This needs to be in a different class.  So a third party can customise it.

include/EntryPointConfirmOptIn.php
```diff
@@ -43,40 +43,104 @@
...
+    }
+    
+    /**
+     * 
+     * @global array $sugar_config
+     * @param array $post
+     * @return string|boolean
+     */
+    private function methodConfirmOptInSelected($post) {
```
daniel-samson: naming a method with the prefix method seems a little redundant.

include/EntryPointConfirmOptIn.php
```diff
@@ -43,40 +43,104 @@
...
 
     /**
-     * Set up
+     * Confirm Opt In User
+     * 
+     * @param array $request
+     * @return string
      */
-    public function pre_display()
+    private function methodConfirmOptInUser($request)
```
daniel-samson: I am concerted that this will leave a nasty merge conflict in git.

In relation to my previous comment, I mentioned that this is a view. Expects a contact, account, lead, or target. So naming it with User at the end is unclear. The tick should not be displayed in users and employees.

include/EntryPointConfirmOptIn.php
```diff
@@ -43,40 +43,104 @@
...
-     */
-    public function display()
-    {
-        
+                
         $template = new Sugar_Smarty();
         $template->assign('FOCUS', $this->emailAddress);
 
-        return $template->fetch('include/SugarObjects/templates/basic/tpls/entrypoints_confirm_opt_in.tpl');
+        return $template->fetch('include/EntryPointConfirmOptIn.tpl');
```
daniel-samson: Again? Why are we moving these files?

include/ListView/ListViewSmarty.php
```diff
@@ -79,14 +79,51 @@ public function __construct() {
...
+        global $app_strings, $sugar_config;
+        
+        $confirmOptInEnabled = isset($sugar_config['email_enable_confirm_opt_in']) && $sugar_config['email_enable_confirm_opt_in'];
+
+        if (!$confirmOptInEnabled) {
+            $this->warn('Confirm Opt In disabled');
+            return false;
+        }
+        
+        $js = "sListView.send_form("
```
daniel-samson: Are we not trying to move away from hard coded content in PHP?

Could we not move this into the template some how?

include/ListView/ListViewSmarty.php
```diff
@@ -79,14 +79,51 @@ public function __construct() {
...
 
     /**
      * Processes the request. Calls ListViewData process. Also assigns all lang strings, export links,
      * This is called from ListViewDisplay
-     *
-     * @param file file Template file to use
-     * @param data array from ListViewData
-     * @param html_public string the corresponding html public in xtpl per row
+     * 
+     * @param file $file Template file to use
```
daniel-samson: :+1:

include/MVC/Controller/entry_point_registry.php
```diff
@@ -58,7 +58,7 @@
 	'WebToPersonCapture' => array('file' => 'modules/Campaigns/WebToPersonCapture.php', 'auth' => false),
     'removeme' => array('file' => 'modules/Campaigns/RemoveMe.php', 'auth' => false),
     'ConfirmOptIn' => array(
-        'file' => 'include/SugarObjects/templates/basic/entrypoints/confirm_opt_in.php',
+        'file' => 'include/entryPointConfirmOptInConnector.php',
```
daniel-samson: see previous comment about moving?

include/EntryPointConfirmOptIn.php
gymad: So this class is a view. - No, not anymore. This class handle the entry points, (see constructor). So it's a controller.

include/EntryPointConfirmOptIn.php
```diff
@@ -43,40 +43,104 @@
...
+        
+        if(is_null($request)) {
+            $request = $_REQUEST;
+        }
+        
+        if(is_null($post)) {
+            $post = $_POST;
+        }
+        
+        $method = isset($request['method']) && $request['method'] ? $request['method'] : null;
```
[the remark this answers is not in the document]
gymad: If you someone create a custom entry point, then add a custom name for it. I don't understand the problem. Please explain.

include/language/en_us.lang.php
```diff
@@ -1559,6 +1559,7 @@
     'LBL_ADD_TO_PROSPECT_LIST_BUTTON_LABEL' => 'Add To Target List',
     'LBL_ADD_TO_PROSPECT_LIST_BUTTON_TITLE' => 'Add To Target List',
     'LBL_ADD_TO_PROSPECT_LIST_BUTTON_LABEL_ACCOUNTS_CONTACTS' => 'Add Contacts To Target List',
+    'LBL_SEND_CONFIRM_OPT_IN_EMAIL' => 'Send Confirm OptIn Email',
```
horus68: Please avoid new words and look for consistency: the word is used in SuiteCRM now in some different forms: Opt In, Opt in and OptIn.
Don't know if also used as Opt-in

gymad: Ok, which one do you prefer? :)

daniel-samson: I can't see that change. Are you referring to an other file?

gymad: I will using "Opt In" anywhere because It is the most commonly used already.

horus68: I'm not an english user... so feel free to pick one!
Usually I fund it funny that english language its so easy to create new concepts and words!
see: Confirmed opt-in (COI) / Double opt-in (DOI)
https://en.wikipedia.org/wiki/Opt-in_email

include/EntryPointConfirmOptIn.php
daniel-samson: "No, not anymore" is not an adequate answer.

It is best practice to separate our business logic from the display logic. We should have a controller and a view class.

I am still not sure why we are moving the logic away from the basic sugar object. Can you please clarify? What was the issue here?

gymad: I don't think there is any reason to add it to a path like 'sugarObject/template'. Even if it is not a sugar object nor template. You added a new PHP class into a template folder which was not a SugarObject. the PHP classes are PHP classes but NOT templates. also it was not a sugar objects.

include/language/en_us.lang.php
gymad: I'm not an english user also, so every time I'm a little bit confused about the language texts :D

include/EntryPointConfirmOptIn.php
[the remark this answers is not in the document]
daniel-samson: My concern is related the inversion of control, do not repeat and that it is uncommon to use a constructor in this fashion.

A third party developer is likely to to want just to call methods from the entrypoint; because it is the usual pattern we use in SuiteCRM. Since we are using classes, they would expect the following pattern:

```php
<?php
$foo = new foo();  // instantiate class
$foo->doBar(): // Do something with class
if ($foo->isSomething()) {
    $foo->doBuzz(): // Do something else with class
}
```

conflating the constructor with the overall process is a confusing pattern, because we would expect the construction to be separate from calling the methods in the "controller" to control the flow.

Just to be clear, there is nothing wrong with calling a method in the constructor. However, you would expect the construct to set up the class, then call something in the class, and then deal with the response. The flow of control should be in the entrypoint file. So that we do not need to repeat the code in the customisation. 

I also have no problem with using a controller. provided that we follow the pattern. This is why I am suggesting to have a separate class for the view. and move any non PHP code into the template.

So can we move some of this code in to other methods eg. execute/process? Can we keep the view and business logic separate?

This is also in relation to the sugar_cleanup call. Please see my comment below regarding the sugar_cleanup. Since github comments are related to the line of code.

include/EntryPointConfirmOptIn.php
daniel-samson: I added sugar_cleanup to the entrypoint, instead of the class. 

Calling sugar_cleanup ends the process. The issue here is what happens if I as a third party wish to add more functionality to this code. As soon as I construct the class, it will end the process, preventing me from adding the functionality. 

It it true that it will call the method in this case. However, in order to customise this, I would have to throw away or replace much of this code. Which is less than ideal.

include/EntryPointConfirmOptIn.php
daniel-samson: i mean that if you are calling a method - why do you need to name it method?

in the same way if you are instatating a class (eg DogClass). Why would you put "class" in the name of the class. It's not a big thing. It's just unnecessary.

If we are making this a controller, the convention in SuiteCRM is to prefix it with "action".

include/SugarObjects/templates/basic/Basic.php
```diff
@@ -113,9 +113,9 @@ public function getOptInStatusFromSugarField($emailField)
 
             if ($emailAddress->confirm_opt_in == '1') {
                 return 'OPT_IN_PENDING_EMAIL_CONFIRMED';
-            } elseif(!empty($emailAddress->opt_in_email_created)) {
+            } elseif(!empty($emailAddress->confirm_opt_in_sent_date)) {
```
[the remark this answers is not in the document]
daniel-samson: Perhaps I could be clearer here. I am concerned because you have renamed and moved the class. You see my pull request expects the ConfirmedOptIn. So I am worried that this is going to clash.

include/EntryPointConfirmOptIn.php
daniel-samson: When I initially added this code, I wasn't sure where to put the entry point code. We don't want to create duplicated code across the accounts, contacts, etc. The entrypoint code is often put in the root folder of a module eg "removeme" => "modules/Campaigns/RemoveMe.php".

You added the method getConfirmOptInTickFromSugarEmailAddressField was added to display the opt in tick. So in order to create a general way to display the tick that allows child classes to display it, I separated how to check the opt in status and the displayOptInFromSugarEmailAddressField method to handle the display logic.  Perhaps this should be moved. 

So the reason I thought it should be in the Basic.php is because it is related to the code we added in the Basic.php. It is the general case for displaying the tick in an email address field. I assumed, when you create the method in the basic case, you where attempting to make possible for all beans to have an email address field.

include/EntryPointConfirmOptIn.php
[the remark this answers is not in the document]
gymad: I believe that we using OOP. Your example isn't but the next one is:
```
Buzzer.php
<?php
class Buzzer {
    public function __construct() {
        $foo = new foo();  // instantiate class
        $foo->doBar(): // Do something with class
        if ($foo->isSomething()) {
            $foo->doBuzz(): // Do something else with class
        }
    }
}

BuzzerCall.php:
<?php
new Buzzer();
```

now you still can make a CustomBuzzer, for e.g:
```
<?php
new CustomBuzzer { ... }
```
or you can implement any custom php as in your example above. Nothing to changed.

include/EntryPointConfirmOptIn.php
gymad: If you want to extend the class, use php `extends`. If you want to a custom, use SuiteCRM custom. see previous comment.

include/EntryPointConfirmOptIn.php
gymad: you can, but the "action" already in use. It's also came from the request: https://github.com/salesagility/SuiteCRM/pull/4952/files#diff-1ca841fc98ed932c05fcbe4880f3f6e8R68

include/SugarObjects/templates/basic/Basic.php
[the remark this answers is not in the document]
gymad: Many times I also have to change my code to make it compatible with the repository: for e.g. last time: https://github.com/salesagility/SuiteCRM/pull/4932/files#diff-25e9e5d4bad721b8ee020c27be3e7851L278
I don't think it's a big deal.

include/EntryPointConfirmOptIn.php
gymad: Ok, but in this case why we needed a new class. Why didn't you just use the exists `Basic`, using new method(s)?

include/EntryPointConfirmOptIn.php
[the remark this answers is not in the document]
daniel-samson: My code example represents entrypoint file.

Controlling the process in a entrypoint file is the pattern used in SuiteCRM. If we are building a controller, we should follow the same pattern. Take the usage SugarApplication::execute as an example.

In the execute method, it set up a controller and executes it's action. This is what I mean by confusing; because in the current convention, the construct and the caller, and the action are separated. However, in your code the are all combined.

Furthermore, the https://github.com/salesagility/SuiteCRM/pull/4932/files#diff-25e9e5d4bad721b8ee020c27be3e7851L278 - could add a check to the switch case so you that it calls the the method dynamically? that would solve this issue.

modules/EmailMan/EmailManDelivery.php
gymad: reformatted

include/ListView/ListViewSmarty.php
gymad: outsourced

include/EntryPointConfirmOptIn.php
gymad: anyway, I outsourced the switch..case to a separated method so if you want to override only the requested methods you can extends the class.

include/EntryPointConfirmOptIn.php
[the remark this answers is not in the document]
gymad: outdated. see previous comment and sourcecode

Dillon-Brown:
Assessed

PR #4954, opened 2018-01-16
Fix #4636 Add template parsing to 7.8.x email

### LINKED ISSUE #4636: email variables are not filled properly in email template...
Using email template in any of the process that we would benefit from such an interresting feature; When we send an email and would like to add variables from the menu they do not get fulfilled.
This problem has been reported really often in the forum but I believe no one  reported here.

how to replicate, simply build an email template and insert variables, 99% of them won't work when you are sending emails...  the variable will not be filled up.

It would such a nice feature to bring to life...

PedroErnst:
Adds email template parsing to 7.8.x
Fixes #4636 
Takes code from aa1577ed62585530d5449aecc967fab60f034e26

## How To Test This

Create an Email Template 
Relate to Cases
Insert a few Case variables
Compose an Email
Load the template
Select related case
Send Email
See that variables are replaced

## Types of changes

### Final checklist

### DIFF
--- a/modules/Emails/Email.php
+++ b/modules/Emails/Email.php
@@ -564,7 +564,38 @@ class Email extends SugarBean {
 		        $this->type = 'out';
 		        $this->status = 'sent';
 			}
-        }
+
+        // This code is 7.8.x LTS specific, from 7.9 onwards it is found in EmailsController and can be deleted here
+        if (!empty($_REQUEST['data_parent_id1'])) {
+            $macro_nv = array();
+            $focusName = $request['parent_type'];
+            $focus = BeanFactory::getBean($focusName, $request['parent_id']);
+            if ($this->module_dir == 'Accounts') {
+                $focusName = 'Accounts';
+            }
+
+            $emailTemplate = BeanFactory::getBean(
+                'EmailTemplates',
+                isset($request['emails_email_templates_idb']) ?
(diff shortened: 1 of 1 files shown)


Dillon-Brown:
Assessed

PaulMcRoar:
Sorry to say that we are still having major problems with this. On SuiteCRM 7.8.12 , the above works when you compose an email and then select a case, but not if you select an opportunity.

It also still does not work for cases if you select "Compose Email" from the activities tab of a case (which is a much more common workflow ).

PaulMcRoar:
Managed to have a glance through what is being posted in the workflow method as opposed to the method where you compose an email.  When clicking from the activities tab, the parameters are data_parent_id0 etc as opposed to data_parent_id1

A change to line 569 to

if (!empty($_REQUEST['data_parent_id1'] || !empty($_REQUEST['data_parent_id0']))) {

and it works OK, although I'm unsure if that's the best way about doing it, as I've not dug under the hood of SuiteCRM much.

(when composing an email and then selecting the opportunity it DOES work, contrary to my initial testing above).

PedroErnst:
Thanks @PaulMcRoar for spotting this. Your proposed solution does fix the issue, so I've made PR for it. It is maybe not the best way to do it, but it works, and this particular piece of code will be made obsolete by newer versions anyway.

PR #4956, opened 2018-01-16
Fix/issue206

Jason-Dang:
## Description

Added conditional statement to check if the table_name is campaign_log and if so use activity_date instead of date_entered in the queries for both new and modified records to prevent query failure. 

## Motivation and Context

Previously the queries failed when workflows were setup to be triggered by Campaign Logs as the date_entered field does not exist in that table and instead there is an activity_date field.

## How To Test This

1. Create an Email Campaign with a tracker URL and send the Campaign.
2. Setup a Workflow that is triggered by either new or modified Campaign Log records. For example to trigger on Campaign click-thrus, set it up for new records and add a condition for activity_type = 'link'. 
3. The Workflow should run without a query failure due to the missing database column 'date_entered'

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOW_WorkFlow/AOW_WorkFlow.php
+++ b/modules/AOW_WorkFlow/AOW_WorkFlow.php
@@ -301,11 +301,19 @@ class AOW_WorkFlow extends Basic
                 switch($this->flow_run_on){
 
                     case'New_Records':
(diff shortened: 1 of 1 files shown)


Dillon-Brown:
Assessed

PR #4965, opened 2018-01-17
Fix studio popup.

e-reeley:
Popup after save click on studio edit was not clickable.

## Description

The index of the mask was reduced to allow the popup to be accessible. 

## How To Test This

1) Navigate to studio module
2) Select a module to edit
3) Navigate to Edit View inside Layouts
4) Select save

## Types of changes

### Final checklist

### DIFF
--- a/themes/SuiteP/css/main.scss
+++ b/themes/SuiteP/css/main.scss
@@ -95,7 +95,7 @@ a:hover {
 // Fix ajax loader in the edit view
 // Since the z-index is inline the only way to override it is to use the !important
 .mask {
-  z-index: 3 !important;
+  z-index: 2 !important;
 }
 
 #ajaxloading_c {


Dillon-Brown:
Assessed

actXc:
I use 7.9.9 fresh install with two languages (EN/DE) 
1) try to change a dropdown list
2) change in a second language
3) try to close the result message
The result message does not close, it hides behind a mask (guessing)
The patch #4965 does not help.

MandieLao:
@actXc  I had the same problem like you, also, #4965 didn't help, but  I think we can  changed another file accrodingly with the same logic. And it works for me!
Maybe you could try to change the file -- cache\themes\SuiteP\css\style.css 
Also, change .mask{z-index:3 !important} to .mask{z-index:2 !important}

e-reeley:
Hi @actXc , @MandieLao 

Sorry to hear that it has not fixed the issue! Is it the case the the mask is still set to 3 as mentioned in the comment directly above?

If this is the case, rebuilding the css should change the cache file for you, you shouldn't need to go in and manually do so :) 

We can talk you through this if need be! 

Thanks.

sasha2002:
@e-reeley And how to clear CSS?
I've tried rebuild and no luck, file is still old :(
I think something in cache clean is broken that's why people need to do this manually.

PedroErnst:
@sasha2002 

To rebuild the CSS you'll need to run sass.

sasha2002:
It's worse, bug from 7.8.X in 7.9.12 .....

PR #4994, opened 2018-01-19
Fix #4984 Make my pipeline dashlet not use labels for drilling down URL

### LINKED ISSUE #4984: My Pipeline Dashlet drilldown fails when sales stage have a different Display Label and Item Name
#### Issue
If a user has their "Sales Stage" objects have different Display Labels than Item Names, the My Pipeline drill-down will fail. 

This appears to be because the filter that clicking on the "My Pipeline" links will be created with a sales stage search that doesn't provide any results, as the sales-stage filter doesn't match anything.

#### Expected Behavior
Clicking the drill-down should apply a filter based on the hidden item name, rather than the displayed filter.

#### Actual Behavior
When using a sales stage of 90, but a display label of 90%, my sales stage drill-down filter discloses no results.

#### Possible Fix
It seems that making sure the Display Label and Item Name match fixes the filter, but I believe making it filter by Item name should result in the intended behavior.

#### Steps to Reproduce
1. Create a custom Sales Stage drop-down list. Use different Display labels than item names
2. Create a "My Pipeline" dashlet
3. Create an opportunity that has one of these mismatched fields
4. Attempt a drill through click from the home My Pipeline dashlet. It should not display any results, and the filter will show the display label.

#### Context
This is likely only a minor bug, as it has a simple fix, but it will likely confuse people, as the regular pipeline appears to work fine either way.

My team had used a display label of 90% for a 90, 80% for 80, etc, and it had completely broken the pipeline for our sales team other than for our "Won" field, which was won/won.

#### Your Environment

* SuiteCRM Version used: Version 7.9.5 Sugar Version 6.5.24 (Build 509)
* Browser name and version: Chrome Version 63.0.3239.132 (Official Build) (64-bit)
I don't run the server, but i think these are relevant if this is important.
MySQL Version: 5.1.36 
PHP Version : 5.3.1  
* Operating System and version: Windows 8.1

PedroErnst:
Makes the URL for drilldown filtering use the item name instead of the label.

## Motivation and Context

Issue #4984 

## Types of changes

### Final checklist

### DIFF
--- a/include/SuiteGraphs/RGraphIncludes.php
+++ b/include/SuiteGraphs/RGraphIncludes.php
@@ -75,7 +75,6 @@ $chart = <<<EOD
                 &&  bar['object']['properties']['chart.tooltips']!== undefined
                 &&  bar['object']['properties']['chart.tooltips'][bar[5]] !== undefined)
                 {
-                    var stage = encodeURI(bar['object']['properties']['chart.labels'][bar[5]]);
                     var graphId = bar[0]['id'];
                     var divHolder = $("#"+graphId).parent();
                     var module = $(divHolder).find(".module").val();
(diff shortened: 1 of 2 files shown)


Dillon-Brown:
Assessed

PR #4998, opened 2018-01-22
fixing Empty email ID

gymad:
## Description

fixing Empty email ID
## Motivation and Context

gdpr
## Types of changes

### Final checklist

### DIFF
--- a/modules/Emails/include/displayEmailAddressOptInField.php
+++ b/modules/Emails/include/displayEmailAddressOptInField.php
@@ -59,8 +59,7 @@ function displayEmailAddressOptInField(Email $focus, $field, $value, $view)
 
     if (empty($focus->id)) {
         $log = LoggerManager::getLogger();
-        $log->error('Email ID is Empty');
-        SugarApplication::appendErrorMessage($app_strings['ERR_EMPTY_EMAIL_ID']);
+        $log->warn('Email ID is Empty');
         return '';
     }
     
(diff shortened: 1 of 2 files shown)


modules/Emails/include/displayEmailAddressOptInField.php
```diff
@@ -59,8 +59,7 @@ function displayEmailAddressOptInField(Email $focus, $field, $value, $view)
 
     if (empty($focus->id)) {
         $log = LoggerManager::getLogger();
-        $log->error('Email ID is Empty');
-        SugarApplication::appendErrorMessage($app_strings['ERR_EMPTY_EMAIL_ID']);
+        $log->warn('Email ID is Empty');
```
horus68: So no translation for the alert?

daniel-samson: The $log will be in the suitecrm.log. The SugarApplication::appendErrorMessage is the message the user will see.

horus68: but the ERR_EMPTY_EMAIL_ID was also removed from include/language/en_us.lang.php

daniel-samson: That is correct. The error typically will show in the emails module. This is due the emails being stored on the IMAP server and not in the CRM. The reason for removing it, is because the indicator will never be able to determine the opt int status of an email which resides on the IMAP server.

However we still want developers to be able to see it, for debugging purposes.

Issue #2307, opened 2016-09-23
Active module won't display further than nth list options down


jshwhitlow:
#### Issue (For my instance)

The Active module is displayed as the last module in each group (unless that module is already in the group or its the ALL group) However, if the group contains more than 6 modules it causes an issue because the active module will always display as the 7th module in the list, no matter how many modules are there.

In 7.7.4 of SuiteCRM, the module list was 10 modules down before the problem occurred.
#### Expected Behavior

The active module should always appear at the end of the list, no matter how many modules are present.
#### Actual Behavior

The active module will not appear further down than the nth option
#### Possible Fix

 (every theme, not just SuiteR)
The fix has to do with modules/themes/SuiteR/tpls/_headermodulelist.tpl on lines 96-170 (could also be between 96-150 as well) Displayed here is lines 96-170
I added the `<script>` tags in this example so that you can refresh the page and see where its stopping (make sure you add more than 6 modules into one group). I'm not too knowledgeable with Smarty, but It appears that `{$modules.modules}` is stopping at the 7th spot and is obtaining that information from the `$groupTabs {foreach}` item.

```
{if $USE_GROUP_TABS}
    <ul class="nav navbar-nav">
        {assign var="groupSelected" value=false}
        {foreach from=$moduleTopMenu item=module key=name name=moduleList}
            {if $name == $MODULE_TAB}
                <li class="topnav">
                    {if $name != 'Home'}
                        <span class="currentTabLeft">&nbsp;</span>
                        <span class="currentTab" style="color:#ffffff !important;">{sugar_link id="moduleTab_$name" module=$name data=$module}</span><span>&nbsp;</span>
                    {/if}
                    <ul class="dropdown-menu" role="menu">
                        {if count($shortcutTopMenu.$name) > 0}
                            <h3 class="home_h3">{$APP.LBL_LINK_ACTIONS}</h3>
                            {foreach from=$shortcutTopMenu.$name item=item}
                                {if $item.URL == "-"}
                                    <li><a></a><span>&nbsp;</span></li>
                                {else}
                                    <li><a href="{$item.URL}">{$item.LABEL}</a></li>
                                {/if}
                            {/foreach}
                        {/if}
                        <h3 class="recent_h3">{$APP.LBL_LAST_VIEWED}</h3>
                        {foreach from=$recentRecords item=item name=lastViewed}
                            {if $item.module_name == $name}
                                <div class="recently_viewed_link_container">
[...]

PR #5035, opened 2018-01-26
Fix Issue #2307

Jason-Dang:
## Description

Ensure the current active module link is displayed at the bottom of any tab group list even when the module list is greater than the max tabs limit.

As the module list is split into two arrays one for the main module list and one for an extra list which gets pushed to the groupTabs array after the main list in which the active module is contained. Added a condition to check if the list is greater than the max tab limit and if so also push the active module to the extra list also.

To ensure the shortcut tab group is still built correctly for the active module, rather than unsetting the current active tab at this point or simply not pushing it into the main list, it is unset after the array used for the shortcut tab functionality is built by looping through the groupTabs array and checking that the tablist is greater than the tab limit, not in the All tab group and present in both the main and extra list arrays.

## Motivation and Context

The max tab limit by default is 10 and this gets set in users' preferences if they do not already have a limit set or if the limit they have set is less than or equal to 0 or greater than 10.

When a group tab list contains more tabs than the max tab limit set by either user preferences and/or the default value in config.php the current active module is not displayed at the bottom of the tab list. Any additional modules after the limit is reached get appended to the bottom of the list as referenced is issue #2307. 

## How To Test This
 
1. Navigate to Admin Console and Configure Module Menu Filters.
2. If the max tab limit is 10 as is default, fill up a menu filter list with a number of modules greater than 10 such as the Support filter. 
3. Navigate to a module that is not contained in the list.
4. The current module should now display at the bottom of the support list.

## Types of changes

### Final checklist

### DIFF
--- a/include/MVC/View/SugarView.php
+++ b/include/MVC/View/SugarView.php
@@ -631,6 +631,9 @@ class SugarView
                     }
                     if (!empty($moduleTab)) {
                         $topTabs[$moduleTab] = $app_list_strings['moduleList'][$moduleTab];
+                        if (count($topTabs) >= $max_tabs - 1) {
+                            $extraTabs[$moduleTab] = $app_list_strings['moduleList'][$moduleTab];
+                        }
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


Dillon-Brown:
Assessed

PR #5077, opened 2018-02-01
Fix #5076 Smtp full log

### LINKED ISSUE #5076: Improve SMTP logging and debugging options
As SMTP gets more complex with increasing anti-spam restrictions, we need to get better error messages from PHPMailer, and shows full logs to the admins in case of error.

See attached PR #5077 

* SuiteCRM Version used: 7.9

pgorod:
## Description
Gets the correct, more informative, exception text from PHPMailer when sending email gives an error.
Let's the admin user see the full SMTP log in case of error.

## Motivation and Context
- SMTP is getting trickier with time, people are facing more restrictions from ISP's, like required certificates, or "From" address checking (must match account).
- our Email problems are compounded when it is not clear whose fault the error is: the SMTP server, the configuration, or some UI or database problem in SuiteCRM. This makes things much more transparent.

## How To Test This

Note: to send unsuccessful emails you can configure a wrong host name, or a wrong password.

1. Send **unsuccessful** emails from all these places, and expect a new message box with detailed error:

    a. Profile->down to the bottom->Settings->Mail Accounts->Mail Accounts (the top half of the screen)->ADD(or EDIT)

    b. Profile->down to the bottom->Settings->Mail Accounts->Outgoing SMTP (lower half, list box)->ADD (or in the list view, EDIT icon). Watch out, this part of SuiteCRM seems to be buggy, you might have trouble sending.

    c. Admin / Outbound email accounts, add

2. Send **successful** emails from all the above, the usual simple old message should appear.

3. Send **unsuccessful** emails from all other places, where my code doesn't do anything, so the expected result is that it doesn't break anything (i.e. just check there are no regressions):

    c. Sending from Compose window.

    d. Sending from Campaigns.

## Authentication data obfuscation:
When a full log is displayed, it should not contain the base64-encoded usernames and passwords of email accounts.

This is what the logs will look like:

    CLIENT -> SERVER: AUTH LOGIN
    SMTP -> get_lines(): $data is ""
    SMTP -> get_lines(): $str is "334 VXNlcm5hbWU6"
    SERVER -> CLIENT: 334 VXNlcm5hbWU6
    CLIENT -> SERVER: bWVAc2l0ZS5jb20=  <---------------
    SMTP -> get_lines(): $data is ""
    SMTP -> get_lines(): $str is "334 UGFzc3dvcmQ6"
    SERVER -> CLIENT: 334 UGFzc3dvcmQ6
    CLIENT -> SERVER: MTIzNDU2     <-----------------------
    SMTP -> get_lines(): $data is ""
    SMTP -> get_lines(): $str is "235 2.7.0 Accepted"

So, seeing `VXNlcm5hbWU6"` and `UGFzc3dvcmQ6`, it just means `User:` and `Password:`, but seeing other strings in place of those marked here with arrows should not happen, they should show as **---obfuscated---**.

## Technical comments

1. There is some repetition of code, but the underlying architecture is like that, and I didn't have time for any serious redesign.
2. The commented out lines I left in the code are on purpose, they offer some extra debugging possibilities which people can use when troubleshooting.

## Types of changes

### Final checklist


modules/OutboundEmailAccounts/OutboundEmailAccounts.php
```diff
@@ -172,15 +182,31 @@ function sendTestEmail()
...
 						success	: function(o) {
 							hideOverlay();
 							var responseObject = YAHOO.lang.JSON.parse(o.responseText);
 							if (responseObject.status)
 								overlay("{$APP['LBL_EMAIL_TEST_OUTBOUND_SETTINGS']}", "{$APP['LBL_EMAIL_TEST_NOTIFICATION_SENT']}", 'alert');
-							else
-								overlay("Send Test Email", responseObject.errorMessage, 'alert');
+       							else {
+                                
+                                                           var dialogBody =
```
gymad: note only: we should refactoring this php file directly contains html, js etc. In future we should outsource these parts to separated template and javascript files. Also as I noticed this piece of code duplicated ish to: https://github.com/salesagility/SuiteCRM/pull/5077/files#diff-40025349886b0dbc1c7c5e25389783efR350

include/SugarPHPMailer.php
```diff
@@ -62,13 +62,13 @@ class SugarPHPMailer extends PHPMailer
     public $disclosureText;
     public $isHostEmpty = false;
     public $opensslOpened = true;
-
+    public $fullSmtpLog='';
     /**
      * Sole constructor
      */
-    public function __construct()
+    public function __construct($exceptions = null)
```
gymad: please add a phpdoc `@param` to this function

include/SugarPHPMailer.php
```diff
@@ -407,4 +407,67 @@ protected function embeddedAttachmentExists($filename)
...
+            $ret = parent::send();
+            $this->exceptions =  $saveExceptionsState;
+        }
+        catch (Exception $e) {
+            $phpMailerExceptionMsg=$e->errorMessage(); //Pretty error messages from PHPMailer
+            if ($phpMailerExceptionMsg) {
+                $GLOBALS['log']->error("send: PHPMailer Exception: { $phpMailerExceptionMsg }");
+            }
+        }
+        catch (\Exception $e) { //The leading slash means the Global PHP Exception class will be caught
```
gymad: class `SugarPHPMailer` has not any namespace so in this context the `Exception` and `\Exception` means same thing, you don't have to handle it separately.

include/SugarPHPMailer.php
```diff
@@ -407,4 +407,67 @@ protected function embeddedAttachmentExists($filename)
...
+            $phpMailerExceptionMsg=$e->getMessage(); //generic error messages from anything else
+            if ($phpMailerExceptionMsg) {
+                $GLOBALS['log']->error("send: PHPMailer Exception: { $phpMailerExceptionMsg }");
+            }
+        }
+
+        /* uncomment this to send full log into suitecrm.log:
+        // TODO: make this dependent on a flag set in config_override.php
```
gymad: Please do not leave TODOs in code, implement it if needed or remove the commented section.

include/SugarPHPMailer.php
pgorod: Hhmm....  I got this code directly from PHPMailer example here: https://github.com/PHPMailer/PHPMailer/blob/master/examples/exceptions.phps

pgorod: Maybe it's for the next version of PHPMailer, they introduced a namespace if I'm not mistaken. I think we should be updating PHPMailer soon, so we could leave this here if it doesn't hurt. What do you think?

modules/OutboundEmailAccounts/OutboundEmailAccounts.php
pgorod: Yeah, I mentioned this in my PR notes: the previous architecture and code are not helping me, and I don't have time to redesign so I just went along with it. I agree this should be improved in the future to avoid repetition, and to better separate HTML from Javascript, etc.

include/SugarPHPMailer.php
gymad: I think, in this example they using a class alias `use PHPMailer\PHPMailer\Exception;` for `Exception` here: https://github.com/PHPMailer/PHPMailer/blob/master/examples/exceptions.phps#L8  consequently there the `Exception` are same as `PHPMailer\PHPMailer\Exception` and it is why they have to mark the original base php `\Exception` with root namespace but here in this file `include/SugarPHPMailer.php` doesn't contains anything like this, so in this context the `Exception` and `\Exception` are same. If you want to using the PHPMailer specified Exception separately as it is in the example, you should `use` it in SugarPHPMailer class. 

Because confusing name similarity I prefer something like this here: `use PHPMailer\PHPMailer\Exception as PHPMailerException;` and after you can handle just like in they example: 
```
} catch (PHPMailerException $e) {
    echo $e->errorMessage(); //Pretty error messages from PHPMailer
} catch (Exception $e) { //The leading slash doesn't need anymore
    echo $e->getMessage(); //Boring error messages from anything else!
}
```
if you want to learn more about namespace importing and aliasing, please see:
http://php.net/manual/en/language.namespaces.importing.php

modules/OutboundEmailAccounts/OutboundEmailAccounts.php
gymad: Yeah I agree, and it was my mention here also as I started with "note only" :)

include/SugarPHPMailer.php
gymad: And last think: you right, if we are using an older PHPMailer version, unfortunately you should wait with this change unit we upgrade the PHPMailer, because this PR doesn't compatible with current SuiteCRM code.

PedroErnst: In this case the class being thrown is 'phpmailerException', and it is being caught by the first catch block, making the second one obsolete.

Dillon-Brown:
Assessed

pgorod:
@Dillon-Brown I don't understand why, but this seems not to be included in 7.10?... 

This was merged into `hotfix`, and I see 2 days ago you also merged `hotfix` into `develop`, so it should be there, but looking at the files I don't see these changes. Any idea why?

It would be important to get this in 7.10.1, I'm already seeing forum threads with Email issues.

PR #5080, opened 2018-02-02
fix for feature/gdpr opt in status

gymad:
## Description

fixes for GDPR branch
- remove '?' char from web form
- rebuild/fixing (confirm)opt in tick after web to person capture
- fixing first/last name in confirm optin email tpl
- action menu opt in email item dynamically toggle when change settings
- web form using selected campaign if exists
- rebuild/fixing opt in tick on edit view
- fixing broken html struct in bulk action / email menu item (it's probably also fix many email field issue)

## Motivation and Context

GDPR buggz

## How To Test This

please, see #5011

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [x] My code follows the code style of this project found [here](https://suitecrm.com/wiki/index.php/Coding_Standards).
- [ ] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://suitecrm.com/wiki/index.php/Contributing_to_SuiteCRM) guidelines.

PR #5087, opened 2018-02-05
Feature/gdpr opt in status

daniel-samson:
## Motivation and Context

These changes correct some minor issues with #5080.

## Types of changes

### Final checklist

### DIFF
--- a/include/language/en_us.lang.php
+++ b/include/language/en_us.lang.php
@@ -3927,7 +3927,7 @@ $app_strings['LBL_OPT_IN_OPT_OUT'] = 'Opted Out';
 $app_strings['LBL_OPT_IN_INVALID'] = 'Invalid';
 
 $app_list_strings['email_settings_opt_in_dom'] = array(
-    '' => 'Disabled',
+    SugarEmailAddress::COI_STAT_DISABLED => 'Disabled',
     SugarEmailAddress::COI_STAT_OPT_IN => 'Opt In',
     SugarEmailAddress::COI_STAT_CONFIRMED_OPT_IN => 'Confirmed Opt In'
 );
--- a/include/language/en_us.lang.php
+++ b/include/language/en_us.lang.php
@@ -42,7 +42,7 @@ if (!defined('sugarEntry') || !sugarEntry) {
     die('Not A Valid Entry Point');
 }
 
-include_once __DIR__ . '/../SugarPHPMailer.php';
++include_once __DIR__ . '/../SugarEmailAddress/SugarEmailAddress.php';
 
 //the left value is the key stored in the db and the right value is ie display value
 //to translate, only modify the right value in each key/value pair
--- a/include/language/en_us.lang.php
+++ b/include/language/en_us.lang.php
@@ -38,12 +38,6 @@
  * display the words "Powered by SugarCRM" and "Supercharged by SuiteCRM".
  */
 
-if (!defined('sugarEntry') || !sugarEntry) {
-    die('Not A Valid Entry Point');
-}
-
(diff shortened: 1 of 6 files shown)


include/language/en_us.lang.php
```diff
@@ -42,7 +42,7 @@
     die('Not A Valid Entry Point');
 }
 
-include_once __DIR__ . '/../SugarPHPMailer.php';
++include_once __DIR__ . '/../SugarEmailAddress/SugarEmailAddress.php';
```
horus68: Will this be backward compatible (the 7.10 language file can be used by oldest SuiteCRM versions?)

Also I consider that including code files in a language files is looking for problems ahead.
If there is a parsing error in some code file it can cause the language file not to load and break the application.
There should be another way! 
I think the language files should be cleaner, slim and neutral on code to reduce errors.

daniel-samson: I agree that the language files should be as clean as possible

Both SugarPHPMailer and SugarEmailAddress have existed since SuiteCRM 7.1. The number of references to SugarEmailAddress and SugarPHPMailer in the languages files, will determine if this is upgrade safe. As far as I can tell, there are 0 references to SugarPHPMailer. So I believe this is correct. It should not cause any issues.

samus-aran: @horus68  - @daniel-samson is looking at this currently and reverting the include.

daniel-samson: I will remove the include completely and then remove the references to SugarEmailAddress. This should keep it backwards compatible.

However, there is a bigger discussion to be had, regarding how long we support language packs, and languages packs should contain code inside them.

horus68: My say on this:
- Language pack support: when backward compatibility is lost we will have a locked language pack for old versions (a zip repo) and maintain/update only the new version. Crowdin can have branch (versions) for language files but I will not use that because then we will lost the auto-package feature from Crowdin (the actual option).

- Languages packs containing code inside them: For me is a big NO. Language files are translators area. Code needs to be packed with core code files. Translations should not be a potential error source, in form of including code files or placing code inside some strings.
Actually there are some code inside language strings and files like the include/ language file causes many errors for parsing code and making the interface not usable for some languages.
Its also too difficult to find errors in language files on languages with different alphabets. 
So less code, less time consuming finding errors errors.

horus68: Forum topic for Language pack support
https://suitecrm.com/suitecrm/forum/international-language-support/17639-language-packs-backward-compatibility-discussion

daniel-samson: OK. Well, in order to remove the code from the languages files. We will need to move the SugarThemeRegistry stuff out into php, smarty templates or JavaScript files. 

What about the labels which contain SQL and HTML?

horus68: If HTML and SQL will be out of languages files even better.

likhobory: I agree with @horus68  about excluding code from language files, but HTML is sometimes used in Russian langpack to highlight text in bold or provide a link to some resources, such as a user's guide for russian users, etc..
 I agree that this is not entirely correct, but sometimes it is convenient to pay user's attention to certain nuances, and in different language packs these nuances may vary. It would be nice if the translator had more tools for a quality translation.

daniel-samson: It is worth mentioning, that we do not support all versions of SuiteCRM. Each version has an end of life. Should the language packs be different?

How far back should the language packs be supported?

likhobory: I think it will be better to work with a single universal package that will work in all supported versions  
(after (1) removing the code from the language files  and  (2) after  patching the corresponding supported SuiteCRM versions) .

Issue #5097, opened 2018-02-06
Number formating of currency fields missing in inline edit


ajedvaj:
#### Issue

_usdollar fields aren't handled properly on inline edit action and they get multiplied by a thousand

Lets say user has following preferences set:

- 1000's separator = dot (.)
- Decimal separator = comma (,)

![User preferenced](https://user-images.githubusercontent.com/18736912/35849936-82aea7e6-0b23-11e8-96e0-b98c46a08099.png)

You have created a quote record in standard way via the create form and  for example these are values in DB: total_amt = 1000 ; total_amt_usdollar = 1000

If you make change to eg. Quote Stage by inline editing (from 'Draft' to 'Negotiation') the total_amt_usdollar will now have value of 1000000000 instead of 1000

![DB screenshot](https://user-images.githubusercontent.com/18736912/35849980-aa3cdf4e-0b23-11e8-8735-9ce1a0bf7529.png)

#### Expected Behavior

The value of _usdollar fields shouldn't change, inline edit action should handle formatting

#### Actual Behavior

The value of _usdollar fields get multiplied

#### Steps to Reproduce

1. Set you user preferences: 1000's separator = dot (.), decimal separator = comma (,)
2. Create quote with some line items 
3. Make change via inline edit
4. The value of _usdollar fields is now scrambled

#### Your Environment

* SuiteCRM Version used: 7.8.12
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): FF 58
* Environment name and version (e.g. MySQL, PHP 7): Mysql, PHP 5.6, PHP 7
* Operating System and version (e.g Ubuntu 16.04): Mint 17, Ubuntu 16.04

PedroErnst:
@ajedvaj 

Thanks for bringing up this issue.

It should be fixed with this.

ajedvaj:
@PedroErnst  
Thanks for fast response! 

Yeah, this looks good. It should also handle situations e.g.  in custom code if you load Bean from db, change somenthing and just call save() which is a good thing! 

I always wondered why SugarBean when it's loaded from from DB (BeanFactory:getBean($module, $id)) doesn't return formatted values?

PR #5098, opened 2018-02-06
Fix/5097 quotes dollar formatting

PedroErnst:
When saving an AOS Quote it would try to convert the currency values into dollars, and if the thousand separator was set to "." then the database value 1.000000 would converted to 1000000.

## Description

Added an extra check, if the record we are working with is fetched from the DB and the value we are trying to convert is the same as the DB row, we do not need to take into account the user format.

## Motivation and Context

Issue #5097 

## How To Test This

1. Set you user preferences: 1000's separator = dot (.), decimal separator = comma (,)
2. Create quote with some line items
3. Make change via inline edit (for example Quote Stage from 'Draft' to 'Negotiation')
4. The value of _usdollar fields is not scrambled in the database

## Types of changes

### Final checklist


Dillon-Brown:
Assessed

PR #5099, opened 2018-02-06
Fix #5095 Pass parameter that gets modified by reference

### LINKED ISSUE #5095: Database failure using custom relate field in where clause
Exporting CSV (but also in AOR Reports) when using a where condition based on a  custom relate field causes a database failure with Unknown column error.

#### Issue
The where clause is incorrect and doesn't refer to the correct column in the joined table.

#### Expected Behavior
The query should work.

#### Actual Behavior

In suitecrm.log  MySQL error 1054: Unknown column <module with relate filed>.<field from related table>

#### Possible Fix
Function getJOIN in modules/DynamicFields/DynamicField.php has code to modify the where clause according to the generated join aliases. See line 350.

The function has the signature
`public function getJOIN($expandedList = false, $includeRelates = false, $where = false)`

but it should be:
`public function getJOIN($expandedList = false, $includeRelates = false, &$where = false)`

i.e. $where passed in by reference ... so the expected side-effect can take effect. This piece of code is annotated with
`//bug 27654 martin`

so the fix can never have worked despite the code doing the right thing but the result never propagated out of getJOIN.

#### Steps to Reproduce

1. Create a custom relate field on Contacts
2. Set a filter based on the value. Subpanel shows filtered results.
3. Select a few records
4. Click Bulk Action/Export
5. Observe the error in the log

#### Context

High Priority - because it has been around for years, It's simple to fix and can be verified by eyeballing the code.

#### Your Environment

* SuiteCRM Version used: 7.8.6

PedroErnst:
A parameter that was being modified inside a function was being passed by value, and therefore the modification lost.

## Motivation and Context

Issue #5095 

## How To Test This

1. Create a custom relate field on Contacts
2. Set a filter based on the value. Subpanel shows filtered results.
3. Select a few records
4. Click Bulk Action/Export
5. Observe there is no error in the log

## Types of changes

### Final checklist

### DIFF
--- a/modules/DynamicFields/DynamicField.php
+++ b/modules/DynamicFields/DynamicField.php
@@ -312,7 +312,7 @@ class DynamicField
      *
      * @return array select=>select columns, join=>prebuilt join statement
      */
-    public function getJOIN($expandedList = false, $includeRelates = false, $where = false)
+    public function getJOIN($expandedList = false, $includeRelates = false, &$where = false)
     {
         if (!$this->bean->hasCustomFields()) {
             return array(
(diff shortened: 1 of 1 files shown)


Dillon-Brown:
Assessed

PR #5112, opened 2018-02-06
Fix #5012 Remove client side password length limitation

### LINKED ISSUE #5012: Suggestion - Increase input maxlength during setup
#### Issue
When setting up an instance as an Azure App Service with hosted MySQL database the requirements for the connection by MS are longer than the allowed characters of the input fields on the configuration page of the setup wizard.
Microsoft requires the username to match the pattern `[username]@[dbname]-mysqldbserver`
Without adding a username or database name the character length is already at 15.

#### Expected Behavior
No (or more reasonable) character length restriction on input fields

#### Actual Behavior
DB Username field is limited to 16 characters

#### Possible Fix
Increase the fields `maxlength` attribute to a higher value (32/64/100) or remove the restriction entirely. 

#### Steps to Reproduce
1. Perform a clean install
2. At the configuration step select "Provide existing user" from the drop down
3. Attempt to put in more that 16 characters into input fields

#### Context
It was simple to get around using the browser developer tools, but would be nice to not have to.
This is probably a low priority issue.

#### Your Environment
* SuiteCRM Version used: 7.9.9
Azure App Service (Windows Server 2016, MySQL 5.7.21, PHP 7.0)

PedroErnst:
Removes an (arbitrary) limitation on the length of DB usernames during SuiteCRM install.

## Description

According to the MySQL manual "MySQL user names can be up to 32 characters long".
MSSQL seems to have 128 characters or no maximum length at all.

It seems unreasonable to limit the user to inputting only a 16 character username if the underlying systems accept much longer ones.

## Motivation and Context

Issue #5012 

## How To Test This

Install a new instance of SuiteCRM
Select 'Provide existing User' for the DB

## Types of changes

### Final checklist

### DIFF
--- a/install/dbConfig_a.php
+++ b/install/dbConfig_a.php
@@ -198,7 +198,7 @@ if($db->supports("create_user")){
     <span id='connection_user_div' style="display:none">
         <span class="required">*</span>
             <label><b>{$mod_strings['LBL_DBCONF_SUGAR_DB_USER']}</b></label>
-            <input type="text" name="setup_db_sugarsales_user" maxlength="16" value="{$_SESSION['setup_db_sugarsales_user']}" />
+            <input type="text" name="setup_db_sugarsales_user" value="{$_SESSION['setup_db_sugarsales_user']}" />
             <label><b>{$mod_strings['LBL_DBCONF_DB_PASSWORD']}</b></label>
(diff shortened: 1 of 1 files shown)


Dillon-Brown:
Assessed

pgorod:
I believe this fix might be incomplete, and the same thing would need to be done here:

https://github.com/salesagility/SuiteCRM/blob/master/install/installConfig.php#L392

Issue #4351, opened 2017-10-01
Notes related to given contact don't appear on his history subpanel


cumanacr:
#### Issue

Notes created using the __Related To:__ field with Contacts module don't appear on the given contact history subpanel

#### Expected Behavior

When you see the __note__ detailview the contact appears and when you check the contact history subpanel note is there too

#### Actual Behavior

When you see the __note__ detailview the contact appears but when you check the contact history subpanel note doesn't show

#### Possible Fix

I think the problem is that history subpanel __is filtering out__ the notes when the related field has __parent module set to Contacts__ because it's assuming that notes already have a `contact_id` field that should be use to link notes and contacts.

imho there's three possible way to fix this:
1. Repair contacts subpanel to show notes when related to module is contacts.
2. Change the notes controller to copy `parent_id` to `contact_id` when `parent_type` is __Contacts__
3. Remove __Contact__ option from the __Related to:__ dropdown so users must use __Contact:__ field

#### Steps to Reproduce

I'm going to create an example with 3 notes and attached some images of it

1. Create a note with a contact and related to Accounts module, Note: __Bad Picture :(__
2. Create a note with a contact and without any relation, Note: __ok picture ¯\_(ツ)_/¯__
3. Create a note without a contact and related to Contacts module, Note: __Great Picture!__
4. You'll be able to see the __Bad Picture :(__ and __ok picture ¯\_(ツ)_/¯__ but not the __Great Picture!__ on the contact history subpanel and if you ask me, that is pretty sad

![selection-03-2017-09-30](https://user-images.githubusercontent.com/6874136/31051796-2cf153f2-a62f-11e7-8839-fe72896b928b.png)

![selection-05-2017-09-30](https://user-images.githubusercontent.com/6874136/31051795-2ce6caf4-a62f-11e7-8830-7c4ca2f4c962.png)

#### Context

I got a report of some missing notes from a client that use them to track several files on their contacts.

To fix it, I updated the notes table copying `parent_id` to `contact_id` when `parent_type` is set to __Contacts__

That took care of the problem in the meanwhile.

#### Your Environment

- SuiteCRM Version used: 7.9.5
- Browser name and version: Google Chrome Version 61.0.3163.100 (Official Build) (64-bit)
- Environment name and version:
  - mysql  Ver 14.14 Distrib 5.7.19, for Linux (x86_64) using  EditLine wrapper
  - PHP Version 5.6.31-6+ubuntu16.04.1+deb.sury.org+1
[...]

PR #5146, opened 2018-02-08
Fix/4351 notes contacts relation

PedroErnst:
Improves handling of Notes which are related to Contacts module and have a different Contact or no Contact set as 'Contact' field.

## Description

Currently the relationship is a 1-M, so there shouldn't be Notes which are related to Contacts, and the relationship is defined on the contact_id field, so to be consistent whenever a Note is related to Contacts, it should also have the related Contact's id in the contact_id field, otherwise they don't count as really being related.

screenshot from 2018-02-08 16-47-09

## Motivation and Context

Issue #4351

## How To Test This

Create a Note which is related to Contacts and fill our related (right side) field but leave Contact (left side) field blank.

## Types of changes

### Final checklist

### DIFF
--- a/modules/Notes/controller.php
+++ b/modules/Notes/controller.php
@@ -61,8 +61,8 @@
 
         // If Note is related to a Contact, we automatically fill the Contact field
         if ($_REQUEST['parent_type'] === 'Contacts'
-						&& !empty($_REQUEST['parent_id'])
-						&& !empty($_REQUEST['parent_name'])) {
+            && !empty($_REQUEST['parent_id'])
+            && !empty($_REQUEST['parent_name'])) {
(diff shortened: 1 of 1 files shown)


modules/Notes/controller.php
```diff
@@ -58,6 +58,12 @@ function action_save(){
 		{
 			$_REQUEST['relate_id'] = false;
 		}
+
+        // If Note is related to a Contact, we automatically fill the Contact field
+        if ($_REQUEST['parent_type'] === 'Contacts') {
```
zoltankocsardi: small remark: you might check parent_id and name if they are not empty

Dillon-Brown:
Assessed

PR #5202, opened 2018-02-14
Updating warning message on the upgrade wizard

daniel-samson:
## Description

This change updates the warning message in the upgrade wizard, reminding users that they will need to run "composer update" if they have customised the composer.json file.

## Types of changes

### Final checklist


horus68:
Can this be done in an additional string just for the extra info and show both in UI?
When updating a translation string value into Crowdin, it will require to delete previous translation.
This is OK when changing the value, but when just adding details into long lines it would be better to keep to old translation intact and adding a new string to translate (with the second sentence) and show both next to each other.

modules/UpgradeWizard/language/en_us.lang.php
```diff
@@ -177,7 +177,7 @@
     'LBL_UW_SKIPPED_FILES_TITLE' => 'Skipped Files',
     'LBL_UW_SQL_RUN' => 'Check when SQL has been manually run',
     'LBL_UW_START_DESC' => 'This wizard will assist you in upgrading this SuiteCRM instance.',
-    'LBL_UW_START_DESC2' => 'Note: We recommend that you backup the SuiteCRM database and the system files (all of the files in the SuiteCRM folder) before upgrading your production system. We highly recommend that you perform a test upgrade on a cloned instance of your production system first.',
+    'LBL_UW_START_DESC2' => 'Note: We recommend that you backup the SuiteCRM database and the system files (all of the files in the SuiteCRM folder) before upgrading your production system. We highly recommend that you perform a test upgrade on a cloned instance of your production system first. You may also need to run "composer update" if you have customised the composer.json file',
```
samus-aran: I think this will scare non tech admin users :smile: Is this the best place to place it or even word it. This message will be BEFORE we have copied anythign over so it really should state that you run the command AFTER the upgrade script yeah?

'We highly recommend that you perform a test upgrade on a cloned instance of your production system first and run a composer update after upgrading if you have a customised composer.json file'

samus-aran: @horus68 I get what you are saying, but doesn't that increase the language array?
Whats the impact of changing the description? It does keep things together (though doesn't justify the START_DESC though, something that may require adjusting) especially when its regarding to providing the user with more information.

horus68: @samus-aran you can apply the change to the value here. I will just remove the approved status on Crowdin so we don't loose translations already done for that string. Translators will be required to verify the translation later.

modules/UpgradeWizard/language/en_us.lang.php
```diff
@@ -177,7 +177,7 @@
     'LBL_UW_SKIPPED_FILES_TITLE' => 'Skipped Files',
     'LBL_UW_SQL_RUN' => 'Check when SQL has been manually run',
     'LBL_UW_START_DESC' => 'This wizard will assist you in upgrading this SuiteCRM instance.',
-    'LBL_UW_START_DESC2' => 'Note: We recommend that you backup the SuiteCRM database and the system files (all of the files in the SuiteCRM folder) before upgrading your production system. We highly recommend that you perform a test upgrade on a cloned instance of your production system first. You may also need to run "composer update" if you have customised the composer.json file',
+    'LBL_UW_START_DESC2' => 'Note: We highly recommend that you create a copy of the SuiteCRM instance you use in production, and test the upgrade package before deploying the new version to production. After the upgrade process has completed, please run <pre>composer update</pre> to ensure that you all the latest.',
```
horus68: the end of the sentence needs a revision: see "to ensure that you all the latest."

Dillon-Brown:
Assessed

PR #5282, opened 2018-02-20
Stop required symbol from wrapping on edit form

isleshocky77:
## Description

Currently because of the smarty templating the label text has text around it which is is playing havoc with the CSS which states wrapping behavior.

This could be fixed by changing the css or by cleaning up the outputted html. I chose to go with the latter route.

P.s. as a note you can see this behavior even in [one of the screenshots](https://suitecrm.com/wiki/index.php/File:Release_notes4.png) of [7.10 Release Notes](https://suitecrm.com/wiki/index.php/Release_notes_7.10.0).

### Before

before

### After
after

## Motivation and Context

The edit form looks bad with the symbol wrapping around and into the label

## How To Test This

1. Checkout branch
2. Clear template cache

## Types of changes

### Final checklist

### DIFF
--- a/include/Smarty/plugins/block.minify.php
+++ b/include/Smarty/plugins/block.minify.php
@@ -1,0 +1,58 @@
+<?php
+function smarty_block_minify($params, $content, &$smarty, &$repeat) {
+    if (!$repeat) {
+        if (isSet($content)) {
+            // HTML Minifier
+            $input = $content;
+            if(trim($input) === "") return $input;
+            // Remove extra white-space(s) between HTML attribute(s)
+            $input = preg_replace_callback('#<([^\/\s<>!]+)(?:\s+([^<>]*?)\s*|\s*)(\/?)>#s', function($matches) {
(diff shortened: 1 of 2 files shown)


Dillon-Brown:
Issue reference: #5412

Dillon-Brown:
Assessed

PedroErnst:
Tested, works as expected

LEAP-nishit:
It would be good is this makes it into the next release.

samus-aran:
Hey @isleshocky77 We've discovered a regression issue from this fix that stops users from creating a target list due to unique case of custom buttons. 
https://github.com/salesagility/SuiteCRM/issues/5702
The[ fix provided here](https://github.com/salesagility/SuiteCRM/pull/5688/files) would resolve that particular scenario but we will now need to investigate the impact that this has caused. As an active contributor I'm addressing this here so we can all learn from this experience and note when raising new PRs consider the possible implications of it. We'll keep this thread updated with our findings :+1:

PR #5323, opened 2018-02-23
Remove word "error" from a debug message

pgorod:
Often when we're `grep`ing a log for errors this comes up. However, I don't think it's really a cause of concern. I am also basing this judgment on the fact that whoever put this line into the code made it a simple `DEBUG` level message.

So this PR is just to remove the word `error` from that message, so that people don't waste time looking for their problems in the wrong place. This has been happening a lot with email troubleshooting, people ask about this.

If this could please go into 7.8, 7.9, and 7.10, I would appreciate it.

### DIFF
--- a/data/SugarBean.php
+++ b/data/SugarBean.php
@@ -1571,7 +1571,7 @@ class SugarBean
                 return true;
             }
         }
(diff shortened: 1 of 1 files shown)


Dillon-Brown:
Assessed

PR #4996, opened 2018-01-22
Creating Case from A contact record's subpanel doesn't send Automatic Email

ijdavie:
## Description

https://suitecrm.com/forum/suitecrm-7-0-discussion/6136-case-creation-auto-email
Creating a case and linking a Contact on the Case Record sends an Email.
Creating a case from the Contact's Record itself does not.

relates to issue #406
## Motivation and Context

required to allow sending of email when case is created in the cases subpanel on a contact. 

moved when the after relationship save hook so it will now fire after the insert of the record into the database and then refreshes the bean itself so any changes e.g case_number is brought back to the bean for the hooks to have access. 

## How To Test This

create a case via the cases subpanel in the contacts record. 
## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [x] My code follows the code style of this project found [here](https://suitecrm.com/wiki/index.php/Coding_Standards).
- [ ] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://suitecrm.com/wiki/index.php/Contributing_to_SuiteCRM) guidelines.

PR #5335, opened 2018-02-26
Fix #5315 Don't cache beans with auto increment fields before saving

### LINKED ISSUE #5315: No CASE number when receiving inbound emails for new Cases
#### Issue
When trying to retrieve the case record a new bean is created instead of retrieving that one that already exists which removes the case number which is assigned in the db I believe. The issue it caused me was not being able to reply to case created emails to update a case.  

#### Expected Behavior
When sending case created emails the outbound email is sent with case number eg: CaseSubject [CASE:7835]

#### Actual Behavior
Email is sent but has no case number eg: CaseSubject [CASE: ]

#### Possible Fix

At line 2800 in modules/InboundEmail/InboundEmail.php & line 106 in modules/InboundEmail/AOPInboundEmail.php

            $c->save(true);
            $caseId = $c->id;
            $c = new aCase();
            $c->retrieve($caseId);
            if($c->load_relationship('emails')) {
                $c->emails->add($email->id);
            }
to 
            $c->save(true);
            $c->retrieve($c->id);
            if($c->load_relationship('emails')) {
                $c->emails->add($email->id);
            }

#### Steps to Reproduce

1. Set up email template that uses the case number
2. Set up to send email on case creation
3. Make case
4. Check history subpanel for emails sent

#### Context

 low/medium priority

#### Your Environment

* SuiteCRM Version used: 7.8.8
* Browser name and version: Chrome Version 64.0.3282.140
* Environment name and version: MYSQL PHP 7

PedroErnst:
Fixes a bug with new Beans that have auto increment fields and the BeanFactory cache.

## Description

BeanFactory holds a copy of 10 last used beans in memory, but does not take into account auto increment fields that change in the database after saving them for the first time.

## Motivation and Context

Potential bugs with any module that uses auto increment fields (Campaigns, Trackers, Prospects, Cases) as well as any auto increment custom fields.

## How To Test This

1. Set up email template that uses the case number
2. Set up a workflow action to send email on case creation
3. Create a new case
4. Check that the email sent contains the correct case number

## Types of changes

### Final checklist

### DIFF
--- a/data/SugarBean.php
+++ b/data/SugarBean.php
@@ -5541,6 +5541,9 @@ class SugarBean
         }
     }
 
+    /**
+     * @return bool
+     */
     protected function hasAutoIncrementFields()
     {
         foreach ($this->field_defs as $key => $field_def){
--- a/data/SugarBean.php
+++ b/data/SugarBean.php
@@ -1894,8 +1894,10 @@ class SugarBean
         }
 
 
-        require_once("data/BeanFactory.php");
-        BeanFactory::registerBean($this->module_name, $this);
+        if (!$this->hasAutoIncrementFields()) {
+            require_once("data/BeanFactory.php");
+            BeanFactory::registerBean($this->module_name, $this);
+        }
 
         if (empty($GLOBALS['updating_relationships']) && empty($GLOBALS['saving_relationships']) && empty($GLOBALS['resavingRelatedBeans'])) {
             $GLOBALS['saving_relationships'] = true;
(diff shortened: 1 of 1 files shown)


mattlorimer:
this is not necessary, it actually caused by the bean been needlessly recreated in At line 2800 in modules/InboundEmail/InboundEmail.php & line 106 in modules/InboundEmail/AOPInboundEmail.php

PedroErnst:
Maybe I got confused.

There is definitely an issue when a bean with an auto increment field is saved.
I'll update the description to match this better, I've followed these steps to reproduce the error (and might have fixed an altogether different issue, because at no point was either of the lines of code mentioned above executed).

1. Set up email template that uses the case number
2. Set up a workflow action to send email on case creation
3. Create a new case
4. Check that the email sent contains the correct case number

mattlorimer:
I see the issue you are fixing, it may be better to merge this with this potential fix https://github.com/salesagility/SuiteCRM/pull/4996/files to globally solve the issues. As currently this would create inconsistentes in the beanfactory - which creates the potential to cause issue in custom overrides to the save.

PR #5355, opened 2018-02-27
Fix is_admin_for_user warning

eggsurplus:
Warning: Use of undefined constant is_admin_for_user

### DIFF
--- a/themes/SuiteP/include/ListView/ListViewGeneric.tpl
+++ b/themes/SuiteP/include/ListView/ListViewGeneric.tpl
@@ -214,7 +214,7 @@
 			<tr height='20' class='{$_rowColor}S1'>
 				{if $prerow}
 				<td>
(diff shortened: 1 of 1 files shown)


samus-aran:
Good Morning Jason,

Saw alot of PRs I'm assuming they are all the same. 
Just a note we don't merge anything directly into master but use hotfix, develop and hotfix-7.8.x or hotfix-7.9.x. If you know the issue exists in all supported versions (7.8.x +) then make the fix into hotfix-7.8.x as when we release we merge upwards hotfix7.8.x into hotfix 7.9.x etc etc

Hope that makes sense.

samus-aran:
Ahh, apologies, they are not all the same.
I would then recommend combining them all together in a single PR. If a commit can be grouped contextually then we can merge it rather than having a single PR for each commit.

Dillon-Brown:
Assessed

Dillon-Brown:
Code review in duplicate #5388.

PR #5372, opened 2018-03-01
Fixed #5174 - Confirmed Opt in tick - Microsoft edge

### LINKED ISSUE #5174: MS Edge: The color of the ticks is allways green , regardless of the Email status (ver.7.10-RC-2)
MS Edge:
default

The same ListView in the FF or IE11:
default

daniel-samson:
## Description

It seems to be a rendering issue in windows instead of a css issue. Edge is correctly selecting the css selector. other browsers such as Firefox have the same issue on windows. The tick does will always be green regardless of the css.

The solution would be to use an icon instead, so that it is not using the unicode value.
## Motivation and Context

- This ensures that the color of the confirmed opt in tick displays the correct colors in web browsers on Microsoft windows.
- Fixed #5174

## How To Test This

- Rebuild sass
- Clear browser cache
- test the color of confirm opt in on Microsoft edge

## Types of changes

### Final checklist


Dillon-Brown:
Assessed

PR #5381, opened 2018-03-02
Feature Anti Malware Scanner Integration

daniel-samson:
## Description

After experimenting with clamav and sophos anti-virus for linux, I came up with a framework to integrate anti-malware software with suitecrm.

Using the PSR-4 autoloader you can configure first-party, custom, and third party packages. SuiteCRM can then autoload those packages and run a command line tool to scan uploads for malware.

av-config
Config.php

The configure can be used to customise the executable path and the option passed into the command line tool.

Developers can create providers which implement the AntiMalwareFileScanner interface and throw a MalwareFound exception. This will make it really easy to support many command line based antivirus programs.

Incidents are recorded in the SuiteCRM.log
av-log

## Motivation and Context

- Scans uploaded content with anti malware software and blocks any issues found
- Adds improved exception handling in the sugar controller, so that better error messages are reported back to the user, if a suitecrm exception is thrown in the GUI.

## How To Test This

- Install [clamav](https://www.clamav.net/) or [sophos](https://www.sophos.com/en-us/products/free-tools/sophos-antivirus-for-linux.aspx) on the web server
- Rebuild the config
- Configure the path location
- Set enabled to true
- Upload a file, create a document/note with the [EICAR](http://www.eicar.org/86-0-Intended-use.html) test file

(Optional)
- Run the automated unit test suite see tests/unit/lib/AntiMalware

## Types of changes

### Final checklist


include/MVC/Controller/SugarController.php
```diff
@@ -411,6 +412,48 @@ protected function handleException(Exception $e)
         }
     }
 
+    /**
+     * Handle SuiteCRM Exceptions
+     * @param \SuiteCRM\Exception\Exception $exception
+     */
+    protected function handleSuiteCRMException(\SuiteCRM\Exception\Exception $exception)
```
PedroErnst: Typehinting is only supported from PHP 7.0 onwards, but SuiteCRM 7.10 should be compatible with 5.5.9 and onwards according to the compatibility matrix.

lib/Utility/AntiMalware/Providers/ClamAV.php
```diff
@@ -0,0 +1,187 @@
...
+        if (!is_string($path)) {
+            throw new InvalidArgumentException(
+                '[ClamAV][scanPathForMalware] $path should be a string'
+            );
+
+        }
+
+        if (!file_exists($this->getExecutablePath())) {
+            throw new InvalidArgumentException(
+                '[ClamAV][scanPathForMalware] $path not found'
```
PedroErnst: $path will not be replaced here, it might be more useful for debugging to have the value of this variable?

Same goes for the similar exception in SophosAntiVirus.php

PedroErnst (changes requested):

include/MVC/Controller/SugarController.php
daniel-samson: Please refer to [php documention](http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration.types). As of php 5.0 you can use classes/interfaces. however primative types like string, int, float are on only for php 7.0 and above. Keep in mind that I have tested this code on PHP 5.5.9 - which is the minimum requirements for SuiteCRM 7.10

PedroErnst: My bad, you are absolutely right

lib/Utility/AntiMalware/Providers/ClamAV.php
daniel-samson: I actually wasn't intending this to be replaced by the value. I guess we could though :+1:

daniel-samson: OK so it looks like I have overlooked something here. So have added an extra check here. Thank you for raising this. :D

PedroErnst (dismissed):

Dillon-Brown:
Assessed

Mac-Rae:

Issue #5433, opened 2018-03-08
admin can't enter an encryption key for LDAP config - even though mcrypt is enabled.


604media:
NOTE: Could not find any similar bug already posted here.

**Expected Behavior**
User should be able to enter an encryption key into LDAP Config in Password Management in order to allow Frontend Implcity Outlook addin to authenticate using LDAP (encryption is required)

**Actual Behavior**
When Admin User attempts to enter an encryption key in the LDAP configuration in Password Management, it is not possible to enter a value - ie. can't click the cursor in the field.

the HELP bubble states "The php_mcrypt extension must be enabled in your php.ini file."

However - we have confirmed that the php_mcrypt is enabled -- another instance running PHP5 has this configured with no problem - but PHP7 seems to be the problem 

**Context**
We are not able to use LDAP authentication with Frontend Implicit Outlook Addin

SuiteCRM Forum reference to this issue 7 months ago --- https://suitecrm.com/suitecrm/forum/installation-upgrade-help/11750-cannot-create-ldap-encryption-key-requires-php-mcrypt-dll

**Environment**
PHP 7.0.25-0
ubuntu 0.16.04.1
SuiteCRM Version 7.9.11

mcrypt support | enabled
Version | 2.5.8
Api No | 20021217

604media:
Please check the below pull request for the fixes
https://github.com/salesagility/SuiteCRM/pull/5442

daniel-samson:
Unfortunately mcrypt_encrypt is also deprecated. We will need to come up more long term solution.

PR #5442, opened 2018-03-08
Fixed 5433 - admin can't enter an encryption key for LDAP config - even though mcrypt is enabled

604media:
Fixed - mcrypt_cbc has been depreciated in php7 so we need to use mcrypt_encrypt instead of mcrypt_cbc


horus68:
Associated issue: https://github.com/salesagility/SuiteCRM/issues/5433

Dillon-Brown:
Hi @604media, this should now be fixed by PR #5479 which uses openssl_encrypt instead of mcrypt_encrypt which is now also deprecated. Thanks.

PR #5488, opened 2018-03-12
fix email addresses in listivew in case of external email client

zoltankocsardi:
## Description
Email addresses are invisible when email client is not Suite

## How To Test This
- go to profile
- change email client to external email client at email settings
- save
- got to contact list view
- check if email addresses are visible
- dance

## Types of changes

### Final checklist

### DIFF
--- a/modules/Users/User.php
+++ b/modules/Users/User.php
@@ -1532,7 +1532,7 @@ EOQ;
             $emailLink = $emailUI->populateComposeViewFields($focus);
         } else {
             // straight mailto:
-            $emailLink = '<a href="mailto:' . $focus->$attribute . '" class="' . $class . '">';
+            $emailLink = sprintf('<a href="mailto:%1$s">%1$s</a>', $focus->$attribute);
         }
 
         return $emailLink;


samus-aran:
:+1: Works for me

modules/Users/User.php
```diff
@@ -1532,7 +1532,7 @@ public function getEmailLink(
             $emailLink = $emailUI->populateComposeViewFields($focus);
         } else {
             // straight mailto:
-            $emailLink = '<a href="mailto:' . $focus->$attribute . '" class="' . $class . '">';
+            $emailLink = sprintf('<a href="mailto:%1$s">%1$s</a>', $focus->$attribute);
```
daniel-samson: You have removed the CSS class attribute. Is this intential. It may effect the confirmed opt in logic.

zoltankocsardi: Yes, it was intentinally, since this method's default class value is an empty string, and there is no method exists which calls this method and set the class parameter. There is no effect for confirmed opt in logic...

daniel-samson: OK. I think we should take a note of this for the future. Perhaps we should remove the parameters? $contact_id, $ret_action, $ret_id and $class have no effect on this method. Or we could update the PHP Doc to state that the params are deprecated.

daniel-samson:
💃 dancing :)

modules/Users/User.php
zoltankocsardi: Ye, we could remove unnecessary params easily. Gonna create a new ticket for refactoring.

PR #5563, opened 2018-03-22
Fix/alert performance

PedroErnst:
Fixes a some bugs affecting Alerts and Reminders performance.

## Description

Fix a small bug causing date_willexecute time not being updated at all, causing every Reminder to be loaded on every page load.
Remove an unnecessary foreach loop and no longer fetches Invitees of a reminder as the data was not being used anywhere.
Only fetch Reminders from the DB that would actually be sent down to the client side.
Prevent JS from running the updateManager function that sends an ajax call more than once per second.

## Motivation and Context

Slow alerts / reminders.

## How To Test This

SuiteCRM should not be slowed down by a large alerts or reminders table.

## Types of changes

### Final checklist


modules/Reminders/Reminder.php
```diff
@@ -110,6 +112,17 @@ private static function saveRemindersData($eventModule, $eventModuleId, $reminde
             $reminderBean->date_willexecute = null;
 
             $reminderBean->save();
+
+            // Delete related alerts
+            $url = Reminder::makeAlertURL(
+                $reminderBean->related_event_module,
+                $reminderBean->related_event_module_id
+            );
+            $db->query("UPDATE `alerts` SET `deleted` = 1 WHERE `url_redirect` = '$url'");
```
zoltankocsardi: Please use $db->quoted for $url.

Issue #5064, opened 2018-01-30
[Suggestion] Collapsed active/with records subpanels - other color of header text or background


Mausino:
#4706

When some collapsed subpanels have records.. will nice if will the color of header or background other as grey... user will know that there are records...  it helps to faster orientations if some user using only collapsed subpanels which are not grouped by menu (also will great for user which have grouped subpanels)

For activity subpanel i am suggesting other color like green.. because you will on first look know that there are active tasks/calls/meetings

![2018-01-30_1145](https://user-images.githubusercontent.com/5773777/35562254-196c8c3c-05b3-11e8-860e-72d9b60e7639.png)

PedroErnst:
Yes, absolutely agree.

I would go another step and actually display the number of records after the name.

ACTIVITIES (0)
ACTIVITIES (5)
ACTIVITIES (25)
ACTIVITIES (100+)

The required COUNT query is not that expensive that it would be noticeable in the load times, and it would save quite a bit of time for the users.

Mausino:
@PedroErnst should you add the link on your pull request for COUNT query ?

PedroErnst:
@Mausino I have a PR for this? I don't think so...

Mausino:
@PedroErnst sorry... i thinked that you created some function abou you wrote 

"The required COUNT query is not that expensive that it would be noticeable in the load times, and it would save quite a bit of time for the users."  

and i can't found it... but i know that is not created yet :)

Mausino:
The main point was fixed in #5577 

Thanks to @PedroErnst

PR #5577, opened 2018-03-23
Feature/collapsed subpanel row count

PedroErnst:
Adds an optional functionality to indicate whether a collapsed subpanel has any content or not.

screenshot from 2018-03-23 10-32-43

## Description

When the user enables this functionality, a quick COUNT query will be run to check if the subpanel would contain any rows.

## Motivation and Context

Suggested here #5064 
Preloading every subpanel is slow, but with collapsed panels the user may waste a lot of time expanding empty subpanels. 
This functionality intends to make the life of the user easier by giving a clear visual hint if a subpanel contains records or not.

screenshot from 2018-03-23 10-32-17

## How To Test This

Go into the user settings, layout options and enable "Show collapsed subpanel hint"
See any record with subpanels

## Types of changes

### Final checklist


Dillon-Brown:
Assessed

Mausino:
@pedroErnst should it show numbers of record instead of plus symbol? It will more usable..

PedroErnst:
Yeah that was what I had indeded to do in the beginning... and it works quite nicely for all 'standard' subpanels, but there are issues with the more complex ones (Activities and History for example). 

It turns out the subpanel will fetch emails related to a Contact for example, and then also emails related to the Account of that Contact, so it could fetch the same email multiple times. The actual subpanel code fetches the duplicates and then filters them out, but the count query simply returns an incorrect number. I don't see a way around this particular issue without fetching the actual rows, which would be slow.

If the subpanel hint shows (17), then you click on it and the subpanel only loads 14 Emails, that's weird. Having a count for some subpanels but not others seemed inconsistent as well.

So in the end the cleanest solution seemed simply indicating if there are records, or not.

Mausino:
@PedroErnst in administration settings you are able to set up how many records should be in subpanel... like 10, 20 or any number... and if you set up 10 for example, you see e.g. in history subpanel if there exist 60 records only number 1 from 10+ records... same system works also on listview

Should be this applied on this subpanels....  if there will for example see Histrory (10+) i will know that there are more as 10 records... 

The main point is that you are able know which subpanels have records and which not.. this you done 👍  and in my case.. not sure as in other user are 90% of subpanels less as number which i setted up to admin settings.... before i setted up very small number because if loaded all subpanels.. it was very slow... you solved this problem that should load only subpanels which are open uncollapsed) based on cokies.... 

My idea doesn't show total number.. nobody needs it.. only use what is working in SuiteCRM... and if are more records as setted up and show 1 from (any number)+ that will great..

PedroErnst:
@Mausino 

I understand what you mean. Still, let's say you set the limit to 10. Now if the subpanels has really 6 records, but the hint shows 8 because of these 'false positives', that would be very confusing to the users. The user would think there is some malfunction, that there are really 8 records but two won't show...

So again, the only real option seems to be not to show any number, just whether there are any record or not.

Mausino:
Hi, i am adding here code which provide us our external company.. maybe can somebody inspire :)



[subpanel.zip](https://github.com/salesagility/SuiteCRM/files/1876316/subpanel.zip)

PedroErnst:
@Mausino 
 
Just had a look at that code, what it does is it fetches the actual records in the subpanel in order to get the count.

This is precisely what I'm trying to avoid, as this solution would have similar performance to having all subpanels expanded. On a good connection and hardware, and with few rows per subpanel, it might not be noticeable at all, but this solution could be a major performance hit for many users.

Thank you anyway for contributing!

Mausino:
@PedroErnst we applied your solution and works perfect

Only one small issue is here... if you expand subpanel and after you collapse it.. the dark color disapper and is there default light grey. On refresh colors of subpanels are fixed

PedroErnst:
Thanks @Mausino for pointing that out. Fixed in my last commit. Needs sass rebuild to see the changes.

Mausino:
@PedroErnst  i checked official demo of SuiteCRM 7.9.5 and there is it broken

Mausino:
@PedroErnst sorry i am overhelmed :(  i thinked 7.10.5

Mausino:
@PedroErnst this morning i saw that official demo is live... it's version 7.10.5 ... gret news 👍 and in the official demo i don't see this feature there ....  there is any (0) or + in rows.... i don't  see changed color of subpanels which contain records....

PedroErnst:
@Mausino 

Yes this PR is not yet merged, so it's NOT included in 7.10.5

As soon as this PR gets merged you can expect this feature to be available in the following release.

Mausino:
sorry my fail.. i don't looks if was this pull merged, because i applied it before few weeks in my instance
but when you will merge it.. this issue will for 100% in campaigns

counting and coloring subpanels also isn't working in ROI of campaigns / videoscreen https://www.screencast.com/t/WnLtMD2h

in detail view of campaign recod subpanes is working.. but when you click on button Status View and you will see chart + sub panels.. there isn't working as i described above

PedroErnst:
I've pushed another commit @Mausino to take into account the Campaign subpanels.

Please note that message queue does not show a hint.

samus-aran:
Codacy rules need tweaked and found this PR was only causing minor issues raised.

PR #5578, opened 2018-03-23
Fixed #5376 - User's Outbound Email Settings Redesign

### LINKED ISSUE #5376: User's Outbound Email Settings redesign
#### Issue

There seem to be two different outbound settings screens, inside the user profile email settings.

1. Profile -> Scroll down to the bottom -> Settings -> Mail Accounts-> Mail Accounts (the top half of the screen)-> ADD(or EDIT) This one sends correctly, and has From name fields.

image

2. Profile -> Scroll down to the bottom -> Settings -> Mail Accounts -> Outgoing SMTP (lower half, list box) -> ADD (or in the list view, EDIT icon) This one doesn't send, and doesn't have these fields

image

#### Expected Behavior
Now, I'm not sure there is actually a sending bug here, maybe it's just so convoluted to set it up that I wasn't able to do it. But that is also a problem, right?

#### Possible Fix
This needs a screen redesign.

1. There should be only one place to set up User's outbound email
2. It should have the `From` and `From name` fields, and they should be on the same screen as the `Send test email` button, since these fields are mandatory for some SMTP servers to allow sending

Since there can be multiple outbound accounts, careful consideration is needed about where to put this.

Another suggestion I have when redesigning this screen is this: 
3. Add some simple text labels with tips. Sentences like `This is the user's personal outbound account, but other system-wide accounts can be set up in Admin / Outbound email settings`. You know, whatever needs to be explained. Let's make this easier to understand.

The redesign needs to work closely with the code that fetches accounts to put in the `Email Compose` window, in the `From` list box. If accounts are discarded from there, for some reason, it should be explicit somewhere. So the settings screens pictured above could signal: `this account won't show up on Compose window because X`

I am not sure what reason this could be, but sometimes the logs have something about it being the same as an admin account, etc. Watch out for users who repeat the same accounts for system account and user's account (typical of admins trying out the system).

#### Environment
* SuiteCRM Version used: 7.9, 7.10

daniel-samson:
## Description

SuiteCRM is not providing a consistent user experience when configuring outbound email addresses. It is not clear what the from name and from address is corresponds to which outbound email account.

See #5376 for details.

## Motivation and Context
- Moves the from name and address to the outbound email account dialog

## How To Test This

- Rebuild JS Minified
- Rebuild JS Grouping
- Quick Repair And Rebuild
- Select profile
- select email settings
- select mail accounts
- configure inbound email accounts

## Types of changes

### Final checklist

### DIFF
--- a/modules/Configurator/tpls/adminwizard.tpl
+++ b/modules/Configurator/tpls/adminwizard.tpl
@@ -342,6 +342,7 @@ function disableReturnSubmission(e) {
                         <table width="100%" cellpadding="0" cellspacing="0">
                             <tr id="mailsettings1">
                                 <td width="20%" scope="row"><span id="mail_smtpserver_label">{$MOD.LBL_MAIL_SMTPSERVER}</span> <span class="required" id="required_mail_smtpserver"></span></td>
+                                <td width="20%" scope="row"><span id="mail_smtpserver_label">{$MOD.LBL_MAIL_SMTPSERVER}</span> <span class="required" id="required_mail_smtpserver"></span></td>
                                 <td width="30%" ><span><input type="text" id="mail_smtpserver" name="mail_smtpserver" tabindex="1" size="25" maxlength="255" value="{$mail_smtpserver}"></span></td>
                                 <td width="20%" scope="row"><span id="mail_smtpport_label">{$MOD.LBL_MAIL_SMTPPORT}</span></td>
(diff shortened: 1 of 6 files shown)


pgorod:
Daniel thanks for working on this! I really appreciate it. 

From the understanding you gained while looking into these inconsistencies, do you see any potential issues when people upgrade? I mean, could they have wrong email accounts data on their database? Do they need to reconfigure their accounts?

Dillon-Brown:
Assessed

PR #5579, opened 2018-03-23
Fix namespace

Dillon-Brown:
## Description

See https://suitecrm.com/suitecrm/forum/suitecrm-beta/18361-7-10-2-breaks-soap-exchange-sync-gmail-sync-and-outlook-integration

## Types of changes

### Final checklist

### DIFF
--- a/service/core/SugarSoapService.php
+++ b/service/core/SugarSoapService.php
@@ -50,7 +50,7 @@ require_once('service/core/SugarWebServiceImpl.php');
  */
 abstract class SugarSoapService extends SugarWebService{
 	protected $soap_version = '1.1';
-	protected $namespace = 'https://suitecrm.com';
+	protected $namespace = 'http://www.sugarcrm.com/sugarcrm';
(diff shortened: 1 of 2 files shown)


pgorod:
+1 for @PedroErnst's suggestion

samus-aran:
We will be changing this in the future - but we need to provide enough warning to that change so portal users have enough time to adapt. That hasn't been scheduled when this will commence.

RichardAtRiva:
Thank you for making this correction.

With regards to samus-aran's comment that this will be changed in the future, I just want to add our comment. We are an ISV with an solution that uses this API and we would prefer you do not make this change in the future. We currently have one code base that uses this API to support both this and SugarCRM. This breaking change will force us to fork the code for both solutions. It is a non-trivial change for us that delivers no value to our end-users. 

The proper way to make this type of change is to introduce a new API with the new namespace that new development can leverage, while maintaining support for the legacy API. You should not be changing namespaces on existing APIs.

Issue #5166, opened 2018-02-09
Your password has expired


horus68:
#### Issue

related to the closed issue #4639 but with some differences.
The message `Your password has expired. Please provide a new password.`  keeps showing in 7.10.RC2. This is a system test, plain stock and upgraded from several versions before. The issue started with 7.10 and its not gone. The password is not generated by the system.

#### Expected Behavior

There are 2 problems:
1- It should'n happen at all. Password expiring should not be a default setting.
2- If you want this to show for expired passwords then the menu should be visible without allowing to see the admin contents as you can see the "recent viewed" menu.

#### Possible Fix

There should be a message displayed when the admin first login to say: 
`please go to Admin / Password Management and change: "System-Generated Password Expiration" to None or else you will have to create a new password each time you login. `
You see my point!
Why use Password Expiration ON by default?!

#### Actual Behavior

```
Please provide a new password.
Your password has expired. Please provide a new password.
```
![2018-02-09 22_10_58-suitecrm php7](https://user-images.githubusercontent.com/3055443/36052696-9ca075ec-0de6-11e8-8f1e-6530c8157e14.png)
#### Your Environment

* SuiteCRM Version used:
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Windows 10 32bit - Chrome / Firefox
* Environment name and version (e.g. MySQL, PHP 7): apache 2.4.18 / PHP Version 7.0.11 / mysql 5.7.17
* Operating System and version (e.g Ubuntu 16.04):

davenz:
Same behaviour observed here in the full 7.10 release, under a similar setup scenario (test system upgraded directly from 7.9.8, which itself was a fresh install). No third-party themes are active nor any modules loaded.

Passwords are not system generated.

Ubuntu 16.04 Server with a stock LAMP stack is the underlying platform.

Landver:
The same problem. Appeared after upgrade from 7.8 to 7.10 few days ago.
Also I can't turn that feature off, because few settings are missing from 'Password Manager' page
![1](https://user-images.githubusercontent.com/15925657/37086937-8e38d4fc-2201-11e8-89fc-9f6850b5b1f6.PNG)

Before typing that message:

- I used `chown www-data:www-data -R /var/www/suitecrm`
- Also I tried to replace `suitecrm/modules/Administration/PasswordManager.tpl` and `PasswordManager.php` with the same files from 'master' branch of GitHub.

But nothing changed.

[...]

PR #5642, opened 2018-04-03
Fix for #5166

rholighaus:
Check for password expiration only when password was system generated

## Motivation and Context

By default, 7.10 would ask users to change their passwords after 7 days even though the password was not system generated. The only way to disable this is to enable the generation of system generated passwords and disable password expiration.
## How To Test This

Don't change any system settings, ie. make sure not to enable system password generation (and keep the default of 7 days password expiration in that section). Set the value of "pwd_last_changed" to a date more than 7 days old. Without my fix, you will be asked to change your password at login; with my change, you won't.
## Types of changes

### Final checklist

### DIFF
--- a/modules/Users/password_utils.php
+++ b/modules/Users/password_utils.php
@@ -85,7 +85,7 @@ function  hasPasswordExpired($username){
 	$current_user->retrieve($usr_id);
     $type='syst';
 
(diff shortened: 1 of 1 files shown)


Dillon-Brown:
Assessed

Issue #4818, opened 2018-01-02
7.10 beta 3 [suggestion] Choose signature in compose email window as in old 7.8.x version


Mausino:
#### Suggestion

The only way how to change singature is that you must go to user settings and to settings of email accounts. 

#### Actual Behaviour

It is not possible to choose singature as was in old type of email system 7.8.x

Dillon-Brown:
I have moved this suggestion to Trello. All github issues that are labeled 'suggestion' will be moved to the new system and closed. Thanks for the suggestion!

Announcement of moving Suggestions:
https://suitecrm.com/forum/suggestion-box/13691-moving-suggestions-from-github-to

New SuiteCRM Suggestion Box
https://trello.com/b/Ht7LbMqw/suitecrm-suggestion-box

PR #5651, opened 2018-04-04
Fix/email quick create

PedroErnst:
Adds Quick Create buttons to Email Imported Detail View.

## Description

The Quick Create buttons existed in the Email module in 7.8.x but are not present in 7.10.
This PR aims to replicate this functionality.

## 7.8.x
screenshot from 2018-04-02 12-02-45

## 7.10.x
image

## How To Test This

Go into the Email module
Import an Email
Create one of each records using the Quick Create buttons

## Types of changes

### Final checklist

### DIFF
--- a/modules/Emails/EmailsController.php
+++ b/modules/Emails/EmailsController.php
@@ -157,6 +157,58 @@ class EmailsController extends SugarController
     }
 
     /**
+     * Creates a record from the Quick Create Modal
+     */
+    public function action_QuickCreate()
+    {
+        $this->view = 'ajax';
+        $originModule = $_REQUEST['module'];
+        $targetModule = $_REQUEST['quickCreateModule'];
+
+        $_REQUEST['module'] = $targetModule;
+
+        $controller = ControllerFactory::getController($targetModule);
+        $controller->loadBean();
+        $controller->pre_save();
+        $controller->action_save();
+        $bean = $controller->bean;
+
+        $_REQUEST['module'] = $originModule;
+
+        if (!$bean) {
+            $result = ['id' => false];
+            echo json_encode($result);
+            return;
+        }
+
+        $result = [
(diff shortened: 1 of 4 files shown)


pgorod:
I've seen users in the Forums extremely disappointed with losing Quick Create from Emails when they upgrade SuiteCRM beyond 7.8. I'm really glad we're getting this back into the product, it's essential to some user's workflows.

Mausino:
@pgorod I've seen users also in the Forums disappointed from #4818 which was also beyond 7.8

But it's great that some functions (ideas) from old email system come back, like this one.

JoeyJoeJoeJrShabadoo:
Very useful and essential functionality. Thanks for bringing this back.

pgorod:
@PedroErnst how does the screen look if you turn off this option

`Admin` / `System Settings` / `Display actions within menus:`

Doesn't that show too many buttons? I always use my SuiteCRM with that option off, saves a _lot_ of clicks.

PedroErnst:
Good point!

Here's a screenshot:

screenshot from 2018-05-08 09-06-35

pgorod:
In a perfect world, I guess each company will have their workflow and always create the same sort of thing: an opportunity, or a contact, or whatever is their initial step in the sales funnel. So it would be nice to go without a dropdown, and just directly press the button you want - if you could only choose which button you want there.

On the other hand, having too many buttons with long names is not very nice, and it takes up a lot of space on top, and makes you have to scroll to see your data.

I don't know what is the best option, unless we would actually want to go and add options in the Settings screens, but I'm not sure we do...

labcerouno:
This will be VERY useful! Thank you guys!

Dillon-Brown:
Assessed

Dillon-Brown:
Assessed

Issue #5659, opened 2018-04-05
Create new case textarea missing


cata-op:
#### Issue
When creating a new case, the description textarea does not show and errors are generated in console.
If a record is edited then the tinyMCE textarea does show up.

#### Expected Behavior
TinyMCE editor should be shown when creating a case.

#### Actual Behavior
Tested on https://demo.suiteondemand.com
Tested on Chrome: Version 65.0.3325.181 (Official Build) (64-bit)
Tested on Firefox : 59.0.2 (64-bit)

#### Your Environment
![error](https://user-images.githubusercontent.com/38098508/38356259-f191e0a2-38c7-11e8-9071-9d57c1128ed8.png)

Tested on the https://demo.suiteondemand.com

crespo5:
When you push "Create new case" you only can create a "closed" case for some record of SuiteCRM, and in the list of cases, you only can edit for modify the resolution.

To create a case, please go to some record (for example, any account) and go to the DetailView of this accoun, in the bots panels, you can create a case for this account with the description of the case. 

Later other user go to cases and edit this case, put the resolution and change the status.

Issue #5094, opened 2018-02-05
Case module layouts inconsistent


Mark-Keating:
Summary:
The defaults for the Case's module have be amiss for a long time, ver 7.8 and up, where the programmed additions and variations on the cases have resulted in various different presentations.
These can not be fixed in studio, and as a default is not very good.
 
Not Seen:  not seen in current list

Security Risk: No
#### Issue

The programmed additions into the cases is different between create/edit/display and quickcreate where fields most notably "description" is not usable in the create, cannot display as full width in  display and in edit mode both description and resolution are switched to HTML fields, with the data still just a text area.

Trying to adjust in studio presents some odd results , and is not fixable without going in and hacking code. 

#### Expected Behavior
The default as shipped should be a consistent working clean presentation.

#### Actual Behavior
Different behavior between functions

#### Steps to Reproduce
1. Install new version
2.create case (notice that description is unavailable)
3.display case (notice that fields are half width - description/resolution) 
4.edit case      ( notice that description/resolution are presented as html fields (in db is html)

#### Context

use the cases "as is" for cases, but it is not very presentable.

#### Your Environment

* SuiteCRM Version used: 7.10 RC - Feb 5th 2018
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): FireFox 64 bit 
* Environment name and version (e.g. MySQL, PHP 7): mysql php7
* Operating System and version (e.g Ubuntu 16.04): Linux Mint

Shot of Create:
![screenshot from 2018-02-05 15-21-28](https://user-images.githubusercontent.com/34941779/35827882-a0292ee2-0a8b-11e8-81eb-e0446d96030d.png)

Shot of Display
![screenshot from 2018-02-05 15-23-13](https://user-images.githubusercontent.com/34941779/35827696-01cad1ce-0a8b-11e8-94fb-1cc61fe05f98.png)

Shot of Quick create
![screenshot from 2018-02-05 15-22-13](https://user-images.githubusercontent.com/34941779/35827743-2d9ea1ae-0a8b-11e8-9e73-4edd53d11c48.png)

Shot of Edit
![screenshot from 2018-02-05 15-23-57](https://user-images.githubusercontent.com/34941779/35827660-e654c6c0-0a8a-11e8-8245-e15329ef4b49.png)

PedroErnst:
Thanks @Mark-Keating for bringing this up.

[...]

PR #5672, opened 2018-04-06
Fix/5094 case description

PedroErnst:
Call tinymce init again after the page is loaded on Cases create view.

## Description

A conflict arises due to the timing of the code which hides the updates field. The updates field is initialized first, but has to wait for a callback from an ajax request. By the time it comes back, the field -has been deleted by the code in Cases/views/view.edit.php L 70-75.

The solution is to hide the field instead of deleting them, and then manually calling tinyMCE init again.

## Motivation and Context

Issue #5094 
Issue #5659 

## How To Test This

1. Go to cases module
2. Create a new case
3. See the description fields displays wysiwyg editor.

## Types of changes

### Final checklist

### DIFF
--- a/modules/Cases/views/view.edit.php
+++ b/modules/Cases/views/view.edit.php
@@ -67,12 +67,13 @@ class CasesViewEdit extends ViewEdit {
             ?>
             <script>
                 $(document).ready(function(){
-                    $('#update_text').closest('.edit-view-row-item').html('');
-                    $('#update_text_label').closest('.edit-view-row-item').html('');
-                    $('#internal').closest('.edit-view-row-item').html('');
-                    $('#internal_label').closest('.edit-view-row-item').html('');
-                    $('#addFileButton').closest('.edit-view-row-item').html('');
-                    $('#case_update_form_label').closest('.edit-view-row-item').html('');
+                    $('#update_text').closest('.edit-view-row-item').hide();
+                    $('#update_text_label').closest('.edit-view-row-item').hide();
+                    $('#internal').closest('.edit-view-row-item').hide();
+                    $('#internal_label').closest('.edit-view-row-item').hide();
+                    $('#addFileButton').closest('.edit-view-row-item').hide();
+                    $('#case_update_form_label').closest('.edit-view-row-item').hide();
+                    tinyMCE.execCommand('mceAddControl', false, document.getElementById('description'));
                 });
             </script>
         <?php


Dillon-Brown:
Assessed

aarondglover:
This works when creating a Case. The wysiwyg editor appears.
The wysiwyg editor also appears when editing a case.

The wysiwyg editor DOES NOT appear if you use the "Save and Continue" or Navigate Next/Previous Buttons

pgorod:
@Dillon-Brown this needs to be reopened because of the above comment by @aarondglover.

Also [on the Forums](https://suitecrm.com/suitecrm/forum/suitecrm-7-0-discussion/18759-case-description-field-text-entry-isn-t-saved-after-upgrading-to-7-10-4?start=15#66604) someone is saying that

> I analize the situation with my user and she register cases using the cases subpanel inside accounts module. When she register using cases module works fine. 

On the other hand, this other PR #5816 seems to overlap with this one? It's changing the same part of the same file.

PedroErnst:
@pgorod 

#5816 addresses the scenario reported by @aarondglover 
It seems to fix the issue. But it's not yet merged.

Then there's also #5840, which replicates both this PR #5672 and #5816, but for 7.8.x, and the changes do not seem to have carried over into 7.10.5

It's a bit confusing.

I'm going to have a look at that forum post.

lw9474:
I just upgraded to 7.10.5 hoping it would resolve this issue.
Unfortunately it  breaks it worse.
7.10.4 would not save description when ticket was created. But would save it if you edited an existing ticket and added the description.

After uppgrade to 7.10.5 and now it does not save when you edit the case either.

Thanks

lw9474:
I just upgraded to 7.10.6 and it still does not save the description of the problem.

[Fri Jun 15 15:25:45 2018] [error] [client 192.168.101.10] PHP Notice:  Undefined property: aCase::$type in /var/www/html/suitecrm/modules/AOP_Case_Events/CaseEventsHook.php on line 74, referer: http://192.168.101.55/suitecrm/index.php

pgorod:
@lw9474 [line 74 on that file](https://github.com/salesagility/SuiteCRM/blob/v7.10.6/modules/AOP_Case_Events/CaseEventsHook.php#L74), in 7.10.6, doesn't seem to have any reference to `$aCase`...

Is your file different? If so, why?

lw9474:
Not to my knowledge.
I just applied the upgrade as normal.
I did check the dates on the modules cases directory and they were
yesterday so I assume the code is the 7.10.6 version.

THX



Wish Markwalter
Karz Systems, Inc.
706-651-0080  XT 209

pgorod:
Well, does your file match the one in the link I gave you? If not, what does it contain?

lw9474:
Sorry.  I did not see the link.  
There is no difference in file in the link and the one in production:

[root@lenny AOP_Case_Events]# diff CaseEventsHook.php /tmp/CaseEventsHook.php
[root@lenny AOP_Case_Events]#

pgorod:
I can reproduce the behaviour of not being able to edit `Description` in a Case, but I can't get that error to show. Maybe it's not strictly related...

pgorod:
@Dillon-Brown is this a matter of copying some PR into the 7.10 branch like @PedroErnst suggests? I am a bit confused...

lw9474:
Do not call tiny mceAddControl on edit view #6165
This PR fixes the edit issue.
But it still does not save the description when a new case is created.
After making the patch there is not an error in the /var/log/httpd/error_log but nothing gets saved in the description.  My line 74 is the same as the one in the fix.

THX

Issue #5692, opened 2018-04-10
'Reply to' email cause error on Send and invalidates session - issue in 7.10.4 and 7.10.3 


LEAP-nishit:
#### Issue

Sending an email by using the following works
Email -> Compose View -> Fill Form -> Send email. 

The following does not work,
Email -> View Email -> Click on Email -> Reply -> Fill Form -> Send email

The following error is generated:

![msg1](https://user-images.githubusercontent.com/9854372/38539608-3b3bde48-3cdc-11e8-8906-b22086a1378f.png)

In both cases the `From:` is set to the same value. 

#### Expected Behavior
Users should be able to replay to an email. The settings work when composing from scratch, fails on reply.  

#### Actual Behavior
A error message (shown above) is displayed, the following error is generated in `suitecrm.log`. 

```log
Tue Apr 10 16:21:04 2018 [18443][1][ERROR] Unable to find relationship emails_email_templates
Tue Apr 10 16:21:04 2018 [18443][1][ERROR] Unable to find relationship emails_email_templates
Tue Apr 10 16:21:04 2018 [18443][1][ERROR] Unable to find relationship emails_email_templates
Tue Apr 10 16:21:11 2018 [18444][1][ERROR] Unable to find relationship emails_email_templates
Tue Apr 10 16:21:11 2018 [18444][1][ERROR] Unable to find relationship emails_email_templates
Tue Apr 10 16:21:11 2018 [18444][1][ERROR] Unable to find relationship emails_email_templates
Tue Apr 10 16:21:11 2018 [18444][1][ERROR] Unable to find relationship emails_email_templates
Tue Apr 10 16:21:31 2018 [18331][1][ERROR] Unable to find relationship emails_email_templates
Tue Apr 10 16:21:31 2018 [18331][1][SECURITY] User Administrator attempted to send an email using incorrect email account settings in which they do not have access to.

```
#### Steps to Reproduce

1. Clean Install of SuiteCRM 7.10.4
2. Setup system email 
3. Setup outbound email
4. Create a case and send email to a client 
5. Wait for client to respond 
6. Open email and reply - Should fail and log you out. 
7. Log back in and Send a new email - should work. 
#### Context

Sending and receiving email for Cases using the CRM requires users to reply to customer responses. In 7.10.x this is not possible.  

#### Your Environment

* SuiteCRM Version used: 7.10.4
* Browser: Firefox 59.0.2 and Chrome 65.0.3325.181
[...]

PR #5734, opened 2018-04-16
Fix/email failed permission check

PedroErnst:
Improved error handling and logging for errors when sending emails. 

## Description

1. No longer logs the user out when sending and email if the  userIsAllowedToSendEmail() function returns false. It is still logged logged with security level. 
This was very annoying when it happens to a user which is not actually trying anything malicious (getting logged out without knowing why), and wouldn't actually stop a malicious actor since they can log in again straight away.

3. Adds more complex logging to the userIsAllowedToSendEmail() function. Currently this function checks whether the user has access to the requested Inbound Account, whether the addresses match and whether he is allowed to use the Outbound Account. But if any of these fails it will make a generic log entry without telling which of the three checks failed. 
So we add a more specific log for each of these cases with extra details.

screenshot from 2018-04-16 14-28-32

## Motivation and Context

Issue #5692 and similar.

## How To Test This

Go to compose Email screen.
Select a from address.
In a different tab, go into the admin or profile settings and change some details from the account that is selected in the first tab.
Try to send the email and fail.
Check the logs.

## Types of changes

### Final checklist

### DIFF
--- a/modules/Emails/EmailsController.php
+++ b/modules/Emails/EmailsController.php
@@ -206,8 +206,6 @@ class EmailsController extends SugarController
                 'title' => $app_strings['LBL_EMAIL_ERROR_SENDING']
             );
             echo json_encode($response);
-            // log out the user
-            session_destroy();
         }
     }
 
--- a/modules/Emails/EmailsController.php
+++ b/modules/Emails/EmailsController.php
@@ -175,21 +175,21 @@ class EmailsController extends SugarController
 
             $this->bean->handleMultipleFileAttachments();
 
-        // parse and replace bean variables
-        $this->bean = $this->replaceEmailVariables($this->bean, $request);
+            // parse and replace bean variables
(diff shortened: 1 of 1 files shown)


modules/Emails/EmailsController.php
```diff
@@ -200,14 +200,12 @@ public function action_send()
             );
 
             $this->view = 'ajax';
-            $response['errors'] = array(
+            $response['errors'] = [
```
samus-aran: Not backwards compatible for PHP 5.5.9

PedroErnst: I think short array syntax was introduced in [5.4](http://php.net/manual/en/language.types.array.php#example-56)

gymad: tested in 5.5.9: it's works for me

Dillon-Brown:
Assessed

PR #5759, opened 2018-04-20
Fixed#2922

### LINKED ISSUE #2922: email to cases failure with email body > 65536 Bytes
suitecrm 7.7.9
email to cases failure with email body > 65536 Bytes

the email body does not fit inside mysql description TEXT column causing a mysql error when case is created. 

solution: email body to be truncated before mysql query.

boombata:
Fixed #2922 - email to cases failure with email body > 65536 Bytes
## Description

Changes the cases vardef to allow the description field of a case to be greater than 65536 Bytes.

Issue ref: #2922
this pr in replacement of https://github.com/salesagility/SuiteCRM/pull/4334 because is a side effect doing changing only vardef, the detail , edit and quick create views do not display description field as a multi lines field but just single line.

## Motivation and Context

  This fixes a case failing to create when using the quick create from the emails module. This was due to an SQL insert query failure when the text of the email was above 65536 Bytes.

## How To Test This

  Create an email with 65536 bytes of text.
    Click on quick create > case

## Types of changes

### Final checklist

### DIFF
--- a/modules/DynamicFields/templates/Fields/TemplateTextLongArea.php
+++ b/modules/DynamicFields/templates/Fields/TemplateTextLongArea.php
@@ -1,0 +1,129 @@
+<?php
+
+/*********************************************************************************
+ * SugarCRM Community Edition is a customer relationship management program developed by
+ * SugarCRM, Inc. Copyright (C) 2004-2013 SugarCRM Inc.
+
+ * SuiteCRM is an extension to SugarCRM Community Edition developed by Salesagility Ltd.
+ * Copyright (C) 2011 - 2019 Salesagility Ltd.
+ *
+ * This program is free software; you can redistribute it and/or modify it under
+ * the terms of the GNU Affero General Public License version 3 as published by the
+ * Free Software Foundation with the addition of the following permission added
+ * to Section 15 as permitted in Section 7(a): FOR ANY PART OF THE COVERED WORK
+ * IN WHICH THE COPYRIGHT IS OWNED BY SUGARCRM, SUGARCRM DISCLAIMS THE WARRANTY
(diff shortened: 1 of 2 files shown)


modules/DynamicFields/templates/Fields/TemplateTextLongArea.php
```diff
@@ -0,0 +1,96 @@
...
+	var $type = 'longtext';
+	var $len = '';
+
+	function __construct()
+	{
+		$this->vardef_map['rows'] = 'ext2';
+		$this->vardef_map['cols'] = 'ext3';
+	}
+
+	function set($values){
```
PedroErnst: Method visibility should be explicitly declared (private / public). Also the methods missing phpdocs.

boombata:
@PedroErnst should be ok now. thanks

Mac-Rae:

pgorod:
I've been playing with textareas and TinyMCE editors recently. I find that the code is a bit of a mess with different field types, and often the implications of moving away from one field type to another are not always completely understood.

We have a user-profile setting for the kind of editor to use (TinyMCE, Mozaik, Direct HTML) that people often complain about, because it affects some rich text editors throughout the product, but not others.
 
My idea for all this was to use a new field type called **editor** which would allow for all those 3 editors (and any future ones) to be used seamlessly in any module/view.

The multiplication of field types is not good for this. 

I think the current problem should be solved by using a **db-type** to force the long text on the database size, but without forking the **`modules/DynamicFields/templates/Fields/TemplateTextArea.php`** file, repeating code. The field type would stay the same as before, only the database would use a longer field.

Sorry for not saying this sooner, but I simply didn't understand this sooner...

Issue #5029, opened 2018-01-26
DateTime calendar popup button & manual edit of data entry doesn't work in 7.9.10


bunk3m:
#### Issue
The pop-up calendar button doesn't function.
You are able to manually add the date portion but not the time.
So you can't add any datetime to any task.

This also happens in the create Call screen.

See picture.
<img width="174" alt="screen shot 2018-01-25 at 17 43 45" src="https://user-images.githubusercontent.com/3494567/35418442-3c7e5be4-0200-11e8-9484-56dc45b90dd3.png">

#### Expected Behavior
As in past versions, you should be able to select the date & time in the pop-up calendar or manually add the date & time.

#### Actual Behavior
Can't add the date or time in a task or call

#### Possible Fix
Haven't found any.

#### Steps to Reproduce
1. Create a new Task (Call) from a drop down or edit a task (call).  You will see that there are }} braces above the datetime entry dialog box.  
2. Try to click on the calendar icon
3. Try to add date & time in the dialog box

#### Context
I can't change or add any calls or tasks.  This is a major issue.

#### Your Environment
This happens using Mac versions of Chrome Version 63.0.3239.132 (Official Build) (64-bit); Firefox 58.0 (64-bit)
SuiteCRM 7.9.10
Debian 9; Apache 2.4.25; mysql  Ver 15.1 Distrib 10.1.26-MariaDB; PHP 7.0.27

ksmith892:
The problem appears to have been caused by additions to the template code in:

/var/www/suitecrm/suitecrm/include/SugarFields/Fields/Datetimecombo/EditView.tpl

Lines 45 - 51 are new additions in version 7.9.10, and I belive the issue is cause by the template delimiters around line 49-50. I don't know enough Smarty Templating to work out the correct fix for this. But I believe it is being caused by a mismatch between single and double curly brackets on thoxse lines.

ksmith892:
This fixes the issue when there is a single DateTimeCombo on the form, however it doesn't fix the issue when there is a second DateTimeCombo box underneath it i.e. on a meeting or task form.

pgorod:
I'm getting this same issue on 7.8.11 in the Create Event screen, which also has two dates and is currently unusable.

okoester:
Also getting this error in Tasks view when trying to create a task.
Still after applying Dillon Browns fix.

In Firefox developer console I see these errors referencing to a cached javascript...

`error adding script
sugar_grp1.js:453:76
[...]

PR #5799, opened 2018-04-27
Fixed #5663 - AOR_Reports Date issues

### LINKED ISSUE #5663: AOR_Reports Date issues
#### Issue
There are still some serious issues with the date in the AOR_Reports module.
Using two dates in Conditions does break the whole function. The second date will not have any hour and minutes fields, and the fields of the first date will be reseted. The datepicker of the second condition additionaly does not work. When deleting the second date field, its not possible to save because of the error message 'Missing required field: Hours'. When only one date is used and it is set as a parameter, the date will be set to a wrong value if date format d.m.Y is used.
Picking date 01.04.2018 and after clicking "update" the date will be 04.01.2018.
Using the date format Y-m-d the does not resolve the issue, but opens another one. The date will be deleted when clicking update on the DetailView of the Reports module.

#### Expected Behavior
Using dates should work in the Reports module.

#### Actual Behavior
When using dates everythings seems to be messed up. 

#### Steps to Reproduce

1. Go to report module
2. Create condition
3. Add a date field and choose a date
4. Add another date field
5. After that, saving is not possible (If only one date field is used it can be saved)
Also, setting one date to parameters does work, but changing the date in DetailView does not take the right date.

#### Context

Unable to use date fields in the Report module as parameters. Unable to use more than one date as conditions.

#### Your Environment

SuiteCRM Version used: 7.10.3 (demo)
Browser name and version: Chrome 65.0.3325.181
Environment: demo.suiteondemand.com

Dillon-Brown:
## Description

Fixes an issue preventing the datetimecombo field from correctly displaying when two date parameters are used as a condition in reports.

## Motivation and Context

Issue reference: #5663 

## How To Test This

1. Go to report module
2. Create condition
3. Add a date field and choose a date
4. Add another date field

Also check the dashlets, workflow and calendar for the datetimecombo field.

## Types of changes

### Final checklist

### DIFF
--- a/include/SugarFields/Fields/Datetimecombo/EditView.tpl
+++ b/include/SugarFields/Fields/Datetimecombo/EditView.tpl
@@ -40,15 +40,22 @@
 *}
 {{capture name=idname assign=idname}}{{sugarvar key='name'}}{{/capture}}
 {{if !empty($displayParams.idName)}}
-    {{assign var=idname value=$displayParams.idName}}
+{{assign var=idname value=$displayParams.idName}}
+{{/if}}
+{{if !empty($displayParams.originalFieldName)}}
+{assign var=idname value=$fields[{{sugarvar key='name' stringFormat=true}}].aliasId}
+{assign var=originalId value=$fields[{{sugarvar key='name' stringFormat=true}}].originalId}
+{{else}}
+{{assign var=idname value={{$idname}}
+{{assign var=originalId value={{$idname}}
 {{/if}}
 
 {{assign var=flag_field value=$vardef.name|cat:_flag}}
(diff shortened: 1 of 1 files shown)


pgorod:
@Dillon-Brown is this just reverting the problematic commit? Or is it really going deeper and making everything right? 

Because that commit was also fixing something, namely this issue #5029, so we should make sure we don't get a regression when fixing the regression that fixed a regression :-)

Dillon-Brown:
@pgorod Ideally we'll need to check every datetimecombo field for any regressions along with including https://github.com/salesagility/SuiteCRM/pull/5792.

PR #5806, opened 2018-04-30
feature/api-refact

zoltankocsardi:
## Description
Implementing changed API

- understandable structure
- using symfony's packages for validating given parameters in middleware
- wrapper classes for reaching core functions more easier

documentation: https://github.com/salesagility/SuiteDocs/pull/90

## How To Test This
Gonna link the doc asap. [tests are also coming soon]

## Types of changes

### Final checklist


tsummerer:
I can't get authentication to work properly with this code. 

The Oauth2Client module uses `crypt()` to store a password. 
```php
// modules/OAuth2Clients/OAuth2Clients.php
public function save($check_notify = false)
{
    if (!empty($_REQUEST['new_secret'])) {
        $this->secret = crypt($_REQUEST['new_secret']);
    }
    $this->setDurationValue();
    return parent::save();
}
```

As you can see, the default is to not set a Salt. Therefore, the selected hashing algorithm is undefined. See [docs](http://php.net/manual/en/function.crypt.php)
 

The API's ClientRepository is explicitly choosing an algorithm (sha256):
```php
//Api/V8/OAuth2/Repository/ClientRepository.php
public function getClientEntity($clientIdentifier, $grantType, $clientSecret = null, $mustValidateSecret = true)
{
    /** @var \OAuth2Clients $client */
    $client = $this->beanManager->getBeanSafe(\OAuth2Clients::class, $clientIdentifier);
    if ($mustValidateSecret && hash('sha256', $clientSecret) !== $client->secret) {
        return null;
    }
```

Obviously the `$client->secret` and `$clientSecret` won't match as they were hashed with different algorithms.

Api/V8/JsonApi/Response/DocumentResponse.php
```diff
@@ -0,0 +1,86 @@
...
+            $this->setMeta(new MetaResponse(['message' => 'Request was successful, but there is no result']));
+        }
+
+        $response = [
+            'meta' => $this->getMeta(),
+            'data' => $this->getData(),
+            'links' => $this->getLinks(),
+        ];
+
+        return array_filter($response);
```
tsummerer: @zoltankocsardi 
If I'm understanding the JSON Api correctly, I believe that this allows an invalid response. 
If you query a module with no records in it (say, Alerts), then an empty array is returned. 

According to the [spec](http://jsonapi.org/format/#document-top-level), we have to return _something_. I think the most logical thing would be to return the "data" key with the value being a blank array.

```
{
    data : []
}
```

zoltankocsardi:
@tsummerer yes, you are right, I haven't synchronized the auth yet. Gonna fix it asap. Thanks!

Api/V8/JsonApi/Response/DocumentResponse.php
zoltankocsardi: Another nice catch. I thought data must contain id and type [otherwise just show a message in the meta object like there is no result], but yes, it could be null or an empty array in case of collections. Gonna refine this as well. Thanks again! :) :beer:

Api/V8/JsonApi/Helper/AttributeObjectHelper.php
```diff
@@ -0,0 +1,48 @@
...
+     *
+     * @return AttributeResponse
+     */
+    public function getAttributes(\SugarBean $bean, $fields = null)
+    {
+        $bean->fixUpFormatting();
+        $attributes = $bean->toArray();
+
+        // using the ISO 8601 format for dates
+        array_walk($attributes, function (&$value) {
```
gymad: array_walk returns could be TRUE on success or FALSE on failure. Your code does not handle the failure.

Api/V8/JsonApi/Response/ErrorResponse.php
```diff
@@ -0,0 +1,84 @@
...
+    {
+        $response = [
+            'errors' => [
+                'status' => $this->getStatus(),
+                'title' => $this->getTitle(),
+                'detail' => $this->getDetail(),
+            ]
+        ];
+
+        return array_filter($response);
```
gymad: array_filter can not change the input array whit this array structure. did you mean array_filter($response['errors']) first?

Api/docs/postman/SalesAgility.postman_collection.json
```diff
@@ -0,0 +1,934 @@
+{
```
gymad: why this json for? where do we use it?

Api/V8/JsonApi/Helper/AttributeObjectHelper.php
zoltankocsardi: Yeah, that's true, but not because of the returns value. I both want to iterate and change the value if it's necessary and array_walk could do the job as well, but you are right.  Gonna use array_map instead,

Api/V8/JsonApi/Response/ErrorResponse.php
zoltankocsardi: Hehe, yeah, my mistake. Was a copy paste. We don;t use filter here.

Api/docs/postman/SalesAgility.postman_collection.json
zoltankocsardi: You can import example endpoints into your postman. Try it :)

Api/V8/JsonApi/Helper/AttributeObjectHelper.php
zoltankocsardi: Fixed.

Api/V8/JsonApi/Response/ErrorResponse.php
zoltankocsardi: Fixed.

tsummerer:
The `page[number]` parameter here is still incomplete.

Two issues:
1. the API only returns a maximum of 20 records. Even if I pass, say 200 to the parameter. 
2. If a number > than 20 is passed, then the pagination calculates incorrectly. 

Steps to Reproduce:
1. Start with an instance with demo data (50 Accounts).
2. `GET Api/V8/module/Accounts?fields[Accounts]=name,account_type&page[size]=200&page[number]=1`

Expected - 200 Accounts
Actual - 20 Accounts and the following pagination details:
```json
    "links": {
        "first": "/V8/module/Accounts?fields[Accounts]=name,account_type&page[size]=200&page[number]=1",
        "prev": null,
        "next": null,
        "last": "/V8/module/Accounts?fields[Accounts]=name,account_type&page[size]=200&page[number]=1"
    }
```

At this point, performing another GET with `page[size]=200&page[number]=2` returns nothing 


IMO, this should either be fixed before it is merged OR the documentation should be updated to clarify these details.

zoltankocsardi:
@tsummerer thank you for your remarks again! :)

tsummerer:
Hey, I haven't been doing rigorous or apples-to-apples tests, but this refactor is looking like it's about twice as fast (at GET requests) as the previous version.. Are you seeing those sorts of numbers too?

Also, do you envision this as becoming a replacement for the V8 API? Or will it be a new "version" (V9)?

zoltankocsardi:
@tsummerer it hasn't been decided it yet :)

ModuleInstall/extensions.php
```diff
@@ -32,37 +32,126 @@
...
+    'aow' => [
+        'section' => 'aow_actions',
+        'extdir' => 'Actions',
+        'file' => 'actions.ext.php',
+        'module' => 'AOW_Actions'
+    ],
+    // API
+    'routes' => [
+        'section' => 'routes',
+        'extdir' => 'Api/lol',
```
samus-aran: Api/lol? :stuck_out_tongue:

composer.json
```diff
@@ -58,7 +58,7 @@
     },
     "classmap": [
       "Api/",
-      "custom/Extension/Api"
+      "custom/Extension/application/Ext/Api"
```
samus-aran: Shouldn't this be
custom/application/Ext/Api
rather than the original file that generates the above file after a repair and rebuild?

ModuleInstall/extensions.php
zoltankocsardi: Ooops, it was for testing purposes.

composer.json
zoltankocsardi: Changed.

PR #5808, opened 2018-04-30
Fixed #5748 - v7.10.4 API v8 filtering using [[in]] operator is broken.

### LINKED ISSUE #5748: v7.10.4 API v8 filtering using [[in]] operator is broken.
#### Issue
Following docs here: https://docs.suitecrm.com/developer/api/version-8/json-api/#_operators
on how to use the [[in]] operator

Two issues when using the in operator:
- When querying a default field (e.g. billing_address_city) the system  will never return results.
- When querying a custom field the system, if there should be multiple results, only one is ever returned.

We can verify that there should be results because querying manually in the Web UI yields expected results

#### Expected Behavior
Expect to receive expected results that meet query criteria.

#### Actual Behavior

*When querying a default field*

we get this in the response
```
    "data": null,
    "meta": {
        "offsets": {
            "current": 0,
            "count": -1
        },
```

this error is generated in the web server error logs

```
[Thu Apr 19 07:44:09.453325 2018] [php7:notice] [pid 2538] [client ********] PHP Notice:  Undefined index: list in <path_to_suitecrm>/lib/API/v8/Controller/ModuleController.php on line 370
```

which points to :
https://github.com/salesagility/SuiteCRM/blob/master/lib/API/v8/Controller/ModuleController.php#L370

When querying a custom field, no error is generated. Only one result is returned even though  there should be multiple

#### Steps to Reproduce
(assumes rest client is setup with authentication)

For default field i was using `billing_address_city` 

1.  Verify existence of records by querying in the web gui
2. in REST client create following queries
`default field`
`<base url>/api/v8/modules/Accounts?filter[Accounts.billing_address_city]=[[in]]London,[[in]]Cardiff`

`custom field`
`<base url>/api/v8/modules/Accounts?filter[Accounts.custom_field_c]=[[in]]100000,[[in]]100001`

3. execute query.

4. cry

#### Context
Bug prevents me from performing simple queries in suitecrm. 

#### Your Environment

* SuiteCRM Version used: 7.4.10
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): irrelevant
* Environment name and version (e.g. MySQL, PHP 7): MySQL, PHP7
* Operating System and version (e.g Ubuntu 16.04): don't know.

s-hal:
## Description
When removing the operator from the filter value to get the comparator, characters was faulty removed from the comparator. Both the operator and the value was converted to arrays. array_diff was used to remove the operator from the value. array_diff removes multiple occurrences. 
operator = [[eq]] 
value = [[eq]]Denver
result = Dnvr

## Motivation and Context
Filter in API v8 do not work properly.

## How To Test This
1.  Navigate to Admin,  System Settings, Logger Settings set  Log Level to Debug
2. Use the JSON API to filter. https://myhost/api/v8/modules/Accounts?filter[Accounts.name]=[[eq]]Denver
3. Check the SQL WHERE clause in the log "where (accounts.name IN ("Dnvr"))"

## Types of changes

### Final checklist

### DIFF
--- a/lib/API/JsonApi/v1/Filters/Parsers/FilterParser.php
+++ b/lib/API/JsonApi/v1/Filters/Parsers/FilterParser.php
@@ -341,9 +341,11 @@ class FilterParser
      * @return string
      */
     private function stringDifference($a, $b) {
-        if (substr($a, 0, strlen($b)) == $b) {
+        if (substr($a, 0, strlen($b)) === $b) {
             return substr($a, strlen($b));
         }
+        
+        return "";
     }
 
     /**
--- a/lib/API/JsonApi/v1/Filters/Parsers/FilterParser.php
+++ b/lib/API/JsonApi/v1/Filters/Parsers/FilterParser.php
@@ -341,10 +341,9 @@ class FilterParser
      * @return string
      */
     private function stringDifference($a, $b) {
-        $aArray = str_split($a);
-        $bArray = str_split($b);
-        $arrayDiff = array_diff($aArray, $bArray);
-        return implode('', $arrayDiff);
+        if (substr($a, 0, strlen($b)) == $b) {
+            return substr($a, strlen($b));
+        }
     }
 
     /**
--- a/tests/unit/lib/SuiteCRM/API/JsonApi/v1/Filters/Parsers/FilterParserTest.php
+++ b/tests/unit/lib/SuiteCRM/API/JsonApi/v1/Filters/Parsers/FilterParserTest.php
@@ -293,6 +293,17 @@ class FilterParserTest extends \Codeception\Test\Unit
         );
     }
 
+    public function testParseFieldFilterWithDuplicateCharactersInTheValue()
+    {
+        $expect = array(
+            '[[eq]]',
+            'denver'
+        );
+
+        $actual = self::$filterParser->parseFieldFilterAdapter('[[eq]]denver');
+        $this->assertSame($expect, $actual);
+    }
+
     public function testParseFilter()
     {
 


lib/API/JsonApi/v1/Filters/Parsers/FilterParser.php
```diff
@@ -341,10 +341,9 @@ protected function isInOperatorsArray($operatorNeedle, array $operatorsHaystack)
      * @return string
      */
     private function stringDifference($a, $b) {
-        $aArray = str_split($a);
-        $bArray = str_split($b);
-        $arrayDiff = array_diff($aArray, $bArray);
-        return implode('', $arrayDiff);
+        if (substr($a, 0, strlen($b)) == $b) {
```
PedroErnst: `stringDifference('php', 'h')` doesn't return anything?

s-hal: I will add the empty string return.

According to you, what should the function do?

s-hal:
This PL does not fix the whole issue # 5748. I read the description again and realized that this do not solve the multiple conditions error.

lib/API/JsonApi/v1/Filters/Parsers/FilterParser.php
PedroErnst: Well, the way I interpret the name is that this function will return the difference between two strings. But then the original body seems to do just that, so I might be wrong. I haven't looked further into this issue, just reviewing this particular method.

Another thing you want to watch out for is the comparison operator, you want strict because otherwise ('01' == '1') equals true, and those are not equal strings.

samus-aran:
Closing as we will be implementing this into V8.

PR #4779, opened 2017-12-24
add support for 'Any Change' condition on related field

YonatanRosemarin:
Fix "Any Change" condition at Workflow on related field

## Description

Bug's steps to reproduce:
1. Create a Workflow on a module that has related field
2. Add condition of type "Any Change"
3. The workflow will always run, even if the related field's value was not changed

## Motivation and Context

fixes the bug described below

## How To Test This

1. Create a Workflow on a module that has related field
2. Add condition of type "Any Change"
3. The workflow will only run if the value of the relation will be changed

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [x] My code follows the code style of this project found [here](https://suitecrm.com/wiki/index.php/Coding_Standards).
- [ ] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://suitecrm.com/wiki/index.php/Contributing_to_SuiteCRM) guidelines.

PR #5818, opened 2018-05-03
Fix/5803 workflow any change related

### LINKED ISSUE #5803: Workflow bug - One field - any change - triggered by any field
#### Issue
It seems like there is a bug in the workflow, because I have never made this one work no matter what version i am on. And pgr confirmed I need to open an issue here. This is very a very important feature of the workflow that needs to work properly.

When I create a workflow for cases, tasks, and other modules. I need a few workflows that get triggered only when one specific field was modified. For example, date or description or user assigned to.

I create a workflow, repeated runs, only on save.
Condition - Field - User Assigned to - Any change 

#### Expected Behavior
This is supposed to trigger ONLY when an assigned user was changed.

#### Actual Behavior
Triggers if any of the fields were modified. And actually triggers even when nothing was modified, just save the record again, and it still triggers that workflow.

#### Possible Fix
Not sure what the possible fix is, but we absolutely need this feature to work! This has not been working for a long time. 

#### Context
Absolutely HIGH priority!
I'm trying to create a few notifications that get triggered only when a specific field was modified. 
When a new user is assigned, the new user will get an email notification (not a default notification)
When the date was changed, the assigned user will get a custom notification about the new date etc.

#### Your Environment
Does not work in any of the versions I tested. Confirmed by other users as well.
Here is my forum post
https://suitecrm.com/suitecrm/forum/suitecrm-7-0-discussion/18783-workflow-notification-when-one-field-was-changed

If this is fixed, I will be beyond thankful!

PedroErnst:
Improves on #4779 
Avoids comparing to an nonexistent array index and always marking a field as changed

## Motivation and Context

Fixes #5803 

## How To Test This

1. Create a workflow related to Contacts or any other module
2. Set repeated runs, on modified records
3. Add conditions: Assigned user, any change
4. Change a Contact without changing the assigned user
5. See the WF didn't run

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOW_WorkFlow/AOW_WorkFlow.php
+++ b/modules/AOW_WorkFlow/AOW_WorkFlow.php
@@ -613,11 +613,12 @@ class AOW_WorkFlow extends Basic
                         break;
 
                     case 'Any_Change':
-			if($data['type'] == 'relate' && isset($data['name'])) {
-				$value = $condition_bean->rel_fields_before_value[$condition->field];
-			} else {
-				$value = $condition_bean->fetched_row[$condition->field];
-			}
+                        if ($data['type'] === 'relate' && isset($data['name'])
+                            && isset($condition_bean->rel_fields_before_value[$condition->field])) {
+                            $value = $condition_bean->rel_fields_before_value[$condition->field];
+                        } else {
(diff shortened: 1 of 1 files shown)


Dillon-Brown:
Assessed

PR #5842, opened 2018-05-08
Do not get the cached version for newly created beans from Factory

### LINKED ISSUE #3468: Email Template - Note Date Entered - Format remains yyyy-mm-dd
#### Issue

We have created a workflow from the module Notes as follows:

Condition:
Module - Notes
Field - Platform
Operator - Equal
Type - One of
Value - Currency Cloud, Ebury

Action:
Send Email to Related Field - Leads : Leads

The email template (Attached) contains the $note_date_entered variable, when the email is received by the lead the varaiable is populated however the data format is not correct according to the system date/time format.
   
#### Expected Behavior

The date in the template should read: dd-mm-yyyy

#### Actual Behavior

The date in the template reads yyyy-mm-dd

#### Steps to Reproduce

1. Create a workflow based on the Notes module and set the action to send an email template to a related lead which includes the the variable $note_date_entered variable.
2.
3.
4.

#### Context

Emails sent to customers have the incorrect date format and changing the System Locale settings does not updated the format.

#### Your Environment

* SuiteCRM Version used: 7.7.9
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Version 57.0.2987.133
* Environment name and version (e.g. MySQL, PHP 7): MySQL 5.6, PHP 5.6 & Apache 2.4
* Operating System and version (e.g Ubuntu 16.04): Linux Centos 7.3

image

image

image

### LINKED ISSUE #5719: 7.10.4 new case number still doesn't show up in the email notification
When you just create a new case, and save it, my work flow email notification that contains the case number variable $acase_case_number is not showing anything. But as soon as the ticket is saved, any other work flow email notification with the same exact variable works fine.

Steps to reproduce

1. Create an email template containing a case number variable $acase_case_number.
2. Create a workflow that would send that email template to a user saying you have been assigned a new case.
3. Case number is not showing in that first email. But is showing in the rest of the emails about that case that is considered already existing.

It seems that it's only not working when you just saved the new case.

### LINKED ISSUE #5814: BUG in workflow emails - date and time showing in UTC format - on save only
When I create a case, I have a workflow notification that shows the scheduled date and time in the email that it sends to the assigned user. It shows time in UTC. This is only happening on save. Any workflow triggered by the scheduler would show correct date and time format.

Example:
1. Create a new case, set time to 4:30pm.
2. A custom workflow sends and email notification that contains that date and time tag in it.
3. The email shows date and time in UTC format 2018-05-09 20:30:00
4. Happens only on save. Emails triggered by the scheduler would show correct date and time format. Using the same date and time variables. 

Here is my post where they suggested me to open a new issue here:
https://suitecrm.com/suitecrm/forum/suitecrm-7-0-discussion/18796-new-case-workflow-notification-time-shows-wrong-time#65094

PedroErnst:
Sometimes when creating a new bean, the data stored in the cached version of the BeanFactory will not be correct. For example auto increment fields will only have correct data once they've been inserted and then retrieved.

## Description

Adds an unregister function to the BeanFactory. Calling it will remove a given Bean from cache, so that next time it will be fetched from the DB.

In the workflow template parser we check for a bean being new, and then make sure it is retrieved at least once from the db.

## Motivation and Context

Fixes #5814
Fixes #3468
Fixes #5719

## How To Test This

1. Create a new workflow
2. Related to cases
3. Run on saving new records
4. Action: send email, case creation template
5. Create a new case

## Types of changes

### Final checklist

### DIFF
--- a/data/BeanFactory.php
+++ b/data/BeanFactory.php
@@ -231,4 +231,24 @@ class BeanFactory
 
         return true;
     }
+
+    /*
+     * Clears a bean from cache so that it will be retrieved from DB next time
+     *
+     * @param $beanId
+     */
+    public static function unregisterBean($module, $id)
+    {
+        if (empty($id)) {
+            return false;
+        }
+        if (!isset(self::$loadedBeans[$module][$id])) {
+            return false;
+        }
+
+        unset(self::$loadedBeans[$module][$id]);
+        unset(self::$touched[$module][$id]);
+
+        return true;
+    }
 }
--- a/modules/AOW_Actions/actions/templateParser.php
+++ b/modules/AOW_Actions/actions/templateParser.php
@@ -36,6 +36,15 @@ class aowTemplateParser extends templateParser {
 			foreach($bean_arr as $bean_name => $bean_id) {
 
 				$focus = BeanFactory::getBean($bean_name, $bean_id);
+
+        if (!$focus->fetched_row) {
+
+            // We do not want the cached version for a newly created bean, as some data such as date fields and
+            // auto increment fields will only be correct after a retrieve operation
+            BeanFactory::unregisterBean($bean_name, $bean_id);
+            $focus = BeanFactory::getBean($bean_name, $bean_id);
+        }
+
 				$string = aowTemplateParser::parse_template_bean($string, strtolower($beanList[$bean_name]), $focus);
 
                 if($focus instanceof Person){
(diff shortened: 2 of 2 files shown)


Dillon-Brown:
Assessed

acoolov:
7.10.6 and the time is still being sent in UTC format when saving a case, task, project etc. When the same notification is sent through the scheduler it shows the correct time. Using the same email template and values.

PR #5913, opened 2018-05-22
fix a bug for product image overwritten by the other product image

alexwang2013:
When the image have the same filename, it would overwriteen the previous
image of another product. So added guid as a prefix to the filename to
differentiate the images with the same filename.


Dillon-Brown:
Assessed

PR #5957, opened 2018-05-30
Fix email detailview performance

daniel-samson:
## Description

These changes address some of the performance issues in the emails modules.

**Issue 1: Attachment Indicator**
List view attachments indicator makes 20 requests to the imap server to open each email. I have removed this column by default and updated inbound email to only run these calls when the column is in the list view.

**Issue 2: Sanitizing HTML**
Detail view calls SugarCleaner twice, one for each body (html, plain). Since we only show one body, I have updated Inbound email to only process the body we will be displaying.

**Issue 3: Too many calls to get Email Headers**
The IMAP functions to get the headers sends too many commands to the imap server. I have reduced down to one call.

**Also:**
I have removed some PHP notices

## Motivation and Context

- up to 50% performance Improvent

## How To Test This

- Set up  emails
- List emails
- View non imported email
- import email

you should notice a speed improvement.

## Types of changes

### Final checklist

### DIFF
--- a/include/formbase.php
+++ b/include/formbase.php
@@ -211,7 +211,7 @@ function getGetToForm($ignore='', $usePostAsAuthority = false)
             }
 	        continue;
         }
-f
+
         if (is_array($value)) {
             if(!empty($value)) {
                 $log->warn('$value must be a string');
--- a/include/formbase.php
+++ b/include/formbase.php
@@ -201,9 +201,23 @@ function getPostToForm($ignore='', $isRegularExpression=false)
 
 function getGetToForm($ignore='', $usePostAsAuthority = false)
 {
+    global $log;
 	$fields = '';
-	foreach ($_GET as $key=>$value)
+	foreach ($_GET as $key => $value)
 	{
+	    if(is_array($key)) {
+            if(!empty($key)) {
+                $log->warn('$key must be an string');
+            }
+	        continue;
+        }
+
+        if (is_array($value)) {
+            if(!empty($value)) {
+                $log->warn('$value must be an string');
+            }
+	        continue;
+        }
 		if($key != $ignore){
 			if(!$usePostAsAuthority || !isset($_POST[$key])){
 				$fields.= "<input type='hidden' name='$key' value='$value'>";
--- a/include/formbase.php
+++ b/include/formbase.php
@@ -207,14 +207,14 @@ function getGetToForm($ignore='', $usePostAsAuthority = false)
 	{
 	    if(is_array($key)) {
             if(!empty($key)) {
-                $log->warn('$key must be an string');
+                $log->warn('$key must be a string');
             }
 	        continue;
         }
-
+f
         if (is_array($value)) {
             if(!empty($value)) {
-                $log->warn('$value must be an string');
+                $log->warn('$value must be a string');
             }
 	        continue;
         }
--- a/include/formbase.php
+++ b/include/formbase.php
@@ -205,14 +205,14 @@ function getGetToForm($ignore='', $usePostAsAuthority = false)
 	$fields = '';
 	foreach ($_GET as $key => $value)
 	{
-	    if(is_array($key)) {
+	    if(!is_string($key) || !is_int($key)) {
             if(!empty($key)) {
-                $log->warn('$key must be a string');
+                $log->warn('$key must be a string or an integer');
             }
 	        continue;
         }
 
-        if (is_array($value)) {
+        if (!is_string($value)) {
             if(!empty($value)) {
                 $log->warn('$value must be a string');
             }
--- a/modules/Emails/include/DetailView/EmailsNonImportedDetailView.php
+++ b/modules/Emails/include/DetailView/EmailsNonImportedDetailView.php
@@ -72,7 +72,7 @@ class EmailsNonImportedDetailView extends EmailsDetailView
      */
     public function populateBean($request = array())
     {
-        global $log;
+        $log = LoggerManager::getLogger();
         if (!empty($request['uid'])
             && !empty($request['inbound_email_record'])
             && !empty($request['msgno'])
--- a/modules/Emails/include/DetailView/EmailsNonImportedDetailView.php
+++ b/modules/Emails/include/DetailView/EmailsNonImportedDetailView.php
@@ -72,15 +72,26 @@ class EmailsNonImportedDetailView extends EmailsDetailView
      */
     public function populateBean($request = array())
     {
-        if (!empty($request['uid']) && !empty($request['inbound_email_record'])&& !empty($request['msgno'])) {
+        global $log;
+        if (!empty($request['uid'])
+            && !empty($request['inbound_email_record'])
+            && !empty($request['msgno'])
+        ) {
             /** @var InboundEmail $inboundEmail **/
             $inboundEmail = BeanFactory::getBean('InboundEmail', $request['inbound_email_record']);
+
+            if (
+                $inboundEmail === false
+                || empty($inboundEmail->id)
+            ) {
+                $log->fatal('Unable to find inbound email account ' . $request['inbound_email_record']);
+            }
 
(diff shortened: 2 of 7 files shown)


include/formbase.php
```diff
@@ -201,9 +201,23 @@ function getPostToForm($ignore='', $isRegularExpression=false)
 
 function getGetToForm($ignore='', $usePostAsAuthority = false)
 {
+    global $log;
 	$fields = '';
-	foreach ($_GET as $key=>$value)
+	foreach ($_GET as $key => $value)
 	{
+	    if(is_array($key)) {
+            if(!empty($key)) {
```
zoltankocsardi: You don't need to use !empty if you just want to warn the user $key must be a string.

include/formbase.php
```diff
@@ -201,9 +201,23 @@ function getPostToForm($ignore='', $isRegularExpression=false)
...
+                $log->warn('$key must be an string');
```
zoltankocsardi: an string type o

modules/Emails/include/DetailView/EmailsNonImportedDetailView.php
```diff
@@ -73,9 +73,10 @@ public function setup(
     public function populateBean($request = array())
     {
         if (!empty($request['uid']) && !empty($request['inbound_email_record'])&& !empty($request['msgno'])) {
+            /** @var InboundEmail $inboundEmail **/
             $inboundEmail = BeanFactory::getBean('InboundEmail', $request['inbound_email_record']);
```
zoltankocsardi: This could be return false.

modules/Emails/views/view.detailnonimported.php
```diff
@@ -86,4 +90,34 @@ public function display()
...
+
+    private function getFieldsInViewDefinitions($metadataFile)
+    {
+        require_once $metadataFile;
+        $fields_in_definition = array();
+        $module_name = 'Emails';
+        if (
+            isset($viewdefs)
+            && isset($viewdefs['Emails']['DetailView'])
+            && isset($viewdefs['Emails']['DetailView']['panels'])
```
zoltankocsardi: The first 2 issets are not necessary.

modules/InboundEmail/InboundEmail.php
```diff
@@ -5401,114 +5403,138 @@ public function returnImportedEmail($msgNo, $uid, $forDisplay = false, $clean_em
 
     /**
      * Used to view non imported emails
-     * @param string $msgNo - imap mesgno
-     * @param string $uid - imap uid
+     * @param string $request - must include the metadata
```
zoltankocsardi: The $request is def not a string.

modules/InboundEmail/InboundEmail.php
```diff
@@ -5401,114 +5403,138 @@ public function returnImportedEmail($msgNo, $uid, $forDisplay = false, $clean_em
...
-            $structure = imap_fetchstructure($this->conn, $uid, FT_UID); // map of email
-            $email->name = $this->handleMimeHeaderDecode($header[0]->subject);
+            $parsedFullHeader = imap_rfc822_parse_headers($fullHeader);
+
+            $email->name = $this->handleMimeHeaderDecode($parsedFullHeader->subject);
             $email->type = 'inbound';
 
+            if(
+                isset($request['metadata'])
+                && isset($request['metadata']['viewdefs'])
```
zoltankocsardi: Same here like above.

modules/InboundEmail/InboundEmail.php
```diff
@@ -5401,114 +5403,138 @@ public function returnImportedEmail($msgNo, $uid, $forDisplay = false, $clean_em
...
+                        str_replace(['D, ', ':s'], '', \DateTime::RFC2822), // day-of-week is optional, seconds are optional
+                        \DateTime::RFC822,
+                        str_replace(['D, '], '', \DateTime::RFC822), // day is optional
+                        str_replace([':s'], '', \DateTime::RFC822), // seconds are optional
+                        str_replace(['D, ', ':s'], '', \DateTime::RFC822), // day is optional, seconds are optional
+                    ];
+
+                    foreach ($possibleFormats  as $possibleFormat) {
+                        $dateTime = \DateTime::createFromFormat( $possibleFormat, $parsedFullHeader->date);
+                        if ($dateTime !== false) {
```
zoltankocsardi: I don't understand this. If you can't parse the date, you just break the iteration?

PedroErnst: If $dateTime is a valid DateTime object we have successfully parsed a string. No need to iterate over further possible formats.

include/formbase.php
daniel-samson: This code is to remove the PHP notice array conversion to string. There is an instance where an empty array is being passed into the $value. If the array is empty, it doesn't really break anything per se, so we can just ignore it. However if a developer is deliberate passing in an array with values in it. We should inform them that this behavior is incorrect.  Then they can decide what are the next steps they need to take. Since this is code needs to be backwards compatible, we cannot use exceptions. So the alternative is to  write it into the suitecrm.log.

include/formbase.php
daniel-samson: OK. Fixed it. :D

modules/Emails/include/DetailView/EmailsNonImportedDetailView.php
daniel-samson: It will only return false it we remove the InboundEmail module from SuiteCRM. The InboundEmail module will still be in SuiteCRM for quite some time.  I will put a check if the bean or it record is missing, and log a warning to the developer.

modules/Emails/views/view.detailnonimported.php
daniel-samson: OK. Fixed

modules/InboundEmail/InboundEmail.php
daniel-samson: OK. Updated to only have on isset

modules/Emails/include/DetailView/EmailsNonImportedDetailView.php
zoltankocsardi: Yeah, and when it might be removed, you will be surprised. You can't be sure it could be renamed either. At least use InboundEmail::class, please.

modules/InboundEmail/InboundEmail.php
daniel-samson: Some IMAP server respond with different data formats. The iteration attempt to use each possible format to decode the detail. The if ($dateTime !== false) means that when the DateTime class successfully decodes the date field it will exit the loop. As we no longer need to continue trying to decode the datetime format.

I have added a comment to help explain this better.

zoltankocsardi: Unnecessary, was my mistake.

modules/Emails/include/DetailView/EmailsNonImportedDetailView.php
gymad:
```
if (!$inboundEmail) {
    LoggerManager::getLogger()->warn('Hoops! Something went wrong, when non imported email detail view trying to populate bean and requested inbound email record id was: ' . $request['inbound_email_record']);
}
````

File: include/formbase.php [hunk 8]
```diff
@@ -201,9 +201,23 @@ function getPostToForm($ignore='', $isRegularExpression=false)
...
+	    if(is_array($key)) {
+            if(!empty($key)) {
+                $log->warn('$key must be a string');
+            }
+	        continue;
+        }
+
+        if (is_array($value)) {
+            if(!empty($value)) {
+                $log->warn('$value must be a string');
```
gymad: I don't really understand this log message. the fact when a $value is a non-empty array is not implicated it to a string. any thing else also could not being a string...

gymad: could you use simply `is_string` function OR change the log message to  what is the problem..

Dillon-Brown:
Assessed

PR #5959, opened 2018-05-30
Feature/composer html purifier

daniel-samson:
## Description

This change moves the third party library (Html Purifier) to composer. I have left the old files intact for for backwards compatibility.

I am using the latest version of Html Purifier.

I have remapped the existing classes to new PSR-4 complaint namespace (SuiteCRM). This allows existing code to continue using the existing classes while new code can leverage the autoloader.

## Motivation and Context

- This adds PHP 7.2 compatibility fixes, which removes the deprecated notices.
- Allows us to use composer to mange this dependancy

## How To Test This

- Use modules which have html fields, eg emails, cases, email templates.

## Types of changes

### Final checklist

### DIFF
--- a/include/HTMLPurifierFilterXmp.php
+++ b/include/HTMLPurifierFilterXmp.php
@@ -39,7 +39,7 @@
  */
 namespace SuiteCRM;
 
-class HTMLPurifier_Filter_Xmp extends \HTMLPurifier_Filter
+class HTMLPurifierFilterXmp extends \HTMLPurifier_Filter
 {
 
     public $name = 'Xmp';
--- a/include/HTMLPurifierFilterXmp.php
+++ b/include/HTMLPurifierFilterXmp.php
@@ -39,11 +39,22 @@
  */
 namespace SuiteCRM;
 
+/**
+ * Class HTMLPurifierFilterXmp
+ * @package SuiteCRM
+ */
 class HTMLPurifierFilterXmp extends \HTMLPurifier_Filter
 {
 
+    /** @var string $name */
     public $name = 'Xmp';
 
+    /**
+     * @param string $html
+     * @param \HTMLPurifier_Config $config
+     * @param \HTMLPurifier_Context $context
+     * @return null|string|string[]
+     */
     public function preFilter($html, $config, $context)
     {
         return preg_replace("#<(/)?xmp>#i", "<\\1pre>", $html);
--- a/include/HTMLPurifierURISchemeCid.php
+++ b/include/HTMLPurifierURISchemeCid.php
@@ -42,7 +42,7 @@ namespace SuiteCRM;
 /**
  * cid: scheme implementation
  */
-class HTMLPurifier_URIScheme_cid extends \HTMLPurifier_URIScheme
+class HTMLPurifierURISchemeCid extends \HTMLPurifier_URIScheme
 {
     public $browsable = true;
     public $may_omit_host = true;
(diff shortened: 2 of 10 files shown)


Dillon-Brown:
Assessed

include/HTMLPurifierFilterXmp.php
```diff
@@ -0,0 +1,62 @@
...
+    /** @var string $name */
+    public $name = 'Xmp';
+
+    /**
+     * @param string $html
+     * @param \HTMLPurifier_Config $config
+     * @param \HTMLPurifier_Context $context
+     * @return null|string|string[]
+     */
+    public function preFilter($html, $config, $context)
```
zoltankocsardi: 2nd and 3rd params are unused.

include/HTMLPurifierURISchemeCid.php
```diff
@@ -0,0 +1,70 @@
...
+    /** @var bool $may_omit_host */
+    public $may_omit_host = true;
+
+    /**
+     * @param \HTMLPurifier_URI $uri
+     * @param \HTMLPurifier_Config $config
+     * @param \HTMLPurifier_Context $context
+     * @return bool
+     */
+    public function doValidate(&$uri, $config, $context)
```
zoltankocsardi: Same here.

include/HtmlSanitizer.php
```diff
@@ -0,0 +1,142 @@
...
+        $this->purifier = new \HTMLPurifier($config);
+    }
+
+    /**
+     * Get cleaner instance
+     * @return HtmlSanitizer
+     */
+    public static function getInstance()
+    {
+        return self::$instance instanceof self ? self::$instance : (self::$instance = new self());
```
zoltankocsardi: if (!self::$instance) {
// populate
}

return self::$instance;

include/HTMLPurifierFilterXmp.php
daniel-samson: The signature must match the parent class. I cannot remove them.

include/HTMLPurifierURISchemeCid.php
daniel-samson: Again the same.

include/HtmlSanitizer.php
daniel-samson: Yes - getInstance is a singleton pattern, I will change this correctly match this pattern. :D

include/HTMLPurifierFilterXmp.php
zoltankocsardi: Actually it was a hint for changing the parent class, since it's look like this.

`
class HTMLPurifier_Filter
{

    /**
     * Name of the filter for identification purposes.
     * @type string
     */
    public $name;

    /**
     * Pre-processor function, handles HTML before HTML Purifier
     * @param string $html
     * @param HTMLPurifier_Config $config
     * @param HTMLPurifier_Context $context
     * @return string
     */
    public function preFilter($html, $config, $context)
    {
        return $html;
    }

    /**
     * Post-processor function, handles HTML after HTML Purifier
     * @param string $html
     * @param HTMLPurifier_Config $config
     * @param HTMLPurifier_Context $context
     * @return string
     */
    public function postFilter($html, $config, $context)
    {
        return $html;
    }
}
`

samus-aran: But we shouldn't be changing the parent class.
If you feel this should be applied then make a PR to the master project.

Dillon-Brown:
Assessed

include/HTMLPurifierURISchemeCid.php
samus-aran: Feel free to open a PR
https://github.com/ezyang/htmlpurifier/issues/182

include/HTMLPurifierFilterXmp.php
zoltankocsardi: You can customize it, meant for "changing", but fine.

Issue #5853, opened 2018-05-10
Email Settings & Fail send email


ijdavie:
#### Issue

It appears that if a SMTP server does not use a email address as the username, the system is able to send a test email via the test option, however once saved and a user tried to send a email via the emails module it will fail with the following error message. 

SugarPHPMailer encountered an error: Invalid address:  (punyEncode) <Username>

From what can be seen the username is trying to be validated as a email address. 
#### Expected Behavior

services like sendgrid should allow sending of email addresses. 

#### Actual Behavior

Email not sent error logged in suitecrm.log file. 

#### Steps to Reproduce

1. setup sendgrid
2. use as email credentials 
3. try and send a email via emails module. 
4.

#### Your Environment

* SuiteCRM Version used: 7.10.2
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): chrome
* Environment name and version (e.g. MySQL, PHP 7): php 7.2
* Operating System and version (e.g Ubuntu 16.04): ubuntu

pgorod:
@ijdavie where exactly did you configure the SMTP Server? If it was in the User profile, please tell me also which of the two screens shown here #5376 you used. Thanks!

ijdavie:
@pgorod It was within the Email settings within the admin panel. (The test settings worked) however when trying to send via the emails module it failed.

samus-aran:
@ijdavie The fix provided we have done our own manual test and can send using a username with a system outbound account if we have a personal inbound linked to it.

Let us know if this isn't the case.

salesagility-davidthomson:
I can confirm that  isleshocky77  https://github.com/salesagility/SuiteCRM/issues/5908 does indeed happen when installed on 7.10.2 as tested by me. 
The email settings disappear and do not display even after reentering them.
It is possible however to continue using the settings when sending mail as this continues to work.
I also have to point out that after merging this fix you also need to rebuild using composer.

LionelBino:
This happen too in 7.9.17

Dillon-Brown:
Hi guys, taking a look at hopefully getting the related PR #6017 merged in. Would anyone be able to assist with testing? just a simple confirmation that it works for you would be greatly appreciated.

PR #5899, opened 2018-05-18
Hotfix 5853 email settings fails

gymad:
## Description

- possible fix for issue #5853
- fixing blank screen issue after switch on display errors in admin
- fixing a ton of php errors in email module

## Motivation and Context

see at issue issue #5853

## How To Test This

see at issue issue #5853

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [x] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

PR #5960, opened 2018-05-30
Hotfix for bug #5853

LionelBino:
Hotfix for bug #5853

This bugfix fix to send mail in the mails module using sendgrid service.

## Description
For an apropiate description of this bug please read [the bug #5853 ](https://github.com/salesagility/SuiteCRM/issues/5853)

## How To Test This
Set up a sendgrid account in inbound emails module and use this.

## Types of changes

### Final checklist

### DIFF
--- a/modules/Emails/EmailsController.php
+++ b/modules/Emails/EmailsController.php
@@ -446,23 +446,45 @@ class EmailsController extends SugarController
         $oe = new OutboundEmail();
         if ($oe->isAllowUserAccessToSystemDefaultOutbound()) {
             $system = $oe->getSystemMailerSettings();
-            $data[] = array(
-                'type' => 'system',
-                'id' => $system->id,
-                'attributes' => array(
-                    'from' => $system->mail_smtpuser,
-                    'name' => $system->name,
-                    'oe' => $system->mail_smtpuser,
-                ),
-                'prepend' => false,
-                'isPersonalEmailAccount' => false,
-                'isGroupEmailAccount' => true,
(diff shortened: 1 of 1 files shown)


Dillon-Brown:
@LionelBino Have you tried #5899? with the email settings bug now resolved that PR seems to solve the issue for me.

LionelBino:
@Dillon-Brown  I have not been able to test hotfix 5853 because it modifies many files ..

modules/Emails/EmailsController.php
```diff
@@ -446,23 +446,45 @@ public function action_GetFromFields()
...
-                    'from' => $system->mail_smtpuser,
-                    'name' => $system->name,
-                    'oe' => $system->mail_smtpuser,
-                ),
-                'prepend' => false,
-                'isPersonalEmailAccount' => false,
-                'isGroupEmailAccount' => true,
-                'outboundEmail' => array(
+
+            if($system->mail_smtpserver != 'smtp.sendgrid.net')
```
PedroErnst: The code seems to work fine, but I think might be done a bit cleaner. Mostly everything in this array is the same, instead of duplicating all this code consider declaring the entire array first, then modifying those keys that should have different values depending on the if statement.

Like:
```
$array = [
    'foo' => 1;
    ...
];

if ($sendgrid == 'bar') {
    $array['foo'] = 'baz';
}
```

I think it will result in a cleaner (and shorter) function.

LionelBino: Yes this is a better approach.

PR #5985, opened 2018-06-04
Fix unified search "no results" page

javitoron:
## Description
The condition if $hits is moved before the table tag and the else after the closing table tag.
The row and cell tags containing the No results messages are replaced by a paragraph tag.

## Motivation and Context
When you get no results in Unified Search you see a "broken" page with a useless table.
Results table should be shown only when there are, actually, results.
If no results are returned it should show just a message.

Now you get something like this (taken from Demo website):
image

After the fix you get this (taken from my SuiteCRM installation):
image

## How To Test This
Make a global search with PEPE. You get an "ugly" no results page.

## Types of changes
- [ X] Bug fix (non-breaking change which fixes an issue)

### Final checklist
- [X ] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).


cameronblaikie:
Assessed :+1:

Issue #6003, opened 2018-06-08
Line items fail to correctly round  0.999 values


LEAP-nishit:
#### Issue
The `function formatNumber`  does not correctly handle the round of numbers that causes an increase in the integral part. 

#### Expected Behavior
Round the value correctly. 

#### Actual Behavior
When passed a value say `475.99999999999994` and round=2 and precision=2  gives `475` instead of `476`. 

#### Steps to Reproduce
1. Have a price that causes the value to give you 0.9999999
2. Round up to 2 digits and precision=2

#### Context
Incorrect calculation of price when code depends on `function formatNumber`  . 

#### Your Environment

* SuiteCRM Version used: 7.10.4
* Browser: Firefox 59.0.2 and Chrome 65.0.3325.181
* Environment name and version (e.g. MySQL, PHP 7): 
    * MariaDB 10.0.34
    * php: PHP 7.0.28
* Operating System and version (e.g Ubuntu 16.04): Ubuntu 16.04.4 LTS

PR #6004, opened 2018-06-08
Fix round up for cases where there is an increase in integral part. 

LEAP-nishit:
The `function formatNumber`  does not correctly handle the round of numbers that causes an increase in the integral part. 

## Description

The function formatNumber does not correctly handle the round of numbers that causes an increase in the integral part. Reported as issue #6003. 

## Motivation and Context
The function should round up values correctly. 

## How To Test This
Apply changes and test in browser console giving number 475.99999999999994, round=2 and precision=2.

## Types of changes

### Final checklist


daniel-samson:
Assessed

daniel-samson:
Hi @LEAP-nishit ,

I think it would be a really good idea if we made an acceptance test to check for this rounding issue. That we we could ensure it the bug never occurs again.

Are you able and willing to write an acceptance test?

We have  a guide on how to set up acceptance test suite https://docs.suitecrm.com/developer/appendix-c---automated-testing/ 

If you need any assistance, I would be happy to help. :D

Issue #5977, opened 2018-06-01
Minor Feature: Set message_id for outgoing email


ApatheticCosmos:
While working on a related project, I realized that outgoing email doesn't have a message_id set in the database.

#### Issue

I'm working on a project to assign incoming mail that are replies to an outgoing email to their respective record, no matter who they are received from. In order to do this, I'll be checking the References: and In-Reply_To: email headers, and matching them against sent email in the DB.

Then I realized that outgoing emails don't have their Message-Id recorded in the database, so this would not be possible.

#### Possible Fix

*PULL REQUEST INCOMING*
Set the message id before sending the message.

#### Steps to Reproduce

1. Send an email from SuiteCRM
2. Check the 'emails' table in the DB and notice that the message_id field is NULL for the email.

#### Your Environment

* SuiteCRM Version used: 7.10.5 (with patch from #5935 applied)
* Browser name and version: Chrome Version 66.0.3359.181 (Official Build) (64-bit)
* Environment name and version: MariaDB (Galera Cluster) v10.1.29-6, Apache 2.4.18-2ubuntu3.8, PHP 7.0.30-0ubuntu0.16.04.1
* Operating System and version: Ubuntu 16.04

Mausino:
@Dillon-Brown  should the issue label by milestone 7.11 ?
This is very usefull feature which save lots of time 👍

Mausino:
@Dillon-Brown @samus-aran  should the issue label by milestone 7.11 or 7.10.12 ?
For us is very usefull feature which save lots of time in future

PR #6009, opened 2018-06-09
#6024 New Feature - Relate/Assign Incoming Email 

ApatheticCosmos:
Change how IncomingEmail handles assigning and relating email retrieved from group mail account.
See the thread at https://suitecrm.com/suitecrm/forum/suitecrm-7-0-discussion/19458-developing-incoming-email-features-request-for-comment for more info and a screenshot of the configuration option.
Issue #6024

## Description
With this patch, all email replies retrieved from a Group Mail account is related with the same record as referenced sent email.
It also provides the option to assign the email to either the owner of the referenced email, or to the owner of the record related with the referenced email.
It sets the message_id on all outgoing emails. See Issue #5977.

## Motivation and Context
Currently, SuiteCRM has no idea what record or user a group incoming email is related with, or is in reply to. The only information is has to go on is the from email address, which must match an email address in a record to even show up there. It currently can't set the parent record of the incoming email properly.
Also, the only way to change ownership of an incoming group email is with workflows, or manually. With this patch, incoming email can be automatically assigned to the user that owns the parent record of the thread, or the user that sent the referenced email.

## How To Test This
When you apply this patch, all outgoing email will get the mesaage_id set in the emails table.
Any email that is retrieved by a group account and references that message_id will be related to the record of the original email. You can also tell it to assign to the owner of that record, or the owner of the referenced email, or do nothing different than before.

## Types of changes

### Final checklist

    (I think it does, although it's documented in the help tip)


modules/InboundEmail/InboundEmail.php
```diff
@@ -5302,7 +5302,39 @@ public function returnImportedEmail($msgNo, $uid, $forDisplay = false, $clean_em
...
+                            $row = $this->db->fetchByAssoc($results);
+                            if(!empty($row)) {
+                                $email->parent_id = $row['parent_id']; // set email parent to parent of referenced email
+                                $email->parent_type = $row['parent_type'];
+                                switch ($this->assignment_behavior) {
+                                    case 'donothing'; // Don't assign the email to a user
+                                        break;
+                                    case 'recordowner'; // Assign the email to the owner of the record the referenced email's parrent
+                                        $parentBean = BeanFactory::getBean($row['parent_type'], $row['parent_id']);
+                                        $email->assigned_user_id = $parentBean->assigned_user_id;
```
PedroErnst: Small check here to make sure $parentBean isn't null could help prevent PHP notices.

PedroErnst (dismissed):
Very nice feature to have!

modules/InboundEmail/InboundEmail.php
ApatheticCosmos: Done!

labcerouno:
After applying this fix and configuring a new mail account, I get the following FATAL ERROR:

MySQL error 1054: Unknown column 'assignment_behavior' in 'field list'

It's quite obvious the reason. So I tried adding this field in the InboundEmail table and it seems to be working fine!

gymad (commented):
I think as it is a new feature, it should be going to the `develop` branch. Can you please change the target branch, please

gymad (approved):

Mausino:
@gymad should the issue to which is this pull request assigned label by milestone 7.11 ? 
This is very usefull feature which save lots of time

ApatheticCosmos:
Can we get the 'wrong branch' tag removed from this please?

samus-aran:
Thanks @ApatheticCosmos We are circulating this internally for review :+1:

modules/Emails/EmailsController.php
```diff
@@ -171,9 +171,13 @@ public function action_send()
     {
         global $current_user;
         global $app_strings;
+        global $sugar_config;
 
         $request = $_REQUEST;
 
+        // Set message_id for email record in DB
+        $request['message_id'] = 'SUITECRM_' . md5('HELLO'.(idate("U")-1000000000).uniqid()) . '@' . $sugar_config['host_name'];
```
gymad: Calling uniqid() without any parameter is a bit dangerous ans not a best option as it's related on a system time clock.
see more in php.net: 
> Warning
This function does not guarantee uniqueness of return value. Since most systems adjust system clock by NTP or like, system time is changed constantly. Therefore, it is possible that this function does not return unique ID for the process/thread. Use more_entropy to increase likelihood of uniqueness.
http://php.net/manual/en/function.uniqid.php#120123

modules/InboundEmail/InboundEmail.php
```diff
@@ -5302,7 +5302,41 @@ public function returnImportedEmail($msgNo, $uid, $forDisplay = false, $clean_em
                     return false;
                 }
                 $email->$rel->add($email->parent_id);
-            }
+            } else {
```
gymad: This function complexity is getting hardly, almost unreadable. Please split out into separated functions.

modules/InboundEmail/InboundEmail.php
```diff
@@ -5302,7 +5302,41 @@ public function returnImportedEmail($msgNo, $uid, $forDisplay = false, $clean_em
...
+                //Assign Parent Values if references header mentions a sent email in the system
+                if(!empty($header->references) || !empty($header->in_reply_to)) {
+                    $references = explode(" ",  $header->in_reply_to . " " . $header->references);
+                    foreach ($references as $reference) {
+                        $referenceId = rtrim(ltrim($reference, '<'), '>');
+                        if (strncmp($referenceId, "SUITECRM_", 9) === 0) {
+                            $query = 'SELECT id, assigned_user_id, parent_id, parent_type FROM emails WHERE emails.message_id = \'' . $referenceId . '\' and type = \'out\' and emails.deleted = 0 LIMIT 1';
```
gymad: Use quote() or quoted() on query parameters.

modules/InboundEmail/InboundEmail.php
```diff
@@ -5302,7 +5302,41 @@ public function returnImportedEmail($msgNo, $uid, $forDisplay = false, $clean_em
...
+                            $row = $this->db->fetchByAssoc($results);
+                            if(!empty($row)) {
+                                $email->parent_id = $row['parent_id']; // set email parent to parent of referenced email
+                                $email->parent_type = $row['parent_type'];
+                                switch ($this->assignment_behavior) {
+                                    case 'donothing'; // Don't assign the email to a user
+                                        break;
+                                    case 'recordowner'; // Assign the email to the owner of the record the referenced email's parrent
+                                        $parentBean = BeanFactory::getBean($row['parent_type'], $row['parent_id']);
+                                        if (isset($parentBean->id)) {
```
gymad: Testing for $parentBean->id but using $parentBean->assigned_user_id. Is it intend?

modules/InboundEmail/InboundEmail.php
```diff
@@ -5302,7 +5302,41 @@ public function returnImportedEmail($msgNo, $uid, $forDisplay = false, $clean_em
...
+                                            $email->assigned_user_id = $parentBean->assigned_user_id;
+                                        }
+                                        break;
+                                    case 'repliedetoowner'; // Assign the email to the owner of the referenced email
```
gymad: Mispelled. Did you test it before made this PR?

modules/InboundEmail/InboundEmail.php
```diff
@@ -5302,7 +5302,41 @@ public function returnImportedEmail($msgNo, $uid, $forDisplay = false, $clean_em
...
+                                        $email->assigned_user_id = $row['assigned_user_id'];
+                                        break;
+                                    default;
+                                        $GLOBALS['log']->error('Inbound Email misconfiguration. Assignment Behavior set to invalid value. Not assigning email.');
```
gymad: Use LoggerManager instead superglobals.

modules/InboundEmail/InboundEmail.php
```diff
@@ -5302,7 +5302,41 @@ public function returnImportedEmail($msgNo, $uid, $forDisplay = false, $clean_em
...
+                                        break;
+                                } // END switch ($this->assignment_behavior)
+                                break;
+                            } // END if(!empty($row))
+                        } // END if (strncmp($referenceId, "SUITECRM_", 9) === 0)
+                    } // END foreach ($references as $reference)
+                } // END if(!empty($header->references) || !empty($header->in_reply_to))
+            } // END else
```
gymad: Tons of IF but no ELSE. If I will call this function how can I determ what happend in this area?

gymad (changes requested):

modules/Emails/EmailsController.php
laszlocsontos: I'm trying to implement open and click tracking with this upcoming feature.

My use case is that I'm using Mailgun that provides callbacks when these emails are opened. Mailgun refers back to emails which were acted upon by their `Message-Id`. If the hostname isn't part of that value which is saved to the DB, an additional processing step is needed to split that value by `@` and do the lookup based on the first part.

From this perspective, it would be more straightforward, if the value of the `message_id` field in the DB was exactly the same as the value of the `Message-Id` field in sent emails.

Currently I'm using this in my custom SuiteCRM installation:

```
$request['message_id'] = '<' . time().'.'.bin2hex(random_bytes(8)).'@'.gethostname().'>';
```

This produces a value, something like `<1549445373.02231e5c88e1a743@typhoon>`. 

What do you think @ApatheticCosmos @gymad?

Mac-Rae:

ApatheticCosmos:
@Mac-Rae Conflict Resolved. I'm going to be doing some more work to clean up the things @gymad has pointed out as well.

serhiisamko091184:
Hello, @ApatheticCosmos 

The PR has been marked as stale because there has been no recent activity. It will be closed if no further activity occurs.
Thanks for your contributions.

Regards,
Serhii

serhiisamko091184:
Hey @ApatheticCosmos 

Thank you.

Regards,
Serhii

PR #6035, opened 2018-06-14
Show only if user has rights to activity/history record

eggsurplus:
On History->View Summary a user with restricted rights is able to see all history items including records they do not have access to.

## Motivation and Context
To ensure that users can only see records that they should have rights to.

## How To Test This
Create a call and assign a user and/or group that with the appropriate roles another user should not have access to it. Then sign in as that other user. That user should not see the call if roles are set to restrict for that specific scenario.

## Types of changes

### Final checklist


eggsurplus:
Edited as requested.

Dillon-Brown:
Assessed

PR #6039, opened 2018-06-15
Fix #5526 Inline Edit not showing updated date time

### LINKED ISSUE #5526: Inline Edit doesn't show new value for date and relate fields
#### Issue
The Inline Edit functionality both on Detail and List Views saves but doesn't show the new value.
Just with date or relate fields.
You have to refresh the page to see result.
1.
screen shot 03-17-18 at 10 36 am
2.
screen shot 03-17-18 at 10 37 am

#### Expected Behavior
To see what i edit inline without the need of refresh, like for other field types.

#### Possible Fix
It worked well at least until v. 7.8.5
This is the error found in Console:

POST https://demo.suiteondemand.com/index.php 500 (Internal Server Error)
sugar_grp1_jquery.js?v=jxk3SiR7fMJxSgg3yIZVzA:4

[Violation] 'keypress' handler took 170ms
sugar_grp1_jquery.js?v=jxk3SiR7fMJxSgg3yIZVzA:3 

#### Steps to Reproduce

Go to the SuiteCRM Demo,
in Activities > Meetings 
in list view try to edit a date start field or a related to field
your new value will be saved but not showed until you refresh the page.
Same on Detail View.

#### Context
SuiteCRM 7.8.15, 7.8.16, 7.8.17, 7.10.2

#### Your Environment
Chrome and Firefox browsers

gitbnw:
## Description

When using Inline Edit on a date or date time field the updated date does not display - the input goes blank.  The updated date only shows on a refresh of the page.

The Inline Edit code was using a hard coded format for the date time which likely did not match the user's preference and causes it to return a "false" leading to this error:

Fatal error: Uncaught TypeError: Argument 1 passed to TimeDate::tzUser() must be an instance of DateTime, boolean given, called in /sugarcrm/include/InlineEditing/InlineEditing.php on line 442 and defined in sugarcrm/include/TimeDate.php on line 894

The correct format for the input was already determined in a previous variable.  If you apply that instead of the hard coded one - and disregard the time zone change - the inline edit works.  I don't think the time zone change is needed because the value is from user input i.e. they are inputting their local time already.

## Motivation and Context

A field going blank after user input is a bit confusing.

## How To Test This

Attempt to use linline edit on a date time field like Tasks and see if the date time appears correctly following the edit.  I additionally tried changing the user preference for date and time formatting and the change still worked.

## Types of changes

### Final checklist


Dillon-Brown:
Assessed

PR #6058, opened 2018-06-20
Fix-Error-Screen-PDF-LIST-SuiteCrm

RafaelRfs:
Fix-Error-Screen-PDF-LIST-SuiteCrm
This code fix the error when suitecrm has a lot of pdfs in the list tutorial: 1 - Add this code inside of detailviewdefs.php in module that u need to fix:

'includes' => array ( 0 => array ( 'file' => 'custom/scripts/fixerrorscreen.js', )

            ),
2 - Add fixerrorscreen.js in your custom/scripts

https://github.com/RafaelRfs/Fix-Error-Screen-PDF-LIST-SuiteCrm


pgorod:
@RafaelRfs I don't understand exactly what you're trying to do in terms of git branches, but you should probably just doing this on your own fork, where you can create a feature sub-branch. Then your PR into SuiteCRM would be directed at `hotfix` (for bugfixes) or `develop` (for new features) branches.

PR #5458, opened 2018-03-09
Fix #5446 make UserViewHelper.php code-customizable

pgorod:
## How To Test This
Check original issue #5446.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [X] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change to the documentation.
- [X] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

PR #6072, opened 2018-06-22
Make some PopupPickers a bit more code-customizable

pgorod:
## Motivation and Context
This comes from a Forum thread
https://suitecrm.com/suitecrm/forum/developer-help/19515-audit-module-is-not-working-in-custom-directory

The user there tried it and it seems to work.

In that case, he was customizing the "View Change log" in some modules, which seems a useful thing to be able to do.

## How To Test This
1. Create a custom copy of `modules/Audit/Popup_picker.php` in `custom/modules/Audit/Popup_picker.php` 
2. Edit that new file with some visible change
3. In some module, click "View Change log" on the Detail view, see if the custom file gets picked up.
4. Maybe also check whether other PopupPickers are unaffected and still load normally?

P.S - I have a couple more similar PR's waiting, if somebody wants to look at all of these together, maybe it's easier: #5458 and #3891

They all arise from real-life users needing to customize...

### DIFF
--- a/tests/unit/include/MVC/View/views/view.popupTest.php
+++ b/tests/unit/include/MVC/View/views/view.popupTest.php
@@ -49,9 +49,7 @@ class ViewPopupTest extends SuiteCRM\StateCheckerPHPUnitTestCaseAbstract
         }
     }
 
-    /** @test */
-
-    public function testdisplay_get_module_popup_picker_if_no_listviews_searchdefs()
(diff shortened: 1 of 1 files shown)


daniel-samson:
Assessed

chris001:
This bug fix needs to be re-added to hotfix/develop/master and be included in the next minor version release.  
It's not currently in the released code.  It seems to have been mistakenly reverted.
See here for proof: https://community.suitecrm.com/t/audit-module-is-not-working-in-custom-directory/59672/17

PR #6105, opened 2018-06-27
Show Audit log even if user does not exist (anymore)

gunnicom:
If a user is missing in the user table the audit log entries are not shown. But even if you do not know the user it is good to know there was a change.

### DIFF
--- a/modules/Audit/Audit.php
+++ b/modules/Audit/Audit.php
@@ -124,7 +124,7 @@ class Audit extends SugarBean {
 
         if($focus->is_AuditEnabled()){
             $order= ' order by '.$focus->get_audit_table_name().'.date_created desc' ;//order by contacts_audit.date_created desc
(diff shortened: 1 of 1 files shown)


gunnicom:
Do you mean testing with an automated test case, or manually. I have no clue how to do automated tests for Suite.

gunnicom:
Steps to reproduce:
- Change any field that is audited
- Check the audit log, the entry should show
- In your database change the "created_by" field of the audit entry to "123xxx"
- Check the audit log, the entry does not show anymore
After the fix:
- The entry will show with an empty user

Motivation for that change:
Although if you only do changes through SuiteCRMyou will not have these problems, there are cases where it is highly useful, to at least see that there were changes:

- You cleanup your user table from very old users and really delete all users marked as deleted
- Your data is imported from another system, but you do not want to import legacy users, but still keep the history of objects.
- Some external program changes values or adds audit log entries with a user that not exists in Suite

Dillon-Brown:
Assessed

Issue #6043, opened 2018-06-15
Feature - Sync with Google Calendar


ApatheticCosmos:
This is a project I'm currently working on. I want to be able to have users sync with their Google Calendar.
I'm creating this issue to gain feedback, and track my progress. If anyone else is working on a similar feature, please chime in so we don't duplicate the effort.

#### Issue

We use Google's Services, as I'm sure a lot of other SuiteCRM users do. Currently the only way to see your CRM calendar in Google Calendar is by subscribing to the iCal URL. The problem is that Google (in their infinite wisdom) has decided that they'll sync that iCal feed very rarely, and only if they feel like it. So, once you subscribe any changes made seem to never show up. It's also a view only option.

I'm working on building a two way calendar sync using Google's API PHP library, and integrating it into SuiteCRM.

Once the calendar sync is working, it can be built on to sync other Records with the user's Google Account, since the API covers most of Google's services.

#### Expected Behavior

When you create or modify an event in SuiteCRM, it should show up on your Google Calendar.

#### Actual Behavior

New calendar events you create are not refreshed once you subscribe to the iCal feed.

#### Possible Fix

Use the Google API PHP library to perform a calendar sync via a scheduled job.
I'm working on this.

#### Steps to Reproduce

1. Subscribe to your SuiteCRM iCal feed in Google Calendar
2. Notice that new events and changes never show up
3. No Profit

#### Context

All our users have a Google account. Keeping the calendar's in sync is currently not a feature of SuiteCRM (without an addon). This is annoying and causes the users to not keep their appointments.

#### Your Environment

* SuiteCRM Version used: 7.10.6
* Browser name and version: Chrome Version 67.0.3396.62 (Official Build) (64-bit)
* Environment name and version: MariaDB (Galera Cluster) v10.1.29-6, Apache 2.4.18-2ubuntu3.8, PHP 7.0.30-0ubuntu0.16.04.1
* Operating System and version: Ubuntu 16.04

acoolov:
That would be amazing. Will cases show up in the calendar as well?

ApatheticCosmos:
Theoretically, any record based on a date/event/etc could be synced... but I'm nowhere close to that yet. I'm still working on the authentication system.

gunnicom:
Does CalDAV help? I have a oneway sync from Sugar to Outlook with sabredav (well, it is not in good code condition, but it works ;) )
Feel free to ask if it can help you.

[...]

PR #6146, opened 2018-07-12
Feature - Google Calendar Sync - Issue #6043

ApatheticCosmos:
New feature. Synchronize SuiteCRM Meetings with Google Calendar. The sync is bidirectional.

## Description

This code introduces a new class, GoogleSync, which is used to synchronize a user's Meetings w/ their Google Calendar.
It includes a function that is run by the Scheduler every quarter hour (by default) that collects a list of all the users that have the feature configured and enabled, and sync's with a non-primary Google calendar called 'SuiteCRM'.
There are also user interface changes in both the Administration panel and User Profile to configure it.

## Motivation and Context

Currently the only built in way to access your calendar from Google is via ical subscription. This is a one way, read only method. Also, it doesn't update nearly often enough to be useful.
See #6043 for more info.

## How To Test This
I've created a pull request updating the official docs for this feature. salesagility/SuiteDocs#127
Attached is a PDF of those docs to follow.
[Google Credentials and Syncing __ Docs.pdf](https://github.com/salesagility/SuiteCRM/files/2189352/Google.Credentials.and.Syncing.__.Docs.pdf)

## Types of changes

### Final checklist

- [X] My change requires a change to the documentation. See salesagility/SuiteDocs#127


ApatheticCosmos:
I'm working on the code style (brackets) and phpDocumentor compatibility. Sorry, I fell into bad habits. It won't take long to fix up.

modules/Users/language/en_us.lang.php
```diff
@@ -679,4 +679,12 @@
     'ERR_USER_FACTOR_SMTP_REQUIRED' => 'SMTP server settings required first.',
     'ERR_USER_FACTOR_CHANGE_DISABLED' => 'Current user is not able to change two factor authentication settings.',
     'LNK_IMPORT_CAMPAIGNS' => 'Import Campaign',
+    // for Google Token/Synchronization settings
+    'LBL_GOOGLE_API_SETTINGS' => 'Google Account Synchronization',
+    'LBL_GOOGLE_API_TOKEN' => 'Google API Token',
+    'LBL_GOOGLE_API_TOKEN_HELP' => 'Use this to authorize SuiteCRM to access your Google Calendar. This is required for synchronization. Don\'t forget to check the \'Enable Calendar Sync\' checkbox when done.',
```
horus68: Instead of `\'Enable Calendar Sync\'` can't we have another sign to avoid escaping? as in `"Enable Calendar Sync"`. 
As a rule of thumb no code or characters that require escaping should be present. If you can't do that at all (as when using apostrophes) at least reduce the use of signs as markers

Language values should not require Escape sequences .
This to avoid bugs caused by translations that sometimes forget to escape things.
@Dillon-Brown Maybe placing this in SuiteCRM good practices?

ApatheticCosmos: @horus68  @Dillon-Brown 
It is in the coding standards:
"Use single and double quotes when appropriate. If you’re not evaluating anything in the string, use single quotes. There should be no need to ever escape quotes within a string."
I'll fix it.

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1230 @@
...
+    public function getAuthJson()
+    {
+        global $sugar_config;
+        if (!$authJson_local = json_decode(base64_decode($sugar_config['google_auth_json']), true))
+        {
+            // The authconfig json string is invalid json
+            $GLOBALS['log']->fatal(__FILE__ . ':' . __LINE__ . ' ' . __METHOD__ . ' - ' . 'Invalid AuthConfig JSON');
+            return false;
+        }
+        else {
```
daniel-samson: Minor Issue: PSR 2 - else must be on the same line as the closing brace '}'. Consider using a psr 2 auto formating tool.

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1230 @@
...
+
+        $this->workingUser->retrieve($id);
+
+        // Retrieve Access Token JSON from user preference
+        $accessToken = json_decode(base64_decode($this->workingUser->getPreference('GoogleApiToken', 'GoogleSync')), true);
+        if (!array_key_exists('access_token', $accessToken))
+        {
+            // The Token is invalid JSON or missing
+            $GLOBALS['log']->fatal(__FILE__ . ':' . __LINE__ . ' ' . __METHOD__ . ' - ' . 'Invalid or Missing AuthToken');
+            return false;
```
daniel-samson: Please update the PHP Doc Block. Return only has Google_Client. Please go through this file an update the doc blocks for methods which return multiple types

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1230 @@
...
+     *
+     * @param string $event_id The Google Event ID
+     *
+     * @return Meeting
+     */
+    public function getMeetingByEventId($event_id)
+    {
+
+        // We do it this way so we also get deleted meetings
+        $query = "SELECT id FROM meetings WHERE gsync_id = '" . $event_id . "'";
```
daniel-samson: Please consider using single quotes or alternatively remove the concatenation. 
`"SELECT id FROM meetings WHERE gsync_id = '{$event_id}'"`

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1230 @@
...
+            return $gEvent;
+        }
+    }
+
+    /**
+     * Get a google event by the event id
+     *
+     * @param string $event_id Google Event ID
+     *
+     * @return Google_Service_Calendar_Event|null Google_Service_Calendar_Event if found, null if not found
```
daniel-samson: This method returns void instead of null. Please update doc block to match.

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1230 @@
...
+        }
+
+        $private['suitecrm_id'] = $event->id;
+        $private['suitecrm_type'] = $event->module_name;
+
+        $extendedProperties->setPrivate($private);
+
+        $event_remote->setExtendedProperties($extendedProperties);
+
+        $greturn = $this->gService->events->update($this->calendarId, $event_remote->getId(), $event_remote);
```
daniel-samson: Variable appears to unused. Please consider remove the variable or adding a test an throwing and exception

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1230 @@
...
+        {
+            $tz = $this->workingUser->getPreference('timezone', 'global');
+            $this->setTimezone($tz);
+        }
+        else {
+            $GLOBALS['log']->fatal(__FILE__ . ':' . __LINE__ . ' ' . __METHOD__ . ' - ' . 'Failed to set the working user and timezone');
+            return false;
+        }
+
+        if (!$this->setGService($this->gClient))
```
daniel-samson: The signature of setGService does not take any arguments. Is this a bug? I think this line needs to be refactored.

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1230 @@
...
+            $reminders_remote->setOverrides($reminders_array);
+            $event_remote->setReminders($reminders_remote);
+        }
+        return $event_remote;
+    }
+
+    /**
+     * Create New Google Event object for SuiteCRM Meeting
+     *
+     * @param $event_local SuiteCRM Meeting Object
```
daniel-samson: Missing class type. I also noticed that event_local is defined as an object in some code blocks. Is the class type a Bean? Please update PHP Doc.

modules/Users/entryPointSaveGoogleApiKey.php
```diff
@@ -0,0 +1,58 @@
+<?php
+if(!defined('sugarEntry') || !sugarEntry) die('Not A Valid Entry Point');
```
daniel-samson: PSR-2 issue, Can you please use braces '{ }' for if statements.

include/GoogleSync/GoogleSync.php
ApatheticCosmos: https://docs.suitecrm.com/community/contributing-code/coding-standards/ should be updated to reflect this. It shows else on a new line.
I'll fix this and follow PSR-2.

daniel-samson: Thank You, We will update the documentation to follow this pattern.

include/GoogleSync/GoogleSync.php
ApatheticCosmos: It should return null. My understanding is that with php void and null are the same thing. Is that incorrect? In any case, I'm setting it explicitly to null so it's consistent.

include/GoogleSync/GoogleSync.php
ApatheticCosmos: Removing the concatenation

include/GoogleSync/GoogleSync.php
ApatheticCosmos: I missed this piece of cruft from older code. I was passing the GClient previously.

modules/Users/entryPointSaveGoogleApiKey.php
ApatheticCosmos: Fixing

include/GoogleSync/GoogleSync.php
ApatheticCosmos: Working on this.

include/GoogleSync/GoogleSync.php
ApatheticCosmos: I was using it for debugging, but I should be checking it. I'll add the check.

include/GoogleSync/GoogleSync.php
ApatheticCosmos: Done.

include/GoogleSync/GoogleSync.php
ApatheticCosmos: Check added.

zahidAbbasAliBaig:
Hi @ApatheticCosmos ,

Does your system only provide one-way sync CRM to GoogleCalendar? Because I tested it seems that way. Am I missing something? Please clarify.

Thanks.

ApatheticCosmos:
@zahidAbbasAliBaig 
It's a 2-way sync. Be sure you're adding the events you want it to pull to the 'SuiteCRM' calendar on Google. It doesn't sync with the users primary Google calendar... for now.

zahidAbbasAliBaig:
I added the event in the google calendar but it didn't sync to the CRM Calendar.

ApatheticCosmos:
@zahidAbbasAliBaig 
Just so we're clear, the event on Google Calendar must be in the 'SuiteCRM' calendar for it to be pulled.
When you click on the Google Event, it shows which calendar it's in like this:
image

zahidAbbasAliBaig:
Yes I know, I have done the same. I selected the suite CRM instead of my own.
Please see the test video 
https://www.dropbox.com/s/f6uf6issifa82xs/Calender%20Sync%20Test.webm?dl=0

ApatheticCosmos:
@zahidAbbasAliBaig 
Well... darn it.
I'll do some digging. Maybe something got missed or juxtaposed in the pull request.
Is there anything in the SuiteCRM log file?

zahidAbbasAliBaig:
Hi @ApatheticCosmos,

Have you found out something about the issue? Thanks for the earlier replies.

ApatheticCosmos:
@zahidAbbasAliBaig 
Not yet. My SysAdmin duties have kept me away from it. :(

zahidAbbasAliBaig:
No worries, please update when you have the time. Thanks for the reply.

daniel-samson:
Hi @ApatheticCosmos,

As part of our goal to improve code quality, we are encouraging SuiteCRM contributors to help write tests. 

Would you be willing to help complete the tests for GoogleSync?

I have added an example test to this pull request. It covers the constructor. I have auto-generated a test for each method in the class. Since GoogleSync connects to an external service we will probably not be able to cover every line. However, it would really good if we could add tests for the methods which do not connect to the google api.

We some documentation on how to set up the automated testing suite.
https://docs.suitecrm.com/developer/appendix-c---automated-testing/

If you need any assistance, please let me know. It would be a fanastic to work with you and help improve the code coverage of this class.

For this pull request, we need only the unit tests. This involves setting up composer with ```composer update```, install an instance of suitecrm and running the unit test eg ```codecept run tests/unit/include/GoogleSync/GoogleSyncTest.php```.

ApatheticCosmos:
Sorry I've been away from this... I'm going on vacation on Friday and things have been insane at work while I try to cram before I leave for two weeks.

@zahidAbbasAliBaig I was able to reproduce the behavior you're seeing. I've got some debugging to do.

@daniel-samson Thank you so much for the template. I absolutely want to learn how to write these tests and provide them with the code I contribute. I'm going to make this as complete as possible. The template and example have already helped me understand a bit. I'll probably not be able to do much until I get back from vacation on Aug 13th, and then after I clean up whatever fires are waiting for me at work.

ApatheticCosmos:
On another of my pull requests, @gymad mentioned that new features should target the 'develop' branch. Should I change that here as well?

ApatheticCosmos:
Crap! When I resolved that single conflict, it merged all of the hotfix branch into mine. I didn't want to do that. :\
Ugh!

ApatheticCosmos:
@daniel-samson 
I've added unit tests for everything (I think) that doesn't require Google Calendar access or a live DB w/ users and events. Please take a look and let me know what you think of it so far.

pgorod:
To answer your old question 

> On another of my pull requests, @gymad mentioned that new features should target the 'develop' branch. Should I change that here as well?

Don't worry, the name of the branch on your fork doesn't matter, as long as you point the PR to the correct `develop` branch on SuiteCRM's original repo.

By the way, this PR is showing 4705 files changed 😱 I guess that means the branch got messed up for some reason... this might be just a bunch of changes done on the origin repo, and you just need to bring your branch up to date with it.

ApatheticCosmos:
@pgorod 
Yeah, It got messed up. I did a conflict resolve, and it somehow pulled all the changes from hotfix to my branch. I reverted the changes (see a601271 above) but the damage was done. I'm not sure what to do now, other than wipe it, apply my changes to a clean branch, and then resubmit the pull request. :\
I really hate to lose the history here, but I don't know another way to fix it. Any thoughts? Can all the comments here be moved to a new pull request or something?

Dillon-Brown:
Hi @ApatheticCosmos, I've gone ahead and fixed the branch. I've also put a backup of the branch before I made changes here: https://github.com/Dillon-Brown/SuiteCRM/tree/feature/google_calendar_backup just in case you need to compare the two.

ApatheticCosmos:
@Dillon-Brown 
I need to figure out how you did that. My Git-fu is weak. :P
Thank you!

pgorod:
@ApatheticCosmos I have a brute-force hack to do that here: https://pgorod.github.io/Git-rebase-atomic/

It's meant to be a one-off command that you do without thinking, so it's slow and stupid. It sets up the whole repo, does the trick, then removes everything. But if you use just the commands in the middle of the script, on a repo you already set up before, it should work.

Dillon-Brown:
Hi @ApatheticCosmos, were you happy to have this PR tested and merged in? or were you still making changes?

ApatheticCosmos:
@Dillon-Brown Yes, I'm satisfied with it. I have some more features in mind, but I would rather get this out so more people can use it. I'm sure that, as with all code, more users will expose issues. I'll add those features for future releases.

gunnicom:
@ApatheticCosmos have you thought about using a CalDAV Library like SabreDAV for your sync?
That could open many more opportunities to sync to different servers/clients.
I use it now in Suite for syncing the calendar to Outlook (Only one-way) with OutlookCalDAVSynchronizer. My implementation is very basic and somewhat hackish, so i did not release it to the public yet.

zahidAbbasAliBaig:
Hi,  @ApatheticCosmos <https://github.com/ApatheticCosmos> can you please
make a package or document of just your original functionality so it can be
embedded in other instance easily.

Thank you.

ApatheticCosmos:
@zahidAbbasAliBaig I've got a pull request for the documentation at salesagility/SuiteDocs#127 and a PDF is linked in the body of this pull request.
If you want to apply this pull request as a patch file, you can append .patch to the url and download it that way. Like this: https://github.com/salesagility/SuiteCRM/pull/6146.patch

ApatheticCosmos:
@gunnicom I had thought about it, but I didn't want to introduce any more layers then were necessary for this functionality. Since Google themselves provide the PHP library for accessing the data, it seemed the better way to go.

modules/Users/language/en_us.lang.php
horus68: @ApatheticCosmos an escaped single quote still existes in  `Don\'t forget to check`

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1167 @@
...
+
+        require_once('modules/Users/User.php');
+        while ($row = $this->db->fetchByAssoc($result)) {
+            $user = new User();
+            $user->retrieve($row['id']);
+
+            if (!empty($user->getPreference('GoogleApiToken', 'GoogleSync')) &&
+            $accessToken = json_decode(base64_decode($user->getPreference('GoogleApiToken', 'GoogleSync'))) &&
+            $user->getPreference('syncGCal', 'GoogleSync') == '1') {
+                $this->addUser($user->id, $user->full_name);
```
gymad: if you pass back the results of addUser method as a return array, it makes your function more sense.

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1167 @@
...
+
+        // First we populate the array of syncable users
+        $this->setSyncUsers();
+
+        // We count failures here
+        $failures = 0;
+
+        // Then we go though the array and sync the users with doSync()
+        if (isset($this->users) && !empty($this->users)) {
+            foreach ($this->users as $key => $value) {
```
gymad: $value seems unused. use array_keys()

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1167 @@
...
+     * have Calendar Sync Configured and Enabled and
+     * Sync them one by one.
+     *
+     * @return bool Success/Failure
+     */
+    public function syncAllUsers()
+    {
+
+        // First we populate the array of syncable users
+        $this->setSyncUsers();
```
gymad: calling the setSyncUsers() method but there is no any check of the end-results.

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1167 @@
...
+            foreach ($this->users as $key => $value) {
+                $return = $this->doSync($key);
+                if (!$return) {
+                    $GLOBALS['log']->error('Something went wrong syncing for user id: ' . $key);
+                    $failures++;
+                }
+            }
+        }
+        if ($failures > 0) {
+            return false;
```
gymad: passing the $failures itself make more informative return value of this function.

modules/Schedulers/_AddJobsHere.php
```diff
@@ -520,6 +521,19 @@ function trimSugarFeeds()
 }
 
 
+/**
+ * + * Job 17
+ * + * this will sync the Google Calendars of users who are configured to do so
+ * + */
+function syncGoogleCalendar()
+{
+    require_once 'include/GoogleSync/GoogleSync.php';
```
gymad: Use autoloader or absolute path starting with `__DIR__` at the top of the file.

modules/Users/UserViewHelper.php
```diff
@@ -473,6 +474,39 @@ protected function setupAdvancedTabUserSettings()
 
         $this->ss->assign("SETTINGS_URL", $sugar_config['site_url']);
 
+        // Set Google Auth variables
```
gymad: Increasing the function complexity. Add it to a separated method.

modules/Users/entryPointSaveGoogleApiKey.php
```diff
@@ -0,0 +1,61 @@
+<?php
```
gymad: missing licence header

gymad: can you please make this entry point to more 'object oriented' using less superglobals

modules/Users/entryPointSaveGoogleApiKey.php
```diff
@@ -0,0 +1,61 @@
...
+}
+
+global $sugar_config;
+
+$client = new Google_Client();
+$client->setApplicationName('SuiteCRM');
+$client->setScopes(Google_Service_Calendar::CALENDAR);
+$json = base64_decode($sugar_config['google_auth_json']);
+if (!$config = json_decode($json, true)) {
+    die('Invalid json for auth config');
```
gymad: use exceptions instead die() method.

modules/Users/entryPointSaveGoogleApiKey.php
```diff
@@ -0,0 +1,61 @@
...
+    $user= new user();
+    $user->retrieve($current_user->id);
+    $accessToken = $client->fetchAccessTokenWithAuthCode($_REQUEST['code']);
+    $user->setPreference('GoogleApiToken', base64_encode(json_encode($accessToken)), false, 'GoogleSync');
+    $accessRefreshToken = $accessToken['refresh_token'];
+    if (isset($accessRefreshToken)) {
+        $user->setPreference('GoogleApiRefreshToken', base64_encode($accessRefreshToken), false, 'GoogleSync');
+    }
+    $user->savePreferencesToDB();
+    $url = "/index.php?module=Users&action=EditView&record=" . $current_user->id;
```
gymad: id should br validated by security reason

modules/Users/entryPointSaveGoogleApiKey.php
```diff
@@ -0,0 +1,61 @@
...
+    SugarApplication::redirect($authUrl);
+}
+
+if (isset($_REQUEST['code'])) {
+    global $GLOBALS;
+    global $current_user;
+    $user= new user();
+    $user->retrieve($current_user->id);
+    $accessToken = $client->fetchAccessTokenWithAuthCode($_REQUEST['code']);
+    $user->setPreference('GoogleApiToken', base64_encode(json_encode($accessToken)), false, 'GoogleSync');
```
gymad: $user->retrieve() could returne null and it could be indicates an error. please handle it.

modules/Users/entryPointSaveGoogleApiKey.php
```diff
@@ -0,0 +1,61 @@
...
+$client->setAuthConfig($config);
+$client->setAccessType('offline');
+
+if (isset($_REQUEST['getnew'])) {
+    $authUrl = $client->createAuthUrl();
+    SugarApplication::redirect($authUrl);
+}
+
+if (isset($_REQUEST['code'])) {
+    global $GLOBALS;
```
gymad: `global $GLOBALS;`  is invalid. $GLOBALS already global

modules/Users/entryPointSaveGoogleApiKey.php
```diff
@@ -0,0 +1,61 @@
...
+    if (isset($accessRefreshToken)) {
+        $user->setPreference('GoogleApiRefreshToken', base64_encode($accessRefreshToken), false, 'GoogleSync');
+    }
+    $user->savePreferencesToDB();
+    $url = "/index.php?module=Users&action=EditView&record=" . $current_user->id;
+    SugarApplication::redirect($url);
+}
+
+if (isset($_REQUEST['setinvalid'])) {
+    global $GLOBALS;
```
gymad: global $GLOBALS; is invalid. $GLOBALS already global

modules/Users/entryPointSaveGoogleApiKey.php
```diff
@@ -0,0 +1,61 @@
...
+    $user= new user();
+    $user->retrieve($current_user->id);
+    $user->setPreference('GoogleApiToken', '', false, 'GoogleSync');
+    $user->savePreferencesToDB();
+    $url = "/index.php?module=Users&action=EditView&record=" . $current_user->id;
+    SugarApplication::redirect($url);
+}
+
+if (isset($_REQUEST['error'])) {
+    global $GLOBALS;
```
gymad: global $GLOBALS; is invalid. $GLOBALS already global

modules/Users/entryPointSaveGoogleApiKey.php
```diff
@@ -0,0 +1,61 @@
...
+    global $current_user;
+    $url = "/index.php?module=Users&action=EditView&record=" . $current_user->id;
+    $exitstring = "<html><head><title>SuiteCRM Google Sync - ERROR</title></head><body><h1>There was an error: " . $_REQUEST['error'] . "</h1><br><p><a href=" . $url . ">Click here</a> to continue.</body></html>";
+    die($exitstring);
```
gymad: use exceptions instead die() method.

modules/Users/entryPointSaveGoogleApiKey.php
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// If we don't get a known return, we just silently return to the user profile.
+global $current_user;
```
gymad: use less global. add it to the top of the method (or file) if it's really necessary.

modules/Users/entryPointSaveGoogleApiKey.php
```diff
@@ -0,0 +1,61 @@
...
+    $url = "/index.php?module=Users&action=EditView&record=" . $current_user->id;
+    SugarApplication::redirect($url);
+}
+
+if (isset($_REQUEST['setinvalid'])) {
+    global $GLOBALS;
+    global $current_user;
+    $user= new user();
+    $user->retrieve($current_user->id);
+    $user->setPreference('GoogleApiToken', '', false, 'GoogleSync');
```
gymad: $user->retrieve() could returne null and it could be indicates an error. please handle it.

modules/Users/entryPointSaveGoogleApiKey.php
```diff
@@ -0,0 +1,61 @@
...
+    $user->savePreferencesToDB();
+    $url = "/index.php?module=Users&action=EditView&record=" . $current_user->id;
+    SugarApplication::redirect($url);
+}
+
+if (isset($_REQUEST['error'])) {
+    global $GLOBALS;
+    global $current_user;
+    $url = "/index.php?module=Users&action=EditView&record=" . $current_user->id;
+    $exitstring = "<html><head><title>SuiteCRM Google Sync - ERROR</title></head><body><h1>There was an error: " . $_REQUEST['error'] . "</h1><br><p><a href=" . $url . ">Click here</a> to continue.</body></html>";
```
gymad: use templates for html outputs.

composer.json
```diff
@@ -27,6 +27,7 @@
     "onelogin/php-saml": "^3.0.0",
     "google/recaptcha": "^1.1",
     "ezyang/htmlpurifier": "^4.10",
+    "google/apiclient": "^2.0",
```
gymad: this lib indicates a composer config in the latest suitecrm version

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1167 @@
...
+            $accessToken['refresh_token'] = base64_decode($this->workingUser->getPreference('GoogleApiRefreshToken', 'GoogleSync'));
+        }
+
+        // New Google Client
+        $client = new Google_Client();
+        $client->setApplicationName('SuiteCRM');
+        $client->setScopes(Google_Service_Calendar::CALENDAR);
+        $client->setAccessType('offline');
+        $client->setAuthConfig($this->authJson);
+        $client->setAccessToken($accessToken);
```
gymad: this method could trow some exception. you sould handle it.

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1167 @@
...
+ * of this program must display Appropriate Legal Notices, as required under
+ * Section 5 of the GNU Affero General Public License version 3.
+ *
+ * In accordance with Section 7(b) of the GNU Affero General Public License version 3,
+ * these Appropriate Legal Notices must retain the display of the "Powered by
+ * SugarCRM" logo and "Supercharged by SuiteCRM" logo. If the display of the logos is not
+ * reasonably feasible for technical reasons, the Appropriate Legal Notices must
+ * display the words "Powered by SugarCRM" and "Supercharged by SuiteCRM".
+ */
+
```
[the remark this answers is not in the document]
ApatheticCosmos: Since this PR has been merged, I'm going to be handling these change requests in my fork and creating a new PR. If there's a better way to do it, please let me know. I can't find any.
This one is fixed.

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1167 @@
...
+    /** @var object A Database Instance */
+    private $db;
+
+    /**
+     * Class Constructor
+     */
+    public function __construct()
+    {
+        if (isset($_SERVER['GSYNC_LOGLEVEL'])) {
+            LoggerManager::getLogger()->setLevel($_SERVER['GSYNC_LOGLEVEL']);
```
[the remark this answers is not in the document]
ApatheticCosmos: Should I take this out of the code altogether? It's for debugging just this class when running from cron.php. I'm not sure how to set the debug level back when its set in the constructor. As it functions now, using this will output debug/info/error/etc level logging for everything in or run from this class. It's useful when I'm trying to figure out what's broken, but I'm not sure it would be useful to anyone else.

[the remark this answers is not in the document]
ApatheticCosmos: An example of how this works from the command line:
$ GSYNC_LOGLEVEL=debug php -f cron.php

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1167 @@
...
+     */
+    public function getAuthJson()
+    {
+        global $sugar_config;
+        $authJson_local = json_decode(base64_decode($sugar_config['google_auth_json']), true);
+        if (!$authJson_local) {
+        // The authconfig json string is invalid json
+            $GLOBALS['log']->fatal(__FILE__ . ':' . __LINE__ . ' ' . __METHOD__ . ' - ' . 'Invalid AuthConfig JSON');
+            return false;
+        } elseif (!array_key_exists('web', $authJson_local)) {
```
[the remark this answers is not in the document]
ApatheticCosmos: Whoops... It's actually testing two different failures, but the log output is the same. I'll fix the output to describe the error better.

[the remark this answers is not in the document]
ApatheticCosmos: Fixed in my fork.

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1167 @@
...
+    {
+
+        // We do it this way so we also get deleted meetings
+        $query = "SELECT id FROM meetings WHERE assigned_user_id = '" . $this->workingUser->id . "' AND date_start <= now() + interval 3 month";
+        $result = $this->db->query($query);
+
+        $meetings = array();
+        require_once('modules/Meetings/Meeting.php');
+        while ($row = $this->db->fetchByAssoc($result)) {
+            $results[] = $row['id'];
```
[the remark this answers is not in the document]
ApatheticCosmos: It's cruft from previous work. Removed in my fork.

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1167 @@
...
+        return $meetings;
+    }
+
+    /**
+     * Set user's google calendar id
+     *
+     *
+     * @return bool Success/Failure
+     */
+    public function setUsersGoogleCalendar()
```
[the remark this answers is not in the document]
ApatheticCosmos: Reducing the complexity of this function is a tall order. It needs to search though the users calendars until it finds the 'SuiteCRM' calendar and then pull the id of that Calendar and set the class variable to it.
If the calendar doesn't exist, it needs to create it, then set the variable to the id of the created calendar.
Then it returns a simple success/failure bool when the id has been set.
I'm not sure what can be simplified.

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1167 @@
...
+     * @return \Google_Service_Calendar_Event|null Google_Service_Calendar_Event if found, null if not found
+     */
+    public function getMissingMeeting($event_id)
+    {
+        // get list of users calendars
+        $calendarList = $this->gService->calendarList->listCalendarList();
+
+        // Search all calendars looking for event
+        foreach ($calendarList->getItems() as $calendarListEntry) {
+            if ($gEvent = $this->gService->events->get($calendarListEntry->getId(), $event_id)) {
```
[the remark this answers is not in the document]
ApatheticCosmos: This entire method is currently unused. I'm not sure it will be used in the future. I created it because I was concerned about what would happen if a user moved a synced event to a different calendar. It turns out that the Google Event ID stays the same even if it's moved, so this was never needed. I kept it around in case I found another use for it in the future. The variable was set for testing, so I could see the Google Event ID in xdebug.
Not sure what, if anything, to do with this. It could be removed, but I may use it in the future when we get more complicated features.

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1167 @@
...
+            return null;
+        }
+    }
+
+    /**
+     * Get a SuiteCRM meeting by Google Event ID
+     *
+     * @param string $event_id The Google Event ID
+     *
+     * @return \Meeting SuiteCRM Meeting Bean
```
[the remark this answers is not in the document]
ApatheticCosmos: Fixed in my fork

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1167 @@
...
+     * Figure out if we need to push/pull an update, or do nothing.
+     *
+     * Used when an event w/ a matching ID is on both ends of the sync.
+     *
+     * @param Meeting|\Google_Service_Calendar_Event $event_1 Either the Meeting Bean or Google_Service_Calendar_Event Object
+     * @param \Google_Service_Calendar_Event $event_2 (optional) Google Event
+     *
+     * @return string push(_delete), pull(_delete), skip
+     */
+    public function pushPullSkip($event_1, $event_2 = null)
```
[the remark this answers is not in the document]
ApatheticCosmos: This is by far the most complicated method in the class, but for good reason. It needs to take two objects, determine which is which, then figure out what event to sync in what direction and whether to delete one of them to match the other. It then returns a simple string with instructions for the next step of the process.
I'd love to simplify it, but I can't find any basic way to do that without complicating something else.
One possibility is to split it into two methods, but the split off method would only ever be called by the original. It's useless without it. I don't think it would actually be simpler, just more difficult to read and follow.
I'm not sure where to go with this. :\

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1167 @@
...
+
+        // Set the SuiteCRM Meeting's last sync timestamp, and google id
+        $this->setLastSync($event_local, $return->getId());
+
+        /* We don't get a status code back showing success. Instead, the return of the
+         * create or update is the Google_Service_Calendar_Event object after saving. 
+         * So we check to make sure it has an ID to determine Success/Failure.
+         */
+        if (isset($return->id)) {
+            return true;
```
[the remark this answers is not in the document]
ApatheticCosmos: This in itself is a workaround for the fact that the google api doesn't return a true/false. I was trying to keep things consistent with a success/failure return. Without this I need to put an isset check elsewhere. the Google ID is saved using the setLastSync call above this, which is the only consistent way to save a SuiteCRM record in the class. Otherwise the time stamp is not set and it doesn't have an accurate last sync to compare.

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1167 @@
...
+     *
+     * If the SuiteCRM Meeting is not provided, a new one will be created
+     * and inserted. If one is provided, the existing meeting will be updated.
+     *
+     * @param \Google_Service_Calendar_Event $event_remote \Google_Service_Calendar_Event Object
+     * @param Meeting $event_local Meeting (optional) \Meeting Bean
+     *
+     * @return bool Success/Failure of setLastSync, since that's what saves the record
+     */
+    public function pullEvent($event_remote, $event_local = null)
```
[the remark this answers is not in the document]
ApatheticCosmos: You mean like this?
pushEvent(Meeting $event_local, Google_Service_Calendar_Event $event_remote = null)
You're right. That would be good. I'll edit and do some testing with it.

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1167 @@
...
+        } else {
+            $event = $this->updateSuitecrmMeetingEvent($event_local, $event_remote);
+        }
+
+        // We need to set the suitecrm_ private properties in the Google event here,
+        // Otherwise it's seen as a new event next time we sync
+
+        // We pull the existing extendedProperties, and change our values
+        // That way we don't mess with anything else that's using other values.
+        $extendedProperties = $event_remote->getExtendedProperties();
```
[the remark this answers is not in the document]
ApatheticCosmos: Yep. It's doing the same thing. I'll see if I can break it out into it's own method without turning the existing methods into spaghetti.

modules/Users/language/en_us.lang.php
ApatheticCosmos: @horus68 Is using `&#39;` for a single quote acceptable here?

include/GoogleSync/GoogleSync.php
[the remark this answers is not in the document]
gymad: Ok, if you need this variable by any reason but in case please.
"Using the assignment operator in conditional expressions frequently indicates programmer error and can result in unexpected behavior."
Use the variable value assignment before the if condition.
Compilant solution:
```
  a = b;
  if (a) {
    /* ... */
  }
```

include/GoogleSync/GoogleSync.php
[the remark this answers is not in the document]
gymad: > One possibility is to split it into two methods, but the split off method would only ever be called by the original. It's useless without it.
Yes, spliting it into two method would be the correct solution.

> I don't think it would actually be simpler, just more difficult to read and follow.
Then, please think about the tests. Each complex function makes more difficult to write proper test to cover the method.

include/GoogleSync/GoogleSync.php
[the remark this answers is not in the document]
gymad: in your case, if I calling your function in the future like this:
```
$ret = $g->pushEvent($x);
if ($ret) {
  echo "Ok, but what is the affected ID?";
} else {
  echo "Fail.";
}
```
if you returns something like this:
`return isset($return->id) ? $return->id : false;`
Then the caller knows and able to handle the information about the affected event.

include/GoogleSync/GoogleSync.php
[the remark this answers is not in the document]
gymad: Yes, exactly this is what I meant.

include/GoogleSync/GoogleSync.php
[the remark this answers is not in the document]
gymad: Less code is less spagetti:)

include/GoogleSync/GoogleSync.php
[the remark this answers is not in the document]
ApatheticCosmos: Done

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1167 @@
...
+
+    /** @var object A Database Instance */
+    private $db;
+
+    /**
+     * Class Constructor
+     */
+    public function __construct()
+    {
+        if (isset($_SERVER['GSYNC_LOGLEVEL'])) {
```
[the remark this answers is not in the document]
ApatheticCosmos: @gymad, do you mean I should take the superglobal and save it as a local variable, and then use it from there?

include/GoogleSync/GoogleSync.php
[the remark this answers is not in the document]
ApatheticCosmos: I see what you mean. If I ever need it, I'll do that. For now I'm going to remove the method from production code, since it's not used anywhere, and is untested.

include/GoogleSync/GoogleSync.php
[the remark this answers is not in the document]
ApatheticCosmos: I set it to return the id instead of true.

include/GoogleSync/GoogleSync.php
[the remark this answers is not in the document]
ApatheticCosmos: Done!

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1167 @@
...
+        // Pull the status code returned to determine Success/Failure
+        $statusCode = $return->getStatusCode();
+
+        if ($statusCode >= 200 && $statusCode <= 299) {
+         // 2xx statusCode = success
+            $GLOBALS['log']->debug(__FILE__ . ':' . __LINE__ . ' ' . __METHOD__ . ' - ' . 'Received Success Status Code: ' . $statusCode . ' on delete.');
+
+            // This removes the gsync_id reference from the table.
+            $sql = "UPDATE Meetings SET gsync_id = '' WHERE id = '" . $meeting_id . "'";
+            $res = $this->db->query($sql);
```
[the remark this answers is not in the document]
ApatheticCosmos: @gymad, thanks for this. I found a bug because of it. :)

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1167 @@
...
+        // Set Start/End/Duration in SuiteCRM Meeting and Assigned User
+        $event_local->date_start = date("m/d/Y h:i:sa", $starttime);
+        $event_local->date_end = date("m/d/Y h:i:sa", $endtime);
+        $event_local->duration_hours = $hours;
+        $event_local->duration_minutes = $mins;
+        $event_local->assigned_user_id = $this->workingUser->id;
+
+        // Disable all popup reminders for the SuiteCRM meeting
+        $sql = "UPDATE reminders SET popup = '0' WHERE related_event_module_id = '" .
+            $event_local->id .
```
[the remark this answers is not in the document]
ApatheticCosmos: @gymad I find myself wishing we had PDO support.
I've changed this to create the string with sprintf and quotes inline.

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1167 @@
...
+
+    /**
+     * Update SuiteCRM Meeting from Google Calendar Event
+     *
+     * @param Meeting $event_local SuiteCRM Meeting Bean
+     * @param \Google_Service_Calendar_Event $event_remote Google_Service_Calendar_Event Object
+     *
+     * @return Meeting SuiteCRM Meeting Bean
+     */
+    public function updateSuitecrmMeetingEvent($event_local, $event_remote)
```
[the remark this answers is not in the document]
ApatheticCosmos: Done.

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1167 @@
...
+        $event_local->date_end = date("m/d/Y h:i:sa", $endtime);
+        $event_local->duration_hours = $hours;
+        $event_local->duration_minutes = $mins;
+        $event_local->assigned_user_id = $this->workingUser->id;
+
+        // Disable all popup reminders for the SuiteCRM meeting
+        $sql = "UPDATE reminders SET popup = '0' WHERE related_event_module_id = '" .
+            $event_local->id .
+            "' AND deleted = '0'";
+        $res = $this->db->query($sql);
```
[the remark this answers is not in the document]
ApatheticCosmos: Done.

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1167 @@
...
+
+        // Mark all reminders where both popup and email are disabled as deleted.
+        $sql = "UPDATE reminders SET deleted = '1' " .
+            "WHERE popup = '0' AND email = '0' AND related_event_module_id = '" .
+            $event_local->id .
```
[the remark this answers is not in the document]
ApatheticCosmos: Done.

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1167 @@
...
+            "' AND deleted = '0'";
+        $res = $this->db->query($sql);
```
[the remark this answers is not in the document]
ApatheticCosmos: Done.

include/GoogleSync/GoogleSync.php
```diff
@@ -0,0 +1,1167 @@
...
+        $query = "SELECT id FROM users WHERE deleted = '0'";
+        $result = $this->db->query($query);
+
+        require_once('modules/Users/User.php');
+        while ($row = $this->db->fetchByAssoc($result)) {
+            $user = new User();
+            $user->retrieve($row['id']);
+
+            if (!empty($user->getPreference('GoogleApiToken', 'GoogleSync')) &&
+            $accessToken = json_decode(base64_decode($user->getPreference('GoogleApiToken', 'GoogleSync'))) &&
```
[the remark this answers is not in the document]
ApatheticCosmos: It's cruft. Fixed.

include/GoogleSync/GoogleSync.php
ApatheticCosmos: It could very well be an empty list of users, if no one is configured to sync.

include/GoogleSync/GoogleSync.php
ApatheticCosmos: Done

include/GoogleSync/GoogleSync.php
ApatheticCosmos: IIRC, this needs to return false on failure so that the cron job can be marked as failed.
I could have just done $failures = 1; instead of $failures++;

modules/Users/entryPointSaveGoogleApiKey.php
ApatheticCosmos: Fixed.

ApatheticCosmos: @gymad I'm not sure what to do here. This code takes all it's data from the _REQUEST after Google redirects to it after authorization. Is there another, better way to get the data?

pgorod:
@ApatheticCosmos does this feature include handling of Event invitations getting accepted?

Suppose User A invites User B to an event. When creating the event, User A's Google calendar receives the new event. 

What about User B, when he Accepts? Will the event be created on his Google Calendar?

ApatheticCosmos:
@pgorod Hmmm... I honestly don't know. If the event is on the SuiteCRM Google calendar, it should be synced. But invites will hit the primary calendar. It would need to be moved before it would sync. The only way to be sure is to test it.

pgorod:
Somebody [on the Forums says it doesn't work](https://suitecrm.com/suitecrm/forum/suitecrm-7-0-discussion/24665-events-not-visible-into-google-calendar), that's why I cam here to ask you about this. I had a hunch this was a peculiar scenario that could have easily been overlooked.

PR #6186, opened 2018-07-20
Feature/robo coding standards

Dillon-Brown:
## Description

Adds a robo task for automatically fixing coding standard issues and a dist file for PhpCsFixer configurations.

## How To Test This

1. Run composer update
2. Run ./vendor/bin/robo configure:phpCSFixer
3. Run ./vendor/bin/robo php:CSFixer

## Types of changes

### Final checklist

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -1,0 +1,43 @@
+<?php
+
+use SuiteCRM\Utility\Paths;
+
+$paths = new Paths();
+
+$finder = PhpCsFixer\Finder::create()
+    ->exclude('cache')
+    ->exclude('service')
+    ->exclude('build')
+    ->exclude('vendor')
+    ->exclude('themes')
+    ->exclude('upload')
+    ->exclude('XTemplate')
+    ->exclude('Zend')
+    ->exclude('include/timezone')
+    ->exclude('include/SuiteGraphs')
+    ->exclude('include/social')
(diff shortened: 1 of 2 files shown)


pgorod:
Cool. Could this tool take care of updating the license header at the top of each file?

Dillon-Brown:
@pgorod That's definitely something i'll look at before finishing with this PR. We can do it through .php_cs.dist but I don't think that's the best approach; for the language files it makes sense to update them all at once but for core files there's no real point in updating them unless the file itself is changed. The php_cs can't really be customised in the way we would need to make it work well with Suite so i'll probably end up creating a more specific robotask.

isleshocky77:
> @pgorod That's definitely something i'll look at before finishing with this PR. We can do it through .php_cs.dist but I don't think that's the best approach; for the language files it makes sense to update them all at once but for core files there's no real point in updating them unless the file itself is changed. The php_cs can't really be customised in the way we would need to make it work well with Suite so i'll probably end up creating a more specific robotask.

I'm confused by "It can't work that way". Why not add the .dist file and only add the modules/files ready to cleaned up and kept clean? I know you working on cleaning up module by module. So you could just add those modules to the .dist file.

connorshea:
This needs a rebase, also shouldn't we have travis fail on failed lints?

jack7anderson7:
Assessed :+1:

PR #6188, opened 2018-07-20
Fixed #5856 - Multiple pages PDF from reports when Total of field is …

### LINKED ISSUE #5856: Multiple pages PDF from reports when Total of field is selected
#### Issue
When creating a report with Total sum, or count of some field you get pdf with multiple pages - report table is broken.
In PDF the first page is ok, but the others each contains only one row from report table. Sometimes it can be PDF of 150pages (mostly blank).

#### Expected Behavior
exported PDF should have unbroken table, which continues on next page

#### Actual Behavior
exported PDF has first page ok, but the report table brokes on other pages - one row of table is displayed on one page, or it isn't displayed at all.
total row is either on the last page, or is missing.

#### Possible Fix
no idea

#### Steps to Reproduce
1. create a report
2. add some fields
3. set Total of one field - sum or count
4. generate PDF export

#### Your Environment
* SuiteCRM Version used: Official demo 7.10.4

ChangezKhan:
## Description

I have fixed some broken HTML content which was causing this issue #5856 .

## Motivation and Context

When creating a report with Total sum, or count of some field you get pdf with multiple pages - report table is broken. In PDF the first page is ok, but the others each contains only one row from report table. Sometimes it can be PDF of 150pages (mostly blank).

## How To Test This

1. create a report
2. add some fields
3. set Total of one field - sum or count
4. generate PDF export

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOR_Reports/AOR_Report.php
+++ b/modules/AOR_Reports/AOR_Report.php
@@ -903,7 +903,7 @@ class AOR_Report extends Basic
         $currency->retrieve($GLOBALS['current_user']->getPreference('currency'));
 
         $showTotal = false;
-        $html = '<table>';
+        $html = '</table><table>';
         $html .= "<thead class='fc-head'>";
         $html .= "<tr>";
         foreach ($fields as $label => $field) {
@@ -929,7 +929,7 @@ class AOR_Report extends Basic
             if ($fieldTotal) {
                 $showTotal = true;
                 $totalLabel = $field['label'] . ' ' . $appListStringsAorTotalOptionsFieldTotal;
-                $html .= "<th>{$totalLabel}</td>";
+                $html .= "<th>{$totalLabel}</th>";
             } else {
                 $html .= '<th></th>';
             }
@@ -988,7 +988,7 @@ class AOR_Report extends Basic
             }
         }
         $html .= '</tr>';
-        $html .= '</tbody></table>';
+        $html .= '</tbody>';
 
         return $html;
     }
--- a/modules/AOR_Reports/AOR_Report.php
+++ b/modules/AOR_Reports/AOR_Report.php
@@ -815,11 +815,11 @@ class AOR_Report extends Basic
 
             $row_class = $row_class == 'oddListRowS1' ? 'evenListRowS1' : 'oddListRowS1';
         }
-        $html .= "</tbody>";
+        $html .= "</tbody></table>";
 
         $html .= $this->getTotalHTML($fields, $totals);
 
-        $html .= '</table></div>';
+        $html .= '</div>';
 
         $html .= "    <script type=\"text/javascript\">
                             groupedReportToggler = {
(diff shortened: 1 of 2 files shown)


ChangezKhan:
Travis CI build is failing. This "AOR_ReportTest: Testget total html" test is causing issue. It is giving "Failed asserting that two strings are identical".  The expected string is itself not correct hence it is failing. Kindly check into this ASAP.

modules/AOR_Reports/AOR_Report.php
```diff
@@ -903,7 +903,7 @@ public function getTotalHTML($fields, $totals)
         $currency->retrieve($GLOBALS['current_user']->getPreference('currency'));
 
         $showTotal = false;
-        $html = '';
+        $html = '</table><table>';
```
daniel-samson: This change is causing the tests to fail. Can you please update the test?
```
1) AOR_ReportTest: Testget total html
 Test  tests/unit/modules/AOR_Reports/AOR_ReportTest.php:testgetTotalHTML
Failed asserting that two strings are identical.
- Expected | + Actual
@@ @@
-<thead class='fc-head'><tr><th>total Sum</td></tr></thead></body><tr class='oddListRowS1'><td>60</td></tr></body>
+</table><table><thead class='fc-head'><tr><th>total Sum</th></tr></thead><tbody><tr class='oddListRowS1'><td>60</td></tr></tbody>
#1  /home/travis/build/salesagility/SuiteCRM/tests/unit/modules/AOR_Reports/AOR_ReportTest.php:272
#2  AOR_ReportTest->testgetTotalHTML
--
```

I think adding the closing table element with the opening table is not an appropriate solution. I think these table elements should be in the build_report_html method instead. That way getTotalHTML method, will output valid or a complete set of tags. Plus it would be more clear to developers. Please removing this line and placing the closing element before the call to getTotalHTML (line 819) and the opening table element after the call to getTotalHTML ( line 821)

ChangezKhan: Hey @daniel-samson , thanks for suggestions. I agree with you to that adding closing table element with opening table element is not an ideal solution. I read your suggestions. But I don't think we should place these table elements in the `build_report_html` method. 

Instead, I propose another better way. Before calling `getTotalHTML', we'll close the table tag and inside `getTotalHTML` method, we'll open the table tag and in the end of that same method, we'll close the table tag. So, all HTML content related to `getTotalHTML` will stays separate from main method. Kindly let me know, what you think about this.

daniel-samson: Hi @ChangezKhan, I think that would a more appropriate solution. We just need update the closing table element in build_report_html. Can you please ensure that both methods output "completed" elements.

Please can you also update the test to so that the expected matches the output of getTotalHTML?

Dillon-Brown:
Assessed

Issue #6142, opened 2018-07-12
Favorites getCurrentUserSidebarFavorites function is too costly


jshwhitlow:
#### Issue
After analyzing my instance to improve performance I have found that a huge burden resources is coming from the `getCurrentUserSidebarFavorites` function within modules/Favorites/Favorites.php

A breakdown of a 4.6 second load times revealed that 3.5 seconds of that time was spent on `getCurrentUserSidebarFavorites ` and the time spent there was so large due to the person in question having 40 favorites.

#### Expected Behavior
This function needs to be optimized, retrieving a user with even just 10-15 favorites is very costly. Let alone 40+ favorites.

#### Actual Behavior
Issue already described above.

#### Possible Fix
Taking a look at the code in modules/Favorites/Favorites.php on line 103-141:

```
public function getCurrentUserSidebarFavorites($id = null)
    {
        global $current_user;
        $db = DBManagerFactory::getInstance();

        $return_array = array();

        if ($id) {
            $query = "SELECT parent_id, parent_type FROM favorites WHERE assigned_user_id = '" . $current_user->id . "' AND parent_id = '" . $id . "' AND deleted = 0 ORDER BY date_entered DESC";
        } else {
            $query = "SELECT parent_id, parent_type FROM favorites WHERE assigned_user_id = '" . $current_user->id . "' AND deleted = 0 ORDER BY date_entered DESC";
        }

        $result = $db->query($query);

        $i = 0;
        while ($row = $db->fetchByAssoc($result)) {
            $bean = BeanFactory::getBean($row['parent_type'], $row['parent_id']);
            if($bean) {
                $return_array[$i]['item_summary'] = $bean->name;
                $return_array[$i]['item_summary_short'] = to_html(getTrackerSubstring($bean->name));
                $return_array[$i]['id'] = $row['parent_id'];
                $return_array[$i]['module_name'] = $row['parent_type'];

                // Change since 7.7 side bar icons can exist in images/sidebar.
                $sidebarPath = 'themes/' . SugarThemeRegistry::current() . '/images/sidebar/modules/';
                if (file_exists($sidebarPath)) {
                    $return_array[$i]['image'] = SugarThemeRegistry::current()->getImage('sidebar/modules/' . $row['parent_type'], 'border="0" align="absmiddle"', null, null, '.gif', $bean->name);
                } else {
[...]

PR #6211, opened 2018-07-31
Prevent massive slowdown if someone accidently added many favorites

gunnicom:
Steps to reproduce:
- Add 200 to 400 Favorites for a user
- SuiteCRM becomes unusable slow ( For me <1sec will become 6-10 sec on each page load)

Solution: There are never more than 30 favorites displayed, so we do not need to get all of them.

Refers to https://github.com/salesagility/SuiteCRM/issues/6142

### DIFF
--- a/modules/Favorites/Favorites.php
+++ b/modules/Favorites/Favorites.php
@@ -110,7 +110,7 @@ class Favorites extends Basic
         if ($id) {
             $query = "SELECT parent_id, parent_type FROM favorites WHERE assigned_user_id = '" . $current_user->id . "' AND parent_id = '" . $id . "' AND deleted = 0 ORDER BY date_entered DESC";
         } else {
-            $query = "SELECT parent_id, parent_type FROM favorites WHERE assigned_user_id = '" . $current_user->id . "' AND deleted = 0 ORDER BY date_entered DESC";
+            $query = "SELECT parent_id, parent_type FROM favorites WHERE assigned_user_id = '" . $current_user->id . "' AND deleted = 0 ORDER BY date_entered DESC LIMIT 30";
         }
 
         $result = $db->query($query);


gunnicom:
@jshwhitlow You can still favorite more than 30, but in the sidebar only maximal 30 will be shown.
In the sidebar of SuiteR there are only 18 shown for me so i think with 30 we are on the safe side.

jshwhitlow:
@gunnicom, @daniel-samson, well on my instance this solution will not help at all, my instance is still taking a long time due to the large quantity of fields for some beans being retrieved. Therefore, my instance will go un-helped to due to this "Hassle-Free" fix. 30 Favorites still takes several seconds on my instance.

There is no reason to slap a bandaid on this issue. The deeper problem is what this issue was opened specifically about which is that the bean is being retrieved every time, this PR does not solve that issue and should not be approved. If my issue is closed, I will simply open another one because this fails to address the actual problem.

PedroErnst:
@jshwhitlow 

This PR aims to fix a very specific issue, that there's no upper limit on favorites being retrieved.

I don't see this fix making the issue you reported worse. It also will not slow down or prevent anyone working on a long term fix. So I don't really see a reason not to merge it - It doesn't fix every problem, but it fixes *a* problem.

samus-aran:
@jshwhitlow We agree this doesn't resolve the fundamental performance issue, but we should be limiting the query if we are not using it. I don't see this as a bandaid, but an immediate step that can be made more configurable at a later date i.e. Admin settings etc and template alterations.

I've commented on the other issue but I wanted to re-iterate this here that if this closes the issue automatically (when we do a release) we will re-open that referenced issue until a more complete solution is found. :smile:

Dillon-Brown:
Assessed

Issue #5176, opened 2018-02-11
Feature request: adds date of birth and horoscope sign to Contacts


horus68:
#### Issue

Adds date of birth and horoscope sign to Contacts. Also adds google search button.
See this for Sugar https://github.com/andrew-styles/sugarcrm-horoscope

#### Context

Can be useful for mailing lists or engaging direct calls with extra information on the contact

PR #6233, opened 2018-08-09
Add Anniversary option to date field AOW

PedroErnst:
## Description

Adds the option to choose Anniversary as date comparison in AOW.
The condition will run if the current month and day are equal to the field's value.
Currently all operators and value comparisons have no effect.

image

## Motivation and Context

Allows easy sending of birthday emails and other workflow related to anniversaries.

## How To Test This

1. Create a workflow.
2. Make a condition of type Equal To, Date, Anniversary
3. See the workflow fires both on save and from the scheduler.

## Types of changes

### Final checklist


horus68:
one step closer to this request? :) https://github.com/salesagility/SuiteCRM/issues/5176

Dillon-Brown:
Assessed

johnM2401:
Changes made on a Local SuiteAssured v2.0.1
Changes work well on this instance, no issues found!

mattlorimer:
I'm not sure it makes sense for anniversary to be the operator, it would make more sense and increase flexibility(even if not fully implemented initially)  for it to be part of the date type comparison options, e.g 'now', 'today', anniversary'.

PedroErnst:
I agree that in principle it's not really an operator, and that it makes more sense to be inside the date type.

The issues I saw with this and the reason I did make it an operator is that it doesn't really require any additional parameters, and in fact it would actually be mostly useless (and very confusing) to have all the other operators in combination with the Anniversary. What does Greater than -> Date -> Anniversary mean? What happens if someone chooses Lesser than or equal -> Date -> Anniversary? The only operator I could see being implemented is Equal To, so the user would have to choose Equal To, then Date, then Anniversary... without having any choices really, so it seemed more straightforward to make it an operator.

image

So that was my reasoning, but if you think that it makes more sense as shown in the second line of the screenshot I'm very ready to change it.

pgorod:
If we go with Pedro's reasoning, we can make the name a bit more operator-like. Maybe `Equal day and month` or `Is on anniversary` or something.

BTW, is it just me with that broken label?

image

pgorod:
If we go with Pedro's reasoning, we can make the name a bit more operator-like. Maybe `Equal day and month` or `Is on anniversary` or something.

BTW, is it just me with that broken label?

image

mattlorimer:
I was thinking more in the context the you want to do something like 7 days before the anniversary it makes more sense to be there. I agree it most common use is equal tobbut you could for instance what to do a workflow runs 7 days in a row leading up to the anniversary with 2 conditions 1 where date is greater than 7 day less than the anniversary and 2 where the date is not greater than the anniversary. I've no doubt this could get complicated so I not suggesting we should implement it all straight away rather put in the right place to easy expanded upon.

modules/AOW_Conditions/AOW_Condition.php
```diff
@@ -141,4 +141,16 @@ public function save_lines($post_data, $parent, $key = '')
             }
         }
     }
+
+    /**
+     * @return bool
```
gymad: This function returns `false` as failure because `$this->value_type !== 'Date'` OR `strtolower($value[0]) !== 'anniversary'`? How do you let to know with caller what happend exactly?

modules/AOW_WorkFlow/AOW_WorkFlow.php
```diff
@@ -75,6 +80,8 @@ class AOW_WorkFlow extends Basic
      */
     private function getSQLOperator($key)
```
gymad: This function complexity getting too large. Please split this function into smaller logical units.

modules/AOW_WorkFlow/AOW_WorkFlow.php
```diff
@@ -75,6 +80,8 @@ class AOW_WorkFlow extends Basic
...
     {
+        global $timedate;
```
gymad: do not use superglobals or globals if it's not necessary, pass it as an (optional) dependency.

modules/AOW_WorkFlow/AOW_WorkFlow.php
```diff
@@ -370,6 +383,18 @@ public function build_query_where(AOW_Condition $condition, $module, $query = ar
                 return $query;
             }
 
+            if ($condition->isDateAnniversaryCondition()) {
+                $this->isAnniversaryWf = true;
+                $where = '('.$field.' '.$this->getSQLOperator('Anniversary');
```
gymad: `$field` value should be escaped before using it in an sql query. You could use the DBManager::quote and quoted methods.

modules/AOW_WorkFlow/AOW_WorkFlow.php
```diff
@@ -370,6 +383,18 @@ public function build_query_where(AOW_Condition $condition, $module, $query = ar
...
+                $leapYearQuery = '';
+                if ($timedate->getNow()->format('m-d') === '02-28' && !$timedate->getNow()->format('L')) {
+                    // If we're not in a leap year then the 29th feb is also celebrated on the 28th
+                    $leapYearQuery = ' OR '.$field.' '.str_replace('28', '29', $this->getSQLOperator($condition->operator));
```
gymad: Please escape the variables before use it in sql queries.

modules/AOW_WorkFlow/AOW_WorkFlow.php
```diff
@@ -755,8 +788,28 @@ public function check_valid_bean(SugarBean $bean)
         return true;
     }
 
+    /**
+     * @param SugarBean $bean
+     * @return bool
+     */
+    protected function hasAlreadyRunToday(SugarBean $bean)
+    {
+        global $db, $timedate;
```
gymad: do not use superglobals or globals if it's not necessary, pass it as an (optional) dependency.

modules/AOW_WorkFlow/AOW_WorkFlow.php
```diff
@@ -755,8 +788,28 @@ public function check_valid_bean(SugarBean $bean)
...
+
+        $day = $timedate->getNow()->format('Y-m-d');
+        $sql = "SELECT id FROM aow_processed WHERE aow_workflow_id = '{$this->id}' " .
+            "AND parent_id = '{$bean->id}' AND status = 'Complete' AND date_entered LIKE '$day%' LIMIT 1";
```
gymad: Please escape the variables before use it in sql queries.

modules/AOW_WorkFlow/AOW_WorkFlow.php
```diff
@@ -755,8 +788,28 @@ public function check_valid_bean(SugarBean $bean)
...
+
+        if ($db->fetchByAssoc($db->query($sql))) {
+            return true;
+        }
+        return false;
+    }
+
     public function compare_condition($var1, $var2, $operator = 'Equal_To')
     {
+        global $timedate;
```
gymad: do not use superglobals or globals if it's not necessary, pass it as an (optional) dependency.

modules/AOW_Conditions/AOW_Condition.php
PedroErnst: It's not as failure, it just means it's not a "Date" value type. It could be "Value" or some other valid data. But unless it is "Date" this condition cannot be anniversary type.

modules/AOW_WorkFlow/AOW_WorkFlow.php
PedroErnst: Could you please give a short example on how you would do this?
I think the function is fairly short and it does a single thing. But I'm open to suggestions.

modules/AOW_Conditions/AOW_Condition.php
gymad: Ok, but the question is still there if someone does the following:
```
$result = $cond->isDateAnniversaryCondition();
if ($result) {
  echo 'Cool, it is a date anniversary condition :)';
  $this->doSomethingWithIt();
} else {
  echo 'Hi, It is not a date anniversary condition but I have no idea why, maybe it's not date maybe it something else, maybe it's a valid data maybe it's invalid... :(';
  $this->IDontKnowWhatShouldIDo();
}
```

modules/AOW_WorkFlow/AOW_WorkFlow.php
gymad: It's ok, my mistake, sorry.

Mac-Rae:
Please rebase this pull request onto hotfix-7.10.x as the hotfix branch is now specifically for 7.11.x issues.

Please take a look at our contribution guidelines for more information.

PR #6238, opened 2018-08-11
Fix #3864 - Deleted and recreated user can't login with system generated password

### LINKED ISSUE #3864: Deleted and recreated user can't login with system generated password
#### Issue

If a  user is deleted and recreated, they are sent a system generated password which is cheked anagist the old instance and therefore cannot log in. If 'Forgot Password' feature is not enabled, it's an SQL job to get them in.

#### Expected Behaviour

Their password should work.

#### Actual Behaviour

The password in checked against the old instance of the user.

#### Possible Fix

In User.php:759
<addr>
		$query = "SELECT * from users where user_name='$name'";
		if(!empty($where)) {
		    $query .= " AND $where";
		}
</addr>
need to add check to exclude deleted users
<addr>
		$query .= " AND deleted=0";
</addr>

#### Steps to Reproduce

1. Create a User
2. Delete the User
3. Create a User with the same name
4. Can't login with the password sent by System Generated Password email

#### Your Environment

* SuiteCRM Version used:

sanchezfauste:
Make sure that `findUserPassword` function only gets non deleted records.

## Description

See #3864

## Motivation and Context

If you delete a user and then you create a new user using the same username, the login fails since function `findUserPassword` is not filtering by `deleted=0` users.

## How To Test This

1. Create a new user.
2. Delete the previously created user.
3. Create a new user with the same username that you used in step 1. Set a diferent password as before.
4. Try to login with the new user and see that you get an error.
5. Apply PR and try to login again and see that now you can login.

NOTE: This error does not always appear. It only appears if the record that retrieves the findUserPassword method is the user that has been deleted instead of the new one.

## Types of changes

### Final checklist

### DIFF
--- a/modules/Users/User.php
+++ b/modules/Users/User.php
@@ -763,7 +763,7 @@ EOQ;
 	{
 	    $db = DBManagerFactory::getInstance();
 		$name = $db->quote($name);
-		$query = "SELECT * from users where user_name='$name'";
+		$query = "SELECT * FROM users WHERE user_name='$name' AND deleted=0";
 		if(!empty($where)) {
 		    $query .= " AND $where";
 		}


Dillon-Brown:
Assessed

PR #6262, opened 2018-08-23
Allow module to be in custom directory

thisboyiscrazy:
Change to Allow module to be only in custom directory

## Description

Changed code to allow module to be in custom directory

## Motivation and Context

It's seem all added module should go in custom and /module should be for SuiteCRM only stuff.

## How To Test This

Put a module only in the custom directory

## Types of changes

### Final checklist

### DIFF
--- a/include/MVC/View/SugarView.php
+++ b/include/MVC/View/SugarView.php
@@ -1185,7 +1185,7 @@ EOHTML;
      */
     protected function _checkModule()
     {
-        if (!empty($this->module) && !file_exists('modules/' . $this->module)) {
+        if (!empty($this->module) && !file_exists('modules/' . $this->module) && !file_exists('custom/modules/' . $this->module)) {
             $error = str_replace("[module]", "$this->module", $GLOBALS['app_strings']['ERR_CANNOT_FIND_MODULE']);
             $GLOBALS['log']->fatal($error);
             echo $error;


Dillon-Brown:
Assessed

samus-aran:
I'm not understanding the context of this? Why would you have a custom module if you haven't built it to go within the module folder and extend it within custom?
I'm not sure the impact of changing this would be throughout the system..

Would be great if you could clarify your end goal here?

thisboyiscrazy:
Well, I may be doing it wrong.

I have a custom internal module I am writing along with other customizations. I was trying to put it in /custom because that is where I am putting all my other customizations. It's seems /modules is used for SuiteCRM modules or modules that a downloaded and installed. 

It keeps it a lot cleaner and easier to put all my stuff in /custom.

pgorod:
@thisboyiscrazy maybe we can close this? Or can you explain better what you want to achieve, and what is the issue that is troubling you? 

Maybe this is just a question of learning where to place the files. It currently is possible to do everything with custom modules and keep all files under the `custom` directory, so this PR shouldn't be necessary as long as you know where SuiteCRM looks for the files.

thisboyiscrazy:
Adapting the directions on [Custom admin config page for new settings | SugarCRM Community](https://community.sugarcrm.com/thread/28405)

Here is my controller: /custom/modules/MyAdminModule/controller.php
```php
<?php

if (!defined('sugarEntry') || !sugarEntry) {
    die('Not A Valid Entry Point');
}

class MyAdminModuleController extends SugarController
{
    public function action_zipfix()
    {
        $this->view = 'zipfix';
    }
}
```

My view /custom/modules/MyAdminModule/views/view.zipfix.php

```php
<?php

if (!defined('sugarEntry') || !sugarEntry) {
    die('Not A Valid Entry Point');
}

global $current_user, $sugar_config;

if (!is_admin($current_user)) {
    sugar_die("Unauthorized access to administration.");
}

require_once('include/MVC/View/SugarView.php');

class Viewzipfix extends SugarView
{

    public function display()
    {
        global $db, $current_user;
    
        $tpl = 'custom/modules/LiveRoofAdmin/tpl/zipfix.tpl';

        $smarty = new Sugar_Smarty();
       
        $smarty->display($tpl);
    }
}
```

But SugarView check to see if the modules exist in _checkModule the problem is it only check /modules

pgorod:
Just from a quick look (I might be wrong, but...) I think you should still be using files in folder `custom/modules/Administration/` exactly like the tutorial says, instead of adapting it to `custom/modules/MyAdminModule/`.

What you're doing is overriding the Administration module so that the Administration page shows some special stuff. You're doing it on behalf of your custom module, but the module you're customizing is still "Administration".

thisboyiscrazy:
Correct in that that tutorial they do, I was trying to break it out into it's own module like [Drag N Drop](https://store.suitecrm.com/addons/drag-n-drop-document) does this. Where there is an Admin link to settings at ```index.php?module=DragNDrop&action=settings```

Drag N Drop put a module in /modules and some stuff in /custom.

But sense I only want to put stuff in /custom to keep it cleaner it doesn't work.

Guess I could to want I want with administration action view maps. But it seems logically if suitecrm is looking for something to exist, it should look first in /custom then in /modules or other places.

pgorod:
For one thing, your proposed change only changes a check for a module file, but it doesn't make SuiteCRM actually load that custom file, does it?

I don't know that add-on, I don't have its code, but if it places stuff in `modules` directory, I would avoid copying it.

But in general I believe you should try to do it like the tutorial you linked first. That sounds correct - a change to the Administration module, in directory `custom/modules/Administration`, and avoids this issue altogether (all you code goes under `custom`).

samus-aran:
The general consensus of this that this completely removes the point of even having custom modules within module folder at all which has not only a big impact on the teachings of developers but other implications as well which although may not be visible from your experience. I don't foresee this to be added in to the legacy SuiteCRM structure, there are a lot of old habits to kill and this I don't think would be as beneficial to include.

Issue #6343, opened 2018-09-21
installer fails, if posix is not installed on linux systems


bjoern86:
installer fails on step 2 with this error:

Fatal error: Uncaught Error: Call to undefined function posix_getpwuid() in /SuiteCRM/include/utils.php:455 Stack trace: #0 /SuiteCRM/install/ready.php(284): getRunningUser() #1 /SuiteCRM/install.php(796): require('/html/SuiteCRM/...') #2 {main} thrown in /SuiteCRM/include/utils.php on line 455

This is because posix is not installed on an linux system with PHP-FPM up and running. The problem is, that there is only a check if the os is windows and if not, the script asumes, that posix is installed. We better check if posix is installed and not for windws in line 452:

 if (!function_exists('posix_getpwuid')) {

instead of

if (is_windows()) {

this should be fixed in update scripts as well.

pgorod:
Actually the script tries a few different methods and is ready to return a value even if none of them work.

https://github.com/salesagility/SuiteCRM/blob/master/include/utils.php#L445

But it seems on some systems, instead of simply returning NULL there, break the script execution with a FATAL. It would be interesting to know more about the specifics of your system.

Anyway, I've seen this happen a couple of times in the forums and I agree it would be better to make it more resilient.

But perhaps the correct way to do it is not replacing it on top of the `is_windows` test - it's simply wrapping the posix call inside a new `if (!function_exists('posix_getpwuid'))`.

pgorod:
Sorry, I hadn't refreshed the page when I wrote that comment - now I see you already provided a fix (and you made it like I said was better, cool). Thanks

pgorod:
New fix proposed in #7398 addressing all the issues raised by @gymad in comments on #6344

PR #6344, opened 2018-09-21
fixes issue #6343 - installer fails, if posix is not installed on linux systems

bjoern86:
fixes issue #6343

## Types of changes

### Final checklist


samus-aran:
Thanks @bjoern86 - Can you point the branch to hotfix rather than master. You can read more on our process at the wiki. Thanks for contributing!

bjoern86:
@samus-aran i changed base to hotfix. everything ok so far?

samus-aran:
It looks like you've included more commits (due to the changes between the two). What I would do is either, create a new branch based on the hotfix branch and apply your fix there and create a new PR (that is the simpliest way if you don't your way around git). Or you can rebase your patch-1 branch with hotfix. I usually actually reset --hard if I've something like this to ensure that the branch I'm working on is exact match to hotfix before working on things.

include/utils.php
```diff
@@ -451,7 +451,7 @@ function getRunningUser()
     if ($runningUser == null) {  // matches null, false and ""
         if (is_windows()) {
             $runningUser = getenv('USERDOMAIN') . '\\' . getenv('USERNAME');
-        } else {
+        } elseif (function_exists('posix_getpwuid')) {
             $usr = posix_getpwuid(posix_geteuid());
             $runningUser = $usr['name'];
         }
```
gymad: What happening in case when it's not a windows system and also posix_getpwuid is not exists too? I think it leads to an error somewhere.
Please handle the final `else` fork,  throw an exception and/or log the problem into suitecrm log so that the and user or caller can figure out what just happend here.

pgorod: @gymad I don't think any more checks are necessary, the function should just return `''` when all else fails. When I wrote this code I knew it wouldn't work on all systems, I just didn't predict (as I should have) that the missing Posix function would crash as a fatal on some systems, instead of just returning `null`. But it is ok for this function to return an empty string.

pribeiro42: @pgorod, @gymad doesn't this function already returns `null` if the posix function doesn't exist ?

pgorod: I think it does, but on some hostings it reacts violently, and instead of returning, FATALs out.

gymad: Okey, let me explane what is the problem here from an different angle:
This function `getRunningUser()` called from many places, let see what happenning for e.g if it's a linux system and the `posix_getpwuid` function does not exists, so this function returns an empty string.
Here is the first case what I just found quickly:
cron.php:57: `$cronUser = getRunningUser();` it returns an empty string and the following code does not work properly, even if the `$sugar_config['cron']['allowed_cron_users']` filled with a correct user but also there is a fatal in the log: "cron.php: running as {$cronUser} is not allowed in allowed_cron_users in config.php. Exiting.". It's confusing, because the log note is not true also this code should not "Exiting".
Here is the second case:
install/performSetup.php:655: `addCronAllowedUser(getRunningUser());` It returns an empty string and passing to addCronAllowedUser() function, so this line does not do anything because `$addUser == ''` is true (see at include/utils.php:475). Also there is no any log note or error message meanwhile the `$sugar_config['cron']['allowed_cron_users']` filled with a correct user but the end-user have no clue what just happend, it seems a possible issue.
Third case: 
install/ready.php:281 `$webServerUser = getRunningUser();` returns an empty string so the help message bellove doesn't contains a correct user message on the screen. see at install/ready.php:286 `>sudo crontab -e -u '.$webServerUser.'</span><br>` - it's also confusing.
Fourth case: 
Scheduler/Scheduler.php:813 `$webServerUser = getRunningUser();` returns an empty string so the help message bellove doesn't contains a correct user message on the screen. Also confusing just like above.
And maybe there is some more other scenario when this change makes nonsense to the end-users causing a possible issue for someone in the future. I think it should be solved before I accept it.

pgorod:
@bjoern86 I will re-do this PR myself, mainly because of the comments @gymad made (he is right, and I wrote that part of the code), but also because your PR has a some commits you probably didn't intend to include.

Thanks for your contribution anyway, and for the other PR's fixing similar issues :)

bjoern86:
@pgorod in master branch this issue still exists. installation and update of suite crm with linux os and posix disabled still not possible. when will you re-do this PR? thanks.

pgorod:
@bjoern86 I'm so sorry, this completely slipped my mind. Thanks for the reminder, I'll try to look into this as soon as I get a chance.

bjoern86:
thank you!!

if you get a chance, please check issue #6349 
changeing include/utils.php fixes point 2 of issue #6349, this issue showes three other files to fix... should just change:

$owner = posix_getpwuid(fileowner($file));
$group = posix_getgrgid(filegroup($file));

to:

if (function_exists('posix_getpwuid')) $owner = posix_getpwuid(fileowner($file));
if (function_exists('posix_getgrgid')) $group = posix_getgrgid(filegroup($file));

???

PR #6374, opened 2018-10-01
Fixed #6046 - SQL "rank" is a reserved word, can't be used as a field…

### LINKED ISSUE #6046: SQL "rank" is a reserved word, can't be used as a field name in MySQL 8.0.2 and later

```
```

#### Your Environment

Dillon-Brown:
## Description

It seems that rank is an SQL reserved word since MySQL 8.0.2:

## Motivation and Context

Issue reference: #6046 

## How To Test This

Install SuiteCRM with MYSQL 8.0.2

## Types of changes

### Final checklist


pgorod:
@Dillon-Brown I can't find any other reference to this `rank` field, but that is a bit strange... is that data never used?

cameronblaikie:
Assessed :+1:

acirulis:
It's March 2020 & still not merged?

PR #6380, opened 2018-10-04
Fix #5701 with SugarTheme Enhancement

### LINKED ISSUE #5701: New SubTheme Changes Break Parent Theme Custom CSS
My SuiteCRM 7.10.2 instance has a custom CSS file `custom/themes/SuiteP/css/style.css` that, in 7.8.17 LTS, would be acknowledged and loaded in `SugarTheme::getCSSURL()`. However, the new subthemes seek CSS files per-subtheme, and so the custom CSS for the SuiteP theme is not loaded.

#### Issue
Existing customizations from prior versions of Sugar/Suite are using this architecture and it should continue to be supported, while also supporting extra CSS for the Dawn/Dark/etc. subThemes.

#### Expected Behavior
Allow use of `custom/themes/SuiteP/css/style.css` and/or `custom/themes/SuiteP/css/Dawn/style.css`

#### Actual Behavior
Only `custom/themes/SuiteP/css/Dawn/style.css` is used, which will force me to issue customization for all subthemes instead of allowing for generic CSS.

#### Possible Fix
Adjustments to `SugarTheme::getCSSURL()` to better handle the subTheme.

#### Environment
* SuiteCRM Version used: 7.10.2, but based on code review likely affects 7.10.x

matthewpoer:
Custom CSS generated for the SuiteP theme was being ignored due to the subTheme system. This enhancement resolves by self-referencing for the current "primary" theme, then descending to get any files for the subTheme.

## Description
Existing customizations from prior versions of Sugar/Suite are using this architecture and it should continue to be supported, while also supporting extra CSS for the Dawn/Dark/etc. subThemes.

Issue logged as https://github.com/salesagility/SuiteCRM/issues/5701

## How To Test This
Install a custom CSS file in `custom/themes/SuiteP/css/style.css` that does something obvious, e.g. 

~~~
h2.module-title-text {
  font-weight: bold;
  color: pink;
}
~~~

QRR and notice that the change isn't seen. Apply the patch and the change will appear because this file is found and used.

## Types of changes

### Final checklist


include/SugarTheme/SugarTheme.php
```diff
@@ -965,12 +965,21 @@ protected function _getImageFileName(
      * @param  bool   $returnURL if true, returns URL with unique image mark, otherwise returns path to the file
      * @return string path of css file to include
      */
-    public function getCSSURL($cssFileName, $returnURL = true)
+    public function getCSSURL($cssFileName, $returnURL = true, $skipSubTheme = FALSE)
```
gymad: I can not find any usages of third parameter `$skipSubTheme`. Is this function still works properly if I call it and pass `true` as `$skipSubTheme` value? Can I Test is some way? I think this function needs a unit and/or acceptance test to see it works properly in each calling situation.
Also: `FALSE` should be lowercase as it is in PSR standards what we use: 
https://www.php-fig.org/psr/psr-2/#25-keywords-and-truefalsenull

gymad (changes requested):

include/SugarTheme/SugarTheme.php
gymad: @samus-aran Sorry, I was not specific enough. I meant I can not find a single call in the codebase something like this:
```
$theme->getCSSURL($cssFile, $retUrl, $thirdParameterHere); // "usages of third parameter"
```

Mac-Rae:

PR #6381, opened 2018-10-04
Time issue fixed in email template for datetime fields

rediansoftware:
Datetime fix in the email template, previously for datetime field it was only taking the date in email templates.

## Description
It was taking the only date in the email for DateTime fields as well because in the logic it was splitting the value with space. In the value first part is date and the second part is time but it was assigning only first part that's why only date value was assigned in a variable and displayed in emails.

## Motivation and Context

Lot of people were facing this issue because they were not able to see time in email.

## How To Test This

1- Create a field as DateTime in any module
2- Create an email template and use this variable in email
3- Send this email template and check if it date and time both

## Types of changes

### DIFF
--- a/modules/AOS_PDF_Templates/templateParser.php
+++ b/modules/AOS_PDF_Templates/templateParser.php
@@ -133,7 +133,7 @@ class templateParser
                 if ($value != '') {
                     $dt = explode(' ', $value);
                     $value = $dt[0];
-                    if(isset($dt[1]) && isset($dt[1]!=''){
+                    if(isset($dt[1]) && $dt[1]!=''){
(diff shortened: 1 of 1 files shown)


samus-aran:
Hi @rediansoftware Have a we read at how to contribute to the core project. I updated the PR so that it faces the hotfix branch rather than master (we don't merge directly to master). https://docs.suitecrm.com/community/contributing-code/bugs/

samus-aran:
Note to @salesagility/suitecrm-core to recreate for hotfix-7.8.x

rediansoftware:
Thanks. Will check the document.

PR #6410, opened 2018-10-11
Fixed #5871 - Cannot save user in custom code.

### LINKED ISSUE #5871: Cannot save user in custom code.
#### Issue

Attempt to call save() on a User object in custom code fails.

### Expected Behavior

It should save the record and not report an error.

#### Actual Behavior

Returns FALSE (indicating error), and reports in suitecrm.log:
> Invalid Referrer: expected the Save action to be called from the User's Profile Edit View

#### Possible Fix

`SugarEmailAddress::saveAtUserProfile()` is rejecting it because the current request isn't coming from User Profile Save.  It's a custom logic hook (in my case, updating a custom field on User).

#### Steps to Reproduce

1. Create any custom logic hook.
2. Create an object in the User class, by fetching any existing user.
3. Change any data element.
4. call `$user->save();`
5. check the return value from the save() call (it will be FALSE), and look in the log.

#### Context

After saving an object, I make a web services call to an external system and get its associated ID, which wants to go into a custom field in Users.  I had to temporarily hack in code to disable the referrer check to make it work.  I feel this is at least a Medium priority.  (This was broken in a recent-ish version; it worked OK in older versions.)

#### Your Environment

* SuiteCRM Version used: 7.9.12
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): any
* Environment name and version (e.g. MySQL, PHP 7): MySQL, PHP 7.
* Operating System and version (e.g Ubuntu 16.04): Debian & Ubuntu

cameronblaikie:
## Description
it was allowing the user to save (via code) but was returning false, now the the user can save and give an accurate result. 

## How To Test This

1. create a logic hook to save a user via the accounts module upon save
2. the return result from calling save() should be the id of the user as expected. 

## Types of changes

### Final checklist

### DIFF
--- a/modules/Users/User.php
+++ b/modules/Users/User.php
@@ -653,10 +653,13 @@ class User extends Person implements EmailInterface
         parent::save($check_notify);
 
         // User Profile specific save for Email addresses
-        if (!$this->emailAddress->saveAtUserProfile($_REQUEST)) {
-            $GLOBALS['log']->error('Email address save error');
-            return false;
+        if ($this->emailAddress->isUserProfileEditViewPageSaveAction($_REQUEST)){
+            if (!$this->emailAddress->saveAtUserProfile($_REQUEST)) {
+                $GLOBALS['log']->error('Email address save error');
+                return false;
+            }
         }
+
 
         // set some default preferences when creating a new user
         if ($setNewUserPreferences) {
--- a/include/SugarEmailAddress/SugarEmailAddress.php
+++ b/include/SugarEmailAddress/SugarEmailAddress.php
@@ -491,7 +491,6 @@ class SugarEmailAddress extends SugarBean
         return $return;
     }
 
-
     /**
      * Check for User Profile EditView / Save action for
      * Email Addresses updates
@@ -501,7 +500,7 @@ class SugarEmailAddress extends SugarBean
      * @param array $request $_REQUEST
      * @return bool
      */
-    protected function isUserProfileEditViewPageSaveAction($request)
+    public function isUserProfileEditViewPageSaveAction($request)
     {
         $return =
             (isset($request['page']) && $request['page'] == 'EditView') &&


modules/Users/User.php
```diff
@@ -653,11 +653,14 @@ public function save($check_notify = false)
         parent::save($check_notify);
 
         // User Profile specific save for Email addresses
-        if (!$this->emailAddress->saveAtUserProfile($_REQUEST)) {
-            $GLOBALS['log']->error('Email address save error');
-            return false;
+        if ($this->emailAddress->isUserProfileEditViewPageSaveAction($_REQUEST)){
+            if (!$this->emailAddress->saveAtUserProfile($_REQUEST)) {
```
Dillon-Brown: The if construct here can be merged with parent.

modules/Users/User.php
```diff
@@ -653,11 +653,14 @@ public function save($check_notify = false)
...
+                $GLOBALS['log']->error('Email address save error');
```
Dillon-Brown: Should use getLogger instead:
```
LoggerManager::getLogger()->error('Email address save error');
```

Dillon-Brown:
Should this go to hotfix-7.10.x?

Mac-Rae:
Please rebase this pull request onto hotfix-7.10.x as the hotfix branch is now specifically for 7.11.x issues.

Please take a look at our contribution guidelines for more information.

modules/Users/User.php
```diff
@@ -653,11 +653,14 @@ public function save($check_notify = false)
...
+                return false;
+            }
```
Dillon-Brown: You can merge these two ```if``` statements together. While there I would also suggest replacing the ```$GLOBALS['log']``` with the ```LoggerManager::getLogger()->error``` leaving you with:
```php
    if ($this->emailAddress->isUserProfileEditViewPageSaveAction($_REQUEST) && 
        !$this->emailAddress->saveAtUserProfile($_REQUEST)) {
        LoggerManager::getLogger()->error('Email address save error');
        return false;
    }
```

chris001 (approved):
**You should keep the code that creates a `calendar_publish_key` for a new user, before you save the user.**

PR #6473, opened 2018-11-07
Fixed #4123 - reports: fix adding parameterized reports to target lists

### LINKED ISSUE #4123: AOR_Reports - "add to target list" not working for PARAMETERIZED reports with non-default values
#### Issue
When I create a parameterized report of contacts with defaults (that for example displays 200 contacts), and change the parameter to non-default values (which may change the list to say just 5 contacts), then click "add to target list" the target list is populated with the original list (all 200) of the contacts, not the 5 contacts that are showing.

#### Expected Behavior
The "add to target list" should add the current set of parameterized records showing in the report (exactly like "export" on the parameterized report works...it will export just the current set of parameterized records showing on the screen.)

#### Actual Behavior
The original set of default records from the report is added to the target list no matter what parameterization is performed to filter the list of contacts.

#### Possible Fix
It should work exactly the same as export does, but instead of exporting the current set of parameterized/filtered records to a file, it should add them to the target list.

#### Steps to Reproduce
1. Create a report based on contacts module, and under report fields, select first name and last name
2. Under report conditions, select "salutation", tick the parameter box, and make it equal to "one of" and then select all salutations
3. Save the report and see it display 200 contacts let's pretend
4. Then change the parameter of salutations to display just "Dr" for example, and then see it display only 5 contacts let's pretend.
5. Select "add to target list" (and create a new list or add to an existing empty list, it doesn't matter)
6. View the target list and notice that there are 200 contacts in the list, NOT the 5 you expected!
7. (You can try to export the list to csv and see that sure enough, only the 5 you expected are exported)

#### Context
I filter contacts by state, by customer type, and by a few other factors by their associated Account status.  So I need to create a target list from reports, not from the contacts module search for example.
The number of combination reports I would have to create to achieve these goals is many, so it would be of medium priority

#### Your Environment

* SuiteCRM Version used: I've tried 7.8.2 and replicated the issue in 7.9.3, and also see it was mentioned a few years back at: https://suitecrm.com/forum/bug-tracker/5363-add-targets-to-target-list-by-report
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Version 60.0.3112.90 (Official Build) (64-bit) on OS/X
* Environment name and version (e.g. MySQL, PHP 7): MySQL, PHP 5.6.30
* Operating System and version (e.g Ubuntu 16.04): Ubuntu 14.04.5 LTS

lazka:
## Description

Given a report with parameters, the action "Add To Target List" did not
take into account the report parameters, resulting in all records getting
added to the target list as if there were no parameters.

Fix this by updating the form with the current parameters and also send
it to the "addToProspectList" controller action. In the controller extract
the parameters before creating the query like with other actions.

This makes the list of records added to the target list match the one the user
sees on the report page.

Fixes #4123

## How To Test This

1. Create a report which has a parameter and results in a record being shown
2. Adjust the parameter so that no record is shown
3. Use the Add To Target List" action and select a target list
4. With this fix applied no record should get added to the target list

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOR_Reports/AOR_Report.js
+++ b/modules/AOR_Reports/AOR_Report.js
@@ -153,9 +153,9 @@ function setProspectReturn(popup_reply_data) {
   var prospect_id = popup_reply_data.name_to_value_array.prospect_id;
   var record = document.getElementsByName('record')[0].value;
 
-  YAHOO.util.Connect.asyncRequest("GET", "index.php?module=AOR_Reports&action=addToProspectList&record=" + record + "&prospect_id=" + prospect_id, callback);
-
-
+  var form = addParametersToForm("addToProspectList");
+  var query = form.serialize();
+  YAHOO.util.Connect.asyncRequest("GET", "index.php?" + query + "&prospect_id=" + prospect_id, callback);
 }
 
 function changeReportPage(record, offset, group_value, table_id) {
--- a/modules/AOR_Reports/controller.php
+++ b/modules/AOR_Reports/controller.php
@@ -139,6 +139,7 @@ class AOR_ReportsController extends SugarController
 
         $key = Relationship::retrieve_by_modules($this->bean->report_module, 'ProspectLists', $GLOBALS['db']);
         if (!empty($key)) {
+            $this->bean->user_parameters = requestToUserParameters();
             $sql = $this->bean->build_report_query();
             $result = $this->bean->db->query($sql);
             $beans = array();


samus-aran:
Thanks for the contribution @lazka !
Note to @salesagility/suitecrm-core to add to hotfix-7.8.x branch

lazka:
\o/ Thanks @samus-aran

samus-aran:
Hey @lazka The issue gets closed automatically when the associated PRs get merged into the master branch! Everything is relatively all automatic ;) What we do though is set the issue as 'Resolved: Next Release'

PR #6486, opened 2018-11-12
Fixed #6485 - Users: Display opt-in ticks also for external email clients

### LINKED ISSUE #6485: Email Links: Opt-In ticks not shown when email client is set to external
#### Issue

When the user has set their email client to "External Email Client" all email
links lose their opt-in information (ticks).

This might be the same issue reported in #5859 where resetting the user
settings fixed the issue. It might be that the proposed workaround of
resetting just reset the email setting to "SuiteCRM Email Client".

sugar client:
good

external client:
bad

#### Expected Behavior

Changing the email client should not change the display of the email links.

#### Actual Behavior

Changing the email client removes the opt-in information.

#### Possible Fix

I'll submit a PR shortly

#### Steps to Reproduce

1. Create a contact
2. Opt-in the contact
3. Opt-in ticks are shown in the contact list
3. Got to your profile and change the email client to external
4. Go back to the contact list

#### Your Environment
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Firefox 63
* Environment name and version (e.g. MySQL, PHP 7): PHP7.2
* Operating System and version (e.g Ubuntu 16.04): Centos 7

lazka:
## Description
This PR contains 3 commits, the first two are refactorings, the last one fixes
the issue.

----

In case the user had selected the external email client all email links would
not display any opt-in information (ticks). This is due to the ticks logic
being implemented in EmailUI which is only used in the sugar client case. The
external client path falls back to a simple mailto link.

This makes things always use EmailUI for rendering email links (including
ticks) and makes EmailUI use simple mailto links in case the user settings are
set to an external client.

## Motivation and Context

Fixes #6485

## How To Test This

Change the email client setting for the active user. The email links should always contain opt-in ticks.

## Types of changes

### Final checklist

### DIFF
--- a/modules/Emails/EmailUI.php
+++ b/modules/Emails/EmailUI.php
@@ -423,6 +423,24 @@ eoq;
     }
 
     /**
+     * @param string $module_name
+     * @param string $record_id
+     * @param string $name
+     * @param string $addr
+     * @param string $text
+     * @return string
+     */
+    private function createEmailLink($module_name, $record_id, $name, $addr, $text) {
+        return '<a class="email-link"'
+            . ' onclick="$(document).openComposeViewModal(this);"'
+            . ' data-module="' . $module_name
+            . '" data-record-id="' . $record_id
+            . '" data-module-name="' . $name
+            . '" data-email-address="' . $addr  . '">'
+            . $text . '</a>';
+    }
+
+    /**
      *
      * @global SugarBean $focus
      * @param SugarBean|null $bean
@@ -458,13 +476,8 @@ eoq;
             $GLOBALS['log']->warn('EmailUI::populateComposeViewFields - $bean is empty');
         }
 
-        $emailLink = '<a class="email-link"'
-            . ' onclick="$(document).openComposeViewModal(this);"'
-            . ' data-module="' . $myBean->module_name
-            . '" data-record-id="' . $myBean->id
-            . '" data-module-name="' . $myBean->name
-            . '" data-email-address="">'
-            . $innerText;
+        $emailLink = $this->createEmailLink(
+            $myBean->module_name, $myBean->id, $myBean->name, '', $innerText);
 
         // focus is set?
         if (!is_object($myBean)) {
@@ -490,25 +503,20 @@ eoq;
 
             foreach ($emailFields as $emailField) {
                 if (!empty($composeData)) {
-                    $emailLink = '<a onclick=" $(document).openComposeViewModal(this);" ' .
-                        'data-module="' . $composeData['parent_type'] . '" ' . 'data-record-id="' .
-                        $composeData['parent_id'] . '" data-module-name="' . $composeData['parent_name'] .
-                        '"  data-email-address="' . $composeData['to_addrs'] . '">';
+                    $emailLink = $this->createEmailLink(
+                        $composeData['parent_type'], $composeData['parent_id'],
+                        $composeData['parent_name'], $composeData['to_addrs'],
+                        '');
                 } elseif (is_object($myBean) && (property_exists($myBean, $emailField))) {
                     $email_tick = $this->getEmailAddressConfirmOptInTick($myBean, $emailField);
                     $optOut = false;
                     $invalid = false;
 
                     if ($enableConfirmedOptIn === SugarEmailAddress::COI_STAT_DISABLED) {
-                        $emailLink = '<a class="email-link"'
-                            . ' onclick="$(document).openComposeViewModal(this);"'
-                            . ' data-module="'
-                            . $myBean->module_name . '" ' . 'data-record-id="'
-                            . $myBean->id . '" data-module-name="'
-                            . $myBean->name . '" data-email-address="'
-                            . $myBean->{$emailField} . '">';
 
-                        $emailLink .= $myBean->{$emailField} . '</a>';
+                        $emailLink = $this->createEmailLink(
+                            $myBean->module_name, $myBean->id, $myBean->name,
+                            $myBean->{$emailField}, $myBean->{$emailField});
                         return $emailLink;
                     }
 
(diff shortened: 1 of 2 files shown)


modules/Emails/EmailUI.php
```diff
@@ -422,6 +422,24 @@ public function generateComposePackageForQuickCreateFromComposeUrl($emailLinkUrl
         return $this->generateComposePackageForQuickCreate($a_composeData, $emailLinkUrl, $lazyLoad);
     }
 
+    private function createEmailLink($module_name, $record_id, $name, $addr, $text) {
```
gymad: Can you please add phpdoc above this PR to see the acceptable type of arguments and return value.

modules/Users/User.php
```diff
@@ -1513,6 +1513,23 @@ public function getEmailLink2(
...
+                $focus->$emailField = $emailAddress;
+                $emailLink = $emailUI->populateComposeViewFields($focus, $emailField);
+                break;
+            }
+        }
+
+        return $emailLink;
+    }
+
+    public function getEmailClient() {
```
gymad: Can you please add phpdoc above this PR to see type of return value.

modules/Emails/EmailUI.php
```diff
@@ -422,6 +422,24 @@ public function generateComposePackageForQuickCreateFromComposeUrl($emailLinkUrl
...
+        global $current_user;
```
gymad: Using globals is never the right way. Pass it as an optional dependecy parameter and use it only if it's necessary.
(btw I like the way how you solve a tons of code duplication here :+1: )

modules/Emails/EmailUI.php
```diff
@@ -422,6 +422,24 @@ public function generateComposePackageForQuickCreateFromComposeUrl($emailLinkUrl
...
+
+        if ($current_user->getEmailClient() == 'sugar') {
+            return '<a class="email-link"'
+                . ' onclick="$(document).openComposeViewModal(this);"'
+                . ' data-module="' . $module_name
+                . '" data-record-id="' . $record_id
+                . '" data-module-name="' . $name
+                . '" data-email-address="' . $addr  . '">'
+                . $text . '</a>';
+        } else {
```
gymad: Using `else` is not necessary after a return statement (and I think it's bit confusing). Also it improves the code complexity. So I don't think it's needed here.

modules/Users/User.php
```diff
@@ -1513,6 +1513,23 @@ public function getEmailLink2(
         $class = ''
     ) {
         $emailLink = '';
+
+        require_once('modules/Emails/EmailUI.php');
```
gymad: Can you please using `require_once` on top of the file. Also please use `__DIR__` constant as a prefix of file path to make sense where is the file exactly.

modules/Users/User.php
```diff
@@ -1513,6 +1513,23 @@ public function getEmailLink2(
...
+        $emailUI = new EmailUI();
+        for ($i = 0; $i < count($focus->emailAddress->addresses); $i++) {
+            $emailField = 'email' . (string) ($i + 1);
+            $optOut = (bool)$focus->emailAddress->addresses[$i]['opt_out'];
```
gymad: I don't see the guarantee for existance of `'opt_out'` and `'email_address'` indexes. Can you please check if array indexes are exists before using it to evolve any additional and possible php error warning or notice in apache log. Use suitecrm log instead.

modules/Emails/EmailUI.php
lazka: Hm, I can only move it to the caller, use global there and then pass it 5 times, which seems worse to me. Any thoughts on this?

modules/Users/User.php
lazka: I just changed the indent on this code (the gihub diff is a bit confusing imo), and I'd like to avoid unrelated changes in this PR. Should I look into this in a separate PR?

modules/Emails/EmailUI.php
gymad: Legacy code using global variables, I am just trying to avoid them as many places I can, doing this in a way like this:
```
public function doStuff(User $currentUser) {
  ...
}
```
(also I working in a wrapper class over the globals and superglobals - unfortunatelly not finished yet)

Of course, if it's not possible to avoid a global variable in a point I will approve it, I just would ask you to help me to avoid these global usages instead adding new if it's possible.
Btw, you'r right, ignore my previous comment.
Thank you

modules/Users/User.php
lazka: > What do you mean "avoid unrelated changes"?

I mean that I tried to not change anything for the "sugar" client code path and I have not changed anything in this loop except removing the "if" surrounding it.

But I'll look into this.

modules/Emails/EmailUI.php
lazka: > Of course, if it's not possible to avoid a global variable in a point I will approve it, I just would ask you to help me to avoid these global usages instead adding new if it's possible.

Sounds good, I'll try to keep that in mind in the future :+1:

modules/Users/User.php
gymad: Yeah, please try to keep the code behaviour as it is, your changes should be backward compatible as we don't want to brake anything else, even if caller function in someones custom folder.

lazka:
(I've only rebased on top of hotfix, no changes)

modules/Users/User.php
gymad: Could this function called from any custom modules?

lazka: Which function?

Maybe looking at the diff of the specific commit makes it clearer (the PR diff is confusing): https://github.com/salesagility/SuiteCRM/pull/6486/commits/001c85061d4f135cb9d9ab441b7b900bf42dc112

Dillon-Brown:
Assessed

PR #6499, opened 2018-11-15
Fix #6498 - handling of Account names and ids in Leads

### LINKED ISSUE #6498: Inconsistencies in Leads-Accounts relationship
#### Issue
The relationship between Leads and Accounts is non-standard. In Leads, we are supposed to allow listing an account name that doesn't exist yet, or possibly never will exist. Two fields in Leads describe this relationship, `account_id` and `account_name`.

Based on the terrible user experience regarding some inconsistencies listed below, I propose to keep the current scheme but with two differences:

a. If the Account name matches exactly and existing account, link also the id.

b. Make sure these two fields don't get out-of-sync.

This addresses the following bugs and behaviors (back from SugarCRM days!):

1. Lead creation: currently fills only name, not id. My change: if Account with that name exists, link id.

1. If I add an account_id from db, it appears on Detail view as "converted" even if "converted=0"

1. Leads detail view shows "Account name" field that matches `account_id` in DB, regardless of empty or different `account_name` (this is _really_ confusing for users)

1. In Leads Edit view, the Account name field is greyed out when converted=1 (but not when it simply says it's converted on top, because of having an id...)

1. In Leads Edit view, changing the Account name field will leave id as it was, likely mismatched.

1. In Leads conversion, the select account screen starts filtered by the `account_name` that matches Lead's `account_id`, but not Lead's actual `account_name` field.

1. ... that same name will be the future Lead's `account_name` (overwriting what was there), while the `account_id` will be the one specified during conversion... whew!

1. From Account, Leads subpanel, "select" the Lead. It will get the correct new id, but the wrong `account_name`, the one from the previous `account_id` it had...

WHAT. A. MESS. Watch how SugarCRM wrote a [lengthy KB article](http://support.sugarcrm.com/Knowledge_Base/Accounts_Contacts_Leads/Troubleshooting_Sticky_Account_Names_on_Leads/) explaining this and suggesting manual fixes in the database... 

This is not as difficult to fix as it sounds, because once the two fields are kept in sync most of those inconsistent behaviors simply evaporate. I will provide a PR.

#### Your Environment
* SuiteCRM Version used: from SugarCRM up to SuiteCRM 7.10.10

pgorod:
Fix #6498 (Inconsistencies in Leads-Accounts relationship)

## Description
See Issue.

## How To Test This
See Issue.

## Types of changes

### Final checklist

### DIFF
--- a/include/InlineEditing/InlineEditing.php
+++ b/include/InlineEditing/InlineEditing.php
@@ -337,6 +337,10 @@ function saveField($field, $id, $module, $value)
             } else {
                 $bean->$field = $value;
             }
+        } else if ($module === 'Leads' && $field === 'account_name') {
+            require_once('modules/Leads/LeadFormBase.php');
+            $bean->$field = $value;
+            $bean->account_id = LeadFormBase::handleLeadAccountName($bean);
         } else {
             $bean->$field = $value;
         }
--- a/include/generic/DeleteRelationship.php
+++ b/include/generic/DeleteRelationship.php
@@ -109,6 +109,17 @@ if($bean_name == 'Team')
 	}
  	$focus->db->query($query);
  }
+if ($bean_name === "Account" && $linked_field === 'leads') {
+    // for Accounts-Leads non-standard relationship, after clearing account_id form Lead's bean, clear also account_name
+    $focus->retrieve($record);
+    $lead = new Lead();
+    $lead->retrieve($linked_id);
+    if ($focus->name === $lead->account_name) {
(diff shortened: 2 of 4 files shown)


pgorod:
@Mausino are you interested in this functionality for your projects? If so, this PR would be a nice one for you to test, it's all testable from the UI. 

See the Issue (#6498) for a list of things this fixes, if you're interested. Thanks.

cameronblaikie:
Assessed :+1:

Issue #5898, opened 2018-05-18
Campaigns/ProcessBouncedEmails does not mark invalid mails as invalid


gody01:
Campaign Bounce Mail processing does not mark bounced emails as invalid.

#### Issue
It seems  that criteria for marikng bounced email as invalid is 

    //do we have the phrase permanent error in the email body.
    if (preg_match('/permanent[ ]*error/',$email_description))

But nobody users such string any more.

It would be better, to check for "550 5.1.1" code in raw email as it seems somehow de-facto standard for Unknown user message.

#### Expected Behavior
If email bounce with User unknown message, SuiteCRM should mark such Email as invalid.

#### Actual Behavior
Practically no bounced Emails are marked as invalid

#### Possible Fix
change criteria or add check for "550 5.1.1" to the if statement.

#### Steps to Reproduce
1. Send email campaigns with not existent email addresses
2. Wait for bounce messages
3. Check, if bounced emails are marked invalid

#### Context
Automatic bounced mail handling was very useful, since emails were automatically marked as invalid in case mail bounced.

#### Your Environment
* SuiteCRM Version used: 7.10.4
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): All
* Environment name and version (e.g. MySQL, PHP 7): All
* Operating System and version (e.g Ubuntu 16.04): Linux - CentOS, OpenSuSE, ...

pgorod:
@amariussi since you do a lot of Campaigns (I believe) can you try this change to see if it works well?

I mean, check for `550 5.1.1` instead of what is currently there.

amariussi:
@pgorod Just noticed this. Although lately I have been using a different method to send campaigns (unfortunately SuiteCRM is far from being fit for Campaigns at the moment) I will try it out and post my comments, if any.

Just a few quick comments on this subject:
. there should be several criteria to detect bounced messages
. admins should be allowed to select (or add) the criterias used (maybe in config_override,php or, even better, through a dedicated section of the Admin page)
. also the actions should be optional and, possibly multiple actions should be allowed (why would we, developers, impose the action to be taken, like mark email as invalid or delete it or put it in a blacklist to be verified: this should be decided autonomously by the single SuiteCRM users/administrators!!!)

lazka:
I've created a fix for this: #6505

samus-aran:
This is now a duplicate and this issue supersedes this.
https://github.com/salesagility/SuiteCRM/issues/6504
[...]

Issue #1578, opened 2016-06-08
Campaign Bounceback not working


gflauder:
Hi there,
Installed 7.6.4, reset all the cache permissions so it loads and worked through a campaign. I created a separate email for the bounce backs and put it to the test. It fails with:

Wed Jun  8 16:48:02 2016 [20289][1][DEBUG] -------------------------------> Email save() done
Wed Jun  8 16:48:02 2016 [20289][1][INFO] Query:SELECT description FROM notes WHERE file_mime_type = 'messsage/rfc822' AND parent_type='Emails' AND parent_id = '19461cdf-4aa7-8359-0d8b-57584cd14815' AND deleted=0
Wed Jun  8 16:48:02 2016 [20289][1][INFO] Query Execution Time:0.00046896934509277
Wed Jun  8 16:48:02 2016 [20289][1][INFO] Warning: skipping bounced email because it does not have the remove me link.
Wed Jun  8 16:48:02 2016 [20289][1][DEBUG] ********************************* InboundEmail finished import of 1 email: Undelivered Mail Returned to Sender

When I check the message source directly, I see:
`<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" lang='en_us'>
	<head>
		<meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
		<title>[Test]: $acase_name [CASE:$acase_case_number]</title>
	</head>
	<body>
		<div class="mozaik-inner" style="font-family:Arial, Helvetica, sans-serif;font-size:14px;line-height:22.4px;color:rgb(68,68,68);padding:0px 30px;margin:0px;">
			<p style="font-family:Arial, Helvetica, sans-serif;font-size:14px;line-height:22.4px;color:rgb(68,68,68);padding:0px;margin:0px;"> </p>
[...]

PR #6505, opened 2018-11-19
Fixed #6504 - Fix and implement proper bounce email handling

### LINKED ISSUE #5980: Bounced Emails are not processed properly in SuiteCRM 7.10.5
When the Email Campaign is executed, it does not process the bounced email in appropriate manner.  

#### Issue
The campaign sent/attempted 534-emails.  There were 66-emails which were returned with the message: "A message that you sent could not be delivered to one or more of its   recipients. This is a permanent error."  However, 2-issues were presented:
1.  All these 66-emails were returned to the email ID which was, e.g. info@yourdomain.com, configured as 'From' & 'Return To' email address.  Note:  The emails were sent via the outbound system account, which was same as info@yourdomain.com.  I believe that these emails should have returned to the email id, 'bounces@yourdomain.com' which was pre-configured as ID meant for bounced emails.
2.  None of these 66-emails which returned were lodged as Bounced Emails in the STATUS REPORT of the campaign.
image

#### Expected Behavior
These 66-emails should have been considered/recorded as bounced emails appropriately, based upon the error code these emails generated.  For examples:
452  1-Email
550  9-Emails
554  7-Emails
550 5.1.1 1-Email
550 5.1.10  1-Email
550 5.7.1  4-Emails
550-5.1.1  7-Emails
550-5.2.1  25-Emails
550-5.2.2  4-Emails
(No code)  7-Emails

The summary compiled of all emails which returned is attached herewith:
[TEMP1.xlsx](https://github.com/salesagility/SuiteCRM/files/2064866/TEMP1.xlsx)

#### Steps to Reproduce

1.
2.
3.
4.

#### Your Environment

* SuiteCRM Version used:  7.10.5
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)):  Latest
* Environment name and version (e.g. MySQL, PHP 7):  PHP 7.1

### LINKED ISSUE #6504: Multiple bounce handling problems
#### Issue

* Imported bounce emails are not always linked with the campaign logs
* Imported bounce emails are not detected as hard bounces
* Marking emails as invalid based on hard bounces doesn't work

Related issues: #5980, #5898, #1578

#### Expected Behavior

When SuiteCRM fetches a bounce email over IMAP it should show up in the
campaign stats and on hard bounces it should mark the email as invalid.

#### Actual Behavior

Even though there is a proper bounce email imported it doesn't show up in the
campaign stats and on hard bounces the email address isn't marked as invalid.

#### Possible Fix

I've opened the PR #6505 to fix this

#### Steps to Reproduce

1. Set up bounce handling in SuiteCRM
2. Create campaign and add a contact with a non-existing email address
3. Send a campaign email
4. Watch the bounce inbox and wait until the email is read
5. Check the the campaign stats

#### Your Environment

* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Firefox 63
* Environment name and version (e.g. MySQL, PHP 7): PHP7.2
* Operating System and version (e.g Ubuntu 16.04): CentOS 7

lazka:
## Description

Bounce handling is currently broken for multiple reasons:

(1) The imported bounce email is saved in the notes description field which is
marked as text and strips away all HTML on save(). This removes/mangles the
link containing the tracking ID so the attachment can't be linked to the
campaign log later on.

(2) SMTP servers usually return the original email as an attachment on a
bounce and include an addition attachment containing the delivery status (RFC
3464). Since this attachment wasn't saved, hard bounces weren't detected.

(3) In case the tracking ID makes it through and the email is considered a
hard bounce then marking the email as invalid failed because
markEmailAddressInvalid() was called with an email object but expected a
string.

To fix all these:

For (1) we don't save the attachment in the description and read it from the
saved attachment file instead. This makes sure we get the original content and
can extract the tracking ID.

For (2) we also save the delivery-status attachments in addition to the email
copy and also evaluate it in the bounce handling code. This allows us to
confidently classify hard bounces by parsing the status and diagnostic fields
in the delivery status attachment.

For (3) fix markEmailAddressInvalid() to take a bounce entry and figure out
the email address from there and mark it as invalid.

## Motivation and Context

To fix #6504

This likely also fixes #5980, #5898 and #1578

## How To Test This

See #6504

## Types of changes

### Final checklist
- [x] I have read the [**How to Contribut

### DIFF
--- a/modules/Campaigns/ProcessBouncedEmails.php
+++ b/modules/Campaigns/ProcessBouncedEmails.php
@@ -53,13 +53,25 @@ if (!defined('sugarEntry') || !sugarEntry) {
  * @param Email $email
  * @return string
  */
-function retrieveErrorReportAttachment($email)
+function retrieveErrorReportAttachment(Email $email)
 {
     $contents = "";
-    $query = "SELECT description FROM notes WHERE file_mime_type = 'messsage/rfc822' AND parent_type='Emails' AND parent_id = '".$email->id."' AND deleted=0";
-    $rs = DBManagerFactory::getInstance()->query($query);
-    while ($row = DBManagerFactory::getInstance()->fetchByAssoc($rs)) {
-        $contents .= $row['description'];
+
+    $email->getNotes($email->id);
+    foreach ($email->attachments as $note) {
+        if ($note->file_mime_type == 'message/rfc822') {
+            $note_content = $note->getAttachmentContent();
+            if ($note_content !== false) {
+                // XXX: we don't know the encoding of the attached email, but
+                // assume it's quoted-printable.
+                $contents .= quoted_printable_decode($note_content);
+            }
+        } else if ($note->file_mime_type == 'message/delivery-status') {
+            $note_content = $note->getAttachmentContent();
+            if ($note_content !== false) {
+                $contents .= $note_content;
+            }
+        }
     }
 
     return $contents;
@@ -88,16 +100,70 @@ function createBouncedCampaignLogEntry($row, $email, $email_description)
     $bounce->related_type='Emails';
     $bounce->related_id= $email->id;
 
-    //do we have the phrase permanent error in the email body.
-    if (preg_match('/permanent[ ]*error/', $email_description)) {
+    if (checkBouncedEmailInvalid($email_description)) {
         $bounce->activity_type='invalid email';
-        markEmailAddressInvalid($email);
+        markBounceEmailAddressInvalid($bounce);
     } else {
         $bounce->activity_type='send error';
     }
 
     $return_id=$bounce->save();
     return $return_id;
+}
+
+/**
+ * Given an bounce entry, mark the related email address as invalid.
+ *
+ * @param CampaignLog $bounce
+ */
+function markBounceEmailAddressInvalid(CampaignLog $bounce)
+{
+    $sea = new SugarEmailAddress();
+    $email_address = $sea->getPrimaryAddress(false, $bounce->target_id, $bounce->target_type);
+    if (empty($email_address)) {
+        return;
+    }
+
+    LoggerManager::getLogger()->info("Marking email address as invalid: ". $email_address);
+    markEmailAddressInvalid($email_address);
+}
+
+/**
+ * Given the email description returns whether the email should be marked invalid.
+ *
+ * @param string $email_description
+ * @return bool
+ */
+function checkBouncedEmailInvalid($email_description)
+{
+    /* Consider as invalid if we get a permanent error status (5.X.X)
+     * and in addition we get an smtp error 550.
+     * https://tools.ietf.org/html/rfc3464#section-2.3.4
+     * https://tools.ietf.org/html/rfc3464#section-2.3.6
+     * https://www.usps.org/info/smtp_codes.html
+     * Example:
+     *  Status: 5.0.0
+     *  Diagnostic-Code: smtp; 550 #5.1.0 Address rejected.
+     * Example:
+     *  Status: 5.5.0
+     *  Diagnostic-Code: smtp; 550 5.5.0 Requested action not taken: mailbox unavailable
+     * Example:
+     *  Status: 5.1.1
+     *  Diagnostic-Code: smtp; 554 5.1.1
+     */
+    if (preg_match('/^Status:\s*([0-9]+)\.([0-9]+)\.([0-9]+)/m', $email_description, $match)) {
+        // 5.1.1 (permanent) Bad destination mailbox address
+        if ($match[1] == '5' && $match[2] == '1' && $match[3] == '1') {
+            return true;
+        }
+
+        // Permanent error with smtp error code for non-existent email address
+        if ($match[1] == '5' && preg_match('/^Diagnostic-Code:\s*smtp\s*;.*550/m', $email_description)) {
+            return true;
+        }
+    }
+
+    return false;
 }
 
 /**
@@ -166,15 +232,9 @@ function campaign_process_bounced_emails(&$email, &$email_header)
     $emailFromAddress = $email_header->fromaddress;
     $email_description = $email->raw_source;
 
-    //if raw_source is empty, try using the description instead
-    if (empty($email_description)) {
-        $email_description = $email->description;
-    }
-
     $email_description .= retrieveErrorReportAttachment($email);
 
     if (preg_match('/MAILER-DAEMON|POSTMASTER/i', $emailFromAddress)) {
-        $email_description=quoted_printable_decode($email_description);
         $matches=array();
 
         //do we have the identifier tag in the email?
--- a/modules/Notes/Note.php
+++ b/modules/Notes/Note.php
@@ -205,6 +205,22 @@ class Note extends File
         return "$this->name";
     }
 
+    /**
+     * Returns the content as string or false if there is no attachment or it
+     * couldn't be located.
+     *
+     * @return bool|string
+     */
+    public function getAttachmentContent()
+    {
+        $path = "upload://{$this->id}";
+        if (!file_exists($path)) {
+            return false;
+        }
+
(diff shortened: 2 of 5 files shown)


pgorod:
Hi @lazka 

just so you don't have only bots 🤖 talking to you, I want to personally thank you for your recent contributions, very valuable.

Thanks!

lazka:
Thanks 😊

If there is anything I should do differently or improve just let me know!

modules/Campaigns/ProcessBouncedEmails.php
```diff
@@ -101,20 +107,58 @@ function createBouncedCampaignLogEntry($row, $email, $email_description)
 }
 
 /**
- * Given an email address, mark it as invalid.
+ * Given the email description returns if the email should be marked invalid.
  *
  * @param $email_address
+ * @return bool
  */
-function markEmailAddressInvalid($email_address)
```
gymad: I understand if you change the function name it makes more logical and readable the code itself but this change could brakes the backward compatbility to any customization which already exists and possible usage of current exists functions.

modules/Campaigns/ProcessBouncedEmails.php
```diff
@@ -101,20 +107,58 @@ function createBouncedCampaignLogEntry($row, $email, $email_description)
...
+            return TRUE;
     }
+
+    return FALSE;
+}
+
+/**
+ * Given an bounce entry, mark the related email address as invalid.
+ *
+ * @param $bounce
```
gymad: Please update the phpdoc with parameter type.

modules/Campaigns/ProcessBouncedEmails.php
```diff
@@ -101,20 +107,58 @@ function createBouncedCampaignLogEntry($row, $email, $email_description)
...
+ */
+function markBounceEmailAddressInvalid($bounce)
```
gymad: Please add class name as typehinting before parameter.

modules/Campaigns/ProcessBouncedEmails.php
```diff
@@ -101,20 +107,58 @@ function createBouncedCampaignLogEntry($row, $email, $email_description)
...
+{
     $sea = new SugarEmailAddress();
-    $rs = $sea->retrieve_by_string_fields(array('email_address_caps' => trim(strtoupper($email_address))));
-    if ($rs != null) {
+    $email_address = $sea->getPrimaryAddress(false, $bounce->target_id, $bounce->target_type);
+    if (empty($email_address))
+        return;
+
+    $GLOBALS['log']->info("Marking email address as invalid: ". $email_address);
```
gymad: You could using LoggerManager::getLogger() method instead superglobals.

modules/Campaigns/ProcessBouncedEmails.php
```diff
@@ -166,15 +210,12 @@ function campaign_process_bounced_emails(&$email, &$email_header)
...
     $email_description = $email->raw_source;
 
-    //if raw_source is empty, try using the description instead
-    if (empty($email_description)) {
-        $email_description = $email->description;
-    }
-
     $email_description .= retrieveErrorReportAttachment($email);
 
+    // TODO: Delete attachments after this?
```
gymad: Do not leave TODO comment in the code. It's implicate you miss something?

modules/Notes/Note.php
```diff
@@ -205,6 +205,24 @@ public function get_summary_text()
         return "$this->name";
     }
 
+    /**
+     * Returns the content or an empty string if there is no attachment or it
+     * couldn't be located.
+     *
+     * @return string
+     */
+    public function getAttachmentContent()
```
gymad: This function returns an empty string in cases:
- file not exists
- file_get_contents() failure
- file is empty
I think it's confusing, the caller does not know why is the empty string passed if any problem occured.

modules/Notes/Note.php
```diff
@@ -205,6 +205,24 @@ public function get_summary_text()
...
+    {
+        $path = "upload://{$this->id}";
+        if (!file_exists($path))
+            return '';
+
+        $result = file_get_contents($path);
+        if ($result === FALSE)
```
gymad: please use PSR standard const names.
see more: https://www.php-fig.org/psr/psr-2/#25-keywords-and-truefalsenull

modules/Campaigns/ProcessBouncedEmails.php
```diff
@@ -56,10 +56,17 @@
 function retrieveErrorReportAttachment($email)
```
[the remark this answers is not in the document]
lazka: Ah, good catch, I've added a type hint, added sql quoting and a small unit test.

modules/Campaigns/ProcessBouncedEmails.php
lazka: ok, I've left the old function as is an made the new one call the old one.

modules/Campaigns/ProcessBouncedEmails.php
lazka: It's just that I searched for related bug reports before I started working on this and this would be a place for fixing it, but I haven't looked into it. But that's another issue. I've removed the TODO

modules/Notes/Note.php
lazka: I've changed it to return false or a string now.

modules/Notes/Note.php
lazka: OK, thanks

lazka:
I think I've addressed all issues raised so far (minus the open conversation)

modules/Notes/Note.php
```diff
@@ -205,6 +205,21 @@ public function get_summary_text()
...
+    /**
+     * Returns the content as string or false if there is no attachment or it
+     * couldn't be located.
+     *
+     * @return bool|string
+     */
+    public function getAttachmentContent()
+    {
+        $path = "upload://{$this->id}";
+        if (!file_exists($path))
```
code-ph0y: Please use { } with if statements.

For-example: 

```
if (!file_exists($path)) {
    return false;
}
```

lazka: thanks, done (also in other places)

MoritzKosubek:
got a question: How often do you run your Schedluer for this? every 15 minutes or less?

cameronblaikie:
Assessed :+1:

mayerelyashiv:
Dear @lazka, thank you very much;
anyway on a 7.10.27 i found that sending Email Campaigns to a not existing email address produces a "Suppressed by Domain" activity instead of "Invalid Email.

I tried to copy your 5 files hotfix but nohing changed.

Have you any hint about it, please?

lazka:
As far as I remember this only handles the case when you get back a bounce email in your bounce imap inbox. Do you see a bounce email for that "message" there? (using an offline email client maybe)

lazka:
This PR only fixed the parsing of those emails fetched from the inbox. If they don't end up there then something else is broken.

mayerelyashiv:
Hi @lazka; i did a lot of tests (on 7.10.27 instances)
Even if i'm receiving a lot of bounce emails in my bounce imap inbox (checking with Gmail and Thunderbird) SuiteCRM is not marking any of those email addresses as invalid, nor it's adding them to the proper Bounced Messages response list.

aimran50:
Email communication via a CRM and proper account management(which includes marking invalid email addresses) is a critical function. Looks like this has been broken for a long time. Is there any plan to fix this problem?

pgorod:
> anyway on a 7.10.27 i found that sending Email Campaigns to a not existing email address produces a "Suppressed by Domain" activity instead of "Invalid Email".

☝️ is this still what we're talking about, or something different?

I'd start by having a look at 
https://github.com/salesagility/SuiteCRM/blob/master/modules/EmailMan/EmailMan.php

Specifically, every call to **`set_as_sent`**. find out which one corresponds to the handling situation you're not happy with.

PR #6523, opened 2018-11-23
Default navigation paradigm should be used

gunnicom:
It is useless if it is not used, and then default defaults to group modules.


samus-aran:
Another one to note for @salesagility/suitecrm-core to re-create for hotfix
Thanks again @gunnicom for the contribution!

PR #6533, opened 2018-11-27
Fix/Individual Flat Subpanels

PedroErnst:
Passes the expected `flat` variable into the function that creates subpanel buttons.

## Description

The function that builds the array of subpanel buttons checks for a variable named `flat` to decide on an individual subpanel basis whether to display the buttons ins drop down menu.
This variable was never being passed into the function from the SubPanel class despite being available.

## Motivation and Context

Now we can define individual subpanels as flat in the defs.

Result: Flat and Dropdown subpanels in the same instance and the same module.
image

## How To Test This

1. Override a subpanel definition (example `custom/Extension/modules/Leads/Ext/LayoutDefs/myCustomization.php`)
2. Add this line: `$layout_defs['Leads']['subpanel_setup']['history']['flat] = 1`
3. Repair and Rebuild
4. Go to Leads module
5. See only the History module now has its buttons displayed not in a dropdown

## Types of changes

### Final checklist

### DIFF
--- a/include/SubPanel/SubPanelTiles.php
+++ b/include/SubPanel/SubPanelTiles.php
@@ -440,10 +440,14 @@ class SubPanelTiles
             }
         }
         require_once('include/Smarty/plugins/function.sugar_action_menu.php');
-        $widget_contents = smarty_function_sugar_action_menu(array(
-            'buttons' => $buttons,
-            'class' => 'clickMenu fancymenu',
-        ), $this->xTemplate);
+        $widget_contents = smarty_function_sugar_action_menu(
+            [
+                'buttons' => $buttons,
+                'flat' => !empty($thisPanel->_instance_properties['flat']),
+                'class' => 'clickMenu fancymenu',
+            ],
+            $this->xTemplate
+        );
         return $widget_contents;
     }
 }


pgorod:
This is cool 😎 Pedro!

I only learned about this `flat` thing recently, when answering a forum thread. Then the user asked to make also the buttons on the far right of the subpanel row flat, namely the `Edit` and `Remove` buttons that you can use to change the relationship of that subpanel row with the main record being shown in the detail view.

These buttons here:
image

Do you know if there's a property for that that can be set in the options, or in `config_override`?

PedroErnst:
I've done a bit of digging, it seems that the parameter "flat" is hardcoded for those buttons in [this file](https://github.com/salesagility/SuiteCRM/blob/2479923a1a71f45143ef8614c7899df11574fbb9/themes/SuiteP/include/SubPanel/tpls/SubPanelDynamic.tpl#L94).

Luckily it's easy to customize, but of course this is going to affect every subpanel on the entire instance.

pgorod:
Nice find. 

And is this difficult to make customizable from the layoutdefs, one subpanel at a time?

PedroErnst:
@pgorod 

One could simply tie it to the flat variable that already exist, with the caveat that then both the top and right side dropdowns wouldn't be independent from eachother, either both flat or none.

Or one could add another variable to the displayParams, I don't think it would be difficult at all, it's just a matter of checking for an existing variable at a single point in the code, if it isn't there it can default to false. So this would probably the better option. I do not want to include that in this PR however because this PR fills a specific need for a project that I work on and I would rather not bloat it with additional features.

pgorod:
I like the idea of putting all these settings to "flatten" action button dropdowns in one place:

image

Maybe in that area signaled in red, we would have:
1. `Display actions within menus:` - already exists, controls action buttons in detail views, on top.
2. We could have another one for the subpanels (left action buttons) and 
3. A third one for the subpanels (right-hand action buttons in each row). We would have to make the names consistent, so people know what it is about :)

But these are generic, system-wide settings, and I think that your mention of adding another variable to the displayParams would be the more powerful solution.

Finally, I agree that this isn't something to do _now_, and _here_. Maybe I'll just start a new issue as a reminder of this discussion. Thanks!

communig8-public:
Hi, It seems to me that the change in commit bb6a0ea partially conflicts with the "Admin/System Settings/Display actions within menus" option which used to allow us to (when unchecked) have flat buttons turned on for ALL SubPanel menus. Where as now we have to also add layout changes for ALL SubPanels, seems rather cumbersome.

In 7.11.2, as a test to see if I could reset the effect of the change, I edited /include/SunPanel/SubPanelTiles.php at line 446 and set 'flat' to 1 but this had an odd effect on the Activities SubPanel...

screen shot 2019-03-04 at 13 51 19

Mausino:
@communig8-public i got same issue in as you in 7.11.2

aureh12:
Still an issue in 7.11.3

PR #6576, opened 2018-12-05

### LINKED ISSUE #6452: Api/V8 create record does not support unicode and space in attributes

capkuan:
## Description

Change `REGEX_FIELD_PATTERN = '/[^\w-,]/' to REGEX_FIELD_PATTERN = '/[^\w-,\s\]/'`
See issue [#6452](https://github.com/salesagility/SuiteCRM/issues/6452)

## Motivation and Context

When create record with Api/V8 it validate attributes with regex pattern const REGEX_FIELD_PATTERN = '/[^\w-,]/', so if you use unicode or space on value will occur error "The option attributes with value array is invalid." in the response.

## How To Test This

```
POST SuiteCRM/Api/V8/module
BODY {
  "data": {
    "type": "Accounts",
    "attributes": {
      "name": "Cocacola company",
    }
  }
}
```

## Types of changes

### Final checklist

### DIFF
--- a/Api/V8/Param/Options/Fields.php
+++ b/Api/V8/Param/Options/Fields.php
@@ -7,7 +7,7 @@ use Symfony\Component\Validator\Constraints as Assert;
 
 class Fields extends BaseOption
 {
-    const REGEX_FIELD_PATTERN = '/[^\w-,]/';
+    const REGEX_FIELD_PATTERN = '/[^\w-,\s\]/';
 
     /**
      * @inheritdoc


tsummerer:
@capkuan - thanks for finding a fix! However, I wonder if `Api\V8\Param\Options\Attributes` needs it's own regex?

It looks like the `Api\V8\Param\Options\Fields` class validates the 'fields' parameter (used on GET requests to choose which fields are returned). This parameter really _shouldn't_ have spaces or unicode. So I don't think we should change that regex. 

I think the developers were incorrect in trying to reuse the `Fields` regex pattern in the `Attributes` class. It would only make sense if they were checking the attribute keys.

Before your fix, `GET /Api/V8/module/Accounts?fields[Accounts]=name,account type` returns:
```json
{
    "errors": {
        "status": 400,
        "title": null,
        "detail": "The option \"fields\" with value array is invalid."
    }
}
```

After your fix it returns:
```json
{
    "errors": {
        "status": 400,
        "title": null,
        "detail": "The following field in Account module is not found: account type"
    }
}
```

I would say that the first error message is more correct - it boots us out right away for sending a misformed request. 

If you give `Attributes` your regex pattern then it will ensure that we're only passing strings which is probably closer to the desired intention. 

**+1** for this issue being made a priority.

capkuan:
@tsummerer I agree with you that regex pattern use for validating GET request. I think developer should separate the regex pattern for each request in this case.

tsummerer:
@capkuan Right. I think that you should leave the regex in `Api/V8/Param/Options/Fields.php` alone, but instead add your Regex to `Api/V8/Param/Options/Attributes.php` and then have line 25 reference `self::REGEX_FIELD_PATTERN` rather than `Fields::REGEX_FIELD_PATTERN`

This would solve the problem identified in #6452, but also allow the GET request validation to remain. 

Are you interested in making this change? Or should I create a PR?

tsummerer:
Ok. Sounds good. I'll probably do it tonight. 
Thanks for getting this started.

tsummerer:
@capkuan  - I guess this is not valid regex. I didn't have error handling turned high enough on my instance to notice, but that regex throws an error. Because of the error, it lets anything through. If you correct it `'/[^\w-,\s]/'` then you'll notice that characters like `.` or unicode are still rejected. 

I have created a PR #6861 which addresses this by removing the regex check on attributes entirely. I have given you credit for getting the solution started. 

I think it would probably be appropriate for you to close this PR at this time. 

Thanks for your help!

capkuan:
@tsummerer I've closed this PR. Thanks for your suggestions.

PR #6596, opened 2018-12-10
Changes for inline edit of date and datetime

gitbnw:
Fixes my attempt to fix issue #5526

## Description

The inline editing of date fields was still not displaying a value after an edit. 

I believe the cause was php datetime format function would incur a fatal error when attempting to format a date value with a datetime string.  So I split the formatting depending on the field type.

## How To Test This

Use inline editing for any date or datetime field and see if the correct value displays after saving.

## Types of changes

### Final checklist

### DIFF
--- a/include/InlineEditing/InlineEditing.php
+++ b/include/InlineEditing/InlineEditing.php
@@ -423,11 +423,13 @@ function formatDisplayValue($bean, $value, $vardef, $method = "save")
             if ($method != "save") {
                 $value = convertDateUserToDB($value);
             }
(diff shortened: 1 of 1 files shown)


gitbnw:
@gymad which branch should I target?

samus-aran:
Hey @gitbnw The original issue was on both 7.8.x and 7.10.x. Ideally if you need to make a fix for both then create two separate PRs (one to hotfix-7.8.x and hotfix). We don't merge directly into master or 7.8.x.

PR #3078, opened 2017-02-09
Fix #3048 - Append and display messages by type

marin-h:
Feature of setting and displaying different types of color-coded messages for the user. 

suitecrm_msgs

## Description

We had implemented this feature in SugarCRM 6.5 a while ago and now adapted the code for the current version of SuiteCRM.
Basically it allows to set messages by using static method SugarApplication::appendMessage, which is meant to be called from controller.php or any other place in the application. We also added the messages rendering to the header in SugarView.php and in the Smarty templates for each theme available, also setting up the code for the different CSS implementations currently used, in order to show color-coded errors with different styling classes.
As a second approach we also found useful to have a JS implementation for this functionality, so we have added that in sugar_3.js, using the functions displayMessage for the header area and displayMessagesAfterSelector for any other area you wish to show the messages on.
Also, this feature offers the possibility for implementing any new type of color-coded message you need. Now to accomplish that you only have to ensure that the corresponding CSS classes for your custom type exist and are available. For example:

```css
.myCustomType {
background-color: #700614;
color: #CCCCCC;
}
```

https://github.com/salesagility/SuiteCRM/issues/3048

## Motivation and Context

Please refer to issue: https://github.com/salesagility/SuiteCRM/issues/3048

## How To Test This
You need to clean application cache. Also remember to repair compiled CSS stylesheets and minifyied JS cache files. Then run SugarApplicationTest.php with PHPUnit and/or call displayMessage function from JS console into to the browser. The type parameter currently has the following allowed values: info, error, alert, okay, working. When you use these remember to quote them as their expected data-type is string.  See some examples below.

In javascript:
```javascript
// Insert this directly into the browser console
displayMessage('working', 'working class hero');
```
In PHP:
```php
// Put this in controller.php or any other PHP file in the view
SugarApplication::appendMessage('some info', 'info');
```

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [x] My code follows the code style of this project found [here](https://suitecrm.com/wiki/index.php/Coding_Standards).
- [x] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://suitecrm.com/wiki/index.php/Contributing_to_SuiteCRM) guidelines.

PR #6602, opened 2018-12-11
FIX #3048 - Append and display messages by type.

### LINKED ISSUE #3048: Implement a function to send messages to the user
#### Expected Behavior
Sometimes is usefull to send messages to the user when something happens.
Drupal has a function called drupal_set_message() for this:

https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/drupal_set_message/7.x

We have implemented this functionality in a SugarCRM CE instance, and use it to warn the user, or to show the user an error happend, or just to send information.

Here are an example showing the 5 types of messages we have implemented
suite_set_message

This feature is manly used in controller.php and views. Here are a real life example:

sugar_set_message_function

Abuelodelanada:
## Description

This PR replaces outdated #3078
Feature of setting and displaying different types of color-coded messages for the user.

captura

We had implemented this feature in SugarCRM 6.5 a while ago and now adapted the code for the current version of SuiteCRM.
Basically it allows to set messages by using static method SugarApplication::appendMessage, which is meant to be called from controller.php or any other place in the application. We also added the messages rendering to the header in SugarView.php and in the Smarty templates for each theme available, also setting up the code for the different CSS implementations currently used, in order to show color-coded errors with different styling classes.
As a second approach we also found useful to have a JS implementation for this functionality, so we have added that in sugar_3.js, using the functions displayMessage for the header area and displayMessagesAfterSelector for any other area you wish to show the messages on.
Also, this feature offers the possibility for implementing any new type of color-coded message you need. Now to accomplish that you only have to ensure that the corresponding CSS classes for your custom type exist and are available. For example:

```css
.myCustomType {
background-color: #700614;
color: #CCCCCC;
}
```
Also, as @samus-aran asked in old PR #3078  we consider to add code for accessible in the html div itself: role=alert

codigo

## Motivation and Context

Please refer to issue: #3048

## How To Test This

You need to clean application cache. Also remember to repair compiled CSS stylesheets and minifyied JS cache files. Then run SugarApplicationTest.php with PHPUnit and/or call displayMessage function from JS console into to the browser. The type parameter currently has the following allowed values: info, error, alert, okay, working. When you use these remember to quote them as their expected data-type is string. See some examples below.

In javascript:

```
// Insert this directly into the browser console
displayMessage('working', 'working class hero');
```

In PHP:

```
// Put this in controller.php or any other PHP file in the view
SugarApplication::appendMessage('some info', 'info');
```

## Types of changes

### Final checklist


Abuelodelanada:
> @Abuelodelanada sorry i lost little bit in this.. i am not developer, please if you will have little bit time.. should you add here some user case example how we can use in real practical life?

Hi @Mausino 

This is very useful to send messages to the user when something happens..... We use this feature in several cases, for example:

* When a request to a webservise fails
* When a validation in a pre_save action fails
* To inform the user the record is visiting is not editable due to some condition
* To inform the user a process is taking place.
* Etc, etc

We implemented 4 type of messages: info, error, alert, okay, working

include/MVC/SugarApplication.php
```diff
@@ -667,22 +667,23 @@ public static function headerRedirect($header_URL) {
...
      *
      * @access	public
-     * @param string $message
+     * @param string $error_message
      */
-    public static function appendErrorMessage($message) {
-        self::appendMessage('user_error_message', $message);
+    public static function appendErrorMessage($error_message)
+    {
+        SugarApplication::appendMessage($error_message, 'error');
```
gymad: Do not change the session key as it already used as it is in many places.

themes/SuiteP/css/suitep-base/messages.scss
```diff
@@ -0,0 +1,28 @@
+/**** Messages ***/
+
+.message {
```
gymad: you don't have to add specific stylesheet for messages as it is already pre-defined by bootstrap. Can you use the bootstrap css selection, please.
https://getbootstrap.com/docs/4.0/components/alerts/

Abuelodelanada: Great! 
Thanks @gymad

include/MVC/SugarApplication.php
Abuelodelanada: As these messages are not only "error messages" but also info, alert, etc.... I think we should stop using the session key "user_error_messages".
In this PR we use: `$_SESSION['suite_messages'][$type]`

I'm going to change this key everywhere:

```
/m/d/t/SuiteCRM git:develop_3048 ❯❯❯ git grep user_error_message                                                                                       
include/UploadFile.php:                if (!is_array($_SESSION['user_error_message'])) {
include/UploadFile.php:                    $_SESSION['user_error_message'] = array();
include/UploadFile.php:                $_SESSION['user_error_message'][] = $error_message;
include/UploadMultipleFiles.php:                if (!is_array($_SESSION['user_error_message'])) {
include/UploadMultipleFiles.php:                    $_SESSION['user_error_message'] = array();
include/UploadMultipleFiles.php:                $_SESSION['user_error_message'][] = $error_message;
modules/FP_events/controller.php:            $_SESSION['user_error_message'] = array();//clear the error message array
modules/FP_events/controller.php:            $_SESSION['user_error_message'] = array();
modules/FP_events/controller.php:            $_SESSION['user_error_message'] = array();
tests/unit/codecept/include/MVC/SugarApplicationTest.php:        //execute the method and check that the method adds the message to user_error_message array.
tests/unit/codecept/include/MVC/SugarApplicationTest.php:        $user_error_message_count = count($_SESSION['suite_messages']['error']);
tests/unit/codecept/include/MVC/SugarApplicationTest.php:        $this->assertGreaterThan($user_error_message_count, count($_SESSION['suite_messages']['error']
tests/unit/codecept/include/MVC/SugarApplicationTest.php:            $user_error_message_count = count($_SESSION['suite_messages'][$type]);
tests/unit/codecept/include/MVC/SugarApplicationTest.php:            $this->assertGreaterThan($user_error_message_count, count($_SESSION['suite_messages'][$typ

```

gymad: You right, we should but we can't as it won't be backward compatible to any possibly exists customizations, so please don't do that.

Abuelodelanada: As I understand the only point that may have backwards incompatibility is where we read the $SESSION keys: user_error_message and user_success_message. And we do this in the method "getMessages()" in SugarApplication.... In develop branch this method is:

```php
    protected static function getMessages($type) {
        
        self::validateMessageType($type);

        if (isset($_SESSION[$type]) && is_array($_SESSION[$type])) {
            $msgs = $_SESSION[$type];
            unset($_SESSION[$type]);
            return $msgs;
        } else {
            return array();
        }
    }
```
And it is called like this in include/MVC/SugarApplication.php:

```php
$messages = self::getMessages('user_error_message');
$messages = self::getMessages('user_success_message');
```


And in my PR is:

```php
    public static function getMessages($type = null, $clear_queue = true)
    {
        $messages = array();
        if (isset($_SESSION['suite_messages'])) {
            if (isset($type)) {
                if (isset($_SESSION['suite_messages'][$type])) {
                    $messages[$type] = $_SESSION['suite_messages'][$type];
                    if ($clear_queue) {
                        unset($_SESSION['suite_messages'][$type]);
                    }
                }
            } else {
                $messages = $_SESSION['suite_messages'];
                if ($clear_queue) {
                    unset($_SESSION['suite_messages']);
                }
            }
        }
        }
        return($messages);
    }
    }
```

I will modify the method in my PR in order to do not break backwards compatibility for custom developments.

samus-aran:
Hey @Abuelodelanada Travis is failling. I looks like its due to the new test.

`Time: 16.03 seconds, Memory: 110.00MB

There was 1 error:

---------
1) [35;1mSugarApplicationTest:[39;22m Testappend message
[37;41;1m Test [39;49;22m tests/unit/codecept/include/MVC/SugarApplicationTest.php:testappendMessage
[37;41;1m                                                              [39;49;22m
[37;41;1m  [PHPUnit_Framework_Exception] Undefined variable: _SESSION  [39;49;22m
[37;41;1m                                                              [39;49;22m
#1  /home/travis/build/salesagility/SuiteCRM/tests/unit/codecept/include/MVC/SugarApplicationTest.php:500
#2  SugarApplicationTest->testappendMessage`

@gymad Are we even able to use Sessions? Shouldn't this be mocked rather than relying on the session of the application during a test?

include/MVC/SugarApplication.php
Abuelodelanada: Hi @gymad 
I agree with you, that's why I [commit some changes](https://github.com/salesagility/SuiteCRM/pull/6602/commits/d642489f9e4ec0d4c9cf13e0106b25026bd0ab3f) after your advice.

Don't you think this commit is working good in order to have backwards compatibility ?
Please let me know if you think is not good enough.

gymad: Yes, maybe it could works but I still don't understand why you don't just following the original concept and adding some extra keys like `user_alert_message`, `user_info_message` and so on.. it makes your code complexity much more simple and you don't have to handle the "old message types"?

include/MVC/SugarApplication.php
```diff
@@ -704,49 +706,77 @@ public static function getSuccessMessages() {
...
+     *
+     * @param string $message the message to be displayed to the user.
+     * For consistency with other messages, it should begin with a capital letter and end with a period.
+     *
+     * @param string $type the type of the message. Values allowed: error, info, alert, okay, warning.
+     *
+     * @param boolean $repeat if this is FALSE and the message is already set, then the message won't be repeated.
+     *
+    */
+    public static function appendMessage($message, $type = 'info', $repeat = true)
```
gymad: Do not change tha places of parameters.

include/MVC/SugarApplication.php
```diff
@@ -704,49 +706,77 @@ public static function getSuccessMessages() {
...
+            SugarApplication::appendMessage($_SESSION['user_error_message'], 'error');
         }
-    }
 
-    /**
-     * 
-     * @param string $type possible message types: ['user_error_message', 'user_success_message']
-     * @throws Exception message type should be valid
-     */
-    protected static function validateMessageType($type) {
```
gymad: Do not remove the validation of message type. (for e.g. nothing to shows up a misspelled index enimore?)

include/MVC/SugarApplication.php
Abuelodelanada: Fixed in commit: https://github.com/salesagility/SuiteCRM/pull/6602/commits/163d288e7218bb236cef64775f5d95d146396826

include/MVC/SugarApplication.php
Abuelodelanada: You are right! Sorry!
Fixed in commit: https://github.com/salesagility/SuiteCRM/pull/6602/commits/7d5473201c4a1d912eb4378418ad8ca73bd3e344

Abuelodelanada:
Hi @gymad 

> I'm in the concept of message types and realy like it however I also think this feature code could be more simply and clean some point (see my requested changes). I suggest to rethink and refactoring these changes, in focus:
> 
>     backward compatibility and upgrade and custom safe way
>     code complexity
>     using exists library (for e.g bootstrap css) instead unique code etc.

- About " backward compatibility and upgrade and custom safe way"  this is solved by the new private function "_appendOldMessageTypes()" 

- About code complexity, I refactored some methods, please see commit https://github.com/salesagility/SuiteCRM/pull/6602/commits/4628815aa82785d4768ef2c52d315a05ec84a88d If you think it could be better, please let me know.

- About "using exists library (for e.g bootstrap css) instead unique code etc."  I fixed this using bootstrap, please see commit https://github.com/salesagility/SuiteCRM/pull/6602/commits/6ea53450cf0557e21ef3abf8230d99b2b98e9efd Do you think there is another existing libraries to use?

include/MVC/SugarApplication.php
Abuelodelanada: When we developed this feature (many years ago), we didn't use the original concept because the messages are not all together in an array, there are as many arrays in $_SESSION as many error types.:

```php
$_SESSION['user_error_message']
$_SESSION['user_success_message']
etc
```
We took the idea from Drupal. As you can see in our original implementation in issue #3048 was a copy of [Drupal implementation](https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/drupal_set_message/7.x). After that, we improved the original implementation in closed PR #3078 and after almost 2 years sleeping, we re-write the PR in order to update it to the new codebase and to get the PR accepted. We also added @samus-aran [suggestion](https://github.com/salesagility/SuiteCRM/pull/3078#issuecomment-445176954)

In this PR the messages are all together in 'suite_messages' key.

`$_SESSION['suite_messages'][$type] = array();`

Managing the old messages types is not complex and also assures backwards compatibility:

https://github.com/salesagility/SuiteCRM/pull/6602/files#diff-3461d64b9728cb94323089f5674179f5R829

Abuelodelanada: @gymad I made a mistake and closed this conversation. I reopen it.

Abuelodelanada:
Hi @gymad 

I left you a comment here:

https://github.com/salesagility/SuiteCRM/pull/6602#issuecomment-448651015

Is there anything else I should do in order to have this PR approved?

Dillon-Brown:
Assessed :+1: . I've also fixed the conflicts and cleaned up some inspections. Travis should hopefully be passing now.

Abuelodelanada:
Hi @samus-aran @gymad @Dillon-Brown 

Having this PR merged into `hotfix-7.10.x` will save us **many** days of work and headaches. If we create another PR, do you think this enhancement could be in 7.10.15?

cameronblaikie:
Assessed :+1:

Mac-Rae:

PR #6641, opened 2018-12-18
Browser search bar support for SuiteCRM

gunnicom:
Superseeds https://github.com/salesagility/SuiteCRM/pull/3677

Firefox allows new search plugins with opensearch ( http://www.opensearch.org/Home ).
This small patch allows sugar as search plugin for browser search bars.


gunnicom:
Wrong branch

PR #6672, opened 2018-12-19
FIX #508 - Minify and concatenate JS files included in metadata files

### LINKED ISSUE #508: Add multiple JS files to a view in only one minifyed JS file.

Abuelodelanada:
## Description

This PR fixes issue #508
After an interesting debate in PR #3009 with @chris001 and @clemente-raposo I closed that PR and open this one.

The way to add JS files to a view is through metadata files, such as editviewdefs.php.

editviewdefs2

And get the files included in the HTML produced:

console1

This is simple, but has two problems:

- Each file is not minifyed.

I modified Smarty function `smarty_function_sugar_include()` and add a method called `joinAndMinifyJSFiles()` in `SugarMin` class.

So if you add one or more JS files in, for example editviewdefs.php, this PR will produce only one concatenated and minified JS file.

console2

## Motivation and Context

This PR will reduce requests in the web server and network traffic.

## How To Test This

Include JS files in metadata file and verify that only one concatenated and minified JS file is served.

## Types of changes

### Final checklist


clemente-raposo:
Hi @Abuelodelanada,

Thank you. 

Looks good. : )

Dillon-Brown:
Assessed

jssource/jsmin.php
```diff
@@ -45,4 +45,50 @@ protected function jsParser() {
...
+     *
+     * @author Jose C. Massón <jose@gcoop.coop>
+     *
+     * @return Minified JS file path
+     */
+    static public function joinAndMinifyJSFiles($jsFiles)
+    {
+        $target = SugarThemeRegistry::current()->getJSPath()
+                . '/' .
+                sha1(implode('|', $jsFiles)) . '.js';
```
clemente-raposo: Hi @Abuelodelanada,

yet again, thanks for this contribution, very cool.

When reviewing the code, two question came to mind:

- If two different modules have a file with the same name, like `edit.js` , won't they both hit the same cached file? thus overwriting one another?

- Another thing, about the ThemeCache, I'm not sure if all files are removed from the theme js cache, when doing a Repair and Rebuild. I just gave a quick look at the repair and rebuild code, so maybe wrong.
    - When testing, did the changes to existing js files get onto the cache and get refreshed on the browser?

jssource/jsmin.php
```diff
@@ -45,4 +45,50 @@ protected function jsParser() {
...
+                $customJSContents .= $jsFileContents;
+            }
+
+            $customJSPath = create_cache_directory($target);
+
+            if ((!inDeveloperMode()) && (!is_file($customJSPath))) {
+                $customJSContents = SugarMin::minify($customJSContents);
+            }
+
+            sugar_file_put_contents($customJSPath, $customJSContents);
```
gymad: sugar_file_put_contents returns: int - Returns the number of bytes written to the file, false otherwise. Please handle the possible error (false) returns.

jssource/jsmin.php
```diff
@@ -45,4 +45,50 @@ protected function jsParser() {
...
+    /**
+     * Join and minify JS files
+     *
+     * @param array $jsFiles an 'array' of js files
+     *
+     * @author Jose C. Massón <jose@gcoop.coop>
+     *
+     * @return Minified JS file path
+     */
+    static public function joinAndMinifyJSFiles($jsFiles)
```
gymad: try to avoid the static calls.

jssource/jsmin.php
```diff
@@ -45,4 +45,50 @@ protected function jsParser() {
...
+            $customJSPath = create_cache_directory($target);
+
+            if ((!inDeveloperMode()) && (!is_file($customJSPath))) {
+                $customJSContents = SugarMin::minify($customJSContents);
+            }
+
+            sugar_file_put_contents($customJSPath, $customJSContents);
+        }
+
+        $customJSURL = sugar_cached($target);
```
gymad: This value already calculated in line https://github.com/salesagility/SuiteCRM/pull/6672/files#diff-bd08177a986e9fc8d14cd7437dbe9195R64. Assign it into a variable and use it instead calculet it twice.

jssource/jsmin.php
```diff
@@ -45,4 +45,50 @@ protected function jsParser() {
...
+            $customJSContents = '';
+
+            foreach ($jsFiles as $jsFileName) {
+                $jsFileContents = '';
+
+                if (is_file($jsFileName)) {
+                    $jsFileContents .= sugar_file_get_contents($jsFileName);
+                }
+
+                if (empty($jsFileContents)) {
```
gymad: $jsFileContents could be empty in each case:
- $jsFileName is not a file
- $jsFileName is a file but sugar_file_get_contents return a false as error
- $jsFileName is a file and sugar_file_get_contents return an empty string as $jsFineName is empty.
Also: in the next line in the log note nothing suggests that which case happend, it's just showing up something went wrong. Please handle the error propery.

jssource/jsmin.php
```diff
@@ -45,4 +45,50 @@ protected function jsParser() {
...
+                sha1(implode('|', $jsFiles)) . '.js';
+
+        if (!is_file(sugar_cached($target))) {
+            $customJSContents = '';
+
+            foreach ($jsFiles as $jsFileName) {
+                $jsFileContents = '';
+
+                if (is_file($jsFileName)) {
+                    $jsFileContents .= sugar_file_get_contents($jsFileName);
```
gymad: You assuming it can not failing but sugar_file_get_contents() function could return `false` as error too. Please handle the error.

jssource/jsmin.php
```diff
@@ -45,4 +45,50 @@ protected function jsParser() {
...
+                }
+
+                if (empty($jsFileContents)) {
+                    LoggerManager::getLogger()->warn(
+                        translate('ERR_FILE_EMPTY') .': '. "{$jsFileName}"
+                    );
+                }
+                $customJSContents .= $jsFileContents;
```
gymad: value of $jsFileContents could came from sugar_file_get_contents(). In case if function sugar_file_get_contents() failing (for e.g because the file is not readable) then the error results (`false`) just concatenated to $customJSContents but the `false` converted to an empty string just like if there is no concatenation. It means if a file permission problem (or anything) occured we never will realise it. Please handle the error results properly.

jssource/jsmin.php
```diff
@@ -45,4 +45,50 @@ protected function jsParser() {
...
+            }
+
+            $customJSPath = create_cache_directory($target);
```
gymad: create_cache_directory() calls sugar_mkdir() and it returns false or throws an exception if an error occured. This error  results assigned into variable $customJSPath. It means we never will realise if something went wrong. Please handle the possible errors.

jssource/jsmin.php
```diff
@@ -45,4 +45,50 @@ protected function jsParser() {
...
+
+            if ((!inDeveloperMode()) && (!is_file($customJSPath))) {
```
gymad: Too many extra parentheses. The following expressions means exactly same thing but simple:
```
-> !inDeveloperMode() && !is_file($customJSPath)
-> !(inDeveloperMode() || is_file($customJSPath))
```
(also I don't really understand why developer mode check is necessary here?)

jssource/jsmin.php
```diff
@@ -45,4 +45,50 @@ protected function jsParser() {
...
+                $customJSContents = SugarMin::minify($customJSContents);
```
gymad: SugarMin::minify could throws an exception at line https://github.com/salesagility/SuiteCRM/blob/master/jssource/jsmin.php#L43 but I can not see the point where it's handled.

include/Smarty/plugins/function.sugar_include.php
```diff
@@ -78,14 +79,22 @@ function smarty_function_sugar_include($params, &$smarty)
 	   	  foreach($params['include'] as $include) {
 	   	  	      if(isset($include['file'])) {
 	   	  	         $file = $include['file'];
-	   	  	         if(preg_match('/[\.]js$/si',$file)) {
-	   	  	            $code .= "<script src=\"". getJSPath($include['file']) ."\"></script>";
-	   	  	         } else if(preg_match('/[\.]php$/si', $file)) {
+                     if (preg_match('/tiny_mce.*[\.]js$/si', $file)) {
```
gymad: Why tiny_mce files are exceptionals?

jssource/jsmin.php
Abuelodelanada: Hi @clemente-raposo 

Thanks for the feedback!

> * If two different modules have a file with the same name, like edit.js , won't they both hit the same cached file? thus overwriting one another?

You are right. I'll fix it!

> * Another thing, about the ThemeCache, I'm not sure if all files are removed from the theme js cache, when doing a Repair and Rebuild. I just gave a quick look at the repair and rebuild code, so maybe wrong.

I'll check it

> * When testing, did the changes to existing js files get onto the cache and get refreshed on the browser?

When testing you should delete cache files

jssource/jsmin.php
```diff
@@ -45,4 +45,50 @@ protected function jsParser() {
...
+            foreach ($jsFiles as $jsFileName) {
+                $jsFileContents = '';
+
+                if (is_file($jsFileName)) {
+                    $jsFileContents .= sugar_file_get_contents($jsFileName);
+                }
+
+                if (empty($jsFileContents)) {
+                    LoggerManager::getLogger()->warn(
+                        translate('ERR_FILE_EMPTY') .': '. "{$jsFileName}"
```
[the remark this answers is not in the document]
Abuelodelanada: Ok!

jssource/jsmin.php
Abuelodelanada: Ok.
We check If you are in developer mode because when developing JS you don't want to have JS concatenated and minifyed...It is not easy to debug JS.

include/Smarty/plugins/function.sugar_include.php
Abuelodelanada: We have tiny_mce files in two editviewdefs.php:

```
modules/AOK_KnowledgeBase/metadata/editviewdefs.php:          'file' => 'include/javascript/tiny_mce/tiny_mce.js',
modules/AOS_PDF_Templates/metadata/editviewdefs.php:          'file' => 'include/javascript/tiny_mce/tiny_mce.js',
```

And Tiny MCE stops working when you concatenate it to another minyfied JS file.... I spent many hours debugging it and I really cannot find the reason of this behavior ... I find this in my first commit because a selenium test fails.

For this I decided to treat this files in an exceptional way... besides tiny_mce files are already minyfied so it is not so bad

Abuelodelanada: We are using it even before initial commit... That comment is there since 06/06/2012, commit c91080dc82 in  SugarCRM CE 6.5.0 code! :-0


```bash
/m/d/t/gca git:release/19 ❯❯❯ git blame jssource/jsmin.php | head                                                                                                                                                                                            
c91080dc82 (John Mertic 2012-06-06 20:57:07 -0400  1) <?php
c91080dc82 (John Mertic 2012-06-06 20:57:07 -0400  2)
c91080dc82 (John Mertic 2012-06-06 20:57:07 -0400  3) class SugarMin {
c91080dc82 (John Mertic 2012-06-06 20:57:07 -0400  4)
c91080dc82 (John Mertic 2012-06-06 20:57:07 -0400  5)     /**
c91080dc82 (John Mertic 2012-06-06 20:57:07 -0400  6)      * jsParser will take javascript source code and minify it.
c91080dc82 (John Mertic 2012-06-06 20:57:07 -0400  7)      *
c91080dc82 (John Mertic 2012-06-06 20:57:07 -0400  8)      * Note: There is a lot of redundant code since both passes
c91080dc82 (John Mertic 2012-06-06 20:57:07 -0400  9)      * operate similarly but with slight differences. It will probably
c91080dc82 (John Mertic 2012-06-06 20:57:07 -0400 10)      * be a good idea to refactor the code at a later point when it is stable.
```

I get your point about the fact that that code is not stable...but we are using that code every day when we execute "Repair JS Files" from the Administration module. 

Anyway I agree with you, we should refactor that code. 
What about using that code for this PR, and refactor that code in another PR? After all we are living with that code since always, and I think there were no big issues with it.

Abuelodelanada:
Hi @gymad 

About TinyMCE and why I treated its files in an exceptional way: I was debugging and find that the problem is that TinyMCE load plugins (.js files) on the fly.

The following screenshot is when the code of this PR do not minify and concatenate tiny_mce.js file:

tiny_mce_ok

As you can see, the plugins are loaded on the fly ok.

But If we minify and concatenate tiny_mce.js file (and thus change the name and location of the resultant js file), the following happens:

tiny_mce_nook

As you can see in the last screenshot, TinyMCE fails to load the plugins on the fly, because fails in discovering the right path to the plugins files.

I think this is a TinyMCE issue.

jssource/jsmin.php
Abuelodelanada: Done in commit 125dadf36c6b4aabf9ccf44e019405f60651c7f2

jssource/jsmin.php
Abuelodelanada: Done in commit e3c2726e86236ba107e2a466406bfb7d081eebfc

jssource/jsmin.php
Abuelodelanada: @gymad when you talk about the "js grouping functionality" you mean the file `jssource/JSGroupings.php` ??

jssource/jsmin.php
Abuelodelanada: Done

jssource/jsmin.php
Abuelodelanada: Done

jssource/jsmin.php
Abuelodelanada: Done

jssource/jsmin.php
Abuelodelanada: Done

jssource/jsmin.php
Abuelodelanada: Done

jssource/jsmin.php
Abuelodelanada: Done

jssource/jsmin.php
Abuelodelanada: Done

include/Smarty/plugins/function.sugar_include.php
Abuelodelanada: Hi @gymad 

> The minified version of js files should doing exactly same as the original, even if the Minifier class works as expected

About TinyMCE and why I treated its files in an exceptional way: I was debugging and find that the problem is that TinyMCE load plugins (.js files) on the fly.

The following screenshot is when the code of this PR do not minify and concatenate tiny_mce.js file:

tiny_mce_ok

As you can see, the plugins are loaded on the fly ok.

But If we minify and concatenate tiny_mce.js file (and thus change the name and location of the resultant js file), the following happens:

tiny_mce_nook

As you can see in the last screenshot, TinyMCE fails to load the plugins on the fly, because fails in discovering the right path to the plugins files.

I think this is a TinyMCE issue not something wrong in joinAndMinifyJSFiles function.

jssource/minify_utils.php
```diff
@@ -504,3 +504,102 @@ function BackUpAndCompressScriptFiles($from_path,$to_path = '', $backup = true){
...
+ *
+ * @param array $jsFiles an 'array' of js files
+ *
+ * @author Jose C. Massón <jose@gcoop.coop>
+ *
+ * @return Minified JS file path
+ */
+function joinAndMinifyJSFiles($jsFiles)
+{
+    $jsFiles = array_map('realpath', array_unique($jsFiles));
```
gymad: More verbose if you do something like this:
```
$jsFilesReal = [];
foreach ($jsFiles as $jsFilesKey => $jsFile) {
  $jsFileReal = realpath($jsFile);
  if (!$jsFileReal) {
    LoggerManager::getLogger()->error('joinAndMinifyJSFiles - Given JS File path is wrong: ' . $jsFile);
  } else {
    if (in_array($jsFileReal, $jsFilesReal)) {
      LoggerManager::getLogger()->warn('joinAndMinifyJSFiles - Given JS File is duplicated: ' . $jsFile);
    } else {
      $jsFilesReal[$jsFilesKey] = $jsFileReal;
    }
  }
}
$jsFiles = $jsFilesReal;
```

include/Smarty/plugins/function.sugar_include.php
```diff
@@ -78,14 +80,26 @@ function smarty_function_sugar_include($params, &$smarty)
 	   	  foreach($params['include'] as $include) {
 	   	  	      if(isset($include['file'])) {
 	   	  	         $file = $include['file'];
```
gymad: I think if you add something like this is more handy to see if something went wrong at this point:
```
$realpath = realpath($include['file']);
if (!$realpath) {
  LoggerManager::getLogger()->error('smarty_function_sugar_include - Given file path is incorrect: ' . $include['file']);
} else {
  $file = $realpath;
  ...
```

include/Smarty/plugins/function.sugar_include.php
```diff
@@ -78,14 +80,26 @@ function smarty_function_sugar_include($params, &$smarty)
...
-	   	  	         if(preg_match('/[\.]js$/si',$file)) {
-	   	  	            $code .= "<script src=\"". getJSPath($include['file']) ."\"></script>";
-	   	  	         } else if(preg_match('/[\.]php$/si', $file)) {
+                     if (preg_match('/tiny_mce.*[\.]js$/si', $file)) {
+                         $code .= "<script src=\"".$file ."\"></script>";
+	   	  	         } else if(preg_match('/[\.]js$/si',$file)) {
+                         $jsFiles[] = $file;
```
gymad: Instead this:
```
if (in_array($jsFiles, $jsFile)) {
  LoggerManager::getLogger()->warn('smarty_function_sugar_include - JS file already added: ' . $file);
} else {
   $jsFiles[] = $file;
}

jssource/minify_utils.php
Abuelodelanada: Done

include/Smarty/plugins/function.sugar_include.php
Abuelodelanada: Done

include/Smarty/plugins/function.sugar_include.php
Abuelodelanada: Done

Abuelodelanada:
Hi @gymad 

I made some commits to follow your advice.
Please let me know if I miss something.

Abuelodelanada:
Hi @gymad 

I left you a comment here:

https://github.com/salesagility/SuiteCRM/pull/6672#issuecomment-452024743

Is there anything else I should do in order to have this PR approved?

Abuelodelanada:
> @Abuelodelanada Did you run into any code not using `$sugar_config['cache_dir']`? I just attempted to change the path of `$sugar_config['cache_dir']` from the default `'./cache'` to anything else, it looks like there are a lot of templates and methods which include javascript from the hard-coded path of `cache/`.
> 
> Thoughts before I open a new issue?

I @isleshocky77 

I do not use hardcoded paths in this PR.
The concatenated and minifyed files are stored in:

`$sugar_config['cache_dir']/themes/SuiteP/js/`

https://github.com/salesagility/SuiteCRM/pull/6672/files#diff-767d4991d10f991262fcc0a503e54dbaR521

I tested this PR changing the  `$sugar_config['cache_dir']` from the default `'./cache'` to anything else and works as expected.

isleshocky77:
@Abuelodelanada 

I don't see you touching the piece of code which was the issue I was running into (I didn't fully debug), but it looks like [include/MVC/View/SugarView.php:927](https://github.com/salesagility/SuiteCRM/blob/80e9e484f7c00340a67a65efbf82aca2e3d5d141/include/MVC/View/SugarView.php#L927) has 'cache/' hard-coded into generation and doesn't get fixed below before calling [include/MVC/View/SugarView.php:947](https://github.com/salesagility/SuiteCRM/blob/80e9e484f7c00340a67a65efbf82aca2e3d5d141/include/MVC/View/SugarView.php#L947) and also isn't changed anywhere in the downstream functions.

Thoughts?  Does your code not touch this js?

Abuelodelanada:
> @Abuelodelanada
> 
> I don't see you touching the piece of code which was the issue I was running into (I didn't fully debug), but it looks like [include/MVC/View/SugarView.php:927](https://github.com/salesagility/SuiteCRM/blob/80e9e484f7c00340a67a65efbf82aca2e3d5d141/include/MVC/View/SugarView.php#L927) has 'cache/' hard-coded into generation and doesn't get fixed below before calling [include/MVC/View/SugarView.php:947](https://github.com/salesagility/SuiteCRM/blob/80e9e484f7c00340a67a65efbf82aca2e3d5d141/include/MVC/View/SugarView.php#L947) and also isn't changed anywhere in the downstream functions.
> 
> Thoughts? Does your code not touch this js?

My code does not touch this JS.

I think this bugs are easy to solve using:

`SugarConfig::getInstance()->get("cache_dir");`

isleshocky77:
> I tested this PR changing the `$sugar_config['cache_dir']` from the default `'./cache'` to anything else and works as expected.

I was more surprised by this statement. Had deleted your `/cache/` directory and checked to see if everything was functioning correctly? I'm seeing the homepage with no styling.

Abuelodelanada:
> > I tested this PR changing the `$sugar_config['cache_dir']` from the default `'./cache'` to anything else and works as expected.
> 
> I was more surprised by this statement. Had deleted your `/cache/` directory and checked to see if everything was functioning correctly? I'm seeing the homepage with no styling.

I mean this PR works as expected.
You are right! I did not delete cache/ directory .... I only changed "cache_dir" in sugar_config and verify that a new directory was created and my PR works

Mac-Rae:

PR #6674, opened 2018-12-20
Fixed #4872 - Fix some async subpanel actions failing if refresh_page=1

### LINKED ISSUE #4872: PHP errors on user page in Security Groups Management tab
#### Issue
Warning: Declaration of SugarWidgetSubPanelEditSecurityGroupUserButton::displayDetail($layout_def) should be compatible with & SugarWidgetField::displayDetail($layout_def) in D:\OSPanel\domains\crm\include\generic\SugarWidgets\SugarWidgetSubPanelEditSecurityGroupUserButton.php on line 78

Warning: Declaration of SugarWidgetSubPanelEditSecurityGroupUserButton::displayHeaderCell(&$layout_def) should be compatible with SugarWidgetField::displayHeaderCell($layout_def) in D:\OSPanel\domains\crm\include\generic\SugarWidgets\SugarWidgetSubPanelEditSecurityGroupUserButton.php on line 78

Warning: Declaration of SugarWidgetSubPanelEditSecurityGroupUserButton::displayList(&$layout_def) should be compatible with SugarWidgetField::displayList($layout_def) in D:\OSPanel\domains\crm\include\generic\SugarWidgets\SugarWidgetSubPanelEditSecurityGroupUserButton.php on line 78

#### Your Environment
* SuiteCRM Version used: 7.9.8
* Browser name and version: Firefox Quantum 57.0.4   64x
* Environment name and version: PHP 7.0.21, Apache 20120211, mysqlnd 5.0.12-dev
* Operating System: OS panel 5.2.7 on Win10

lazka:
## Description

In case the subpanel layout uses refresh_page=1 then showSubPanel() will start an
async request and after that reload the page. In many cases this will lead to the
just started request being aborted and the page gets reloaded without anything changing.

This was introduced in 829a06f5b8af where the previously sync call was replaced with an
async one.

This is most visible for me in the "Users" module were I can't remove security groups
through the subpanel and it just refreshes the page without removing the group.

Fix this my moving the page refresh check in the ajax callback so the page gets only reloaded
once the request is finished.

## Motivation and Context

Make adding and removing security groups and roles in the Users module work
for example.

## How To Test This

(Using Firefox, in case that makes a difference)

1) Admin -> User Management
2) Select a user
3) Scroll down to the security groups subpanel
4) Add a security group
5) Click "Edit" on the newly added group and then "Remove"

Without this patch the page reloads and the security group is still there.

## Types of changes

### Final checklist

### DIFF
--- a/include/SubPanel/SubPanelTiles.js
+++ b/include/SubPanel/SubPanelTiles.js
@@ -70,13 +70,13 @@ function got_data(args,inline){var list_subpanel=document.getElementById('list_s
 SUGAR.util.evalScript(args.responseText);subpanel.style.display='';set_div_cookie(subpanel.cookie_name,'');if(current_child_field!=''&&child_field!=current_child_field){}
 current_child_field=child_field;$("ul.clickMenu").each(function(index,node){$(node).sugarActionMenu();});}}
 function showSubPanel(child_field,url,force_load,layout_def_key){var inline=1;if(typeof(force_load)=='undefined'||force_load==null){force_load=false;}
+function checkRefreshPage(url){if(typeof(url)!='undefined'&&url!=null&&url.indexOf('refresh_page=1')>0){document.location.reload();}}
 if(force_load||typeof(child_field_loaded[child_field])=='undefined'){request_map[request_id]=child_field;if(typeof(url)=='undefined'||url==null){var module=get_module_name();var id=get_record_id();if(typeof(layout_def_key)=='undefined'||layout_def_key==null){layout_def_key=get_layout_def_key();}
 url='index.php?sugar_body_only=1&module='+module+'&subpanel='+child_field+'&action=SubPanelViewer&inline='+inline+'&record='+id+'&layout_def_key='+layout_def_key;}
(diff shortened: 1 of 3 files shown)


lazka:
@PedroErnst since you made the async change, could you have a quick look maybe? thanks!

pgorod:
Is this the same Issue as #4872 ?

In one of the comments a user claims he can't delete security groups from the user's detail view.

pgorod:
@lazka it seems your patch solves the problem mentioned in a comment in #4872.

If you can add a fix for the PHP Warnings that would be great, we could close that Issue. It's normally very easy, it's just a matter of matching parameter definitions. I am assuming you get the same errors on your installation?

lazka:
> @lazka it seems your patch solves the problem mentioned in a comment in #4872.

Thanks. I've added `Fixed #4872` to the first commit and the PR title.

> If you can add a fix for the PHP Warnings that would be great, we could close that Issue. It's normally very easy, it's just a matter of matching parameter definitions. I am assuming you get the same errors on your installation?

I've added a second commit fixing the warning.

Dillon-Brown:
Assessed

PR #6714, opened 2019-01-04
fix last page when modulus is zero

gymad:
## Description

pager fix (last page could contains zero element)
## Types of changes

### Final checklist

### DIFF
--- a/lib/Search/UI/SearchResultsController.php
+++ b/lib/Search/UI/SearchResultsController.php
@@ -99,7 +99,7 @@ class SearchResultsController extends Controller
                 'prev' => $page > 1,
                 'next' => $total - $from > $size,
                 'page' => $page,
-                'last' => (int)($total / $size) + 1,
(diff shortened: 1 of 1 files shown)


lib/Search/UI/SearchResultsController.php
```diff
@@ -99,7 +99,7 @@ public function display()
                 'prev' => $page > 1,
                 'next' => $total - $from > $size,
                 'page' => $page,
-                'last' => (int)($total / $size) + 1,
+                'last' => (int)($total / $size) + ($total%$size === 0 ? 0 : 1),
```
code-ph0y: This will throw an error if $size is 0

Example:
index.php?action=Search&search-query-string=Asia&search-query-size=0&search-query-from=0&search-engine=ElasticSearchEngine

gymad: Yes, It's right, however `search-query-size` query parameter never can be 0. Also I added a commit to handle it in back-end too.

PR #6744, opened 2019-01-09
Fix emails losing confirmed opt-in when converting a lead to a contact

lazka:
## Description

When confirmed opt-in is enabled then converting a lead with a confirmed email
address to a contact makes the address lose its confirmation.

AddUpdateEmailAddress only gets a flag if opt-in is set or not, and when trying to update
the existing email record it replaces the COI_STAT_CONFIRMED_OPT_IN state with just
COI_STAT_OPT_IN because of that.

To fix this, in case the passed opt-in flag is 1 and the existing record has some kind
of opt-in state (!COI_STAT_DISABLED) we use that existing state to update the record instead
of forcing COI_STAT_OPT_IN.

In case of a lead conversion where nothing is changed this result in no DB update being
performed on the address.

## Motivation and Context

I'd prefer if converting a lead wouldn't change the opt-in status of the
linked email address.

## How To Test This

* Create a lead, set email and opt-in flag
* Send a confirmation email and wait until confirmed
* Convert the lead to a contact

## Types of changes

### Final checklist

### DIFF
--- a/include/SugarEmailAddress/SugarEmailAddress.php
+++ b/include/SugarEmailAddress/SugarEmailAddress.php
@@ -1268,7 +1268,15 @@ class SugarEmailAddress extends SugarBean
             $isValidEmailAddress
             && (int)$optInFlag === 1
         ) {
-            $new_confirmed_opt_in = self::COI_STAT_OPT_IN;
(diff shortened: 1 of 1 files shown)


cameronblaikie:
Assessed :+1:

PR #6745, opened 2019-01-09
Fixed #6037 - Reports: Fix adding relations from the first tree node

### LINKED ISSUE #6037: AOR Reports - Issue with related records in reports
#### Issue
Report will return no records if either the report fields or conditions have a record from a related module.
E.g. If the report module is Contacts and the report field is Contacts:Email Address the report will return no records.
image

#### Expected Behavior
The report should return first name, last name and contacts email address.

#### Actual Behavior
Report return no records.
SQL Error: Unknown column 'contacts.email_address' in 'field list'.

#### Possible Fix

Issue is when saving conditions and fields in AOR_Condition.php and AOR_Field.php
Issue may be when module_path is encoded.
$postField[$i] = base64_encode(serialize(explode(":", $postField[$i])));

In previous versions: $postField[$i] = 'email_address'
In 7.10.6:  $postField[$i] = ':email_address'
Colon appear to be added twice.

#### Steps to Reproduce

1. Create new AOR Report
2. Make the report module Contacts
3. Choose a related record to contacts like email
4. Run Report

#### Context

Medium priority. 

#### Your Environment

* SuiteCRM Version used: 7.10.6
* Browser name and version: Chromium Version 66.0
* Environment name and version: MySql - PHP 7.1

lazka:
## Description

The main report module is added to the module tree as well, but with an empty
module_path. When selecting related modules the module paths get concatenated
with the root module which is empty in this case and results in a wrong module
path being stored in the report.

This patch works around the issue by checking for an empty module_path before adding
the next module to the path.

Note that this wont fix existing reports that are broken because of this because
the wrong path is saved in the database and persists.

There is really no difference in the children of the first module and the other
toplevel nodes, so ideally the main module wouldn't have any child nodes in the first
place, but this would require more work and can be changed another time.

## Motivation and Context

It's currently easy to create reports which just show empty results without any
feedback because of wrong module paths which result in broken SQL.

Should fix #6037

## How To Test This

* Create a report
* Select Leads
* Select the first tree module node ("Leads")
* Add first name, last name to the report
* Select the "Email Address" sub-node of the "Leads" node
* Add "Email Address" to the report
* Save

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOR_Reports/AOR_Report_Before.js
+++ b/modules/AOR_Reports/AOR_Report_Before.js
@@ -172,7 +172,9 @@ $(document).ready(function(){
           if(relData[field]['type'] == 'relationship') {
             modulePath = field;
             if (node) {
-              modulePath = node.module_path + ":" + field;
+              if (node.module_path) {
+                modulePath = node.module_path + ":" + field;
+              }
               modulePathDisplay = node.module_path_display + " : "+relData[field]['module_label'];
             }else{
               modulePathDisplay = $('#report_module option:selected').text() + ' : ' + relData[field]['module_label'];


jack7anderson7:
Hi @lazka,
 were you able to replicate this as I wasn't able to in 7.10.7

lazka:
> Hi @lazka,
> were you able to replicate this as I wasn't able to in 7.10.7

Yes, I have the problem with 7.10.11

jack7anderson7:
Accessed :+1:

willrennie:
This fix is still failing in MSSQL, although it appears the query failure isn't related to this fix specifically, and with this working in MySQL it may be ok to merge? Error thrown in MSSQL when testing: 

```
Tue Mar 26 12:08:54 2019 [3629][1][FATAL]  Query Failed:SELECT count(*) c FROM (SELECT [leads].first_name AS N'First_Name0', [leads].last_name AS N'Last_Name1', [leads].email_address AS N'Email_Address2', [leads].id AS N'leads_id' FROM [leads]  WHERE leads.deleted = 0 ) as n::: [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Invalid column name 'email_address'.
Tue Mar 26 12:08:54 2019 [3629][1][FATAL]  Query Failed:DECLARE @topCount INT SET @topCount = 20 SELECT TOP (@topCount)  [leads].first_name AS N'First_Name0', [leads].last_name AS N'Last_Name1', [leads].email_address AS N'Email_Address2', [leads].id AS N'leads_id' FROM [leads]  WHERE leads.deleted = 0 ::: [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Invalid column name 'email_address'.
```

lazka:
Note that the wrong SQL query is saved in the DB, so with this fix applied you have to create a new report record for testing.

modules/AOR_Reports/AOR_Report_Before.js
```diff
@@ -172,7 +172,8 @@ $(document).ready(function(){
           if(relData[field]['type'] == 'relationship') {
             modulePath = field;
             if (node) {
-              modulePath = node.module_path + ":" + field;
+              if (node.module_path)
```
code-ph0y: Please use if statements with starting and ending curly braces. 

For-example:

```
if (node.module_path) {
    modulePath = node.module_path + ":" + field;
}
```

lazka: thanks, done.

Issue #6615, opened 2018-12-13
Workflow Process Audit Subpanel


lhsatsuma:
#### Issue

Parent name from WorkFlow Process Audit it's not showing correctly.
#### Expected Behavior

It should be visible parent's name of all process audit from workflow actions.
#### Actual Behavior

The parent name it's showing ONLY in first record of audit.
![image](https://user-images.githubusercontent.com/44283104/49957570-63cfc600-feef-11e8-9731-1c0dbb15c7f1.png)

#### Steps to Reproduce

1. Create a workflow with any conditions and actions. Example:
![image](https://user-images.githubusercontent.com/44283104/49957661-aa252500-feef-11e8-8eb0-2f02f25d6105.png)

2. Trigger the WorkFlow with two or more records.
3. See the subpanel Process Audit from WorkFlow.

#### Context

You can see the test that I make in demo: https://demo.suiteondemand.com/index.php?module=AOW_WorkFlow&action=DetailView&record=9ae0da7b-7cf3-0a1e-3e31-5c129bf2114d
#### Your Environment

SuiteCRM Version used: 7.10.11
Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Chrome Version 70.0.3538.110 (32 Bit)
Environment name and version (e.g. MySQL, PHP 7): MySQL 5.6.37 PHP: 7.0 Apache: 2.4.27
Operating System and version (e.g Ubuntu 16.04): Bitnami (Debian 8).

iamandrew:
Any ideas about where the issue might be?  We'd be willing to put some work into fixing this.

johnM2401:
Hey!

Thank you for raising this issue.

I've tested this in the latest release (7.14.6) and this appears to have been since resolved.

See my Process Audit Log, listing multiple records:

![Image](https://github.com/user-attachments/assets/64015928-8ead-4450-af88-5c956e623232)

<br>

I'm going to close this off for now.
However, if you're still seeing this issue in the latest releasese, please get in touch and we'd be happy to investigate further.

Thank you again!

PR #6746, opened 2019-01-09
#6615 - Fixed issue with parent_field query generation

iamandrew:
## Description
At some point, this `else` statement was removed from the retrieve_parent_fields loop for query generation,  meaning that only the first result was ever returned.

Adding it back in fixed the issue with the workflow audit log only showing 1 link and name for audit records in the workflow subpanel.

### DIFF
--- a/data/SugarBean.php
+++ b/data/SugarBean.php
@@ -1595,6 +1595,10 @@ class SugarBean
                                 " FROM " . $templates[$child_info['parent_type']]->table_name .
                                 " WHERE id IN ('$childInfoParentId'";
                         }
+                    } else {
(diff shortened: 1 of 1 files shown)


Dillon-Brown:
Hi @iamandrew, can you make this for the hotfix branch? we don't accept direct PR's to master. Thanks!

iamandrew:
@Dillon-Brown sorry about that. Changed it over to hotfix

jack7anderson7:
Hi @iamandrew,
the original issue has already been resolved and has been merged into hotfix and thus this pr is no longer needed.

PR #6329, opened 2018-09-19
Allow attachments in email templates when editor is not Mozaik

QuickCRM:
Added support for attachments when editor is TinyMCE or Direct HTML

## Description

variable mozaik was not defined.
Just define it!

Go to https://demo.suiteondemand.com.
Try to add attachments to an email template. 
These attachments won't be saved.

#5961 

## Motivation and Context

Adding attachments when editor is TinyMCE or Direct HTML

## How To Test This

Change editor to TinyMCE in profile and add attachments to an email template

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [ ] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change to the documentation.
- [ ] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

PR #6757, opened 2019-01-14
Fixed #6484 - Fix image insertion with mozaik

### LINKED ISSUE #6484: email templates: inserting images in mozaik broken since 7.10.10
#### Issue

1) Open template editor (mozaik)
2) Right click, insert image
3) Select image on local disk

Expected:

4) Source gets filled out

Actual:

4) Nothing happens

#### Possible Fix

I've bisected this to e840ea8968caf5c27bd
Reverting this commit on top of 7.10.10 makes things work again.

lazka:
## Description

In #6329 mozaik was set to a dummy object to make it work when not using
the mozaik editor. But the added variable declaration shadows the global
variable breaking the case where mozaik is actually used, setting it to
undefined.

This breaks inserting images into emails, the file gets uploaded but the
URL field stays empty.

Fix this by explicitly checking if we are using mozaik instead of using
a dummy object.

## Motivation and Context

Fixes #6484 which is a regression introduced by #6329

## How To Test This

* Use mozaik
* Insert > Insert/edit image
* Select a file and upload
* The "Source" field should contain a path to the added image file

## Types of changes

### Final checklist

### DIFF
--- a/modules/Emails/javascript/Email.js
+++ b/modules/Emails/javascript/Email.js
@@ -213,22 +213,18 @@ function multiFiles( list_target){
             element.onchange = function() {
                 var url = 'index.php?to_pdf=1&module=EmailTemplates&action=AttachFiles',
                     new_element;
-
-                // allow attachments with TinyMCE or Direct HTML;
-                if (typeof mozaik == 'undefined') {
-                    var mozaik={};
-                }
+                var isMozaik = typeof mozaik !== 'undefined';
 
                 //AJAX call begins
                 YAHOO.util.Connect.setForm(document.getElementById("upload_form"), true, true);
                 YAHOO.util.Connect.asyncRequest('POST', url, {upload: function(e) {
-					if(mozaik.uploadPathField) {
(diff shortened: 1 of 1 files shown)


cameronblaikie:
Hi @lazka,
After testing the PR, I was just wondering if the intention is to make the image appear as a attachment or in the template? As for me it's appearing as an attachment

lazka:
>  After testing the PR, I was just wondering if the intention is to make the image appear as a attachment or in the template? As for me it's appearing as an attachment

With mozaik it doesn't appear as an attachment, with the other editors it does. I don't know what the intention behind this is.

cameronblaikie:
You're right, my mistake the PR works as said.

cameronblaikie:
Assessed :+1:

lazka:
I've created #6775 for 7.10 (I assume this is what the "Needs Duplicated: LTS " label means)

Dillon-Brown:
@lazka Perfect, thanks!

PR #6770, opened 2019-01-16
Don't delete the composer.lock when running tests on travis.

### LINKED ISSUE #6759: Removing composer.lock in travis?
In the `.travis.yml` there are these lines:

https://github.com/salesagility/SuiteCRM/blob/80e9e484f7c00340a67a65efbf82aca2e3d5d141/.travis.yml#L60-L62

This means the tests can fail whenever a package within the ranges defined by the `composer.json` is updated, without any change in the SuiteCRM codebase. Is there a specific reason we do this in CI? It doesn't seem like a good idea, and doesn't really test the actual behavior of the code that you'd get by installing SuiteCRM locally, since Composer installs versions based on the lockfile.

connorshea:
## Description
Fixes #6759, removing the `composer.lock` has the side-effect of making tests less consistent. This PR changes the `.travis.yml` to make sure the lockfile isn't deleted.

## Motivation and Context
We should always be testing on the same base of dependencies. As long as the commit is the same, we should always have the same dependencies whether it's 2019 or 2029. Deleting the lockfile lets composer essentially install whatever it wants.

## How To Test This
Make sure Travis continues to pass.

## Types of changes

### Final checklist

- [x] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,6 @@ before_script:
   - phpenv config-add travis.php.ini
   - sudo service apache2 restart
   # Install composer
-  - rm composer.lock
   - composer install
   - ./vendor/bin/codecept build
 


connorshea:
@Dillon-Brown I've updated this to be merged into hotfix now :)

connorshea:
Rebased and force-pushed so the tests will hopefully pass now

cameronblaikie:
Assessed :+1:

PR #6774, opened 2019-01-17
EmailTemplates: Use html2text library for generating the text only body

lazka:
## Description

When creating an email template with the editor and saving then SuiteCRM
will generate a text version automatically by stripping away tags and decoding
html entities (it also gets regenerated if the text field is cleared before saving)

This currently has the downside that many features of the email are lost,
like basic structure through headings/lists and, more importantly, links.

This patch adds the soundasleep/html2text library and uses it instead to generate the
initial text version. It also adjusts the regex for separating variables from surrounding
text to be less strict to allow using variables in HTML links (see the tests)

This for example makes the generated text from the included opt-in confirmation email usable
without having to manually fix it.

Also adds a few tests for common edge cases.

## Motivation and Context

I'd like the auto converted email body to be good enough so that you normally
can just use it as is.

## How To Test This

* Create an email template with HTML
* Save
* Look the the template preview and click "Show Plain Text"
* The text version should somewhat resemble the HTML version

## Types of changes

### Final checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,8 @@
     "symfony/options-resolver": "^3.4",
     "symfony/validator": "^3.4",
     "ezyang/htmlpurifier": "^4.10",
-    "leafo/scssphp": "^0.7.7"
+    "leafo/scssphp": "^0.7.7",
+    "soundasleep/html2text": "~0.5"
   },
   "require-dev": {
     "mikey179/vfsStream": "1.6.*",
--- a/composer.lock
+++ b/composer.lock
@@ -1049,6 +1049,56 @@
             "time": "2018-09-16T10:54:21+00:00"
         },
         {
+            "name": "soundasleep/html2text",
+            "version": "0.5.0",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/soundasleep/html2text.git",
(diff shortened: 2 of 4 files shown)


lazka:
I don't know if this is something you want (or if the new dependency is acceptable). I can also move it to another branch if wanted.

Feedback welcome.

pgorod:
This looks useful to me. If I understand correctly, 

- we send two versions of the email - HTML and text
- this only affects the text version

Is that right?

Another thing that I believe needs improvement is the actual HTML email we send; on one hand, we should let people make their HTML as complex as they want, to make really nice-looking newsletters; on the other hand, security demands that we protect against malicious HTML (with two threats: that it has injection attacks on our own database, or that it has browser attacks for the person receiving the email).

We probably could do with a better handling of our HTML content in emails, from what I gather in the Forums.

samus-aran:
I'm restarting that 5.6 travis environment incase it was just a fluke.

lazka:
@jack7anderson7 did you run "composer install" to install the new dependency?

jack7anderson7:
@lazka I have now and it works fine, thank you.

jack7anderson7:
Assessed :+1:

PR #6794, opened 2019-01-21
Reports: fix "Illegal string offset 'record_id'" errors

lazka:
## Description

The 'params' value was changed from an array to a string in 6bdf937e9a07cd55e8
but not all usages fixed. This results in the "$att['params']['record_id']" assignment failing
and in lots of "Illegal string offset 'record_id'" errors when viewing a simple report.

This patch renames the 'params' key to 'format', because it's used only for the field format string
at this point, and creates a new $params array for passing the record_id to getModuleField().
From what I see the field format is checked everywhere because the reports module implements
its own date formatting and avoids calling into getModuleField() when a field has its own
a date format set.

This gets rid of the "Illegal string offset" errors and results in the correct record_id getting
passed to getModuleField().

## Motivation and Context

Viewing a simple report resulted in lots of PHP errors in the logs (one per report line)

Some reports about the errors in the forum:
* https://suitecrm.com/suitecrm/forum/suitecrm-7-0-discussion/20869-how-to-solve-warning-illegal-string-offset
* https://suitecrm.com/suitecrm/forum/installation-upgrade-hilfe/21022-creating-a-report-error-message-illegal-string-offset-record-id-in-aor-report-php-on-line-822

## How To Test This

* Create a report for "Leads"
* Add a field for "First Name", change nothing
* Save and view the record
* Check the PHP logs

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOR_Reports/AOR_Report.php
+++ b/modules/AOR_Reports/AOR_Report.php
@@ -233,7 +233,7 @@ class AOR_Report extends Basic
             $fields[$label]['total'] = $field->total;
 
 
-            $fields[$label]['params'] = $field->format;
+            $fields[$label]['format'] = $field->format;
 
             // get the main group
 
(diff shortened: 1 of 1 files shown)


jack7anderson7:
Assessed :+1:

Issue #4084, opened 2017-08-03
Option "contains" not available in MultiSelect Field


fgiehlede:
#### Issue
A user is not able to select the option "contains" in a multi select field in a report. The search condition is very complicated to use, when we want to search different values in this field.

#### Expected Behavior
It should be able to use "contains" for a multiselect field

#### Actual Behavior
We are not able to use "contains" as operator for a multiselect field in the report module

#### Steps to Reproduce
1) We have a multi select field with 3 options: A, B and C

2) We are now searching accounts with option B.

3) We must select any combination of A, B and C to find every account with the option B. Because we have only the option to search "value" for a multiselect field.

#### Context
-

#### Your Environment
SuiteCRM 7.8.3

lazka:
I've created #6811 which allows "equal to +  one of" which translates to "contains at least one of the selected values". "not equal + one of" translates to "does not contain any of the selected values".

Issue #3386, opened 2017-04-11
MultiSelect Parameter In Reports


CamoMacdonald:
#### Issue

If you have a report with a multi select field condition as a parameter and update that field through detail view the update does not take affect
#### Expected Behavior

The report should be updated to represent the new multiselect condition value

#### Steps to Reproduce

1. Create a report with a multi select field condition (parameter) and set the default to null
2. On detail view select a value that should display the additional results and press update
3. The result don’t update

#### Context

This stops users from making parameter reports with multi select fields
#### Your Environment

* SuiteCRM Version used: 7.8.3
* Operating System and version (e.g Ubuntu 16.04): Ubuntu 14.04

Dillon-Brown:
@CamoMacdonald I can't seem to replicate this on 7.8.3. 

Multi-select field with null as default:
![issue-3386_1](https://cloud.githubusercontent.com/assets/26431166/25134166/ccb41e3a-2446-11e7-9744-6034bf969c25.png)

After selecting value and hitting update:
![issue-3386_2](https://cloud.githubusercontent.com/assets/26431166/25134190/ddec4b82-2446-11e7-9f5f-a8a7b74cbf29.png)

CamoMacdonald:
@Dillon-Brown If you have an actual multi select field rather than the 'Type' 'One of'. Example is below:
![dillion](https://cloud.githubusercontent.com/assets/20798044/25134773/5afa546a-2448-11e7-869e-647e1db46efb.png)

belchfireid:
Bug is also in 7.10.7.  

Changing the parameter and Update does not work whether the Type is Value or One Of.

lazka:
I've created #6811 for the "One of" case

PR #6811, opened 2019-01-23
Reports: Fix "One of" operator for multi select fields

lazka:
## Description

MultiSelect fields are encoded like "^foo^,^bar^" in the DB so we need
to encode the values  we want to match against first and have to use
LIKE to match a substring because the field can contain multiple values.

This just copies the logic from the WorkFlow module where this was already
implemented this way:

https://github.com/salesagility/SuiteCRM/blob/ad0f598cd6d561c92f2f8d3ebeae519a12472e6f/modules/AOW_WorkFlow/AOW_WorkFlow.php#L523

This does not fix the other operators which are still broken.

## Motivation and Context

I'd like to create a report which shows all records that have one of the
values in the multienum selected.

This likely helps with #3386 and #4084

## How To Test This

* Create a report
* Add a condition on a multiselect field (equal to, one of, select one value)
* Change a record to have all values selected
* Update the report, it should show the record

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOR_Reports/AOR_Report.php
+++ b/modules/AOR_Reports/AOR_Report.php
@@ -1735,15 +1735,25 @@ class AOR_Report extends Basic
                             $multi_values = unencodeMultienum($condition->value);
                             if (!empty($multi_values)) {
                                 $value = '(';
-                                foreach ($multi_values as $multi_value) {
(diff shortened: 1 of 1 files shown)


jack7anderson7:
Assessed :+1:

PR #6850, opened 2019-01-28
SAML2: Use php-saml from composer

lazka:
## Description

Instead of using the outdated php-saml 2.11.0 that is included in tree
use 3.0.0 from composer.

To avoid breaking any existing code as much as possible leave the settings.php
where it is and add class aliases that map the old 2.x class names to the new namespaced
class names.

## Motivation and Context

Upstream is currently at 2.14.0/3.0.0 while the in-tree version is at 2.11.0.
We shouldn't use an outdated version of a security sensitive library.

## How To Test This

* SAML login/logout should work as before.

## Types of changes

### Final checklist


connorshea:
Shouldn't this also go into hotfix?

Dillon-Brown:
Just to note, we haven't been merging the branches together for the past few releases as it tends to cause more problems than it solves (Instead we are just making duplicate PRs). In this case we'll be merging hotfix-7.10.x into hotfix as they are pretty much even anyway.

lazka:
It's been a month and none of the pinged SAML users complained so far, so I think this is good to go.

connorshea:
needs a rebase

Dillon-Brown:
Assessed :+1: . Just needs conflicts resolved and then a quick code-review and then this should be good to merge.

lazka:
Sorry, was busy last week. Thanks for the rebase/merge!

pgorod:
Someone on the Forums had to revert to an older version of Onelogin to get things working...

https://suitecrm.com/suitecrm/forum/developer-help/25855-suitecrm-saml-configuration/

ebogaard:
Please don't downgrade this library!

We had issues with SAML as well, after upgrading to the latest SuiteCRM versions (7.10.21/22 in our case), which contain php-saml library version 3.3.x or higher.
Due to setting "strict" to "true" as a default, all SAML communications has to be signed. See the release notes: github.com/onelogin/php-saml/releases/tag/3.3.0

To disable this check, add "'strict' => false," to "$settingsInfo" in "custom/saml/modules/Users/authentication/SAML2Authenticate/lib/onelogin/settings.php"
Like this, for example.
```
$settingsInfo = array(
    // Turn off strict mode
    'strict' => false,
    // End
    'sp' => array(
```

That should restore the functionality to previous versions of the library.

You really shouldn't downgrade this library to anything lower than 3.3.1, as that version contains a fix for a critical security issue.

The best way to manage this setting, would probably be adding a setting to config_override.php, which enables/disables this strict check.

lazka:
(this PR updated to 3.1.0 while the strict change was added in 3.3.0)

For completeness, to quote the php-saml README:

> In production, the strict parameter MUST be set as "true" and the signatureAlgorithm and digestAlgorithm under security must be set to something other than SHA1 (see https://shattered.io/ ). Otherwise your environment is not secure and will be exposed to attacks.

vladaman:
After upgrading to 7.11.10 the SAML2 stopped working and it's in redirect loop with SAML2 IdP. Previously it worked.

Issue #5497, opened 2018-03-14
Currently, We could not hide Reportable some modules via in Report modules.


anhnhatuit:
#### Issue

Some cases, admin need hide some modules from Report module, However, SuiteCRM not support. I think this feature is necessary. 
![2018-03-14_141026](https://user-images.githubusercontent.com/15109575/37388118-be1dd006-2791-11e8-8480-a8dc70b973d6.png)

#### Possible Fix

We can base on reportable attribute in bean to check module need display or not when create Report.
#### Steps to Reproduce

1.
2.
3.
4.

#### Your Environment

* SuiteCRM Version used:
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)):
* Environment name and version (e.g. MySQL, PHP 7):
* Operating System and version (e.g Ubuntu 16.04):

anhnhatuit:
Hi @Dillon-Brown ,
Can you confirm this feature, If you are ok, I'll take care this feature.
Thanks,
Nhat

lazka:
I've created #6860 which makes it take into account user permissions. Maybe this works for your use case too.

PR #6860, opened 2019-01-30
Reports: Hide inaccessible modules in the reports editor

lazka:
## Description

The reports module already hides all fields (in the field list in the bottom left corner)
for modules not accessible to the user but still shows them in the "Module tree" and the
"Report Module" dropdown.

This leads to the users seeing lots of modules and potential relations which are not
usable in the end. To make this clearer also hide inaccessible modules in the module tree
and the module dropdown.

The call to load_report_beans() is moved to fill_in_additional_detail_fields() because
calling into ACLController in the constructor results in an infinite recursion as it
instantiates a AOR_Report itself.

### Before/After

report-filter

## Motivation and Context

* Users should not see modules they don't have access to.
* Might be a good enough solution for #5497

## How To Test This

* Log in as a user that only has access to some modules
* Create a report
* Only the modules available for the user should be visible

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOW_WorkFlow/aow_utils.php
+++ b/modules/AOW_WorkFlow/aow_utils.php
@@ -182,7 +182,7 @@ function getModuleTreeData($module)
                         'module_label'=> $module_label)
     );
 
-    if ($module != '') {
+    if ($module != '' && ACLController::checkAccess($module, 'list', true)) {
         if (isset($beanList[$module]) && $beanList[$module]) {
             $mod = new $beanList[$module]();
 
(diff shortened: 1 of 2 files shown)


modules/AOW_WorkFlow/aow_utils.php
```diff
@@ -193,6 +193,9 @@ function getModuleTreeData($module)
                     $rel_module = $mod->$name->getRelatedModuleName();
                 }
 
+                if (!ACLController::checkAccess($rel_module, 'list', true))
```
Dillon-Brown: Missing curly braces

amariussi:
Sorry but I am not sure I understand completely the issue.
Specially:
> * Users should not see modules they don't have access to.

Do you mean:
1. "if these modules are shown here, then the user will be able to use them, although he shouldn't be allowed to use them"
or
2. "even the user tries to use them his attempt will have no effect"?

If the answer is 2. I agree with this code change.
However if it is 1. I would add a config variable to be checked before removing these moduels from the available ones, By doing so you make the Reports module more powerful and flexible, while if yuo just remove them you are just reducing its functionality, which, in my opinion, should not be forced.

lazka:
> If the answer is 2. I agree with this code change.

Currently users can select the modules they don't have access to but they wont see any fields and thus can't add them.

(Besides, they don't have rights to see the content, so that would be a security issue if possible)

willrennie:
Assessed and confirmed as working.

PR #6861, opened 2019-01-30

### LINKED ISSUE #6452: Api/V8 create record does not support unicode and space in attributes

tsummerer:
## Description
The `Attributes.php` parameter previously used the `Fields.php` parameter's regex pattern to verify that attributes were formatted correctly. 

This was improper because it greatly limited the available characters for field values. A field (especially one like `description`) should accept a wide range of characters - including spaces, periods, brackets, etc.

This PR was based (and with permission) on #6576. The OP requested that I create a PR to correct for a slight omission. Thanks to @capkuan for identifying and finding the original solution. However, I have removed this code because it was not properly formatted and was causing a false positive (as shown by failing Travis tests). 

I have removed the regex check because I cannot think of a valid pattern to exclude. I believe that the application sanitizes the input before writing to db. 
. 

Original bug: #6452 

## Motivation and Context

It is currently impossible to create/edit a record with a space character in any of the fields. 

## How To Test This

```
POST SuiteCRM/Api/V8/module
BODY {
  "data": {
    "type": "Accounts",
    "attributes": {
      "name": "Cocacola company",
    }
  }
}
```

## Types of changes

### Final checklist

### DIFF
--- a/Api/V8/Param/Options/Attributes.php
+++ b/Api/V8/Param/Options/Attributes.php
@@ -9,7 +9,6 @@ use Symfony\Component\Validator\Constraints as Assert;
 
 class Attributes extends BaseOption
 {
-    const REGEX_FIELD_PATTERN = '/[^\w-,\s\]/';
 
     /**
      * @inheritdoc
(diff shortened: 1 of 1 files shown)


tsummerer:
I'm actually not entirely convinced that we need to exclude blank values either (line ~24 of `Attributes.php`) - it seems like there could be completely valid use cases where a person may want to clear out a value. 
Should I remove that constraint? 
Are there any constraints we need to add to these values? I know that it doesn't currently have any way to understand arrays - but I don't know which Constraint would represent that.

tsummerer:
Can someone restart Travis? Seems that it choked for reasons unrelated to this PR.

samus-aran:
We suggest using REG so we can customise it so removing it isn't preferred. However we will provide a solution. Thanks for your contribution.

PR #6404, opened 2018-10-10
Mass update on user preferences doesn't save 7.10

JackBuchanan:
Moved code from save.php to user.php and deleted save.php removed calls
to deleted file in controller.php

## Description

Added function saveFormPreferences() to User.php whcih contains code from deleted file save.php and removed call to save.php in controller.

Issue: #6402
## Motivation and Context

Mass update on user preference fields 
## How To Test This

-Navigate to admin panel
-Select users to mass update
-Change Email Client field
-Check if change has saved in user edit view

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [x] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

PR #6898, opened 2019-02-04
WIP: Fixed #6869 - Users: fix saving user records resetting user preferences

### LINKED ISSUE #6869: Saving user resets preferences
#### Issue

Since #6404 `$current_user->save()` calls `saveFormPreferences()` which when not called in the context of a form resets user settings to their "off" state.

#### Expected Behavior

Calling `$current_user->save()` shouldn't change user preferences.

#### Actual Behavior

Calling `$current_user->save()` sets things such as `use_real_names` to off, even if the default is different.

#### Steps to Reproduce

1. Go to "Admin" -> "User Management"
2. Inline edit any of the user fields in the list
3. The users preferences are reset after that (the most visible one is that in the top right corner the username is shown instead of the fullname)

lazka:
## Description

In #6404 the code to apply the user edit view form to the preferences
was moved into User::save() unconditionally. This results in the preferences
being reset in case save() is called in a different context than the form.

For example when inline editing a user in the user list view the preferences
for that user get set as if there was an empty edit form.

Fix this by only applying the form updates in case we are in the context of
saving in the edit view page.

## Motivation and Context

#6869

## How To Test This

* Go to "Admin" -> "User Management"
* Edit a user and set "Show Full Names:" to true.
* Go back to the list view and inline edit any of the user fields in the list
* The "Show Full Names" value should still be true after that

## Types of changes

### Final checklist

### DIFF
--- a/modules/Users/User.php
+++ b/modules/Users/User.php
@@ -675,7 +675,9 @@ class User extends Person implements EmailInterface
             }
         }
 
-        $this->saveFormPreferences();
+        if (isset($_POST['module']) && $_POST['module'] == 'Users' && isset($_POST['action']) && $_POST['action'] == 'Save') {
+            $this->saveFormPreferences();
+        }
 
         $this->savePreferencesToDB();
 


lazka:
This seems a bit hacky, but keeps things simple. Ideas welcome if there is a better way.

lazka:
Looks like this breaks some tests :/

PR #6975, opened 2019-02-19
FIX #6950 - Add wikimedia/composer-merge-plugin

Abuelodelanada:
## Description

Fix #6950

## Motivation and Context

Fix #6950

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [x] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

PR #6976, opened 2019-02-19
FIX #6950 - Add wikimedia/composer-merge-plugin

### LINKED ISSUE #6950: We should have a way to add composer dependencies safe-upgrade
#### Issue

When developing it is very useful to add new dependencies to composer.json.
But adding your custom dependency to composer.json file is not safe-upgrade.

#### Expected Behavior

Would be great to have a way to add custom dependencies in a safe-upgrade mode

#### Actual Behavior

Modifying composer.json file is not safe-upgrade

#### Possible Fix

May be we could use [composer-merge-plugin](https://github.com/wikimedia/composer-merge-plugin)

#### Context

Fixing this would make developers life happier.

#### Your Environment

* SuiteCRM Version used: 7.10.x - 7.11.x

Abuelodelanada:
## Description

Fix #6950 

## Motivation and Context

Fix #6950

## Types of changes

### Final checklist


PedroErnst:
Looks like a nice addition. Should `composer.ext.json` maybe reside in the `custom` folder?

Jason-Dang:
Hi, I think this would be a really cool and useful feature for upgrade safe composer packages. Though I seen the include file is `composer.ext.json`. Would it be possible to add this to the extensions framework, so that files can be added to `custom/Extension/application/Ext/Composer/` and on QR+R get compiled down to `custom/application/Ext/Composer/composer.ext.json`?.

PedroErnst:
@Jason-Dang 

Really good point.
It would appear that updating the following line in composer.json will take care of compiling all the files from the Extension framework (and any subdirectory in the Composer folder):

```
      "merge-plugin": {
          "include": [
              "composer.ext.json",
              "custom/Extension/application/Ext/Composer/*/*.json"
          ],
```

Maybe it would be a good idea to include this in the PR?

Jason-Dang:
@PedroErnst That sounds like the best solution to me as it eliminates the need for a lot of extra work building the extension file etc. 

Last thing I think needs to be considered is version conflicts for the same package. Assuming composer uses the highest version, what if something in the application requires an older version?

This also appears to be a duplicate of https://github.com/salesagility/SuiteCRM/pull/6975

Abuelodelanada:
> @Jason-Dang
> 
> Really good point.
> It would appear that updating the following line in composer.json will take care of compiling all the files from the Extension framework (and any subdirectory in the Composer folder):
> 
> ```
>       "merge-plugin": {
>           "include": [
>               "composer.ext.json",
>               "custom/Extension/application/Ext/Composer/*/*.json"
>           ],
> ```
> 
> Maybe it would be a good idea to include this in the PR?

Hi @PedroErnst @Jason-Dang 

Do you think I should modify the PR with this?

I was looking for something like the config_override.php but for composer plugins... that's why I only set one file called: composer.ext.json in root directory.

PedroErnst:
@Abuelodelanada 

Yes I think as Jason pointed out this would keep in line with the Extension framework... different customizations could add composer requires without needing to be aware of eachother.

I've tested and if the folder in question is missing or empty it doesn't throw an error. So if someone wants to go with the simple solution you propose with the `composer.ext.json` that works fine, but it also leaves the door open for more complex setups down the line. Best of both worlds.

Abuelodelanada:
@PedroErnst 

Great! So, do you need I made a fix to this PR?

PedroErnst:
Ultimately it's up to you, if you think it makes sense what we suggest then you're welcome to include it in your PR. I would say yes, but I'm just another dev :)

Abuelodelanada:
> Ultimately it's up to you, if you think it makes sense what we suggest then you're welcome to include it in your PR. I would say yes, but I'm just another dev :)

Done! Thanks @PedroErnst and @Jason-Dang for the suggestion!

Jason-Dang:
Assessed & Approved.

Issue #7030, opened 2019-03-08
Workflow Module - Start with operator not working


pavithra-selvam:
Workflow module

#### Issue
 Workflow module - contains , start with and end with operator not working in all versions. 

#### Steps to Reproduce

1.Create workflow
2.add condition and select operator is any one like start with,end with or contains.

#### Your Environment
SuiteCRM Version: 7.9.7
PHP: 7

pavithra-selvam:
i found solution for this can you please test and is it okay can you please commit to your next release.
in modules/AOW_WorkFlow/AOW_WorkFlow.php file can you please repleace these lines.
 **case "Starts_With" : return strrpos($var1,$var2, -strlen($var1));
 case "Ends_With" : return strpos($var1,$var2,strlen($var1) - strlen($var2));**
insted of 
**case "Starts_With" : 
                $length = strlen($var2);
                return (substr($var1, 0, $length) === $var2);
case "Ends_With" : 
                $length = strlen($var2);
                if ($length == 0) {
                    return true;
                }
                return (substr($var1, -$length) === $var2);**

pgorod:
@pavithra-selvam thanks for that. 

Do you want to try and make a Pull Request? For such a simpe change, it's easy to do from Github web interface.

Just go here (notice that I selected the branch for you):
https://github.com/salesagility/SuiteCRM/blob/hotfix-7.10.x/modules/AOW_WorkFlow/AOW_WorkFlow.php

And click the ✏️ icon to edit the file, then save your changes and write up the text for the PR. I can help you if you run into problems. Thanks!

pavithra-selvam:
@pgorod ,
Thanks for your help, i followed your instruction and updated file. is everything fine or need to change anything?.
Thanks!

pgorod:
When you edit the file, Github automatically forked the SuiteCRM repo into your own Github area, and created a commit there. Then you were given the option to create a Pull Request (asking the SuiteCRM maintainers to integrate your patch), but it seems you didn't finish that step.

The full process is described here:
https://help.github.com/en/articles/editing-files-in-another-users-repository

You don't have to repeat everything, you can just go here
https://github.com/salesagility/SuiteCRM/compare/hotfix-7.10.x...pavithra-selvam:patch-1?expand=1

and start the PR. Thanks

pavithra-selvam:
@pgorod 
Thanks for your help pgorod.. I created pull request.
Thanks!

PR #7042, opened 2019-03-11
Update AOW_WorkFlow.php

pavithra-selvam:
Workflow - Startwith and End with fixes.

## Description
Its only fixes for workflow module start with and end with operator changes.

## How To Test This
Create new workflow with any module.
Add condition with start or end with operator.

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOW_WorkFlow/AOW_WorkFlow.php
+++ b/modules/AOW_WorkFlow/AOW_WorkFlow.php
@@ -832,8 +832,15 @@ class AOW_WorkFlow extends Basic
             case "Greater_Than_or_Equal_To": return $var1 >= $var2;
             case "Less_Than_or_Equal_To": return $var1 <= $var2;
             case "Contains": return strpos($var1, $var2);
-            case "Starts_With": return strrpos($var1, $var2, -strlen($var1));
-            case "Ends_With": return strpos($var1, $var2, strlen($var1) - strlen($var2));
+            case "Starts_With": 
+                $length = strlen($var2);
+                return (substr($var1, 0, $length) === $var2);
+            case "Ends_With": 
+                $length = strlen($var2);
+                if ($length == 0) {
+                    return true;
+                }
+                return (substr($var1, -$length) === $var2);
             case "is_null": return $var1 == '';
             case "One_of":
                 if (is_array($var1)) {


pgorod:
This is regarding #7030

samus-aran:
Travis is failling. Since you've changed the return values from those functions the existing tests are failing.
There was 1 failure:

1) AOW_WorkFlowTest::testcompare_condition
Failed asserting that true matches expected 0.

Have a wee look at the tests and see if they need adjusting (I suspect) or the proposed code requires changes

pavithra-selvam:
@samus-aran
is any issue ?

samus-aran:
> @samus-aran
> is any issue ?

Yes, So if you are aware Travis CI run tests against the current code/functions which confirm that existing functionality (if covered by tests) work the same as it did previously. Now if people make changes to current code that have tests then that code could cause the tests to fail because the tests need updated to reflect the new code (i.e. your PR) or the new code is not suitable.

So I suggest you have a look at the tests that are associated with AOW

/home/travis/build/salesagility/SuiteCRM/tests/unit/phpunit/modules/AOW_WorkFlow/AOW_WorkFlowTest.php:315
/home/travis/build/salesagility/SuiteCRM/vendor/phpunit/phpunit/phpunit:52

https://github.com/salesagility/SuiteCRM/blob/master/tests/unit/phpunit/modules/AOW_WorkFlow/AOW_WorkFlowTest.php#L315

^^
These two tests are expecting a return value to which is throwing up an error. Can you ensure your code is passing in the right values as I can see it is returning a number of different types of variables.

Originally it is expecting that when it makes a comparison then it should be "0" however that must not be the case now due to your changes with code.

As I said it may need the tests to also be updated or your code isn't needs updated.

samus-aran:
Ok had a wee look and your code has now changed the returns to be a boolean (either true or false) whereas before it was just a strpos.

So you will need to update the test to reflect that change. I suggest you look at the above tests like e.g. https://github.com/salesagility/SuiteCRM/blob/master/tests/unit/phpunit/modules/AOW_WorkFlow/AOW_WorkFlowTest.php#L308

They use the assertTrue indicating that in order to pass it must return TRUE which in your new code returns that. So I would replace Starts_with and Ends_with tests with assertTrue now.

pavithra-selvam:
okay fine.

samus-aran:
Hi @pavithra-selvam Have you had a chance to supply a new test for this PR. We can't accept this until you provide an updated test.

Any other community members willing to add a test here to merge this in?

pavithra-selvam:
Hi @samus-aran , I modified the code. I think this code satisfy your all test case. how to commit this code? here i paste my code. Can you please test this code and  let me know still you are facing any test case issue.
**My Code**
case "Starts_With" : 
                $length = strlen($var2);
                if(substr($var1, 0, $length) === $var2){
                    return true;
                }
                else{
                    return false;
                }
            case "Ends_With" : 
                $length = strlen($var2);
                if ($length == 0 || (substr($var1, -$length) === $var2)) {
                    return true;
                }
                else {
                    return false;
                }
Thanks 
Pavithra

cameronblaikie:
Assessed :+1:

Dillon-Brown:
Hi @pavithra-selvam, thanks! I can commit that provided code to this PR for you but we'd need you to sign the clahub (It's the first red "X" below this comment, you just click on "Details").

pavithra-selvam:
Hi @Dillon-Brown , as per your instruction i followed and after submit the CLA form i got this error but whether is submitted or not. Here i have attached your image for your reference. 
  
error

SinergiaCRM:
The filter "Contains" also fails when the needle is at the beginning of the haystack.

We can open a new PR if needed. We would apply this patch:

```
// Fixing conditions
// case "Contains": return strpos($var1, $var2);
case "Contains": return strpos($var1, $var2) !== false;
// case "Starts_With": return strrpos($var1, $var2, -strlen($var1));
case "Starts_With": return substr( $var1, 0, strlen($var2) ) === $var2;
// case "Ends_With": return strpos($var1, $var2, strlen($var1) - strlen($var2));
case "Ends_With": return substr( $var1, -strlen($var2) ) === $var2;
// End Fix
```

DJUserkent:
The patch works here with us - (from @SinergiaCRM  ). We checked all 3 cases (contains /starts with / ends with) on
- Version 7.11.21
- Sugar Version 6.5.25 (Build 344)

I'm not a dev.... but the change will SIMPLIFY the test code  - as  ALL the cases can be tested for boolean  - whereas before these 3 cases needed special handling.

>  public function compare_condition($var1, $var2, $operator = 'Equal_To')
    {
        switch ($operator) {
            case "Not_Equal_To": return $var1 != $var2;
            case "Greater_Than":  return $var1 >  $var2;
            case "Less_Than":  return $var1 <  $var2;
            case "Greater_Than_or_Equal_To": return $var1 >= $var2;
            case "Less_Than_or_Equal_To": return $var1 <= $var2;
            **case "Contains": return strpos($var1, $var2) !== false;
            case "Starts_With": return substr( $var1, 0, strlen($var2) ) === $var2;
            case "Ends_With": return substr( $var1, -strlen($var2) ) === $var2;**
            case "is_null": return $var1 == '';

pgorod:
Thanks for that report of your testing.

This code probably worked fine at some point, so I wonder what is now causing it to fail... perhaps it's a new PHP version that has some changes to the way these types are handled?

Or possibly some change in the function that calls this one, so that the variables are now reaching this code with different values?

The reason I ask these things is because I fear some recession when we change them, if we are not understanding why it failed...

DJUserkent:
> Thanks for that report of your testing.
> 
> This code probably worked fine at some point, so I wonder what is now causing it to fail... perhaps it's a new PHP version that has some changes to the way these types are handled?

It does seem hard to believe that such a blatant fail, could exist for long - un-noticed; so your idea of a recent PHP version makes sense.

Checking PHP.net - PHP  has no change applicable here for function 'strpos' - https://www.php.net/manual/en/function.strpos.php

Looking into the function that needed changing is used: it is ONLY in the same AOW_Worklow.php; like this

`       if (!($this->compare_condition($field, $value, $condition->operator))) {
                    return false;`

> Or possibly some change in the function that calls this one, so that the variables are now reaching this code with different values?

I'm not a coder  -my first time looking into Github - but History seems so show no changes applicable for 3 years... (maybe I;m using gitHub wrong). 


> **The reason I ask these things is because I fear some regression when we change them, if we are not understanding why it failed...**

Indeed.
If there were live demo sites running older SuiteCRM versions  -I could have tried them out. Although.... because Workflow is powerful it is not exposed on the current  suitecrm demo site... so probably too dangerous to expose on a public website...

PS I'm not a coder... ( but have some logic ability...)...
I looked in AOR_Reports as it is very similar in user-function  -  I mean it gives the user the same filter choices (contains, starts with, ends with) -yet  it does NOT contain this function NOR use of var1 var2 that it uses.

- because AOW appears to use it in a pre-check type thing( check_valid_bean)  - before running in earnest   

>if ($flow->check_valid_bean($bean)) {
>                    $flow->run_actions($bean, true);
>                }

But AOR shares with AOW this code - which looks  like it is code explicitly intended to handle the filtering  - 

>  //handle like conditions
            switch ($condition->operator) {
                case **'Contains**':
                    $value = "CONCAT('%', ".$value." ,'%')";
                    break;
                case **'Starts_With':**
                    $value = "CONCAT(".$value." ,'%')";
                    break;
                case '**Ends_With**':
                    $value = "CONCAT('%', ".$value.")";
                    break;

and both have similar code to handle 'cases to logic' conversion

>         $aow_sql_operator_list['Equal_To'] = '=';
>         $aow_sql_operator_list['Not_Equal_To'] = '!=';
>         $aow_sql_operator_list['Greater_Than'] = '>';
>         $aow_sql_operator_list['Less_Than'] = '<';
>         $aow_sql_operator_list['Greater_Than_or_Equal_To'] = '>=';
>         $aow_sql_operator_list['Less_Than_or_Equal_To'] = '<=';
>         $aow_sql_operator_list['Contains' = 'LIKE';
>         $aow_sql_operator_list['Starts_With'] = 'LIKE';
>         $aow_sql_operator_list['Ends_With' = 'LIKE';

so... not sure why 'check_valid_bean' is needed (it is long and complex).... could it be dropped in favour of tweaking 'run_actions'  with a toggle for 'dry_run' mode; to be used first time round?
I guess that it is not possible, for ways beyond me.

DJUserkent:
@SinergiaCRM 's patch does work in 7.
And the code for Suite 8 show t s needed there too (not tested myself).

**Case sensitivity**
The code, though is case sensitive (as it was before).
That is kind of confusing for users...    

To make it case insensitive would need a little more work:
A)  - to use stripos for strpos /  and make the equalities work by lower-casing both sides: not just 

> return substr( $var1, 0, strlen($var2) ) === $var2;)

B)  update the documentation page to flag they are are insensitive for:

- position
- replace

at: https://docs.suitecrm.com/user/advanced-modules/workflow-calculated-fields/

jack7anderson7:
Hi @pavithra-selvam,

Thank you for your contribution. However we are closing this PR as a fix has been merged via the [9614 PR](https://github.com/salesagility/SuiteCRM/pull/9614).

Thanks,
Jack

Issue #2464, opened 2016-10-21
Logo upload function is not working


saleembdr:
#### Issue

The logo upload function is not working. A message with the type of supported images is shown even though a supported image is uploaded.
#### Expected Behavior

The logo image should be uploaded.
#### Actual Behavior

A message indicating supported image types is shown.
#### Possible Fix

The variable '$upload->temp_file_location' includes a path with the extension 'tmp' in the file 'modules/Configurator/UploadFileCheck.php'
#### Steps to Reproduce
1. Login
2. Go to the administration section
3. Click on system settings
4. Attempt to upload a image with a supported extension

![image](https://cloud.githubusercontent.com/assets/10287173/19597205/d3343e4e-979b-11e6-8fea-3902ba641192.png)
#### Context

I was trying to change the logo that is shown on the login screen of SuiteCRM.
#### Your Environment

SuiteCRM 7.7.4 using Microsoft SQL Server and PHP 7 on IIS.

shogunpol:
The issue has been tested in recent version of SuiteCRM(7.7.6), and it not appear,
@saleembdr , can you confirm is this not issue any more in recent version of SuiteCRM, if yes can you close an issue.
Thanks for your contribution.

Ninidelaogong:
Hi,
I am using SuiteCRM V7.7.6. and the problem still occurs. I have tried  .jpeg and  .png.
The images have a max. height aof 170px and a width of 368px and even smaller and the message described shows.
![error_message](https://cloud.githubusercontent.com/assets/23031353/19646129/62813a6c-99f8-11e6-90b0-93511e5c7e92.JPG)
![version](https://cloud.githubusercontent.com/assets/23031353/19646128/627fdc8a-99f8-11e6-9f45-be6fb61df497.JPG)

saleembdr:
I have installed the version 7.7.6 for testing. The same message is encountered while trying to upload a JPEG image for the logo. The following screen-shot provides more information.

![image](https://cloud.githubusercontent.com/assets/10287173/19674773/f2b7e55a-9a92-11e6-954c-0ec9aa0a40a2.png)

shogunpol:
@saleembdr , just quick question, did you try with another images, and extensions? like take some pictures with different size .jpg .jpeg .png, and try out?
Thanks for your contribution.

saleembdr:
[...]

Issue #1863, opened 2016-08-03
Regression: images not showing with PHP versions where mime_content_type is deprecated


pgorod:
#### Issue

Not my problem, I'm just copying over an issue from the forums:
https://suitecrm.com/forum/suitecrm-7-0-discussion/10348-image-field-not-working

It seems that in this fix 
https://github.com/salesagility/SuiteCRM/pull/1509

Some code was introduced that causes this error in some systems:
`Call to undefined function mime_content_type() in /home/domain/public_html/crm/download.php on line 174`
The browser shows an HTML error 500 on the console, and the image doesn't show.

It's confusing, from what I gathered, because mime_content_type was deprecated (around version 5.5?) but then was un-deprecated in 7.0. Cool, I didn't even know they did "undeprecations".

But we need to support some PHP versions that can't handle this code...
#### Expected Behavior

Image should show; code should run on all our supported PHP versions.
#### Actual Behavior

Image doesn't show on detail views, etc.
Notice that now the problem is with every browser, not just IE and Edge like the original problem...
#### Possible Fix

There are a couple of useful StackOverflow questions with several alternatives.
#### Steps to Reproduce

See the forum post above. 
But essentially: 
1. add an image field to Accounts or Contacts. 
2. upload an image for a given record.
3. see that record's detail view.
#### Your Environment

- SuiteCRM Version used: 7.6 and 7.7
- Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): IE 9, Firefox and Chrome
- Environment name and version (e.g. MySQL, PHP 7): PHP 5.6
- Operating System and version (e.g Ubuntu 16.04): ?

604media:
@pgorod Can you please enable php_fileinfo extension in your php.ini file and check.

Add the below line in your php.ini file

extension=php_fileinfo.dll

pgorod:
@604media sorry, I can't... it's not my bug, I'm just reporting the experience of another user from the forums (see link at the beginning of the issue description).

But even if that fixes it, the point here is to make the code work out-of-the-box for everybody within the supported PHP versions group.

samus-aran:
Hi @pgorod 
Depending on the OS the originally commentor had I'm pretty usre this is a duplication of this issue.

https://github.com/salesagility/SuiteCRM/issues/1725

And as @604media is suggesting would resolve it.  We do support that PHP version, but (un)suprisingly Windows seem to have different PHP packages and require extra steps because of those differences.

[...]

PR #7044, opened 2019-03-11
Fix Content-Type header missing in some cases for the getImage entry point

lazka:
## Description

The getImage entry point only handled setting the correct mime type for
gif and png images. In case of svg images (for example the icon left of attachments in
the email template editor) the returned content type was text which made firefox
show an error and not display the image.

Fix this by adding mime types for all supported image extensions in SugarTheme
and using that in the getImage end point.

This also adds some basic tests for this new API.

## Motivation and Context

In the email template editor the attachments had broken images

## How To Test This

* Access `index.php?entryPoint=getImage&themeName=SuiteP&imageName=Accounts.svg`
* Check the returned content type

## Types of changes

### Final checklist

### DIFF
--- a/include/SugarTheme/SugarTheme.php
+++ b/include/SugarTheme/SugarTheme.php
@@ -278,13 +278,29 @@ class SugarTheme
     private $_clearCacheOnDestroy = false;
 
     private $imageExtensions = array(
-            'svg',
-            'gif',
-            'png',
-            'jpg',
-            'tif',
-            'bmp',
+            'svg' => 'image/svg+xml',
+            'gif' => 'image/gif',
+            'png' => 'image/png',
+            'jpg' => 'image/jpeg',
+            'tif' => 'image/tiff',
+            'bmp' => 'image/bmp',
     );
+
+    /**
+     * Returns the mime type for the image extension in case it is supported.
+     * In case the extension isn't supported returns null.
+     *
+     * @param $extension The extension name, e.g. 'png'
+     * @return string|null
+     */
+    public function getMimeType($extension)
+    {
+        if (!isset($this->imageExtensions[$extension])) {
+            return null;
+        }
+
+        return $this->imageExtensions[$extension];
+    }
 
     /**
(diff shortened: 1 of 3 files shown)


pgorod:
I wonder if this also fixes #2464

and if the discussion in #1863 about including all mime types at once could be useful here...

lazka:
> I wonder if this also fixes #2464

I can't reproduce the error mentioned there sadly, so not sure.

> and if the discussion in #1863 about including all mime types at once could be useful here...

You mean using mime_content_type() or something else? I tried to keep the changes small.

pgorod:
I mean this solution proposed [here](https://github.com/salesagility/SuiteCRM/issues/1863#issuecomment-302541886), along with [this simplified method](https://github.com/salesagility/SuiteCRM/issues/1863#issuecomment-371433719) of keeping the official list of MIME types handy.

Anyway I appreciate your option to keep changes minimal, that makes this PR easier to get through and be merged. I was just thinking if this could be the moment to make a deeper fix into MIME type handling, it affects several different functions in SuiteCRM...

willrennie:
@lazka I've assessed this and I'm sure the fix is good, but I couldn't actually replicate the issue raised. I tested in firefox both directly entering the URL and also creating an email template and checking images displaying there. It would be good to have more detailed steps and screen shots so I can try to replicate this to confirm your fix addresses the bug.

lazka:
@willrennie thanks for testing. You are right, after renaming the suitecrm cache dir the image resolved to an actual .gif file. Somehow the image cache stored the .svg path for that image name, not sure why.

The only other place I can see .svg files being used directly is in advanced list filter popup the button at the bottom. But it has a .png fallback there, so there is no improvement. A manual test URL would be `index.php?entryPoint=getImage&themeName=SuiteP&themeName=SuiteP&imageName=AOR_Reports.svg` (check the returned content type)

Note that I can always move PRs to hotfix instead if it is considered too risky for the LTS.

PR #7048, opened 2019-03-13
Specify that the methods are static

isleshocky77:
## Description

During a code cleanup commit, a visibility declaration was added to the
method without taking into a account that the methods are documented as
being static in the phpDoc.  This leads code using them statically to
throw a warning

## Motivation and Context

I like being able to leave traces on to see what's happening.

## How To Test This

Turn on traces. 
Have a plugin which is calling these methods statically.

## Types of changes

### Final checklist

### DIFF
--- a/modules/ACLRoles/ACLRole.php
+++ b/modules/ACLRoles/ACLRole.php
@@ -104,10 +104,10 @@ class ACLRole extends SugarBean
      * static  getUserRoles($user_id)
      * returns a list of ACLRoles for a given user id
      *
-     * @param GUID $user_id
-     * @return a list of ACLRole objects
+     * @param string $user_id GUID
+     * @return array a list of ACLRole objects
      */
-    public function getUserRoles($user_id, $getAsNameArray = true)
+    public static function getUserRoles($user_id, $getAsNameArray = true)
     {
 
         //if we don't have it loaded then lets check against the db
@@ -138,10 +138,10 @@ class ACLRole extends SugarBean
      * static  getUserRoleNames($user_id)
      * returns a list of Role names for a given user id
      *
-     * @param GUID $user_id
+     * @param string $user_id GUID
      * @return array a list of ACLRole Names
      */
-    public function getUserRoleNames($user_id)
+    public static function getUserRoleNames($user_id)
     {
         $user_roles = sugar_cache_retrieve("RoleMembershipNames_".$user_id);
 
@@ -172,9 +172,9 @@ class ACLRole extends SugarBean
      * static getAllRoles($returnAsArray = false)
      *
      * @param boolean $returnAsArray - should it return the results as an array of arrays or as an array of ACLRoles
-     * @return either an array of array representations of acl roles or an array of ACLRoles
+     * @return array either an array of array representations of acl roles or an array of ACLRoles
      */
-    public function getAllRoles($returnAsArray = false)
+    public static function getAllRoles($returnAsArray = false)
     {
         $db = DBManagerFactory::getInstance();
         $query = "SELECT acl_roles.* FROM acl_roles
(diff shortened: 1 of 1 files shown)


Dillon-Brown:
Assessed

gymad:
It is an interesting PR and makes a dilemma to me.

> This leads code using them statically to throw a warning

I get your point in this PR but also we trying to avoid all of the `public static function` calls as many as possible instead adding more and more. (Explanation according to https://stackoverflow.com/questions/33705976/when-should-i-use-static-methods)

> Have a plugin which is calling these methods statically.

Thats why I think, making "plugins" which calls these function statically is not a right direction, instead maybe better just simply removing the "static" from the php-doc (which step was missing from the code cleanup commit in my opinion) and do not use it statically. What do you think @isleshocky77 ?

isleshocky77:
> It is an interesting PR and makes a dilemma to me.
> 
> > This leads code using them statically to throw a warning
> 
> I get your point in this PR but also we trying to avoid all of the `public static function` calls as many as possible instead adding more and more. (Explanation according to https://stackoverflow.com/questions/33705976/when-should-i-use-static-methods)
> 
> > Have a plugin which is calling these methods statically.
> 
> Thats why I think, making "plugins" which calls these function statically is not a right direction, instead maybe better just simply removing the "static" from the php-doc (which step was missing from the code cleanup commit in my opinion) and do not use it statically. What do you think @isleshocky77 ?

I'm fine with that as well; however, it will undoubtedly break plugins (I know you guys are big on b.c. with old plugins and mods). Maybe make a deprecated set for next release?

Just to note, my current instance of this issue was from the [Analytic Reporting](https://store.suitecrm.com/support/analytic-reporting-tool-pro) module.

samus-aran:
Can I ask the question that should we not be able to encourage the proper behaviour in third party plugins rather than adapting core code to suit them? Obviously we are happy to declare what is required but also state that if a plugin throws up the warning.. is it the plugin at fault or core?

So my understanding here is that we throw up a deprecated warning for next release, but we still need to reach out to plugin makers that they are possibly using incorrect code and need to update.

isleshocky77:
Of I completely agree that you should strive for core to be the best it can be and give third-parties direction on how to best use it. 
Which is why I'm first proposing making the method is as it's documented (static).

If you believe static is a bad design decision and want to change that then, yes, I would still fix it, mark it as deprecated, and let third-parties parties know to fix it before a next major release.

isleshocky77:
@Dillon-Brown Rebased on `hotfix`

isleshocky77:
@Dillon-Brown Rebased on `hotfix-7.10.x`

Dillon-Brown:
Hey @isleshocky77, so after taking a look at this and reading some of the comments here I think our preference should probably be to update the docblocks to remove the static.

My thinking here is that either way people are going to be calling this function incorrectly as they either went with the docblock and are currently getting hit with a PHP deprecation warning or they went with what is actually there and are getting hit with a static method warning from the docblock.

As the worst thing that can happen here is that you get hit with a warning, it would probably make more sense to head towards good architecture design instead since any change we make either to the function or docblock will end up making people change their code anyway and that way they wouldn't need to update twice if we deprecated and replaced these functions. 

Please do let me know what you think though.

isleshocky77:
@Dillon-Brown The reason I went with making them static was that everything inside the methods and all methods used had no reason to not be static.

serhiisamko091184:
Hello @isleshocky77, 

This PR has been marked as stale because there has been no recent activity. It will be closed if no further activity occurs.
If that's possible, please,  update your branch and point it to the hotfix.
Many thanks!

Regards,
Serhii

serhiisamko091184:
Hey @isleshocky77 

Thank you.

Regards,
Serhii

PR #7058, opened 2019-03-14
Fixed #7054 - Email body blank when sent as plaintext only

### LINKED ISSUE #7054: Email body blank when sent as plaintext only
#### Issue
Some non imported emails are showing a blank body in the detail view even though they have a body when viewed in my usual email client. 

#### Expected Behavior
When viewing emails in the detail view the body should be displayed.

#### Actual Behavior
When the email is sent as plain-ASCII only and has no HTML component nothing is displayed in the body text.  :(

I am also seeing the following entry in the error_log which I think is related:
PHP Notice:  Undefined property: stdClass::$parts in /web/modules/InboundEmail/InboundEmail.php on line 3974

#### Possible Fix
I have done some debugging into the issue and looks like the root cause is that the description field is not defined in the metadata for the non imported detail view.

If you look at line 5642 of the returnNonImportedEmail function in /web/modules/InboundEmail/InboundEmail.php  you can see that it is checking if the 'description' field is in $fields_selected.  These $fields_selected are ultimately defined by the metadata of the non imported detail view in /modules/Emails/metadata/nonimporteddetailviewdefs.php.

When I add a new metadata entry below the description_html field I am now seeing the body of the email as expected. 
This is what I have added around line 120 of /modules/Emails/metadata/nonimporteddetailviewdefs.php:

`
array(
    'body' => array(
        'name' => 'description',
        'label' => 'LBL_BODY'
     ),
),
`

I am not sure if this is the correct fix for this problem and would appreciate if someone more knowledgeable than me could advise if there are any unintended side effects of this fix. 

#### Steps to Reproduce
1. Send an email with plaintext only to an email inbox 
2. Open the detail view from the email list.
3. See that the email has no body

#### Context
Without the fix in place this is causing issues in my logged emails as anyone that is sending as plaintext only is not getting the body of the email, which is the most important part!

#### Your Environment
I am running in a shared environment managed by CPanel with the following versions:
SuiteCRM v7.11.2
PHP - v7.1.26
MariaDB - 10.2.22
Browser - Firefox - 65.0.2 (64-bit)

Dillon-Brown:
## Description

Fixes an issue where non-imported emails display a blank body due to always being processed as HTML. This fix will detect what type of email is received and only run the SugarCleaner::clean once.

## Motivation and Context

Issue reference: #7054 

## How To Test This

1. Setup inbound/outbound email.
2. Send a plaintext and html email to yourself.
3. View both the plaintext and html email, check that the body is visible.
4. Import the emails and check the body is still visible.

## Types of changes

### Final checklist

### DIFF
--- a/modules/InboundEmail/InboundEmail.php
+++ b/modules/InboundEmail/InboundEmail.php
@@ -38,8 +38,6 @@
  * display the words "Powered by SugarCRM" and "Supercharged by SuiteCRM".
  */
 
-use SuiteCRM\Imap;
-use SuiteCRM\ImapInterface;
 use SuiteCRM\StateSaver;
 
 if (!defined('sugarEntry') || !sugarEntry) {
@@ -5612,8 +5610,9 @@ class InboundEmail extends SugarBean
                 $oldPrefix = $this->imagePrefix;
 
 
-                if (in_array('description_html', $fields_selected)) {
-                    $structure = $this->getImap()->fetchStructure($uid, FT_UID); // map of email
+                $structure = $this->getImap()->fetchStructure($uid, FT_UID);
+
+                if ($structure->subtype === 'HTML') {
                     $email->description_html = $this->getMessageTextWithUid(
                         $uid,
                         'HTML',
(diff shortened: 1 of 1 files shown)


cameronblaikie:
Assessed :+1:

PR #7062, opened 2019-03-15
FIX #7051 - Fixing a limit of 2.147.483 seconds for autoRefresh

### LINKED ISSUE #7051: getAutoRefresh() method can return a number bigger than 32 bits and freeze SuiteCRM
#### Issue

When setting up a very high number (more than 32 bits) in `autoRefresh` key for a Dashlet user's preferences, give us an unusable SuiteCRM instance. This kills the browser, the web server and the database.

When setting up the user's preferences for a Dashlet, there is one key named: `autoRefresh`. This key stores the seconds for autorefreshing the Dashlet, for example:

```php
       '5af7a2bb-b419-f463-855d-545a1f519a63' => array (
           'className' => 'age_permisoDashlet',
           'module' => 'age_permiso',
           'options' => array (
               'filters' => array (),
               'title' => 'Mi Permisos',
               'myItemsOnly' => false,
               'displayRows' => '5',
               'displayColumns' => array (
                   0 => 'name',
                   1 => 'numero_agente',
                   2 => 'punto_venta_numero',
                   3 => 'local',
                   4 => 'persona',
               ),
               'autoRefresh' => '4294967295',
           ),
           'fileLocation' => 'modules/age_permiso/Dashlets/age_permisoDashlet/age_permisoDashlet.php',
       ),
```

The problem lies in the "autoRefresh" key that had a value bigger than 32 bits. That value represents every few **seconds**  should be refreshed the Dashlet with the javascript function `setInterval()`

This function `setInterval()` receives 2 parameters: A function to be called and the interval
of time in **milliseconds** between calls.

To reconcile the value in seconds that are set as preferences for the Dashlet and the value in milliseconds that the function `setInterval()` requires, SuiteCRM multiplies that value by 1000, in the function `getAutoRefresh()` from the file `include/Dashlets/Dashlet.php`.

Multiplying 4294967295 per 1000 gives us a very high value that you could think this would result in the dashlet almost never being refreshed. The problem is that `setInterval()` function does not work like this, and when we pass a number of [more than 32 bits](https://stackoverflow.com/questions/12633405/what-is-the-maximum-delay-for-setinterval/12633556#comment78208539_12633488), it overflows and it is set to 0. Which meant that every 0 milliseconds the Dashlet tries to refresh itself.

This kills the browser, the web server and the database.

#### Expected Behavior

The function `getAutoRefresh()` should not return a number bigger than 32 bits.

#### Actual Behavior

The function `getAutoRefresh()` can return a number bigger than 32 bits.

#### Possible Fix

Patch `getAutoRefresh()` function to never return a number higher than 32 bits.

#### Steps to Reproduce

1. Set the value '4294967295' for the key autoRefresh in a Dashlet preference for a known user
2. Login SuiteCRM.
3. Open the browser Network console.
4. Verify that SuiteCRM tries to refresh the Dashlet every 0 seconds.

#### Your Environment

* SuiteCRM Version used: Every version of SuiteCRM
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Firefox 63, Chromium 72
* Environment name and version (e.g. MySQL, PHP 7): PHP 7.1

Abuelodelanada:
## Description

Read #7051

## Types of changes

### Final checklist

### DIFF
--- a/include/Dashlets/Dashlet.php
+++ b/include/Dashlets/Dashlet.php
@@ -373,7 +373,22 @@ class Dashlet
             $autoRefresh = $this->autoRefresh;
         }
 
-        return $autoRefresh * 1000;
+        $ret = $autoRefresh * 1000;
+
+        /**
+           This number is used by setInterval() function in JS
+           We should consider a limit of 2**31 -1
+           https://stackoverflow.com/questions/12633405/what-is-the-maximum-delay-for-setinterval/12633556#comment78208539_12633488
+         */
+        if ($ret > (pow(2, 31) - 1)) {
+            $ret = pow(2, 31) - 1;
+            LoggerManager::getLogger()->warn(
+                "The value of autoRefresh key in Dashlet: {$this->title} must be less than 2.147.483 seconds."
+                ."{$autoRefresh} was configured. Using 2.147.483 seconds instead."
+            );
+        }
+
+        return $ret;
     }
 
     /**


willrennie:
Assessed and tested this fix. It definitely resolves the issue with continual loading. Will hand over to another team member for code review.

PR #7068, opened 2019-03-18
Fix random unittest error in SugarControllerTest [hotfix] 

lazka:
Same as #7043 for the hotfix branch


Dillon-Brown:
Hi @lazka, thanks for this but you don't actually need to duplicate PR's for the hotfix branch, only the 7.8.x branch if necessary. We'll merge them across in #/7014.

lazka:
oh, I somehow remembered that differently, thanks :)

PR #7077, opened 2019-03-19
[optimize] max(id) and joining with self will always return one row

gunnicom:
"... WHERE id = (SELECT MAX(id) as id FROM ..."
MAX(id) will always be one row, so the query will be always one row. 
Rewriting the query without subquery is 10x faster for me.

It seems like in history it was used differently, but i can not see at all where this one row of the tracker is used. Maybe these code parts are totally useless, but still get called.

### DIFF
--- a/modules/Trackers/Tracker.php
+++ b/modules/Trackers/Tracker.php
@@ -113,8 +113,8 @@ class Tracker extends SugarBean
 	           $module_query = is_array($modules) ? ' AND module_name IN (\'' . implode("','" , $modules) . '\')' :  ' AND module_name = \'' . $modules . '\'';
 	        }
 
-	        $query = 'SELECT item_id, item_summary, module_name, id FROM ' . $this->table_name . ' WHERE user_id = \'' . $user_id . '\' AND deleted = 0 AND visible = 1' . $module_query . ' ORDER BY id DESC LIMIT 1';
-	        $result = $this->db->query($query,true,$query);
+	        $query = 'SELECT item_id, item_summary, module_name, id FROM ' . $this->table_name . ' WHERE user_id = \'' . $user_id . '\' AND deleted = 0 AND visible = 1' . $module_query . ' ORDER BY id DESC ';
+	        $result = $this->db->limitQuery($query,0,1,true,$query);
 	        while(($row = $this->db->fetchByAssoc($result))) {
 	               $breadCrumb->push($row);
 	        }
--- a/modules/Trackers/Tracker.php
+++ b/modules/Trackers/Tracker.php
@@ -110,14 +110,11 @@ class Tracker extends SugarBean
 			$breadCrumb = $_SESSION['breadCrumbs'];
 	        $module_query = '';
 	        if(!empty($modules)) {
-	           $history_max_viewed = 10;
 	           $module_query = is_array($modules) ? ' AND module_name IN (\'' . implode("','" , $modules) . '\')' :  ' AND module_name = \'' . $modules . '\'';
-	        } else {
-	           $history_max_viewed = (!empty($GLOBALS['sugar_config']['history_max_viewed']))? $GLOBALS['sugar_config']['history_max_viewed'] : 50;
 	        }
 
-	        $query = 'SELECT item_id, item_summary, module_name, id FROM ' . $this->table_name . ' WHERE id = (SELECT MAX(id) as id FROM ' . $this->table_name . ' WHERE user_id = \'' . $user_id . '\' AND deleted = 0 AND visible = 1' . $module_query . ')';
-	        $result = $this->db->limitQuery($query,0,$history_max_viewed,true,$query);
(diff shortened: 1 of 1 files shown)


Dillon-Brown:
Works for me, much faster and the history_max_viewed functionality is still intact. Nice find!

modules/Trackers/Tracker.php
```diff
@@ -110,14 +110,11 @@ function get_recently_viewed($user_id, $modules = '')
...
 	        if(!empty($modules)) {
-	           $history_max_viewed = 10;
 	           $module_query = is_array($modules) ? ' AND module_name IN (\'' . implode("','" , $modules) . '\')' :  ' AND module_name = \'' . $modules . '\'';
-	        } else {
-	           $history_max_viewed = (!empty($GLOBALS['sugar_config']['history_max_viewed']))? $GLOBALS['sugar_config']['history_max_viewed'] : 50;
 	        }
 
-	        $query = 'SELECT item_id, item_summary, module_name, id FROM ' . $this->table_name . ' WHERE id = (SELECT MAX(id) as id FROM ' . $this->table_name . ' WHERE user_id = \'' . $user_id . '\' AND deleted = 0 AND visible = 1' . $module_query . ')';
-	        $result = $this->db->limitQuery($query,0,$history_max_viewed,true,$query);
+	        $query = 'SELECT item_id, item_summary, module_name, id FROM ' . $this->table_name . ' WHERE user_id = \'' . $user_id . '\' AND deleted = 0 AND visible = 1' . $module_query . ' ORDER BY id DESC LIMIT 1';
```
samus-aran: shouldn't remove limitQuery and shouldn't add LIMIT due to MSSQL. But I think perhaps the select MAx can be removed...

willrennie:
@gunnicom This fix works for MySQL but I have tested this in an MSSQL environment and it fails with the following error: 

Fri Mar 22 16:44:11 2019 [3537][1][FATAL] SELECT item_id, item_summary, module_name, id FROM tracker WHERE user_id = '1' AND deleted = 0 AND visible = 1 ORDER BY id DESC LIMIT 1 Query Failed:SELECT item_id, item_summary, module_name, id FROM tracker WHERE user_id = '1' AND deleted = 0 AND visible = 1 ORDER BY id DESC LIMIT 1::: [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Incorrect syntax near 'LIMIT'.
Fri Mar 22 16:44:11 2019 [3537][1][FATAL] Exception in Controller: Database failure. Please refer to suitecrm.log for details.

modules/Trackers/Tracker.php
gunnicom: Ah, ok. Did not think about mssql. I just thought query should be faster than limitQuery so it was an additional optimization. But you are right maybe using a limitQuery with limit 1 is the better way to go.

pribeiro42: You can. maybe, use something like:

switch ($db_type) {
  case 'mysql':
    $query = .... // Your  proposed query
    break;
  case 'mssql':
    $query = "SELECT .... ORDER BY id DESC OFFSET 0 ROWS FETCH FIRST ROW ONLY" ;
    break;
}

This gives the correct syntax for the query, allowing for future expansion...

gunnicom: Can you check again if the new patch is ok with mssql?

gunnicom:
Did some changes for mssql.

modules/Trackers/Tracker.php
```diff
@@ -110,14 +110,11 @@ function get_recently_viewed($user_id, $modules = '')
...
 	        if(!empty($modules)) {
-	           $history_max_viewed = 10;
 	           $module_query = is_array($modules) ? ' AND module_name IN (\'' . implode("','" , $modules) . '\')' :  ' AND module_name = \'' . $modules . '\'';
-	        } else {
-	           $history_max_viewed = (!empty($GLOBALS['sugar_config']['history_max_viewed']))? $GLOBALS['sugar_config']['history_max_viewed'] : 50;
 	        }
 
-	        $query = 'SELECT item_id, item_summary, module_name, id FROM ' . $this->table_name . ' WHERE id = (SELECT MAX(id) as id FROM ' . $this->table_name . ' WHERE user_id = \'' . $user_id . '\' AND deleted = 0 AND visible = 1' . $module_query . ')';
-	        $result = $this->db->limitQuery($query,0,$history_max_viewed,true,$query);
+	        $query = 'SELECT item_id, item_summary, module_name, id FROM ' . $this->table_name . ' WHERE user_id = \'' . $user_id . '\' AND deleted = 0 AND visible = 1' . $module_query . ' ORDER BY id DESC ';
```
code-ph0y: Please look over queries taking into consideration SQL Injection - using the quoted() function where appropriate.

gunnicom: I do not see that there ist more SQL injection possible than before. But feel free to adjust this.

Mac-Rae:
Please rebase this pull request to the hotfix-7.10.x branch as version 7.8.x of SuiteCRM has reached its end of life and therefore will no longer be receiving updates. 

For more information about 7.8.x end of life please visit our [blog post](https://suitecrm.com/7-8-x-end-of-life/)

gunnicom:
No, wont do.

PR #7082, opened 2019-03-20
EmailTemplates: Fix undefined property error when creating a new template

lazka:
## Description

When a non-admin user creates a new template the "type" property isn't set resulting
in a PHP error.

Add a check to to prevent that and merge the two identical branches while at it.

## How To Test This

1) Log in as a non-admin user
2) Go to Email Templates
3) Try creating a new template
4) Check PHP logs

## Types of changes

### Final checklist

### DIFF
--- a/modules/EmailTemplates/EditView.php
+++ b/modules/EmailTemplates/EditView.php
@@ -110,10 +110,7 @@ if (empty($focus->id)) {
 
 echo getClassicModuleTitle($focus->module_dir, $params, true);
 
-if (!$focus->ACLAccess('EditView')) {
-    ACLController::displayNoAccess(true);
-    sugar_cleanup(true);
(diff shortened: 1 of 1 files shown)


modules/EmailTemplates/EditView.php
```diff
@@ -113,7 +113,7 @@
 if (!$focus->ACLAccess('EditView')) {
     ACLController::displayNoAccess(true);
     sugar_cleanup(true);
-} elseif (!is_admin($current_user) && $focus->type === 'system') {
+} elseif (!is_admin($current_user) && isset($focus->type) && $focus->type === 'system') {
```
Dillon-Brown: Works for me but it should be combined with the previous if statement as the code in the body is duplicated.

PR #7087, opened 2019-03-21
StateChecker: Don't save hash debug traces (unused, slow and memory intensive)

lazka:
## Description

On every state hash check the StateChecker saved the whole debug_backtrace() and never
removed it again, resulting in slow execution and ever growing memory usage.

Since it wasn't used anywhere just remove it.

This reduces the test time by ~10% and the memory usage by ~70% (on travis-ci)

The second commit removes the now unneeded "redefineMemoryLimit" option in "StateCheckerConfig".

## Motivation and Context

I couldn't run the full test suite on my local machine because they needed too much
memory (>16GB)

## How To Test This

Run the unit tests.

## Types of changes

### Final checklist

This requires documentation changes for https://github.com/salesagility/SuiteDocs/blob/master/content/developer/Appendix%20C%20-%20Automated%20Testing.adoc

I will look into that if this change gets approved.

### DIFF
--- a/include/StateChecker.php
+++ b/include/StateChecker.php
@@ -112,24 +112,9 @@ class StateChecker
             throw new StateCheckerException('Incompatible DB type, only supported: mysqli');
         }
         $this->resetHashes();
-        
-        if (StateCheckerConfig::get('redefineMemoryLimit')) {
-            $this->memoryLimit = ini_get('memory_limit');
-            ini_set('memory_limit', -1);
-        }
                 
         if (StateCheckerConfig::get('autoRun')) {
             $this->getStateHash();
-        }
-    }
-
-    /**
-     *
-     */
-    public function __destruct()
-    {
-        if (StateCheckerConfig::get('redefineMemoryLimit')) {
-            ini_set('memory_limit', $this->memoryLimit);
         }
     }
 
--- a/include/StateChecker.php
+++ b/include/StateChecker.php
@@ -86,13 +86,7 @@ class StateChecker
      * @var string
      */
     protected $lastHash;
-    
-    /**
-     *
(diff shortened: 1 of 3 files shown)


lazka:
I've created https://github.com/salesagility/SuiteDocs/pull/251 for the docs change

Dillon-Brown:
@lazka this one good to close now that #7329 has been merged?

lazka:
#7329 doesn't help me as it only changes travis-ci.

You can of course close this one, then I'll have to keep this patch in my fork.

Dillon-Brown:
@lazka Ah I see, have you tried setting ```save_traces``` and ```redefine_memory_limit``` to false in your config? that should fix the memory issue locally without needing to make a core change.

lazka:
Disabling saveTraces also disables printing of exception stack traces which makes testing hard.

Dillon-Brown:
Makes sense, perhaps we need a third config value then, to allow printing the exception stack trace but not saving the entire ```debug_backtrace```.

lazka:
Why are stack traces disabled in this case anyway?

What create a new PR where I change save_traces to default to false and enable printing stack traces in all cases?

Dillon-Brown:
Tbh I'm not sure that how it currently functions is the original intention. I think having ```saveTraces``` set to true might actually be an accidental commit, according to the docs it should be set to false by default. 

I don't see any issues with having ```saveTraces``` set to false and printing stack traces in all cases, seems like the best option.

lazka:
> I don't see any issues with having `saveTraces` set to false and printing stack traces in all cases, seems like the best option.

ok, perfect, I'll open a new PR tomorrow

Issue #5936, opened 2018-05-24
Date filter not being saved


chogori92:
#### Issue
When using date field as filter, it works but it's not saved. For example filter meetings by date start (ex. today), click on the first record and try to get again to meetings: you'll see no filter applied.

This issue occours when using date formats different from database one. If you use YYYY-mm-dd date format, it works fine.
You can reproduce in the online demo.

#### Expected Behavior
Filter should be saved.

#### Actual Behavior
Missing datetime fields from filter, after moving to record's detail or another module. 

#### Possible Fix
Check date format prior conversion at modules/MySettings/StoreQuery.php:77 
$db_format = $timedate->to_db_date($value, false);
This line causes conversion exception when using d/M/Y format.

#### Steps to Reproduce
1. Go to Meetings module.
2. Apply a filter to date start field (ex. today).
3. Click on a record to get detail view.
4. Come back to Meetings.
5. Filter lost!

#### Your Environment
* SuiteCRM Version used: 7.8.8 LTS
* Browser name and version: Chrome 66.0.3359.181 (64 bit)
* Environment name and version: Apache/2.4.18 + PHP 7 + MySQL 5.7.21
* Operating System and version: Ubuntu Server 16.04 LTS

connorshea:
I'm seeing the same thing with a date range filter. SuiteCRM version is 7.9.17, PHP 7.1.

connorshea:
It seems like this issue is the same as #3857, which suggests the same fix.

Dillon-Brown:
Duplicate of #3857

PR #7090, opened 2019-03-21
Preserve dates properly when restoring a filter's query parameters

### LINKED ISSUE #3857: When using filters date/time fields do not get restored after going back to list view
#### Issue

When doing a basic or advanced search, if your query includes a date/time column (for example "Date Created" or "Date Modified") the results returned are as expected,
screen shot 2017-07-09 at 4 00 51 pm

however if you click on the results and then go back to the list view, 
screen shot 2017-07-09 at 4 01 46 pm

the filter no longer displays the value for the date/time field .. it gets lost or is not stored properly?

upon returning to list view, because the filter for date/time does not work, the result list will be larger:
(you can't really see it in the screenshot with demo data because "Date Created" is the same across all records)
 if you searched for records created after July 9th and you got 25 results, when you go back to list view you will have results since the beginning of time because the filter is broken

screen shot 2017-07-09 at 4 01 56 pm

I noticed that this only happens to date/time type fields, other fields like "Assigned user" or "Lead Status" will be restored properly; 

It works as expected only when user's date format is set to '2010-11-23' but does not work with '11/23/2010' or other user date format

This bug is not present in 7.6.3 or in 7.7.9
the bug is present in 7.8.2, 7.8.5, 7.9.2

#### Expected Behavior

All query parameters should be restored upon returning to list view

#### Actual Behavior

Values for date/time fields are lost upon returning to list view

#### Possible Fix

Don't know, but I think it has to do with converting from the user set date format to mysql date format to run the query, and then restoring it back;

If you look at the first screenshot, notice how the columns have "Date Created" in 11/23/2010 format,
however the tooltip that is shown when you point at the filter button displays the "2010-11-23" format for mysql

#### Steps to Reproduce

1. install latest 7.9.2 or LTS 7.8.5 with demo data
2. in admin panel go to "developer tools",  Studio > Contacts > Layouts > Filter > Quick Filter
3. add "Date Entered" or "Date Modified" to default, click "Save and Deploy"
4. go to the Contacts module, do a basic search/filter and set any a value to "Date Created" click on any result
5. go back to Contacts module, click on filter button again... Date Created is now empty (but will still show condition Equals, Before, After, etc)

#### Context
our company has a custom date/time field to mark our client's requested service date,
it is important to us that we're able to efficiently search through our records
i think this is a low/medium priority type bug

#### Your Environment

* SuiteCRM Version used: 7.6.3, 7.7.9 (works as expected); 7.8.2, 7.8.5, 7.9.2 (does not work)
* Browser name and version: Chrome Version 59.0.3071.115 (Official Build) (64-bit)
* Environment name and version (e.g. MySQL, PHP 7): mysql 5.7.15
* Operating System and version (e.g Ubuntu 16.04): ubuntu 16.04 (server/crm), os x 10.10.5 (client)

connorshea:
## Description
Fixes #3857 and #5936. Credit to @frappuccino284 for suggesting the fix to this issue in #3857.

This adds proper handling in the `storeQuery.php` file for the case where the date is already provided in 'database' format (e.g. '2019-03-21').

I applied this fix to our fork of the CRM and can confirm it fixes the problem :)

## Motivation and Context
Date-related filters weren't preserved because the `TimeDate->to_db_format` method returns `''` if you give it a date that's already in db format. (Arguably that should be what's fixed here, but I'm not sure if that'd be a big breaking change so I'm just fixing this specific issue instead).

## How To Test This
You can test this with any module that has a filterable date property by filtering by that date property, clicking on a result, then returning to the listview for the module and checking that the filter retains the date filter info.

Or, if you want a specific example:
- In the Admin panel go to Studio > Contacts > Layouts > Filter > Quick Filter
- Add "Date Entered" or "Date Modified" to default, click "Save and Deploy"
- Go to the Contacts module, do a basic search/filter and set "Date Created" to any value, and then click on any result.
- Go back to the Contacts list view, hover over the filter button and note that the date property has been preserved in the filter.

## Types of changes

### Final checklist

### DIFF
--- a/modules/MySettings/StoreQuery.php
+++ b/modules/MySettings/StoreQuery.php
@@ -53,8 +53,8 @@ class StoreQuery
     /**
      * SaveQuery
      *
-     * This function handles saving the query parameters to the user preferences
-     * SavedSearch.php does something very similar when saving saved searches as well
+     * This function handles saving the query parameters to the user preferences.
+     * SavedSearch.php does something very similar when saving saved searches as well.
      *
      * @see SavedSearch
      * @param $name String name  to identify this query
@@ -73,7 +73,14 @@ class StoreQuery
                             $type = $bean->field_defs[$field]['type'];
 
                             if (($type == 'date' || $type == 'datetime' || $type == 'datetimecombo') && !preg_match('/^\[.*?\]$/', $value)) {
-                                $db_format = $timedate->to_db_date($value, false);
+                                // If the value is already in the db date format (e.g. '2019-03-21'), don't re-convert
+                                // it as that causes $db_format to be set to nothing. If the value isn't in
+                                // the format that the db wants (e.g. '3/21/2019'), then we can convert it.
+                                if (preg_match('/^\d{4}-\d{2}-\d{2}$/', $value)) {
+                                    $db_format = $value;
+                                } else {
+                                    $db_format = $timedate->to_db_date($value, false);
+                                }
                                 $this->query[$key] = $db_format;
                             } elseif ($type == 'int' || $type == 'currency' || $type == 'decimal' || $type == 'float') {
                                 if (preg_match('/[^\d]/', $value)) {


connorshea:
Force pushed to get the tests to re-run and hopefully succeed this time.

Dillon-Brown:
@connorshea Looks good, does this need a hotfix-7.8.x and hotfix-7.10.x PR as well?

willrennie:
@connorshea I have tested this in both MySQL and MSSQL with your fix. On the initial filter save, the date appears to be preserved and the results are filtered but if I navigate away from the list view and return, the filter is not kept. I have ensured I am on a clean copy of your fix with fresh cache, correct permissions and having followed your steps to reproduce/test.

gymad:
(I approved it as the code seems ok but I will be more happy if travis also says it's the right move here. :)

connorshea:
@willrennie you're on MySQL 5.7 and this branch of the CRM? Odd, the fix definitely works fine on my CRM but it seems like you're still getting the old behavior. I'm admittedly on an older version of the CRM (7.9.17), but I'd assumed the fix would still work. How annoying. I guess I'll see if I can get a CRM set up running on 7.11.

connorshea:
@Dillon-Brown it definitely needs a 7.10 PR, I haven't used 7.8 but based on the age of the issues for this I'd say that's probably necessary too.

willrennie:
@connorshea I directly checked out your branch. I say MySQL, I stand corrected I'm actually running MariaDB and MSSQL. Let me go check the versions and get back to you. Definitely on a fresh clone/checkout of your fix branch.

Lehnerr:
Problem is still there in 7.8.31. Fix helps in this Version too.
Pls test and include it in next 7.8.32 patch

Mac-Rae:
Assessed :+1: Not sure why this wasn't merged already? @Dillon-Brown Any ideas?

Dillon-Brown:
@Mac-Rae Hasn't been merged due to the feedback from @willrennie. Did you encounter the same issue or did everything work as expected? (Make sure to test on MSSQL as well).

connorshea:
I'm not sure what the best way to write tests for this would be 🤔

connorshea:
I've rebased this on top of the latest hotfix-7.10.x and changed the target branch to match.

connorshea:
I've opened #8032 as a replacement since rebasing this PR seems to have confused Travis, so I'm closing this.

PR #7118, opened 2019-03-27
Fixes #7111 - composer: include mpdf/mpdf 6.1 and drop in-tree version (5.7)

### LINKED ISSUE #7111: Update mPDF and move it to Composer
#### Issue
https://github.com/mpdf/mpdf

#### Problems with PHP 7.x

lazka:
## Description

Remove the included version of mpdf (in modules/AOS_PDF_Templates/PDF_Lib) and
add it to composer instead.

This also updates the version from 5.7.1 to 6.1.0 to get some fixes that were already
applied manually to the version in SuiteCRM. There are many newer versions available,
but they require more changes and this is a good first step (and this potentially
could be backported to 7.10)

The version in tree had one manual fix applied (d1279ca) which is included in
6.1.0 already, so the upstream version can be used as is.

The required changes to make things compatible with 6.1 were
* mPDF_exception is now called MpdfException
* SetAutoFont() no longer exists, the docs suggest to use autoLangToFont instead
  to get the same results as before.

Fixes #7111

## Motivation and Context

Less code to maintain, better PHP compatibility.

## How To Test This

There are three places for generating PDFs:

1) Create a report and generate PDF
2) Create a PDF template for leads, go to leads and generate a PDF
3) Create a PDF template for contracts, create a contract and generate a PDF

I've tested all three with some basic reports/templates.

## Types of changes

### Final checklist


Dillon-Brown:
@lazka This would need to go to develop since it's a major change.

lazka:
ok, I'll have a look

lazka:
done

pgorod:
Thanks for this @lazka!

cameronblaikie:
Assessed :+1:

connorshea:
How realistic would it be to backport this to 7.10?

connorshea:
So, FWIW I backported this change to 7.10 and lots of errors/notices/warnings have been fixed by it, but there was one weird change that required changing a PDF template we have. It made the font sizing much smaller, but I managed to fix it by removing a containing div and adding `font-size: x-small; width: 100%;` to the table element. We haven't deployed it to production quite yet, but it looks like it works pretty well.

Dillon-Brown:
@connorshea Agreed this would be a good one to get into hotfix-7.10.x. I've shoved up a quick backported PR here: #7948

PR #7130, opened 2019-03-29
Update composer.json

Dillon-Brown:
## Description

Updating composer.json + composer.lock with a few small changes:

- Moving leafo/scssphp to require-dev: leafo/scssphp is a dev package and shouldn't be pulled down when running ```composer install```.
- Add optimize-autoloader flag: Convert PSR-0/4 autoloading to classmap to get a faster autoloader.
- Add sort-packages flag: Keep packages sorted in composer.json.
- Updating browserstack/browserstack-local to use ```^1.1``` instead of ```dev-master```: [Unbound version constraints should be removed](https://getcomposer.org/doc/faqs/why-are-unbound-version-constraints-a-bad-idea.md).
- Updating the autoloader, changing ```tests/unit/lib/SuiteCRM``` to ```tests/unit/phpunit/lib/SuiteCRM``` as the previous path no longer exists.
- Add "roave/security-advisories" to composer: [Roave Security Advisories](https://github.com/Roave/SecurityAdvisories) is a simple composer exclusion list to ensure that we don't installed dependencies with known security vulnerabilities.
- Add project support information: links to the documentation site, support forum, issue tracker, developer chat, project source and main website.
- Remove duplicate: symfony/yaml was in both require and require-dev (Can result in unexpected behaviour).
- Move "filp/whoops" to require-dev: dev packages should not be in "require".
- Add -f flag to "post-install-cmd": this is to prevent ```composer install``` from failing when the Elasticsearch/Tests directory has already been removed.

## How To Test This

Functionality should remain the same, the only thing to really test is that composer install works and the support information links are correct. 

## Types of changes

### Final checklist


willrennie:
Assessed the composer install and update. Install succeeds and no longer fails on elastic search removal.

Issue #6455, opened 2018-10-30
The V8 API does not allow filtering by custom fields


ajprouty87:
#### Issue

When using the version 8 API to list out module entities (for example, accounts) you cannot filter by custom fields. 

#### Expected Behavior

When running a GET request against /api/V8/module/Account?filter[my_custom_field_c][EQ]=MyCustomValue it would return all Account entities where the custom field 'my_custom_field_c' is equal to 'MyCustomValue'.

#### Actual Behavior

When running a GET request against /api/V8/module/Account?filter[my_custom_field_c][EQ]=MyCustomValue it returns a 400 error along with the detail message of:

> Filter field my_custom_field_c in Account module is not found

This happens even through my_custom_field_c was already added as a custom field to the account module.

#### Possible Fix

Here is the original code that I believe caused this problem:
https://github.com/salesagility/SuiteCRM/blob/475f732cef5b0e02b0f7208446c40b5a985aa406/Api/V8/JsonApi/Repository/Filter.php#L47-L69

You can see that it only checks to see if the field exists on the bean. Custom fields do not get added directly to the bean. The second problem is that the table is hard coded to filter against $bean->getTableName();

If you replace that section of code with the following it seems to be an acceptable solution:

```
if ( property_exists($bean, $field) ) {
    $tableName = $bean->getTableName(); 
} else if ( $bean->custom_fields->fieldExists($field) ) {
    $tableName = $bean->getTableName() . "_cstm";
} else {
    throw new \InvalidArgumentException(sprintf(
        'Filter field %s in %s module is not found',
        $field,
        $bean->getObjectName()
    ));
}

if (!is_array($expr)) {
    throw new \InvalidArgumentException(sprintf('Filter field %s must be an array', $field));
}

foreach ($expr as $op => $value) {
    $this->checkOperator($op);
    $where[] = sprintf(
        '%s.%s %s %s',
        $tableName,
        $field,
        constant(sprintf('%s::OP_%s', self::class, strtoupper($op))),
        $this->db->quoted($value)
    );
}
```

I switched out the table name into a local variable. If the property exists on the bean it uses the bean table name (previous logic). If it does not exist on the bean, however, it does exist in the custom fields it will add '_cstm' to the table name. This works because the query already automatically adds the custom table to it. If it does not exist in either place it will still throw the exception. 

#### Steps to Reproduce

1. Create a custom field on a module (Accounts for instance).
2. Create an access token through an API request (I use postman).
[...]

PR #7142, opened 2019-04-04
Fix custom fields in sort/filter in the API

JimMackin:
## Description

Fix the check for existing fields in the filtering and sorting in the API.

Previously this was checking for the fields as properties on the bean. However custom fields are not explicitly defined as properties. Instead we check the field defs which is a more reliable indicator of whether or not a field exists.

## Motivation and Context

Without this fix it is not possible to filter or sort on any custom field.

## How To Test This

Calling the API with a custom field as a filter will fail before this fix and work as expected afterwards.
The related issue (https://github.com/salesagility/SuiteCRM/issues/6455) provides information. 
## Types of changes

### Final checklist

### DIFF
--- a/Api/V8/JsonApi/Repository/Filter.php
+++ b/Api/V8/JsonApi/Repository/Filter.php
@@ -45,7 +45,7 @@ class Filter
 
         $where = [];
         foreach ($params as $field => $expr) {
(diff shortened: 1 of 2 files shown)


cameronblaikie:
Assessed :+1:

Issue #6379, opened 2018-10-03
Unable to GET deleted records through API


tsummerer:
#### Issue

The REST API does not provide a way to access soft-deleted records. 

The code sort of exists, but there's no way to access it. (see samples below)

#### Expected Behavior

If an API request has the `deleted` flag set to `1`, then deleted records should be included in the response

#### Actual Behavior

The flag is ignored and only non-deleted records are returned.

In `lib/API/v8/Library/ModulesLib.php` under the `getModuleList` function, the following code exists:
```php
protected function getModuleList(Request $req, \SugarBean $module, array $args = array())
    {
        /** @var array $page */
        $page = $req->getParam('page');

        // Order by (sorting)
        $orderBy = $this->getSorting($req);

        // Pagination (offset)
        $currentOffset = isset($page['offset']) ? (integer)$page['offset'] : -1;

        // Pagination (page limit)
        $limit = isset($page['limit']) ? (integer)$page['limit'] : -1;

        // Maximum results (-1 === Unlimited)
        $maximumResults = -1;

        // Show deleted results
        $show_deleted = 0;
        if (isset($filter['deleted'])) {
            $show_deleted = (integer)$filter['deleted'];
        }
```
Under "Show deleted results", you'll notice that the `$filter` variable is never defined. So of course the `isset` condition will always return false. 

The `$show_deleted` variable is later passed to the `get_list` function. Due to the code above, it is always false. 

#### Possible Fix

Consider adding a 'deleted' flag. In which case you would refactor to something like this:
```
if ( $req->getParam('deleted')){
   $show_deleted = (integer) $req->getParam('deleted');
}
```
Alternatively, refactor further and pull it from the `filters` parameter. 

#### Steps to Reproduce

1. Start with a Suite instance with Demo Data (or create a few records)
2. Delete a record
3. Perform a GET to the corresponding module (`/api/v8/modules/[MODULE]?deleted=1`) with the `deleted` flag set

#### Context

An external connector may need to know if a record is deleted. This can be important if two systems are trying to stay in sync. 

The functionality should either be added, or the code cleaned up.

#### Your Environment

* SuiteCRM Version used: I confirmed in 10.7, and the code appears to be the same in 10.9
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): n/a
* Environment name and version (e.g. MySQL, PHP 7): LAMP with PHP 7
* Operating System and version (e.g Ubuntu 16.04):

[...]

PR #7144, opened 2019-04-04
Fix API list total pages display

PedroErnst:
Fixes an issue where API filtering includes deleted records by default.

## Description

Adds a default filter parameter to use when no filters are specified.
Injects a deleted = 0 filter parameter when the list of filter parameters doesn't specifically include deleted records.

## Motivation and Context

Most API operations are unaffected by this issue because we use a Bean function to actually retrieve records, so deleted records are filtered.

However, for the `[meta][total-pages]` a count query is run which does not include the deleted = 0 parameter, and thus returns a number of pages based on the total rows in the db, deleted or not.

## How To Test This

1. Make sure your Accounts db table contains deleted records as well as non deleted records.
2. Do a call to `Accounts?fields[Account]=name,account_type&page[number]=1&page[size]=1`
3. Observe the `[meta][total-pages]` is accurate for the data in the db.

## Types of changes

### Final checklist

### DIFF
--- a/Api/V8/JsonApi/Repository/Filter.php
+++ b/Api/V8/JsonApi/Repository/Filter.php
@@ -80,7 +80,7 @@ class Filter
      * @param array $params
      * @return array
      */
-    protected function addDeletedParameter(array $params): array
+    protected function addDeletedParameter(array $params)
     {
         if (!array_key_exists('deleted', $params)) {
             $params['deleted'] = [
--- a/Api/V8/JsonApi/Repository/Filter.php
+++ b/Api/V8/JsonApi/Repository/Filter.php
@@ -43,6 +43,8 @@ class Filter
             unset($params['operator']);
         }
 
+        $params = $this->addDeletedParameter($params);
+
         $where = [];
         foreach ($params as $field => $expr) {
             if (!property_exists($bean, $field)) {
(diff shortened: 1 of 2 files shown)


isleshocky77:
I just ran across this exact issue. The patch definitely fixes the page count for case where you're only requesting records with no filter and would not expect to see deleted results; however,

Although I don't need it, I also decided to test this code with getting deleted records. I'm unable to get deleted records. I tried the following requests

```
# Returns a response stating no result
Api/V8/module/Accounts?fields[Account]=name,account_type&page[number]=1&page[size]=50&filter[deleted][eq]=1
```

```
# Returns the same result as not specifying a filter at all
Api/V8/module/Accounts?fields[Account]=name,account_type&page[number]=1&page[size]=50&filter[deleted][eq]=1&filter[operator]=or&filter[deleted][eq]=0
```

PedroErnst:
Yeah the fix is only for the page count.

Not being able to retrieve deleted records is still an open issue.

isleshocky77:
@PedroErnst Just noting that there is a separate issue in for retrieving deleted record #6379

Dillon-Brown:
Issue reference: #7144

willrennie:
Assessed as working on the fix branch using example test within Postman from the issue this resolves.

Issue #6893, opened 2019-02-03
API V8 unable to update record with decimals


LaoSteve:
#### Issue
API V8 unable to update record with decimals

#### Expected Behavior
record updated

#### Actual Behavior
ProtocolErrorThe remote server returned an error: (400) Bad Request

#### Steps to Reproduce
This works:
https://XXXXXXXXX/SuiteCRM/Api/V8/module
{"data":{"type":"Opportunity","id":"38cac79d-b3dd-3bf4-a9cd-5c56da62666f","attributes":{"description":"test","amount":4444}}}

This doesn't
https://XXXXXXXXX/SuiteCRM/Api/V8/module
{"data":{"type":"Opportunity","id":"38cac79d-b3dd-3bf4-a9cd-5c56da62666f","attributes":{"description":"test","amount":4444.55}}}
ProtocolErrorThe remote server returned an error: (400) Bad Request

This doesn't
https://XXXXXXXXX/SuiteCRM/Api/V8/module
{"data":{"type":"Opportunity","id":"38cac79d-b3dd-3bf4-a9cd-5c56da62666f","attributes":{"description":"test","amount":"4444.55"}}}
ProtocolErrorThe remote server returned an error: (400) Bad Request. 

#### Context
seems to affect every module numeric fields, whole numbers update OK but not decimals

#### Your Environment

* SuiteCRM Version 7.11.1
* Browser name and version Firefox 65.0
* Environment name and version VirtualBox 5.2.20
* Operating System and version Ubuntu 18.04

PR #7161, opened 2019-04-08

### LINKED ISSUE #6452: Api/V8 create record does not support unicode and space in attributes

Dillon-Brown:
## Description

The `Attributes.php` parameter uses the `Fields.php` parameter's regex pattern to verify that attributes were formatted correctly. I've added a new regex to `Attributes.php` as the `fields` parameter validation shouldn't be re-used in the attributes class since while the `fields` parameter needs to restrict spaces and Unicode, attributes doesn't. 

I've also removed the ```Assert\NotBlank``` validation as we shouldn't need to exclude blank values.

## Motivation and Context

It is currently impossible to create/edit a record with a space or Unicode character in any of the fields due to the attributes validation using ```REGEX_FIELD_PATTERN = '/[^\w-,]/'```.

Issue reference: #6452 
Issue reference: #6893 

## How To Test This

```
POST SuiteCRM/Api/V8/module
BODY {
  "data": {
    "type": "Accounts",
    "attributes": {
      "name": "Cocacola company",
    }
  }
}
```

## Types of changes

### Final checklist

### DIFF
--- a/Api/V8/Param/Options/Attributes.php
+++ b/Api/V8/Param/Options/Attributes.php
@@ -10,7 +10,7 @@ use Symfony\Component\Validator\Constraints as Assert;
 
 class Attributes extends BaseOption
 {
-    const REGEX_ATTRIBUTE_PATTERN = '/[\b\B]/';
+    const REGEX_ATTRIBUTE_PATTERN = '/\b\B/';
 
     /**
      * @inheritdoc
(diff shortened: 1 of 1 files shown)


icegit:
Hi gentlemen,
`const REGEX_ATTRIBUTE_PATTERN = '/\b\B/';` doesn't work for me 

However this one does:
`const REGEX_ATTRIBUTE_PATTERN = '/[\b\B]/';`

Thanks for the fix.

willrennie:
Assessed as working on the fix branch using example test within Postman from the issue this resolves.

PR #7173, opened 2019-04-10
Fix V8 API authorization header passing with apache+php-fpm

lazka:
## Description

With apache+php-fpm the authorization header gets stripped and never
reaches the app. The .htaccess already includes some rules to pass them through
with an env var, but because of the rule order this is never applied.

Fix the rewrite rules by moving the env var rules before the rules marked as [L]ast.

Passing an env var with a rewrite rule results in apache prefixing the env
var with "REDIRECT_". To work around this copy REDIRECT_HTTP_AUTHORIZATION to
HTTP_AUTHORIZATION in case only the former is set.

## Motivation and Context

I wanted to try the API, but it didn't work.

I also found similar issues in the forum (which inspired this patch):

* https://suitecrm.com/suitecrm/forum/installation-upgrade-help/21465-missing-authorization-header
* https://suitecrm.com/suitecrm/forum/suitecrm-7-0-discussion/19783-missing-authorization-header-issue

## How To Test This

* Use the API with APache and php-fpm

## Types of changes

### Final checklist

### DIFF
--- a/Api/Core/app.php
+++ b/Api/Core/app.php
@@ -10,6 +10,13 @@ if (!defined('sugarEntry')) {
 }
 // @codingStandardsIgnoreEnd
 
+// For php-fpm we pass the "Authorization" header through HTTP_AUTHORIZATION
+// using .htaccess rewrite rules. The rewrite rules result in apache prefixing
+// the env var which gives us REDIRECT_HTTP_AUTHORIZATION.
+if (!isset($_SERVER['HTTP_AUTHORIZATION']) && isset($_SERVER['REDIRECT_HTTP_AUTHORIZATION'])) {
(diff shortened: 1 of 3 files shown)


cameronblaikie:
Assessed :+1:

Dillon-Brown:
Hi @lazka, would ```CGIPassAuth on``` work here? (See [docs](https://httpd.apache.org/docs/2.4/en/mod/core.html#cgipassauth)) your solution looks good for a quick-fix but it would be nice if we could come up with something a little more elegant. Thanks!

lazka:
I'm using centos 7.6.1810 which includes apache 2.4.6 and CGIPassAuth is only available with 2.4.13+

`.htaccess: Invalid command 'CGIPassAuth', perhaps misspelled or defined by a module not included in the server configuration`

PR #7194, opened 2019-04-17
Fixed callable in loop condition

Dillon-Brown:
## Description

Removed callable calls from loop termination condition, this will speed up those loops without any changes to existing functionality.

## How To Test This

See Travis, this should not change existing functionality.

## Types of changes

### Final checklist


include/php-sql-parser.php
```diff
@@ -2059,7 +2059,7 @@ private function load_reserved_words()
             'zerofill'
             );
 
-        for ($i=0;$i<count($this->reserved);++$i) {
+        for ($i=0, $iMax = count($this->reserved); $i< $iMax; ++$i) {
```
connorshea: This doesn't really matter, but some of these are formatted as `$i< $iMax` and some are formatted as `$i < $iMax`. I'm surprised the linter doesn't enforce the latter.

modules/AOD_Index/LuceneUtils.php
```diff
@@ -244,7 +244,7 @@ function rtf2text($filename)
                     $param = null;
 
                     // Start reading characters after the backslash.
-                    for ($k = $i + 1, $m = 0; $k < strlen($text); $k++, $m++) {
+                    for ($k = $i + 1, $m = 0, $kMax = strlen($text); $k < $kMax; $k++, $m++) {
```
connorshea: Does this loop having an `$m` value as well make any difference for the code here?

include/php-sql-parser.php
Dillon-Brown: Thanks! I had assumed that the linter would have taken care of the formatting. Will update this PR.

modules/AOD_Index/LuceneUtils.php
Dillon-Brown: Nope, the behaviour should be the same as before with $k and $m getting returned and incremented by one every loop.

Mac-Rae:
Assessed by Travis +1

However, Needs conflicts sorted before merge

PR #7201, opened 2019-04-17
Fixed #6455 - The V8 API does not allow filtering by custom fields

### LINKED ISSUE #6455: The V8 API does not allow filtering by custom fields

Dillon-Brown:
## Description

As custom fields are not explicitly defined as properties, you were unable to filter with them. With this fix, we use the field definitions instead of the properties.

## Motivation and Context

Issue reference: #6455 

## How To Test This

1. With the V8 API, check that using a custom field as a filter works as expected.

## Types of changes

### Final checklist

### DIFF
--- a/Api/V8/JsonApi/Repository/Filter.php
+++ b/Api/V8/JsonApi/Repository/Filter.php
@@ -45,7 +45,7 @@ class Filter
 
         $where = [];
         foreach ($params as $field => $expr) {
-            if (!property_exists($bean, $field)) {
+            if (empty($bean->field_defs[$field])) {
                 throw new \InvalidArgumentException(sprintf(
                     'Filter field %s in %s module is not found',
                     $field,
--- a/Api/V8/JsonApi/Repository/Sort.php
+++ b/Api/V8/JsonApi/Repository/Sort.php
@@ -18,12 +18,12 @@ class Sort
     public function parseOrderBy(\SugarBean $bean, $value)
     {
         $orderBy = self::ORDER_BY_ASC;
-        if ($value[0] === '-') {
+        if (strpos($value, '-') === 0) {
             $orderBy = self::ORDER_BY_DESC;
             $value = ltrim($value, '-');
         }
 
-        if (!property_exists($bean, $value)) {
+        if (empty($bean->field_defs[$value])) {
             throw new \InvalidArgumentException(sprintf(
                 'Sort field %s in %s module is not found',
                 $value,


connorshea:
Should this include a test case? 😇

Dillon-Brown:
@isleshocky77 I'm using ```empty``` which will also check ```isset```.

cameronblaikie:
Assessed :+1:

PR #7216, opened 2019-04-24
Fix travis due date

Dillon-Brown:
## Description

Adding a 5-second wait to the subpanel due date test to hopefully resolve failing Travis.

## How To Test This

See Travis.

## Types of changes

### Final checklist

### DIFF
--- a/tests/acceptance/modules/Activities/ActivitiesCest.php
+++ b/tests/acceptance/modules/Activities/ActivitiesCest.php
@@ -83,10 +83,11 @@ class ActivitiesCest
         $listView->waitForListViewVisible();
         $listView->clickNameLink($account_name);
 
-        //Click on Activites subpanel
+        //Click on Activities subpanel
         $I->click(['id'=>'subpanel_title_activities']);
         $I->waitForElementVisible('#Activities_createtask_button', 60);
         $I->expect('the due date is visible');
+        $I->wait(5);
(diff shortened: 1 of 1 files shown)


connorshea:
@Dillon-Brown does Codeception have anything like Capybara, where it can [automatically wait until an element appears on the page](https://github.com/teamcapybara/capybara#asynchronous-javascript-ajax-and-friends) instead of waiting a fixed amount of time? e.g. if the element appears after 3 seconds it only has to wait 3 seconds, but if it takes 5 seconds it'll wait 5 seconds, etc. (up to a configured max limit of X seconds).

connorshea:
Oh actually it seems like `$I->waitForElementVisible()` is exactly what we'd want, is there a reason that isn't used instead of a wait?

Dillon-Brown:
@connorshea I used wait here because it was slightly faster to type :). I agree we should be using waitForElementVisible instead of wait unless there is some edge case where wait is required (Wasn't planning on actually merging this PR, just wanted to kick off a Travis build).

connorshea:
@Dillon-Brown is there an actual fix for this? I'm suffering from it now :P

Dillon-Brown:
@connorshea This should work #7263 (Or at least it's passing on my Travis....). Will need to spend some time to look into the 7.8.x branch though, seems to be passing both locally and on my forked repo's Travis but then failing on SalesAgility/SuiteCRM.

PR #7227, opened 2019-04-26
Fixed #7188 - ACL doesn't work on JSON API V8

### LINKED ISSUE #7188: ACL doesn't work on JSON API V8
#### Issue
I am using JSON API authentication that will be used by some other external system to verify the user from a mobile App.
I have created few users inside CRM and generated oauth client for them using grant_type = "client_credentials", I have also assigned a role to the user and set up few permission, that seems to work fine inside CRM but not within JSON API v8 
I am able to get the token and APIs are giving the response but i am not able to restrict the user as whatever token i send it will give me all the data where it should return an HTTP forbidden 

#### Expected Behavior
JSON API should check ACL and return the data for allowed modules only
#### Actual Behavior
JSON API is returning back all the data without considering ACL
#### Possible Fix
I tried to play around with ModuleController and for each action there can be a function to check for ACL for the target module. e.g. if the user is not allowed access to Accounts module then it should return HTTP Code 403 etc.

This can also be added inside ModuleService getBean method so that whenever the API tries to access a bean it should throw and exception

#### Steps to Reproduce
1.Create a normal user in CRM.
2.Create a role and set the module permissions.
3.Assign the role to created user.
4.Generate OAuth client for the created user.
5.Call Api/access_token API to get a JWT.
6.With this JWT try to access the restricted module.

#### Context
Documentation states that ACL will work for API calls also but this didn't work and i have an urgent requirement to restrict the access of users accessing via mobile. I was under impression that Roles will work just fine and committed the timeline on the basis of authentication only, Now i am stuck,

#### Your Environment
* SuiteCRM Version used: 7.11.0
* Sugar Version 6.5.25 (Build 344)
* Browser name and version: Postman native app v6.7.4
* Environment name and version: PHP 7.1, Mysql 5.7.25
* Operating System and version: Ubuntu 18.04.01

Dillon-Brown:
## Description

Adds ACL functionality to the V8 API.

## Motivation and Context

Issue reference: #7188

## How To Test This

1. Create a user + role and set the module permissions.
2. Assign the role to the created user.
3. Generate OAuth client for the created user.
4. Call Api/access_token API for an access token.
5. Try to view/update/create records in restricted modules..

## Types of changes

### Final checklist

### DIFF
--- a/Api/V8/Service/ModuleService.php
+++ b/Api/V8/Service/ModuleService.php
@@ -14,6 +14,7 @@ use Api\V8\Param\GetModuleParams;
 use Api\V8\Param\GetModulesParams;
 use Api\V8\Param\UpdateModuleParams;
 use Slim\Http\Request;
+use SuiteCRM\Exception\AccessDeniedException;
 
 class ModuleService
 {
@@ -57,9 +58,9 @@ class ModuleService
 
     /**
      * @param GetModuleParams $params
-     * @param string $path
-     *
+     * @param $path
      * @return DocumentResponse
+     * @throws AccessDeniedException
      */
     public function getRecord(GetModuleParams $params, $path)
     {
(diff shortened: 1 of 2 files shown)


cameronblaikie:
Assessed :+1:

PR #7245, opened 2019-05-02
Image not found because "s" is added

gunnicom:
I do not know, if anybody else has this.
For me, Suite tries to load an  image "Meetingss.gif". So double ss in the end. The gif of course is named "Meetings.gif"
I know that at some places (at least in the past), plural of module names were handled differently, but do not know if that is the reason for the additional "s".

Same problem should exist in every version of SuiteCRM.
I found that problem by looking at the webserver log, or the network log in the browser.


Dillon-Brown:
Hi @gunnicom, thanks for the contribution! are you ok to close this one in favour of #7659? both look like they resolve the issue but #7659 is a little more general.

PR #7246, opened 2019-05-02
Add return to function

gunnicom:
Did not find at which point it was fixed for 7.10, but should be backportet to 7.8.x

https://github.com/salesagility/SuiteCRM/blob/1892b13ecfcd9a9e3cd575c8b4b3336188e86f28/include/SubPanel/SubPanelDefinitions.php#L432


cameronblaikie:
Assessed :+1:

PR #7257, opened 2019-05-06
Favorite link in active Module does not get generated

gunnicom:
The variable name is "module_name" not "module"


Mac-Rae:
Assessed by Travis :+1:

PR #7275, opened 2019-05-10
API - Enable getting deleted records in Get List

### LINKED ISSUE #6379: Unable to GET deleted records through API

PedroErnst:
Enable retrieving deleted records in Get List call

Does a quick check to see if deleted records are being requested.
Only works for Get List call.

## Motivation and Context

Potential fix for #6379 

## How To Test This

1. Make sure you have a deleted record in CRM
2. Query the API such as this `/V8/module/Accounts?fields[Accounts]=name,account_type&filter[operator]=and&filter[deleted][eq]=1`

## Types of changes

### Final checklist


Jason-Dang:
Assessed. Code looks good.

cameronblaikie:
Assessed :+1:

PR #6815, opened 2019-01-23
unittests: Fixes for PHP 7.3

lazka:
## Description

This fixes two types or new errors/warnings in PHP 7.3:

(1) PHP Warning:  "continue" targeting switch is equivalent to "break".
    Did you mean to use "continue 2"?

In all cases the continue was used as a break anyway, so just replace it
with the less confusing break statement.

(2) preg_replace(): Compilation failed: invalid range in character class at offset 4

Fix by properly escaping the "-" in the regex range.

This makes the unit tests pass with PHP 7.3 (the codeception tests still fail)

## Motivation and Context

Debian has php 7.3 now and this allows me to run the tests there.

## How To Test This

Run the unit tests with php 7.3

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist
- [x] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

PR #7303, opened 2019-05-21
Update SugarBean.php

kaleidoc0de:
although using continue within a switch is equivalent to break, this is causing a warning message to be printed.
as the switch is nested within a for each and continue is being used with breaks, it seems as if the original contributor intended to continue the iteration and not break out the switch.

adding "continue 2" improves readability, understanding of intent, and eliminates the warning message.

## Types of changes

### Final checklist

### DIFF
--- a/data/SugarBean.php
+++ b/data/SugarBean.php
@@ -2548,7 +2548,7 @@ class SugarBean
                         case 'currency':
                         case 'float':
                             if ($this->$field === '' || $this->$field == null || $this->$field == 'NULL') {
-                                continue;
+                                continue 2;
                             }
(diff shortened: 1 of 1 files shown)


lazka:
This was already fixed in #6815 using break.

>  as the switch is nested within a for each and continue is being used with breaks, it seems as if the original contributor intended to continue the iteration and not break out the switch.

True, but the switch is at the end of the loop anyway, so there shouldn't be any difference if I'm not mistaken.

Dillon-Brown:
Closed as equivalent fix already merged into hotfix-7.10.x here: #6815

PR #7322, opened 2019-05-29
*WIP Fixed #7111 - PDF Notices

### LINKED ISSUE #7111: Update mPDF and move it to Composer

Dillon-Brown:
## Description

Prevents seeing warnings in PDF Templates:

```
<br />
<b>Warning</b>:  Illegal string offset 'ID' in <b>/var/www/html/suitecrm/modules/AOS_PDF_Templates/PDF_Lib/classes/cssmgr.php</b> on line <b>1070</b><br />
<br />
```

## Motivation and Context

Issue reference: #7111 
https://suitecrm.com/suitecrm/forum/feedback/22105-print-as-pdf-error

## How To Test This

1. Create a PDF template with display_errors turned on.
2. Download the PDF and check for errors.

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOR_Reports/controller.php
+++ b/modules/AOR_Reports/controller.php
@@ -254,6 +254,7 @@ class AOR_ReportsController extends SugarController
         $stylesheet = file_get_contents(SugarThemeRegistry::current()->getCSSURL('style.css', false));
         ob_clean();
         try {
+            ob_start();
             $pdf = new mPDF('en', 'A4', '', 'DejaVuSansCondensed');
             $pdf->SetAutoFont();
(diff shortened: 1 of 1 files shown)


pgorod:
@Dillon-Brown according to [this](https://suitecrm.com/suitecrm/forum/suitecrm-7-0-discussion/25658-pdf-template-fails-to-load#85186), maybe something more is needed to fix it.

I suggest saving the `display_errors` value into a variable, and then turning it off during PDF generation, and then setting it back the way it was.

PR #7325, opened 2019-05-30
Feature: Ability to add custom code to bootup

grahambrown11:
## Description
The ability to add customisations in 2 places: in the beginning after the config is loaded and once the user is loaded.

## Motivation and Context
1. We need to execute some code after the config is loaded in order to get better reporting from our application performance monitoring system, so following the `custom/include` pattern: if
 `custom/include/entryPoint.php` exists it'll be `required` .
2. Implements a `after_load_user` hook so custom code can be executed on the loaded user before continuing with the application logic, this is different from the `after_login` as that only fires once when the user performs the login, where this will fire for every request when the user is retrieved from the session

## How To Test This
Both need to be tested manually as there are no unit tests available for the `entryPoint.php` and `SugarApplicationTest::testloadUser` is commented out due to `die` logic...
1. Add custom code in `custom/include/entryPoint.php` and make sure it gets executed.
2. Adding an application hook registration for `after_load_user` and ensuring the function is executed
```php
if (!isset($hook_array['after_load_user'])) {
    $hook_array['after_load_user'] = [];
}
$hook_array['after_load_user'][] = [
    1,
    'Check User',
    'custom/<HookClass>.php',
    '<HookClass>',
    '<checkUserFunction>'
];
```
I've manually tested these both with the file and hook present and not present.

## Types of changes

### Final checklist


samus-aran:
Hey @grahambrown11 Thanks for your contribution. We'll have a wee look :)

grahambrown11:
Any reason this has not been merged?
I understand if it's priorities, just want to make sure I've done nothing wrong...

samus-aran:
> Any reason this has not been merged?
> I understand if it's priorities, just want to make sure I've done nothing wrong...

From what I can see you haven't done anything wrong. Just circulating if there is any issues. :)

grahambrown11:
> > Any reason this has not been merged?
> > I understand if it's priorities, just want to make sure I've done nothing wrong...
> 
> From what I can see you haven't done anything wrong. Just circulating if there is any issues. :)

Thanks for the feedback

Dillon-Brown (dismissed):

samus-aran:
Hey @grahambrown11 
Sorry for the delayed response. The biggest issue we have is that this hook appears to be from Sugar propriety and we would rather not encourage that behaviour. We can see that based on the recent codebase that its a sound hook and does offer value to the users. So our suggestion is to change the hook's name itself to differentiate it from the current implementation that Sugar has.
We felt that after_user_load would be acceptable and also feel its more inline with the other hook's name structure (state_subject_action)

That be suitable for you and your implementation?

grahambrown11:
> Hey @grahambrown11
> Sorry for the delayed response. The biggest issue we have is that this hook appears to be from Sugar propriety and we would rather not encourage that behaviour. We can see that based on the recent codebase that its a sound hook and does offer value to the users. So our suggestion is to change the hook's name itself to differentiate it from the current implementation that Sugar has.
> We felt that after_user_load would be acceptable and also feel its more inline with the other hook's name structure (state_subject_action)
> 
> That be suitable for you and your implementation?

Sure @samus-aran, sorry for what seemed to be a copy of Sugar propriety function - it kinda was, I was searching the Google for how to implement and came across the Sugar documentation, and saw it was not implemented in Suite, so yes the name was copied but not the code as I have no access to that code. I've updated my branch, and it seems to have updated this PR - hope you find this is order...

Dillon-Brown (dismissed):

samus-aran:
Restarting Job :+1:

grahambrown11:
I don't understand why the build fails... I made this change so long ago and I believe it was approved just never merged...
Will this ever get merged or should I give up on this PR?

Issue #6993, opened 2019-02-25
Consider moving away from Travis CI


connorshea:
With the recent news of the [Travis acquisition](https://techcrunch.com/2019/01/23/idera-acquires-travis-ci/) last month followed by [the significant layoffs last week](https://www.reddit.com/r/devops/comments/at3oyq/it_looks_like_ibera_is_gutting_travis_ci_just_a/) I think it's worth discussing options for moving to a different CI provider if the need arises in the near future.

My two personal preferences would be either [GitLab CI](https://about.gitlab.com/solutions/github/) or [CircleCI](https://circleci.com/):

### GitLab CI

[GitLab CI](https://about.gitlab.com/solutions/github/): GitLab CI provides 50k build minutes/month (Note that the page says that the GitHub integration will be moved from the GitLab.com Free plan to the Silver (paid) plan, but my understanding is that open source projects have all Gold features available to them by default, we can get that clarified if we're concerned about it).

Job concurrency is dependent on the availability of the GitLab.com Shared Runners, which can vary somewhat ([there's a public Grafana instance with this kind of info available in realtime if that's interesting to anyone](https://dashboards.gitlab.com/d/000000159/ci?refresh=30m&orgId=1&from=now-3h&to=now)).

You can also use your own runners you host yourself, if that's of any interest.

- [GitLab CI Documentation](https://docs.gitlab.com/ee/ci/)
- [The `.gitlab-ci.yml` config file documentation](https://docs.gitlab.com/ee/ci/yaml/)

### CircleCI

[CircleCI](https://circleci.com/open-source/): CircleCI offers 4 Linux containers with no build limits for open source projects. I used CircleCI a few years ago – in 2015/16ish – and enjoyed it at the time, though I don't have as much recent experience with it.

- [CircleCI Documentation](https://circleci.com/docs/)

---------

Either way, we'd need to get SuiteCRM running in a Docker container to run tests on it (GitLab CI supports non-Docker environments but I'm not familiar with them and I'm not sure if you can do that kind of thing with the Shared Runners provided by GitLab.com).

Both GitLab CI and CircleCI allow you to provide your own containers for running your test suite on. GitLab has a Container Registry you can upload custom images to, I think CircleCI requires you to upload the container to Docker Hub. Both can use container images from Docker Hub.

[...]

PR #7328, opened 2019-05-31
Proof of Concept: GitLab CI

connorshea:
## Description
See #6993. This is a proof of concept for moving the project's CI infrastructure from Travis to GitLab. 

## Motivation and Context
Travis has been a pretty big pain for a number of reasons. Its environment is limited to what Travis itself provides you (you can run Docker containers but they're not really first-class citizens), everything else needs to be installed on every build. The job concurrency is pretty minimal, only 5 jobs can run at a time. The company has also been acquired and had significant layoffs in the past few months, and there's not really a very good guarantee that the company will still support the product properly in the future.

In #6993 I proposed moving to GitLab CI, and this PR is a result of that effort. I'm partially doing this because we want to run CI locally ourselves, and we're working on having GitLab CI set up on our own GitLab instance so this has to be done by us eventually anyway, if we want CI to happen :)

### Benefits of GitLab

- Docker containers are first-class citizens, meaning we can optimize the build time by creating Docker images that are pre-built with ChromeDriver, PHP, Composer, etc. installed already. We also get more control over the underlying environment the tests are run on, and can upgrade things as we like.
- GitLab Container Registry, so we can host the container images in the same place as the tests are run (I'm currently using my own container in the Container Registry that was built from the Dockerfile this branch).
- The concurrency limits are mostly dependent on the current load of GitLab.com's fleet of Runners, rather than a hard limit of 5 jobs at a time. This means we don't have to limit ourselves to as small a number of jobs. In the future, I'd potentially like to split our current jobs into smaller pieces (e.g. have "PHP 7.2 - acceptance tests", "PHP 7.2 - unit tests", "PHP 7.1 - acceptance tests", etc.) to make the test suite run much faster.
- We can host our own GitLab Runners if we want to have dedicated CI runners. This can be used to improve job concurrency and avoid potential bottlenecks if GitLab.com is under higher-than-normal load.
- GitLab CI has [artifacts](https://docs.gitlab.com/ce/user/project/pipelines/job_artifacts.html), which are a way to store things like log files, screenshots, etc. when a test is run. I'm already using them, and you can see an example of job artifacts [here](https://gitlab.com/connorshea/SuiteCRM/-/jobs/221756545/artifacts/browse). They're pretty useful for debugging failed tests. We already have these in Travis, but they're uploaded to a site called `https://www.simpleupload.co.uk/` and it's not really clear how to access them?
- [A ton of other useful features](https://docs.gitlab.com/ee/ci/yaml/README.html): Built-in code coverage tracking, jobs that are allowed to fail, scheduled jobs (e.g. run x job once a week, rather than on every push), separate stages, manually-triggerable jobs, caching between jobs, etc.

### Downsides

- GitLab CI integrates with GitHub and can be used to limit merging until CI passes, but by default it's not able to run tests on PRs from forks. This is discussed in #6993 and some bots have been made by other people that can be used to fix this problem. It's not ideal, but it should be a workable solution.
- Using GitLab Shared Runners means that, sometimes, there are backups in the CI job queue and we could end up waiting a few minutes to potentially hours (this is rare and usually means that GitLab CI is being abused by a malicious party, usually bitcoin miners 🙄). This shouldn't be a huge problem, but it is something to potentially keep in mind.
- Maintaining Docker containers ourselves can be a bit of a pain, but that hopefully won't be enough of a problem to matter too much.
- GitLab.com limits open source projects to 50,000 minutes of CI time per month. I'm not sure if we'd get anywhere near that number, and I assume if we hit it I can probably talk to some people I know at GitLab to get it raised, but it's worth noting.
- GitLab needs to mirror the repository from GitHub.com to pull the code into GitLab CI. In my experience, this happens within a few seconds so it's not a big issue. The only big problem occurs when the GitHub.com webhooks service is down. When that happens, the repo can still be told to manually sync by anyone that has permissions to do so, but pushing to the GitHub repository won't trigger a webhook, which means GitLab won't mirror the repository changes automatically and CI won't run. This isn't particularly common, so it's probably not a significant problem. I don't think Travis works without webhooks either(?).
- It can be a huge pain to get an environment set up locally that matches the one in GitLab CI, meaning it can be difficult to debug some CI issues. This can potentially be fixed by creating a docker-compose setup configuration that closely matches the way GitLab CI works. You can also run a GitLab Runner locally, though I admittedly haven't gone as far as doing that.
- No way to SSH into Runners to debug CI failures. (I'm not sure if this is possible if you host your own?).

## How To Test This

It's not really done yet, only some of the test suites are actually running, let alone passing. [You can see the job logs here](https://gitlab.com/connorshea/SuiteCRM/-/jobs).

### Current Progress

- MySQL and ChromeDriver are set up.
- 'Install' test suite runs and passes.
- PHPUnit test suite runs and almost passes, there are about a dozen failing tests, half of which are caused by ElasticSearch not being set up in the CI environment currently.
- Acceptance test suite runs but fails on one of the first tests, I haven't figured that issue out yet.
- API tests fail due to an error with the routes not working correctly(?)

### TODO

- [x] Set up ElasticSearch for use in CI.
- [ ] Get all the test suites passing.
  - [x] Install tests
  - [x] PHPUnit test suite
  - [ ] API tests
  - [ ] Acceptance tests
- [ ] Set up the test suite to run on all the different supported PHP versions.
- [ ] Figure out how to get the CI working for PRs from forks.
- [x] Clean up the mess that is this branch's commit history.
- [x] Change the target branch of this PR to develop (or hotfix?).
- [ ] Figure out what we want to do about CI on the hotfix-7.10.x branch. I assume 7.8.x will be EOL before this is done, so I won't bother with that.

### Setting up GitLab CI yourself

If you want to set up GitLab CI yourself on GitLab.com and try to help me (I'd appreciate it a lot!):

- Create an account on GitLab.com
- Fork SuiteCRM on GitHub.com and push a copy of my gitlab-ci branch to it.
- Go to https://gitlab.com/projects/new > CI/CD for external repo > connect your GitHub account and import the SuiteCRM fork you've made.
- The import should have set up the GitHub integration automatically. If not, see the ['GitHub project integration' docs](https://docs.gitlab.com/ee/user/project/integrations/github.html#github-project-integration-premium). **Make sure to have the integration set to 'Static static check names'** so you can require the check to pass for merging. (if the check name changes per-branch, as it does with the dynamic setting, it won't work be possible for GitHub to 'require' the check).

## Types of changes

### Final checklist


.devcontainer.json
```diff
@@ -0,0 +1,7 @@
+// See https://aka.ms/vscode-remote/devcontainer.json for format details.
+{
```
connorshea: This file is for use with the [VS Code Remote - Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) extension. I was using it to run some tests in the Docker container locally.

I'll probably delete it before this is merged.

connorshea:
Here's an example job of the current progress. All the dependencies install, the Install test suite passes, and the PHPUnit suite almost passes: https://gitlab.com/connorshea/SuiteCRM/-/jobs/221756545

@lazka could you take a look at that and tell me if you have any ideas on how to fix the state issues that are causing failures? (Sorry if I asked this already elsewhere, this whole thing has been super complicated so I've lost track of some things.)

connorshea:
With the `config_override` change that changes the elasticsearch host to `elasticsearch`, the PHPUnit test suite passes now \o/

connorshea:
I think this may be causing a significant burden on the CI Runnner, which is why tests are taking quite a while to run: https://github.com/salesagility/SuiteCRM/pull/7087

connorshea:
I'm down to a few remaining failures for the acceptance tests.

A few modules are failing to be created, which causes other cascading failures. Notably, `FileModuleCest`, `CompanyModuleCest`, and IssueModuleCest which all fail in these two job runs:
- https://gitlab.com/connorshea/SuiteCRM/-/jobs/223288568
- https://gitlab.com/connorshea/SuiteCRM/-/jobs/223296145

I suspect the problem is caused by a wait timeout that isn't long enough, but I don't know for sure.

connorshea:
Current status of API Tests:

- Latest job showing API test failures: https://gitlab.com/connorshea/SuiteCRM/-/jobs/228643851
- All the failures are caused by 404 errors.
- I'm unsure if there's a problem with the Docker container (missing an OpenSSL-related extension for PHP?), with the configuration of the CRM, with the Apache config, with the PHP configuration, or with something entirely different. Does anyone happen to know if the API fails on non-FastCGI servers? 

The specific error looks like this:

```
  I send post "http://localhost/...",{"username":"admin...}
  [Request] POST http://localhost/Api/access_token {"username":"admin","password":"admin1","grant_type":"password","scope":"","client_id":"API-4c59-f678-cecc-6594-5a8d9c704473","client_secret":"secret"}
  [Request Headers] []
  [Page] http://localhost/Api/access_token
  [Response] 404
  [Request Cookies] []
  [Response Headers] {"Date":["Mon, 10 Jun 2019 21:43:11 GMT"],"Server":["Apache/2.4.25 (Debian)"],"Content-Length":["289"],"Content-Type":["text/html; charset=iso-8859-1"]}
  [Response] <!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
  <html><head>
  <title>404 Not Found</title>
  </head><body>
  <h1>Not Found</h1>
  <p>The requested URL /Api/access_token was not found on this server.</p>
  <hr>
  <address>Apache/2.4.25 (Debian) Server at localhost Port 80</address>
  </body></html>
```

connorshea:
I cherry-picked #7173 to see if that'll fix the API issues.

lazka:
Here is my minial unittest-only gitlab config:

```yaml
image: debian:buster

variables:
  MYSQL_DATABASE: automated_tests
  MYSQL_USER: automated_tests
  MYSQL_PASSWORD: automated_tests
  MYSQL_ROOT_PASSWORD: automated_tests

stages:
  - test

test:
  stage: test
  services:
    - name: mysql:5.7
  script:
    - apt update
    - apt install -y composer php-curl php-xml php-ldap php-zip php-gd php-intl php-mysql
    - composer install --no-progress
    - cp tests/travis_config_si.php config_si.php
    - echo '$sugar_config_si["setup_db_host_name"] = "mysql";' >> config_si.php
    - php tests/testinstall.php
    - ./vendor/bin/robo test:unit
  artifacts:
    paths:
      - suitecrm.log
    when: always
```

Depends on #7773 and this patch to skip inboundemail:

```patch
diff --git a/tests/unit/phpunit/modules/InboundEmail/InboundEmailTest.php b/tests/unit/phpunit/modules/InboundEmail/InboundEmailTest.php
index e236830ce..ded64cffa 100644
--- a/tests/unit/phpunit/modules/InboundEmail/InboundEmailTest.php
+++ b/tests/unit/phpunit/modules/InboundEmail/InboundEmailTest.php
@@ -8,6 +8,11 @@ class InboundEmailTest extends SuiteCRM\StateCheckerPHPUnitTestCaseAbstract
 {
     public static $class_state = null;
 
+    protected function setUp()
+    {
+        $this->markTestSkipped('broken');
+    }
+
     public static function setUpBeforeClass()
     {
         // FIXME: The tests here depend on the email_cache table staying around and the right test order.
```

connorshea:
...huh. The unit tests seem to pass as they are now without skipped InboundEmail 🤔

connorshea:
Ok, so I've got the 1/3 ElasticSearch tests passing (the other 2 still fail 😞 ) and all of the tests/acceptance/Core/ tests passing (except maybe 1), and I haven't changed the modules tests at all yet. So that should be about 5 or 6 failures left, down from 24 earlier today.

connorshea:
Oh and PHPUnit passes. Sadly, it doesn't seem that the API suite is working yet :/

lazka:
I get lots of ` Hash doesn't match at key "database::aod_index".` with 7.10.21 now. Any idea what could be causing this?

Issue #7344, opened 2019-06-06
Automated Testing improvements


connorshea:
I didn't want to open a bunch of issues at once, so I figured I could just write out my suggestions as one larger issue.

I've been working on migrating the test suite to GitLab CI and I've noticed some problems while doing it. I've also had significant trouble setting up the test environment locally. I intend to fix some of these myself, but I wanted to discuss them to see if I'm missing anything important that these changes would break.

- [x] In `./vendor/bin/robo driver:run-chrome`, the ChromeDriver version that's installed should always be the latest version to make sure it remains compatible with newer versions of Chrome. (#7345)
- [x] Remove most `I->wait(x)` from the acceptance test suite. Use `waitForElementVisible` instead. This makes the test suite take longer than necessary. (#7442)
- [x] Remove all specified timeouts from `waitForElementVisible` and similar. If a test fails, this unnecessarily adds 30-120 seconds to the test run. In special cases we can add a longer timeout, but for the most part we should always use the default 10 seconds timeout. (#7414)
- [x] Remove all the extra test environments (Browser Stack, probably also Selenium?) (#7347)
- [x] Remove the functional test suite (it's unused) (#7346)
- [x] Remove the demo test suite (no one uses it, I think?) (#7352)
- [ ] Add support for a `config.test.php` file, which would be used to define configuration values for the test environment specifically (e.g. so you can run the tests on a separate database from the development site). Alternatively, maybe just `config_override.test.php`? (#7621)
- [ ] Make the `code:coverage` Robo command work outside of CI. (#7622)
- [x] Add a `config.codeception.yml.dist` file and gitignore the file `config.codeception.yml` in the `tests/_env` directory. These can be used to set up a test environment locally without needing to mess with any files that'd be committed to the repo. (#7433)
- [ ] Adopt `.env` files for use with testing, either [phpdotenv](https://github.com/vlucas/phpdotenv) or [Symfony Dotenv](https://github.com/symfony/dotenv). This is mostly useful for running multiple CRMs side-by-side on the same computer, and also because it makes it easier to figure out what the local environment variables are. (Issue: #7620)
- [x] Add Robo tasks for running the test suite instead of needing to use the full command paths. (#7374)
[...]

PR #7347, opened 2019-06-06
Remove support for automated testing with Browserstack.

connorshea:
As far as I can tell, BrowserStack is unused and adds unnecessary complexity to the codebase.

If people do use, I can close this PR. Otherwise I think it should be removed.

Part of #7344.


cameronblaikie:
Assessed :+1:

PR #7352, opened 2019-06-07
Remove demo test suite.

connorshea:
The demo test suite isn't used and can probably be deleted. The purpose - as far as I can tell - was to validate that your configuration worked, but since nobody tests that it works it could be broken for all we know. Its usefulness as a configuration validator seems pretty minimal to me. The docs also don't mention it for this purpose.

Part of #7344.

If this gets merged I'll open a PR in the docs to remove the demo test suite from the automated tests documentation.


connorshea:
Seems like flaky tests. Could someone restart them? :)

PR #7359, opened 2019-06-08
fixes issue #6343

bjoern86:
fixes issue #6343 - installer fails, if posix is not installed on linux systems

check if posix is installed before useing it. if getRunningUser() returns null log an error.

### DIFF
--- a/include/utils.php
+++ b/include/utils.php
@@ -448,15 +448,17 @@ function get_sugar_config_defaults()
 function getRunningUser()
 {
     // works on Windows and Linux, but might return null on systems that include exec in
-    // disabled_functions in php.ini (typical in shared hosting)
+    // disabled_functions in php.ini (typical in shared hosting) or if posix is not installed on Linux
     $runningUser = exec('whoami');
 
     if ($runningUser == null) {  // matches null, false and ""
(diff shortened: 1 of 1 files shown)


pgorod:
I just made a more complete fix in #7398

samus-aran:
> I just made a more complete fix in #7398

It would be much appreciated if you were able to push up to @bjoern86 rather than creating a separate PRs as it misses the chance to include bjoern's contribution via commit. 

This is something we definitely need to document as I feel that would discourage members to contributing if we inadvertently exclude their effort.

regards to the changes though, we have missed the Global Logging. https://github.com/salesagility/SuiteCRM/pull/7359/files#diff-b97be8484e5a5757aa6d4372838d140eR460 If I close it would you @bjoern86 be able to create a new PR for that?

pgorod:
The logging command was deliberately removed from getRunningUser to be placed in `cron.php` (which calls this function) so that all messages are coming from [the same place](https://github.com/pgorod/SuiteCRM/blob/posix_patch/cron.php#L60)

Thanks to @bjoern86 who is the real hero of this fix. I just put the pieces together, especially regarding the observations that gymad made, which were very pertinent.

PR #7382, opened 2019-06-12
WIP: Upgrade to ChromeDriver 75 in Travis

connorshea:
I want to see if this causes the same kinds of failures as I'm seeing with #7328, which would help me narrow down why these tests are failing.

The same tests fail in Travis as GitLab CI, but they fail less frequently. I'm not really sure why, or what causes the problem exactly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ env:
 before_install:
   # Install chrome driver#
   - whereis google-chrome-stable
-  - wget https://chromedriver.storage.googleapis.com/2.36/chromedriver_linux64.zip
+  - wget https://chromedriver.storage.googleapis.com/75.0.3770.90/chromedriver_linux64.zip
   - unzip -o chromedriver_linux64.zip
   - ./chromedriver --url-base=/wd/hub &
   # Lint PHP
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ env:
 before_install:
   # Install chrome driver#
   - whereis google-chrome-stable
-  - wget https://chromedriver.storage.googleapis.com/75.0.3770.90/chromedriver_linux64.zip
+  - wget https://chromedriver.storage.googleapis.com/77.0.3865.40/chromedriver_linux64.zip
   - unzip -o chromedriver_linux64.zip
   - ./chromedriver --url-base=/wd/hub &
   # Lint PHP
--- a/tests/_envs/travis-ci-hub.yml
+++ b/tests/_envs/travis-ci-hub.yml
@@ -10,8 +10,8 @@ modules:
       window_size: false
       clear_cookies: true
       capabilities:
-        chromeOptions:
-            args: ["--headless", "--disable-gpu", "window-size=1920x1080"]
+        "goog:chromeOptions":
+            args: ["--headless", "--no-sandbox", "window-size=1920x1080"]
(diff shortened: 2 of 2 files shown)


connorshea:
Okay, so a few things.

1) I don't understand how ChromeDriver 2.36 has worked with Chrome 74/75 for this long, it's only supposed to be compatible with Chrome 62-64. (they changed the versioning system a few releases ago so now ChromeDriver versions match the Chrome version it supports).
2) I don't really understand why, but ChromeDriver 75 fails on the install suite while ChromeDriver 74 works fine. This could be a bug in ChromeDriver 75, but I'm honestly not sure. I tried disabling the w3c compliance mode since that's the only major change in ChromeDriver 75 but that didn't seem to do anything (or my change didn't actually disable the mode properly).
3) Travis upgrades their version of Chrome automatically. This is a problem because once it gets a few versions behind, ChromeDriver will throw errors about how it's incompatible with the current version of Chrome. So we either need to determine the ChromeDriver version based on the Chrome version (this can't be done right now anyway because ChromeDriver 75 is broken(?)) or keep it updated manually.

Unfortunately, this doesn't really tell me anything about why the failures are so much common on GitLab CI. I suppose it could simply be caused by GitLab's machines having less RAM than Travis CI's? Or there's some other difference I'm not noticing.

connorshea:
ChromeDriver 73 is incompatible with Chrome 75, so it looks like they require that you stay within 1 version of Chrome.

connorshea:
Interestingly, with ChromeDriver 74 in 4ca1002 the PHP 5.6 and 7.1 builds both failed on `IssueModuleCest`. Which also frequently fails on GitLab CI. That's probably just a coincidence, though. None of the other tests failed for any of the other PHP versions.

connorshea:
Interestingly, the tests do seem to fail more often than before (but not quite 100% of the time) on ChromeDriver 74/Chrome 75: https://travis-ci.org/salesagility/SuiteCRM/builds/544888492

This suggests a few potential things:
- The releases of ChromeDriver since 2.36 either introduced inconsistent behaviors or uncovered bad code in our tests. I'm going to assume it's the latter. A lot of the tests already failed, just with less consistency before. The tests that fail on Travis seem to essentially match those that fail on GitLab, although it _seems_ like on GitLab CI they fail more frequently?

Solutions to these failures:
- Get to the bottom of the flaky behavior and fix it
- Disable the tests for now.

Personally, I'm leaning toward wanting to go with the latter option, primarily because disabling the tests is easier and I've already spent quite a while trying to determine the source of flakiness without much success. I'm not sure whether we can disable these individual tests or if the other tests in the same module depend on them.

Ultimately, in my opinion it's far more valuable to have a test suite you can actually trust than it is to have a test suite with slightly better coverage. Also, the flakiness is a huge pain in the butt for anyone opening PRs because they can't restart the failed jobs themselves. I'm trying to determine what tests should be commented out, but it looks like FileModuleCest (`testScenarioViewFileTestModule`), IssueModuleCest (`testScenarioViewRecordFromListView`), and AccountsCest (`testScenarioCreateAccountChild`) are the most-frequently flaky.

connorshea:
@Dillon-Brown I'm guessing dropping PHP 5.5 support isn't an option we can consider for 7.10?

connorshea:
Oh nevermind, I guess PHP 5.5 will continue to work? Weird.

PR #7389, opened 2019-06-13
Avoid caching incomplete beans in during SugarBean->fill_in_relationship_fields

PedroErnst:
Fix an elusive bug which sporadically causes incomplete beans to be cached by BeanFactory, resulting in variables not pulling through to email templates, pdf templates, and generally causing malfunctions.

## Description

While filling relationship fields during SugarBean->retrieve, a check is in place to avoid infinite recursion.
Only top level beans are fully retrieved, while the beans we retrieve to fill out the top level relationship fields don't have their own relationship fields filled out.

1. Load an Account bean
2. `Account->retrieve` calls `Account->fill_in_relationship_fields`
3. `Account->fill_in_relationship_fields` calls `Contact->retrieve` to set any relate fields there might be
4. `Contact->retrieve` calls `Contact->fill_in_relationship_fields`, but since it's nested the check kicks in and returns the function early without actually filling in any relationship fields the Contact might have
5. Since it used BeanFactory to retrieve the Contact bean, this incomplete bean is now cached
...
99. Somewhere else during the execution, a function calls BeanFactory->get and retrieves the incomplete Contact record. Bugs occur.

Solution is to unregister the incomplete beans during the call to `fill_in_relationship_fields` so that they'll only be cached when they've been properly retrieved.

## Motivation and Context

Incomplete beans can lead to bugs.

## How To Test This

It's a bit tricky to reproduce, this bug is really elusive
1. Have a module (Accounts for example) with a relate field to another module (Contacts for example)
2. Have an additional custom non-db relate fields that work off the related id, such as 'contact_birthdate' for example
3. Include the relate field in an email template
4. Make sure you trigger Account->retrieve first due to logic hooks, so that by the time the function that replaces variables in the EmailsController gets the incomplete cached version of the Contact
5. See in debugger the Contact record is missing all relate field values
6. See the email variables do not get replaced correctly

## Types of changes

### Final checklist

### DIFF
--- a/data/SugarBean.php
+++ b/data/SugarBean.php
@@ -4983,6 +4983,7 @@ class SugarBean
                                         $this->$name = $mod->name;
                                     }
                                 }
+                                BeanFactory::unregisterBean($related_module, $this->$id_name);
                             }
                         }
                     }
--- a/data/SugarBean.php
+++ b/data/SugarBean.php
@@ -4983,6 +4983,7 @@ class SugarBean
                                         $this->$name = $mod->name;
                                     }
                                 }
+                                // The related bean is incomplete due to $fill_in_rel_depth, we don't want to cache it
                                 BeanFactory::unregisterBean($related_module, $this->$id_name);
                             }
                         }


Mac-Rae:
Assessed :+1:

PR #7394, opened 2019-06-14
Fixed #7393 - Displaying dropdown db value instead of dropdown label in group header in Reports module

### LINKED ISSUE #7393: Displaying dropdown db value instead of dropdown label in group header in Reports module
Displaying dropdown db value instead of dropdown label in group header in Reports module

#### Expected Behavior
Group header should display dropdown label instead of value
drodpown_fix

#### Actual Behavior
Group header displaying dropdown value instead of label
dropdown_error

#### Steps to Reproduce
For eg
1. Create new dropdown field in lead module from studio
2. Created new report for lead module and group values based on newly created dropdown field
3. Run the report and check group header

I issue will persist in all module. I referred lead module for example

#### Your Environment
* SuiteCRM Version used: 7.10.x
* Environment name and version (e.g. MySQL, PHP 7): PHP5.6 and MYSQL 5.7
* Operating System and version (e.g Ubuntu 16.04):Ubuntu 16.04.3 LTS

604media:
Fixed #7393 - Displaying dropdown db value instead of dropdown label in group header in Reports module


Mac-Rae:
Assessed :+1:

PR #7398, opened 2019-06-14
Fix #6343 handling posix not allowed on some Linux systems

### LINKED ISSUE #6343: installer fails, if posix is not installed on linux systems

pgorod:
## Description
According to the previous discussion on #6344, I examined 8 places in the code and decided to update 5 of them, the others aren't necessary.

## Instances of `posix` calls, and of places calling `getRunninguser`

1. https://github.com/salesagility/SuiteCRM/blob/master/modules/UpgradeWizard/systemCheck.php

2. https://github.com/salesagility/SuiteCRM/blob/master/modules/UpgradeWizard/uw_ajax.php

3. https://github.com/salesagility/SuiteCRM/blob/master/modules/UpgradeWizard/uw_utils.php

These 3 instances are only using this to output a message to the user. I guarded the code for the case where `posix` functions are disabled, and made sure the output made sense.

4. https://github.com/salesagility/SuiteCRM/blob/master/cron.php
Added an if before the others, so that a proper message is logged.

5. https://github.com/salesagility/SuiteCRM/blob/master/install/performSetup.php
No change done here. The call won't have any effect. When cron.php checks, it will report with the new log message added in 4.

6. https://github.com/salesagility/SuiteCRM/blob/master/install/ready.php
No change. Will output a generic message: `sudo crontab -e -u  <web_server_user>`. This was already there before the `allowed_cron_users` mechanism was introduced.

7. https://github.com/salesagility/SuiteCRM/blob/master/modules/Schedulers/Scheduler.php
Same as above.

8. https://github.com/salesagility/SuiteCRM/blob/master/include/utils.php
Added a check for the existence of the `posix` functions.

## How To Test This
For each situation below, test **with** and **without** `posix` functions active in Linux. This can be done with `disable_functions` in php.ini, or with `--disable-posix` flag in command-line.

1. Installation

2. Post-install, visit screens
  - Admin / Schedulers
  - Admin / Upgrade Wizard, run permissions check

3. Execute cron.php and check logs

## Types of changes

### Final checklist

### DIFF
--- a/cron.php
+++ b/cron.php
@@ -55,8 +55,11 @@ if (substr($sapi_type, 0, 3) != 'cli') {
 if (!is_windows()) {
     require_once 'include/utils.php';
     $cronUser = getRunningUser();
-
-    if (array_key_exists('cron', $sugar_config) && array_key_exists('allowed_cron_users', $sugar_config['cron'])) {
+ 
+    if ($cronUser == '') {
+        $GLOBALS['log']->warning('cron.php: can\'t determine running user. No cron user checks will occur.');
+    } 
+    elseif (array_key_exists('cron', $sugar_config) && array_key_exists('allowed_cron_users', $sugar_config['cron'])) {
         if (!in_array($cronUser, $sugar_config['cron']['allowed_cron_users'])) {
             $GLOBALS['log']->fatal("cron.php: running as $cronUser is not allowed in allowed_cron_users ".
                                    "in config.php. Exiting.");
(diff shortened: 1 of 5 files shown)


pgorod:
@bjoern86 I'd appreciate it if you could help wit the testing of this PR. Thanks!

Mac-Rae:
Assessed :+1:

PR #7405, opened 2019-06-15
Securitysuite 3.1.16 updates

eggsurplus:
This brings SecuritySuite up to 3.1.16 on all bug fixes.

## Description

Includes fixes for the blank rows in the My Activity Stream, changing over to static functions as appropriate, and group selector on new record creation.

## Motivation and Context

Long overdue and also due to this issue: https://github.com/salesagility/SuiteCRM/issues/1491

## How To Test This

Do groups still work?
Is the blank row problem solved?

## Types of changes

- [X ] Bug fix (non-breaking change which fixes an issue)

### Final checklist

- [X] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [X] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

PR #7406, opened 2019-06-16
Info about making PR's in GitHub project README

pgorod:
This follows [a suggestion](https://github.com/salesagility/SuiteCRM/pull/7405#issuecomment-502365685) from a contributor who came here looking for this information and couldn't find it...

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,10 @@ This repository has been created to allow community members to collaborate and c
 
 ### Contributing to the project ###
 
+#### Making a Pull Request ####
+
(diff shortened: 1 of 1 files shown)


Dillon-Brown:
@pgorod Should probably go into hotfix-7.10.x so it ends up in all the hotfix branch README too.

pgorod:
Ok, I made a new one, it was easier than re-basing.

Please close this one if it's not needed.

But I notice that there are small differences between the READMEs in the different branches, in the initial parts: title and codecov links, etc. So maybe this file should be altered in each branch, instead of merged from one to the other?

Dillon-Brown:
@pgorod It'll appear as a conflict and get resolved in the merge PR.

PR #7417, opened 2019-06-17
Fix flaky tests

connorshea:
## Description

This PR has two major changes that are done to reduce test suite flakiness at the cost of speed:

- Remove login snapshots
- Clear cookies between tests

This adds somewhere between 1 and 3 minutes to the length of the acceptance suite, but it should significantly reduce the amount of times we need to re-run the tests in CI due to flaky test failures. I think that's an acceptable trade-off.

It removes a bunch of conditional logic that assumed logins could be skipped in certain situations (the cookie clearing causes these to no longer work). It also disables an AOW_Workflow test that was dependent on the state of a previous test, since I didn't want to deal with fixing that problem right now.

Note for reviewers: Disable showing whitespace changes in the diffs to make them a bit easier to follow.

## Motivation and Context

Logins were causing problems because they used codeception's session snapshots feature. Essentially, Codeception lets you save a browser session's state into a 'snapshot', which lets you reuse it to speed up the test suite. So if there was an existing session saved, the test would reuse that session to skip the login process. 

My best explanation as to why this was a problem is that the session token eventually expired. When the login snapshot was reloaded and the session happened to expire before loading the snapshot, the user would be logged out of the CRM and the test would fail.

There are a few ways to fix this:

- Hack around the expiration (e.g. by setting the token to never expire?)
- Remove the snapshots entirely and just login on every test. (This is the solution I'm currently using in this PR.)
- Add conditional logic to the login function that checks whether the user needs to login (e.g. because their session expired). I tried this, but I couldn't figure out a good way to do it with the tools provided by Codeception.
- Create a function to generate the session token on the backend. This would speed up the tests while also fixing the login problem, unfortunately I don't know how to do this so *shrug*.

I had to reset the cookies between every test because of a problem that showed up after I removed the snapshots. The tests started failing because each test would start from wherever the previous test left off. In that case, the test was in a situation where the user was already logged in.

My options here were to:

- Add a Logout function at the end of every test.
- Rewrite most of the tests to assume that a previous test had logged them in (bad for randomizing the test order in the future, would also suffer from problems with test expiration
- Clear the cookies between every test, which would force the user to be logged out.

So I chose to clear the cookies, since it's simple and makes the tests less dependent on each other's state.

## How To Test This
Make sure Travis passes.

## Types of changes

### Final checklist

### DIFF
--- a/tests/_envs/travis-ci-hub.yml
+++ b/tests/_envs/travis-ci-hub.yml
@@ -8,6 +8,7 @@ modules:
       port: 9515 # ChromeDriver port
       browser: chrome
       window_size: false
+      clear_cookies: true
       capabilities:
         chromeOptions:
             args: ["--headless", "--disable-gpu", "window-size=1920x1080"]
--- a/tests/_support/AcceptanceTester.php
+++ b/tests/_support/AcceptanceTester.php
@@ -39,16 +39,12 @@ class AcceptanceTester extends \Codeception\Actor
     public function login($username, $password)
     {
         $I = $this;
-        if ($I->loadSessionSnapshot('login')) {
-            return;
-        }
-        // Log In
-        $I->seeElement('#loginform');
+
+        $I->waitForElementVisible('#loginform');
         $I->fillField('#user_name', $username);
         $I->fillField('#username_password', $password);
         $I->click('Log In');
         $I->waitForElementNotVisible('#loginform', 120);
-        $I->saveSessionSnapshot('login');
     }
 
     public function loginAsAdmin()
--- a/tests/acceptance/Core/BasicModuleCest.php
+++ b/tests/acceptance/Core/BasicModuleCest.php
@@ -123,17 +123,14 @@ class BasicModuleCest
     ) {
         $I->wantTo('Create Basic Test Module Record');
 
-        if ($this->lastView !== 'ListView') {
-            $I->amOnUrl(
-                $webDriverHelper->getInstanceURL()
-            );
+        $I->amOnUrl(
+            $webDriverHelper->getInstanceURL()
+        );
+        $I->loginAsAdmin();
 
-            $I->loginAsAdmin();
-
-            // Go to Basic Test Module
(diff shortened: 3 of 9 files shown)


connorshea:
So it looks like there aren't actually any uses of `login()` other than for the admin user, so the snapshots would always be an admin anyway (this change is probably still worth having in case someone uses it as a non-admin user).

My current working theory is that the tests are failing because the snapshots get stale and the session token expires. (e.g. https://www.simpleupload.co.uk/uploads/OhTcnYFileModuleCest.testScenarioViewFileTestModule.fail.png).

It may be best to just remove the login snapshots entirely, though I don't know how much longer it'd cause tests to take. Or, we can potentially work around the problem by creating an escape hatch for the case when the login snapshot is loaded and the session token has expired.

connorshea:
Removing the snapshot loading shows that we don't properly log out between tests, which causes these tests to fail :/

connorshea:
Interesting, I just noticed that the inconsistent test cases all use the conditional logic stuff like `if ($this->lastView !== 'DetailView') {`. I wonder if that's relevant.

connorshea:
You can see the tests all passing here: https://travis-ci.org/salesagility/SuiteCRM/builds/547346533

connorshea:
I'll open a PR later to remove all the `lastView` stuff, since it's no longer useful.

I'm not sure if 7.11 has any significant amount of new tests that'll need to be updated for this change. It looks like there are only 4 or 5 new acceptance tests in `hotfix` vs. `hotfix-7.10.x`, so I'm not too worried.

connorshea:
@Dillon-Brown we'll probably need to figure out how we want to port this change forward to `hotfix`, but other than that I'd like to have this merged ASAP if possible. It'd make my life a _lot_ easier :)

Mac-Rae:
Assessed by Travis :+1:

Dillon-Brown:
@connorshea Will make my life much easier as well :D. Seriously nice work.

connorshea:
🎉

PR #7429, opened 2019-06-19
 Fixed #3157 - SuiteCRM fails to adequately housekeep session files potentially resulting in server failure

### LINKED ISSUE #3157: SuiteCRM fails to adequately housekeep session files potentially resulting in server failure
There may be implications with Custom Session Directory #877 as that exacerbates the problem, but system concerned is unusable and it is not possible to test this..

#### Issue

session files are not adequately managed. This can result in huge numbers of session files accumulating on the server e.g:

ls -al session_dir | wc

after 6 months use gave:

483419 4350764 41090483

which is over half a million session files. Attempting to remove them:

# rm sess_*
bash: /bin/rm: Argument list too long

as by this point the damage had been done as bash relies on temporary files in order to do this, and as no spare inodes were available, the removal process failed. You can do each one individually.....but you can't keep up with the creation rate!

#### Expected Behavior

SuiteCRM *AND the server it is on* should not be brought to a halt! SuiteCRM can actually disable everything on the server.

#### Actual Behavior

All of the inodes on the mounted drive are used up preventing the writing of any data to the drive, preventing any application that needs to write to that drive from functioning correctly. Signs of this happening are:

pje@archimedes:~$ df -h
Filesystem      Size  Used Avail Use% Mounted on
udev            4.0G  8.0K  4.0G   1% /dev
tmpfs           802M  244K  802M   1% /run
/dev/xvda2       99G   37G   57G  39% /
none            4.0K     0  4.0K   0% /sys/fs/cgroup
none            5.0M     0  5.0M   0% /run/lock
none            4.0G     0  4.0G   0% /run/shm
none            100M     0  100M   0% /run/user
/dev/xvda1      232M   82M  138M  38% /boot
pje@archimedes:~$ echo 1 > test.txt
-bash: test.txt: No space left on device

(i.e. you can't write to a drive that clearly has sufficient space on it)

In order to identify that you are affected by this issue, you can:
df -i
and you can find the offending directories with:
find / -xdev -size +100k -type d

#### Possible Fix

Add to the cron process that a command to housekeep the session folder on a regular basis. There are probably better ways within the framework to carry out housekeeping on the sessions directory. 

I have introduced a cron task to delete all of the session files at 1am as a temporary measure.

#### Steps to Reproduce

1.     Have a suiteCRM installation with a custom folder for session files.
2.     Wait 6 months for all of the session files to accumulate and use up all of the inodes on the storage device. Information above gives an understanding of the end result and points to the severity of the problem.

#### Context

Server failure

Probably a medium->high issue as it not only brings down the suiteCRM instance but also the rest of the server. 

There is also the failure that has occurred not being immediately obvious due to the logging needing to be improved on the operating system itself:
https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1665833

#### Your Environment

* SuiteCRM Version used:
Multiple versions of suiteCRM on the system, which started as 7.7.5, and have now all been updated to 7.7.8 and 7.7.9
mySQL/PHP - not important can't create temporary file: No space left on device when trying to diagnose anything!
Ubuntu 14.04

Dillon-Brown:
## Description

Implemented a new session handler to fix an issue with custom session paths resulting in huge numbers of session files.

## Motivation and Context

Issue reference: #3157 

## Types of changes

### Final checklist

### DIFF
--- a/include/MVC/SugarApplication.php
+++ b/include/MVC/SugarApplication.php
@@ -598,9 +598,14 @@ class SugarApplication
 
     public function startSession()
     {
+        if (file_exists('custom/include/SessionHandler.php')) {
+            require_once __DIR__ . '/../../custom/include/SessionHandler.php';
+        } else {
+            require_once __DIR__ . '/../../include/SessionHandler.php';
+        }
+
         $sessionIdCookie = isset($_COOKIE['PHPSESSID']) ? $_COOKIE['PHPSESSID'] : null;
         if (isset($_REQUEST['MSID'])) {
-            session_id($_REQUEST['MSID']);
             session_start();
             if (!isset($_SESSION['user_id'])) {
                 if (isset($_COOKIE['PHPSESSID'])) {
@@ -610,10 +615,8 @@ class SugarApplication
                 session_destroy();
                 exit('Not a valid entry method');
             }
-        } else {
-            if (can_start_session()) {
-                session_start();
-            }
+        } elseif (can_start_session()) {
+            session_start();
         }
 
         //set the default module to either Home or specified default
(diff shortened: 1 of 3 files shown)


pgorod:
Great work Dillon!

I noticed you don't have much exception handling around file operations. Maybe it isn't necessary - but still I would like to ask _how_ this breaks if the session dir is not writable, for example?

I understand that it's ok to just FATAL out when that dir is not writable, but since I've seen it happen (including in installations where it was working well, and then got misconfigured), I'd like to ask if you tested for that case and if you think it will provide adequate troubleshooting messages. Thanks.

tsummerer:
If you're going to be editing SugarApplication.php anyways, could you check for a custom SessionHandler? 
It'd be nice to be able to extend your SessionHandler for two reasons:

1. It might be interesting to swap for a Redis or other non-file based implementation
2. Some Ubunto stacks use a cron-based session garbage collection - this ensures that the User experience is not hampered by waiting around on GC. For these implementations, it's unnecessary to be editing the ini settings.

samus-aran:
Hey everyone. It would be great if users who had this issue were able to test it and confirm it works for them.
@tsummerer @personman @dschroth

tsmgeek:
Why not just drop in Zebra_Session and push towards mysql sessions?

pgorod:
I guess that wouldn't be compatible with the other databases, with MariaDB perhaps yes, but not with SQL Server...

tsmgeek:
@pgorod have a look at PR14 on that project, seems someone refactored it to allow for multiple backends, maybe fork that and extend it mysql/sqlserver/redis etc.

AlbertoSTIC:
Is it OK to add this PR in a SuiteCRM 7.11.15 installation?

pgorod:
You'd have to test it, but on a first look, I don't see any reason why it wouldn't work just fine.

PR #7433, opened 2019-06-19
Clean up codeception environments

connorshea:
## Description
This significantly simplifies the available Codeception environments by removing unused environments and creating a single customizable configuration to run the test suite locally. Part of #7344.

- Add a `custom.dist.yml` environment with a bunch of comments and a proper setup for running the install and acceptance suites locally. It's supposed to be copied to custom.yml so it can be customized to work for the user's setup (by default this config should pretty much work without any changes).
- Ignore `custom.yml` so it can't be committed to the repository.
- Remove `chrome-driver.yml` since it was effectively supposed to cover what `custom.dist.yml` is now for.
- Remove the various Selenium configurations since, based on the responses in #7344, no one really uses it.

## Motivation and Context

I wasn't able to run the test suite with a recent version of Chrome/ChromeDriver unless I messed with the `chromedriver.yml` a lot, and that caused Git to see changes since the file is tracked in the repo. And a lot of the defaults in `chromedriver.yml` didn't really make sense.

These changes serve a few purposes:
- Makes it obvious which environment should be used locally.
- Makes it possible to configure the test environment to work for your setup without needing to workaround problems with git.
- Simplifies the environments that we need to maintain going forward.

## How To Test This
- See if Travis passes
- Install ChromeDriver with the Robo task (and Chrome, assuming you don't have it already)
- Copy the `tests/_envs/config.dist.yml` file to `tests/_envs/config.yml` and customize it as-needed.
- Run ChromeDriver in another terminal so the test suite can use it.
- Try to run the acceptance suite locally (`./vendor/bin/codecept acceptance -d -vvv --env custom`), it still probably won't pass unfortunately but it's at least a step in the right direction.

For instructions on setting up the test suite locally, see https://github.com/salesagility/SuiteDocs/pull/290.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist
- [x] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [x] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

PR #7436, opened 2019-06-19
Simplify the acceptance and install suite configs.

connorshea:
These are unnecessary since they're handled by the travis config.

The only problem this would cause is if a custom env isn't used, but the acceptance and install suites using different settings by default when run locally is also bad. I'd say this is probably better than the current situation, and #7433 will help with this as well.


connorshea:
This is ready to review now.

cameronblaikie:
Assessed :+1:

PR #7437, opened 2019-06-19
Remove wait from Codeception Travis env

connorshea:
This doesn't speed up the test suite like I had hoped - or at least not by any noticeable amount - but it should make the tests more consistent across environments.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,6 @@ matrix:
       dist: xenial
       script: ./vendor/bin/codecept run install --env travis-ci-hub -f --ext DotReporter; echo "\$sugar_config['state_checker']['test_state_check_mode'] = 1;" >> config_override.php  ; ./vendor/bin/robo code:coverage; cat codeception.yml; bash <(curl -s https://codecov.io/bash) -f ./_output/coverage.xml
 
-
 services:
   - mysql
 addons:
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -5,30 +5,9 @@ sudo: required
 matrix:
   fast_finish: true
   include:
-    - name: "PHP 5.5 / MySQL 5.6"
-      php: "5.5"
-      dist: trusty
-    - name: "PHP 5.6 / MySQL 5.6"
-      php: "5.6"
-      dist: trusty
-    - name: "PHP 7.0 / MySQL 5.7"
```
Dillon-Brown: I wonder if we should actually just remove PHP 7.0. Is there really much benefit from testing 7.0 alongside 7.1?

I guess ideally I'd like to split things up so we maybe do a daily build that runs through absolutely everything while still having a fairly speedy process for PRs.

connorshea: I think Travis has some form of scheduled jobs that'd allow this? I certainly wouldn't be opposed.

Can we remove 5.5 while we're at it ;)

connorshea:
I think this can be merged now.

Mac-Rae:
Assessed by Travis :+1:

PR #7442, opened 2019-06-20
Replace most instances of $I->wait(n) with waitForX. 

connorshea:
## Description
Part of #7344.

## Motivation and Context
This makes the tests faster by removing a lot of usages of `$I->wait(x)` with `$I->waitForElementVisible(x)` or `$I->waitForText(x)`. Essentially, the problem with using `$I->wait` is that it waits for the full x number of seconds even if the page loads in the mean time.

```php
// Waits 10 seconds no matter what.
$I->wait(10);
$I->click('button');

// Will wait up to 10 seconds if necessary, otherwise it will skip to the click as soon as the element is visible. 
$I->waitForElementVisible('button');
$I->click('button');
```

Individually, these usually only add a few seconds each. Once you add two dozen or more, though, it can unnecessarily add over a minute to the test suite.

It also introduces potential flakiness if the tests take longer on another machine than on Travis, because a page load may take less than a second on Travis, but on a slower laptop it could take more than a second, and the `$I->wait(1)` wouldn't be sufficient in that case. This is also true of the `waitForElementVisible` functions, but to a lesser degree since they wait for up to 10 seconds if they can't find the element. If a lot of tests fail, this can add a lot more time to the test suite, but generally since the tests _usually_ pass, this shouldn't be a concern.

This PR doesn't remove all the instances of `$I->wait(x)`, I tried removing as many as I could, but for some of them I wasn't able to figure out how to get the tests to pass.

## How To Test This
See if the tests continue to pass on Travis.

## Types of changes

### Final checklist

### DIFF
--- a/tests/_support/Step/Acceptance/Calls.php
+++ b/tests/_support/Step/Acceptance/Calls.php
@@ -75,7 +75,6 @@ class Calls extends \AcceptanceTester
         $I->fillField('#parent_name', $module_name);
 
         $I->wait(2);
-
         $I->clickSaveButton();
         $DetailView->waitForDetailViewVisible();
     }
--- a/tests/_support/Step/Acceptance/Campaigns.php
+++ b/tests/_support/Step/Acceptance/Campaigns.php
@@ -57,7 +57,7 @@ class Campaigns extends \AcceptanceTester
         $I->click('#wiz_submit_finish_button');
 
         // Step 4
-        $I->wait(3);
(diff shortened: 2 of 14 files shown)


connorshea:
Based on some quick math, this PR removes up to 103 seconds worth of waiting (realistically it'll probably save closer to half of that, since the replacement waitForX functions will still involve _some_ waiting).

connorshea:
This is ready to go, and it looks like it does indeed save at least a minute of time in the acceptance tests :D

Mac-Rae:
Assessed by Travis :+1:

PR #7374, opened 2019-06-11
Robo test-running commands

connorshea:
## Description
Part of #7344. This adds Robo tasks for easily running the four different test suites in a consistent way. 

- `./vendor/bin/robo tests:install`
- `./vendor/bin/robo tests:acceptance`
- `./vendor/bin/robo tests:api`
- `./vendor/bin/robo tests:unit`

## Motivation and Context

The test commands are pretty long and annoying to type, and also hard to discover. This PR makes Robo commands that make running the tests simple and consistent.

I intend to update the automated testing documentation to incorporate these changes and simplify the general process of a developer running the test suite locally.

You'll need to set up the `tests/_envs/custom.yml` codeception config file (see https://docs.suitecrm.com/developer/automatedtesting/). You'll also need to either uncomment the part of the config with the background ChromeDriver process to have it run automatically with the tests, or otherwise run ChromeDriver in the background with `./vendor/bin/robo chromedriver:run` when running the install/acceptance tests.

## How To Test This
Pull down this branch and try running the various test suites:

- `./vendor/bin/robo tests:install`
- `./vendor/bin/robo tests:acceptance`
- `./vendor/bin/robo tests:api`
- `./vendor/bin/robo tests:unit`

Each of these have `--verbose` and `--fail-fast` flags that allow you to run the given test suite with extra debug output or with the command stopping as soon as an error or failure occurs.

They also all allow you to specify what tests you want to run - either by file or by directory - e.g. `./vendor/bin/robo tests:unit ./tests/unit/phpunit/modules/`.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist
- [x] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

PR #7452, opened 2019-06-21
Disable stopOnFailure and stopOnError in PHPUnit config.

connorshea:
## Description
We use these options explicitly in Travis, and they make running the tests without stopOnFailure/Error enabled unnecessarily difficult.

Also, note that this is essentially a backport from #7374. If that's merged before this, this PR won't really be necessary.

## Motivation and Context
I want to run the PHPUnit suite to completion and it's not possible to do so locally without changing this file, which is annoying to do whenever I switch branches. Plus, we explicitly use the flags in CI so there's not really a reason to set them in the PHPUnit config.

## How To Test This
Run PHPUnit locally without `--stop-on-error` or `--stop-on-failure` and see if it stops on either of those cases occurring.

## Types of changes
This is technically a breaking change because people could have relied on the old behavior in scripts they'd written?

### Final checklist

### DIFF
--- a/tests/phpunit.xml.dist
+++ b/tests/phpunit.xml.dist
@@ -8,8 +8,8 @@
     convertWarningsToExceptions     = "true"
     convertErrorsToExceptions       = "true"
     processIsolation                = "false"
-    stopOnFailure                   = "true"
-    stopOnError                     = "true"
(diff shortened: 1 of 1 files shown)


cameronblaikie:
Assessed :+1:

PR #7456, opened 2019-06-23
Fix #7455 Leads conversion - handle copying photo file, if present

### LINKED ISSUE #7455: Keep Lead photo when converting to Contact
#### Issue
When companies work with Leads including the photo field, this information is only half-converted to the Contact - the `photo` field which contains a file name will be converted, but the actual photo file won't be copied, so the Contact will have a broken `photo` field after Lead conversion.

#### Expected Behavior
It is debatable whether the photo should be copied or moved to the Contact. I opted for copying because it is less destructive, and it's easier to delete the Leads or the Lead's photos afterwards.

#### Possible Fix
I will provide a PR, I had this fixed in some code I made in 2016...

#### Steps to Reproduce
1. Enable photo field in Leads, and add it to the Layouts.
1. Enable photo field in Contacts, and add it to the Layouts.
1. Create a Lead and add a photo to it.
1. Convert the Lead to a contact.
1. Check the new Contact, photo is missing.

#### Context
When your company works with smaller groups of Leads, and they are not _that_ anonymous, and you use photos, the photos become an important part of your handling of customers. 
In my case this was a selection process that involves personal interviews. So people naturally want the converted Leads (admitted candidates) to keep their photos.

#### Your Environment
* SuiteCRM Version used: any

pgorod:
## Description
Please see Issue #7455

## How To Test This
See the steps to reproduce in the Issue.

## Types of changes

### Final checklist

### DIFF
--- a/modules/Leads/views/view.convertlead.php
+++ b/modules/Leads/views/view.convertlead.php
@@ -515,7 +515,7 @@ class ViewConvertLead extends SugarView
             if (!empty($lead->id) && !empty($lead->photo) &&
                 !empty($beans['Contacts']->id) && !empty($beans['Contacts']->photo)) {
                 $bCopied = false;
-                if (($lead->photo == $beans['Contacts']->photo) && is_readable('upload/' . $lead->id . '_photo')) {
+                if (($lead->photo === $beans['Contacts']->photo) && is_readable('upload/' . $lead->id . '_photo')) {
                     $bCopied = copy('upload/' . $lead->id . '_photo',
                                    'upload/' . $beans['Contacts']->id . '_photo');
                 }
@@ -523,7 +523,7 @@ class ViewConvertLead extends SugarView
                     $beans['Contacts']->photo = $lead->photo;
                     $beans['Contacts']->save();
                 } else {
-                    $GLOBALS['log']->warning('Lead conversion: failed copying upload/' .
+                    LoggerManager::getLogger()->warning('Lead conversion: failed copying upload/' .
                         $lead->id . '_photo to upload/' . $beans['Contacts']->id . '_photo');
                 }
             }
--- a/modules/Leads/views/view.convertlead.php
+++ b/modules/Leads/views/view.convertlead.php
@@ -499,6 +499,7 @@ class ViewConvertLead extends SugarView
             $this->copyAddressFields($bean, $beans['Contacts']);
 
             $bean->save();
+
             //if campaign id exists then there should be an entry in campaign_log table for the newly created contact: bug 44522
             if (isset($lead->campaign_id) && $lead->campaign_id != null && $bean->object_name == "Contact") {
                 campaign_log_lead_or_contact_entry($lead->campaign_id, $lead, $beans['Contacts'], 'contact');
@@ -509,6 +510,23 @@ class ViewConvertLead extends SugarView
             $lead->converted = '1';
             $lead->in_workflow = true;
             $lead->save();
+
+            // handle copying photo file, if present:
+            if (!empty($lead->id) && !empty($lead->photo) &&
+                !empty($beans['Contacts']->id) && !empty($beans['Contacts']->photo)) {
+                $bCopied = false;
+                if (($lead->photo == $beans['Contacts']->photo) && is_readable('upload/' . $lead->id . '_photo')) {
+                    $bCopied = copy('upload/' . $lead->id . '_photo',
+                                   'upload/' . $beans['Contacts']->id . '_photo');
+                }
+                if ($bCopied) {
+                    $beans['Contacts']->photo = $lead->photo;
+                    $beans['Contacts']->save();
+                } else {
+                    $GLOBALS['log']->warning('Lead conversion: failed copying upload/' .
+                        $lead->id . '_photo to upload/' . $beans['Contacts']->id . '_photo');
+                }
+            }
         }
 
         $this->displaySaveResults($beans);


Mac-Rae:
Assessed :+1:

modules/Leads/views/view.convertlead.php
```diff
@@ -509,6 +510,23 @@ protected function handleSave()
...
+                $bCopied = false;
+                if (($lead->photo == $beans['Contacts']->photo) && is_readable('upload/' . $lead->id . '_photo')) {
+                    $bCopied = copy('upload/' . $lead->id . '_photo',
+                                   'upload/' . $beans['Contacts']->id . '_photo');
+                }
+                if ($bCopied) {
+                    $beans['Contacts']->photo = $lead->photo;
+                    $beans['Contacts']->save();
+                } else {
+                    $GLOBALS['log']->warning('Lead conversion: failed copying upload/' .
```
Dillon-Brown:
```suggestion
                    LoggerManager::getLogger()->warning('Lead conversion: failed copying upload/');
```

modules/Leads/views/view.convertlead.php
```diff
@@ -509,6 +510,23 @@ protected function handleSave()
             $lead->converted = '1';
             $lead->in_workflow = true;
             $lead->save();
+
+            // handle copying photo file, if present:
+            if (!empty($lead->id) && !empty($lead->photo) &&
+                !empty($beans['Contacts']->id) && !empty($beans['Contacts']->photo)) {
+                $bCopied = false;
+                if (($lead->photo == $beans['Contacts']->photo) && is_readable('upload/' . $lead->id . '_photo')) {
```
Dillon-Brown: Can ```===``` be used here?

modules/Leads/views/view.convertlead.php
```diff
@@ -509,6 +510,23 @@ protected function handleSave()
...
+                    $bCopied = copy('upload/' . $lead->id . '_photo',
+                                   'upload/' . $beans['Contacts']->id . '_photo');
```
Dillon-Brown: I think it might be worth validating the ID here.

modules/Leads/views/view.convertlead.php
pgorod: There isn't much question about the data types of these Bean fields, but ok, I'll change it to `===`

modules/Leads/views/view.convertlead.php
pgorod: Ok. Let me just ask you a few question for my education on this...

1. Is this the general rule now - always use the new syntax?
2. And whenever I touch a file that has the old syntax, change all occurrences to the new?
3. I assume we don't just find-and-replace this all over the code, because in some places it might break?

modules/Leads/views/view.convertlead.php
pgorod: Sorry, I don't understand this one. We're in the middle of the conversion process, there is a lot of code managing these `ids` and I don't think we can reach this point without valid `ids`, can we?

Thanks Dillon for the attentive review, I appreciate it!

modules/Leads/views/view.convertlead.php
Dillon-Brown: 1. I don't think it's actually in the style guide but I would say yes. We should keep it consistent.
2. You wouldn't necessarily need to change everywhere in the file unless you really wanted to. In general, I would say you only really need to do so for the functions you touch (Same for PSR-2, etc).
3. I highly doubt find and replace would cause any issue, these are functionally identical. It's simply that nobody has got around to doing it yet afaik.

modules/Leads/views/view.convertlead.php
Dillon-Brown: Yeah it doesn't really matter here, it's simply to resolve the inspection from appearing.

modules/Leads/views/view.convertlead.php
Dillon-Brown: There is indeed a lot of places managing these ```ids```. I haven't looked through to see if there is anywhere that someone could potentially exploit this but my thinking is that whenever we are doing file operations like this we should try and do some sort of validation to prevent someone from modifying the paths to do something they shouldn't.

There's a built-in function for this in ```/lib/Utility/SuiteValidator.php``` called ```isValidId```.

pgorod: This is safe. We are only duplicating a file that already existed in our system (it's not being uploaded at this time), so all the necessary checks for photo files have already been made.
And the convention is that the file is named as `<beanId>_photo` so we're using the valid Lead id and Contact id that are being handled for the previous 200 lines in this `handleSave` function. If the `ids` weren't valid lots of things would have broken long before this point.

About that `isValidId` function - I have lobbied for its removal, or at least make it clear that it's just a convenience `isValidUUID` function, but that SuiteCRM can use any string format for an `id`. I think checking with that function breaks existing installations and is an unjustified violation of a basic convention since SugarCRM days... 🤷‍♂

Dillon-Brown: In that case, can we at least use ```is_uploaded_file()```? ```copy()``` just has so many security implications imo....

pgorod: I believe it's already done, for that file, when it was uploaded: https://github.com/salesagility/SuiteCRM/blob/master/include/UploadFile.php#L285, assuming the Upload photo function passes there.

I really don't see the case for doing it here. If the file needs to be checked (and it does) then that definitely has to happen when uploading the file from the user, that's when the security boundary is crossed. Once it is in the `upload` directory, with the name and the permissions we assigned to it, then duplicating it cannot possibly create a new security risk, as far as I can see... and the `copy` command here is very focused, a specific dir, and a specific file name structure...

As an exercise, we can try thinking about this in paranoid terms: suppose an attacker has arranged to place a malicious file as a Lead photo in `upload` dir. I don't know how he could explore that, but even so I would consider that as a potentially serious breach. But what would be the advantage of turning it into a Contact photo instead? And if we had that vulnerability, we would definitely need to fix it before this point.

But if you are still unconvinced by my ramblings, I don't mind adding another validation call (assuming it returns `true` in normal cases).

PR #7477, opened 2019-06-26
Remove unused webDriverHelper variables

connorshea:
## Description
Throughout the acceptance test suite the webDriverHelper was unused. After #7418 removed repetitive URL visits throughout the test suite, it was no longer necessary in most cases.

This cleans all of those up.

## Motivation and Context
Getting rid of dead code.

## How To Test This
Make sure Travis CI passes.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist
- [x] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

PR #7479, opened 2019-06-26
Add a new visitPage helper for tests

connorshea:
## Description
Part of #7344. 

Add a new `visitPage` helper for tests.

```php
# Example of usage
$I->visitPage('Campaigns', 'index');
```

```php
/**
 * Helper for navigating to a page.
 * 
 * @param string $module SuiteCRM module name
 * @param string $action View action name, e.g. index, EditView, DetailView.
 * @param string|null $record The id of a record, used for EditView and DetailView routes.
 */
public function visitPage($module, $action, $record = null)
{
    $I = $this;
    $url = $I->getInstanceURL();
    if ($record !== null) {
        $url .= "/index.php?module={$module}&action={$action}&record={$record}";
    } else {
        $url .= "/index.php?module={$module}&action={$action}";
    }
    $I->amOnUrl($url);
}
```

The goal of this helper is to make it easier to navigate through the site without needing to interact with as much of the site. Instead of needing to click through the navigation bar to reach a page in a test, you can do it by going to the page directly.

I changed a bunch of tests already to replace the `gotoModule()` methods with the `visitPage()` helper, including Accounts, Documents, Contacts, Contracts, Products, Product Categories, Invoices, etc.

## Motivation and Context

This saves a decent amount of time (20-30 seconds total, as far as I can tell) since navigating to a URL directly is inherently faster than interacting with the page. It also gets rid of the repetitive goto helpers for various modules.

The reason to do this instead of going through the interface is because the interface elements (e.g. the navigation bar) should be tested separately from other components, we don't need to make sure the navigation bar links are working in dozens of separate tests, we only need to do it once.

## How To Test This
Make sure Travis continues to pass and the code makes sense.

## Types of changes

### Final checklist

### DIFF
--- a/tests/_envs/chrome-driver.yml
+++ b/tests/_envs/chrome-driver.yml
@@ -12,5 +12,3 @@ modules:
         chromeOptions:
             args: ["window-size=1920x1080"]
             binary: "/usr/bin/google-chrome"
-
-
--- a/tests/_support/AcceptanceTester.php
+++ b/tests/_support/AcceptanceTester.php
@@ -82,4 +82,23 @@ class AcceptanceTester extends \Codeception\Actor
         $I->dontSee('error');
         $I->dontSee('PHP');
     }
+
+    /**
+     * Helper for navigating to a page.
+     * 
+     * @param string $module SuiteCRM module name
+     * @param string $action View action name, e.g. index, EditView, DetailView.
+     * @param string|null $record The id of a record, used for EditView and DetailView routes.
+     */
+    public function visitPage($module, $action, $record = null)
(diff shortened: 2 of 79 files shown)


connorshea:
Maybe I should remove all the empty files in `tests/_support/Step/` now 🤔

connorshea:
I'll wait for #7477 to be merged before removing those empty classes, since it'll require removing those classes from the tests.

connorshea:
Failure is a flaky test, looks like.

Mac-Rae:
Assessed by Travis :+1:

Issue #7445, opened 2019-06-20
Run PHPUnit with StateChecker enabled in CI


connorshea:
#### Issue
The StateChecker should be run as a job in CI to prevent state-poisoning tests from making it into the test suite.

Part of #7344, probably.

#### Expected Behavior
I should be able to run the PHPUnit test suite locally with the StateChecker enabled, and it should pass.

#### Actual Behavior
It fails instead, because some tests don't properly clean up after themselves.

#### Possible Fix
The specific tests that modify state and cause errors should be fixed, but more importantly there should be a CI job that checks to make sure the state checker is passing, to make sure it stays that way. We only need to run it once, on one PHP version.

#### Context
It makes it harder to run the PHPUnit test suite locally.

#### Your Environment
* SuiteCRM Version used: 7.10.x
* Environment name and version (e.g. MySQL, PHP 7): MySQL 5.7, PHP 7.2
* Operating System and version (e.g Ubuntu 16.04): macOS Mojave (10.14.5)

connorshea:
No more state checker, so I'm closing this :)

PR #7480, opened 2019-06-26
WIP: Try running the StateChecker in CI

connorshea:
## Description
This is an attempt to run PHPUnit with the StateChecker enabled in CI, to make sure that the PHPUnit suite doesn't regress with state problems.

See #7445.

## Motivation and Context
Running the unit test suite with state checks enabled causes over 100 errors right now, many of them statechecker-related. This makes running the suite locally quite difficult.

I don't want to enable the StateChecker for every CI job, but having it enabled for just one should be sufficient to ensure we don't have state issues.

Right now, this PR removes most of the PHP versions we're testing to improve the speed of the feedback loop.

## How To Test This
See if Travis passes.

## Types of changes

### Final checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,21 +5,18 @@ sudo: required
 matrix:
   fast_finish: true
   include:
-    - name: "PHP 5.5 / MySQL 5.6"
-      php: "5.5"
-      dist: trusty
-    - name: "PHP 5.6 / MySQL 5.6"
-      php: "5.6"
-      dist: trusty
-    - name: "PHP 7.0 / MySQL 5.7"
-      php: "7.0"
-      dist: xenial
-    - name: "PHP 7.1 / MySQL 5.7"
-      php: "7.1"
-      dist: xenial
-    - name: "PHP 7.2 / MySQL 5.7"
+    - name: "StateChecker PHPUnit"
       php: "7.2"
       dist: xenial
+      script:
+        - echo "using install wizard"
+        - ./vendor/bin/codecept run install --env travis-ci-hub -f --ext DotReporter
+        - ./build/push_output.sh
+        # Run the unit tests
+        - sudo chmod -R 777 .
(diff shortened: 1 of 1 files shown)


connorshea:
Unsurprisingly, it fails when the state checker is enabled. There are only 17 errors though, most of them seem related to the InboundEmails tests.

@lazka any ideas on how to fix these?

lazka:
> @lazka any ideas on how to fix these?

Yes, I added a comment there when I tried to fix RUN_PER_TESTS  https://github.com/salesagility/SuiteCRM/blob/83ba237a4eb48255b1e9e6eb62e37c2a046a4406/tests/unit/phpunit/modules/InboundEmail/InboundEmailTest.php#L13

Note that you are setting test_state_check_mode to 2 (PER_CLASSES) and then set the developer mode which overrides it to PER_TESTS. Which one do you want ? :)

connorshea:
Ah, probably per-test :D

The documentation should really associate the numbers with the state they represent, also maybe have the StateChecker print out a message about what mode it's running in when running the unit tests?

I guess the InboundEmail tests need to be at least partially rewritten to be independent :/

connorshea:
@lazka any chance you could fix these in the near future? I think this'd be important for making sure the tests can be run locally, and you're definitely more experienced than me when it comes to fixing these state checker issues. 🙇

lazka:
The main problem in this case is understanding the code and to know what it is trying to test in the first place, and I gave up there last time. But I'll have another look.

lazka:
> The main problem in this case is understanding the code and to know what it is trying to test in the first place, and I gave up there last time. But I'll have another look.

I've fixed all the InboundEmail tests and will open a PR tomorrow.

lazka:
See #7818

connorshea:
I'm gonna close this in favor of #7807.

PR #7525, opened 2019-07-05
Fix - API Create Relationship via Link

Jason-Dang:
## Description

- Add link option to create relationship params and route.
- Use given link instead of guessing from beans.
- Update Postman collection with new path part to specify link name.

## Motivation and Context

Previously unable to specify link and logic just grabbed the last link it found between two modules.

## How To Test This

Use the updated Create Relationship call in the Postman collection.
Example: For `Accounts` use the `contacts` link.

## Types of changes

### Final checklist


Api/V8/Config/routes.php
```diff
@@ -98,7 +98,7 @@
          */
         $app
             ->post(
-                '/module/{moduleName}/{id}/relationships',
+                '/module/{moduleName}/{id}/relationships/{linkFieldName}',
```
PedroErnst: Would it maybe make sense to leave the old route active as well?  
I think by changing the route it becomes a breaking change that's not backward compatible. Existing applications that use this API might stop working.  
It might be good to have both routes.

Jason-Dang: That’s a good point, although personally I would argue it not to be a breaking change as the current route never worked as it should have. I don’t know if anyone would want an endpoint that creates a random link between two modules if they have more than one relationship. Well I believe it just uses the first/last link it finds. 

I do see the sense in maintaining backward compatibility and will be happy with whatever decision is made.

Dillon-Brown:
@Jason-Dang Failing tests/api/V8/CreateRelationShipCest.

Jason-Dang:
@Dillon-Brown I forgot about the test, I believe it should be updated to use an out of the box link name such as `contacts` for the relationship between Accounts and Contacts.

I'll push up a commit with this update.

Yeah that is strange as IMO it's how it should be.

Jason-Dang:
@Dillon-Brown Finally made the time to fix this properly. 
I fixed the test and it seemed to pass travis so I've now added some extra meta info and changed it to use module labels. 
Currently with travis now so hopefully all good and this can be merged.

pgorod:
@Jason-Dang or @PedroErnst, if any of you guys can please have a quick look at this Documentation PR, it's also about using relationships via the API so it's similar to what you're doing here and you should be able to review that in just a minute.

https://github.com/salesagility/SuiteDocs/pull/318/files

Please just leave a positive review there if it looks ok. Thanks!

Jason-Dang:
@Dillon-Brown I've reverted the old endpoint and added a new one with the changes. It's going through travis now. I'll leave the deprecation of the old one up to your team. I've also updated the Postman collection to remove all the dummy access token values and replaced ID values with the placeholder `{ID}`.

Jason-Dang:
@pgorod I'll take a look soon as I get some free time, would you mind nudging me with a reminder in case I forget down the line please.

Dillon-Brown:
Assessed

PR #7536, opened 2019-07-09
Cleanup files created by acceptance tests between test runs

### LINKED ISSUE #7419: Acceptance tests need to clean up after themselves
#### Issue
This is part of #7344, but it's a big enough problem that I figured opening a separate issue was warranted.

Essentially, the acceptance test suite leaves behind a bunch of files locally that cause the test behavior to differ if you do multiple runs on the same CRM.

#### Expected Behavior
The acceptance tests wouldn't pollute the state of my CRM and I would be able to re-run the test suite without any differences in the test results.

#### Actual Behavior
The acceptance tests create modules and files locally that aren't cleaned up. Then, when we re-run them they exhibit different behavior. 

#### Possible Fix
Does Codeception have a before/after hook for tests? We could save the current file list in `before` and delete any files created since then in the `after` hook. Unfortunately I assume there are tests that implicitly rely on each other so this may not be a simple fix :/

#### Steps to Reproduce
1. Clone the repository
2. Set the test suite up.
3. Run the acceptance suite locally.
4. Run it again, note that the results differ (assuming the first run actually got far enough to create modules, etc.)

#### Context
I need to be able to run the test suite locally because debugging issues with the test failures in CI is miserable. I have no choice but to wait 20-30 minutes between every change to see if it was effective in fixing the issue.

#### Your Environment
* SuiteCRM Version used: 7.10.16
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Firefox 67
* Operating System and version (e.g Ubuntu 16.04): macOS Mojave (10.14.5)

connorshea:
## Description
Fixes #7419! Now the acceptance tests can be run as many times as you want, without needing to worry about the files that are left behind by previous test runs.

This doesn't take very long, so it shouldn't add any noticeable runtime to the acceptance suite.

Unfortunately, there are still two files modified by the acceptance tests, which I didn't fix: `Api/Core/Config/ApiConfig.php` and `custom/modules/unified_search_modules_display.php`. The former is modified to include an `OAUTH2_ENCRYPTION_KEY` (can we make that an environment variable or something?) and the latter is just some spacing changes (should I just commit the changed version of this file so it doesn't get changed every time we run the tests?).

## Motivation and Context

This allows me to run the acceptance tests locally without worrying about the modules being left behind by a previous test run. We could also ignore the relevant directories if we want to, to prevent `git status` from returning ~130 files while the test suite is still running.

## How To Test This
Clone the repository locally, follow the automated testing instructions (latest docs are [here](https://github.com/salesagility/SuiteDocs/pull/300)), and run the acceptance tests locally to verify that the files are deleted before and after the tests are run. Make sure you can re-run the acceptance tests.

You might want to use something like `./vendor/bin/codecept run acceptance tests/acceptance/Core/CompanyModuleCest.php -vvv -d --env custom` (assuming you set up the custom.yml env) to run just some of the acceptance tests, instead of all of them.

## Types of changes

### Final checklist

### DIFF
--- a/tests/_support/Helper/Acceptance.php
+++ b/tests/_support/Helper/Acceptance.php
@@ -18,4 +18,56 @@ class Acceptance extends \Codeception\Module
         }
         return true;
     }
+
+    public function _beforeSuite()
+    {
+        $directories = array(
+            "custom/modulebuilder/builds/CompanyTestModule",
+            "custom/modulebuilder/packages/CompanyTestModule",
+            "modules/Test_CompanyTestModule"
+        );
+        
+        foreach ($directories as $_index => $directory) {
+            if (is_dir($directory)) {
+                $this->getModule('Filesystem')->deleteDir($directory);
+            }
+        }
+
+        $files = array(
+            "custom/application/Ext/Include/modules.ext.php",
+            "custom/Extension/application/Ext/Include/CompanyTestModule.php"
+        );
+
+        foreach ($files as $_index => $file) {
+            if (file_exists($file)) {
+                unlink($file);
+            }
+        }
+    }
+
+    public function _afterSuite()
+    {
+        $directories = array(
+            "custom/modulebuilder/builds/CompanyTestModule/",
+            "custom/modulebuilder/packages/CompanyTestModule/",
+            "modules/Test_CompanyTestModule/"
+        );
+
+        foreach ($directories as $_index => $directory) {
+            if (is_dir($directory)) {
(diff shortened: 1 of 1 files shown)


connorshea:
Flaky HomeCest 🙄

connorshea:
@Dillon-Brown ready for review/merge, this should make it possible to run the acceptance tests repeatedly locally without any trouble :)

PR #7538, opened 2019-07-10
Fixed #7304 - ListView: Fix selection count for the "Select All" case

### LINKED ISSUE #4297: Error in Listview with previously selected rows and changing to next page (pagination) - Bulk Action disabled and wrong "Selected" count
#### Issue

In Accounts-Listview i select e.a. the first to rows (for using with Bulk-Action).
Then change to page number 2 using paginagion button (Next).
On page 2 select some more Rows.
Change back to page number 1 using pagination button (prev).

the 2 previously selected rows are still selected (correct), but its not possible to trtigger any "Bulk Action" because the button is disabled.

if i deselect one of the selected rows it shows "Selected: -1".

#### Expected Behavior

the number of selected items/rows should be correct even when changing to next page.
after changing to nexxt page (using pagination) the "Bulk ActioN" should not be disabled if previously selected any item/rot.

#### Actual Behavior

see description on top. "Bulk Action" is disabled on second page and "Selected" shows wrong counter.

#### Steps to Reproduce

1. select rows in accounts list
2.go to page number 2
3.select some more items and go back using paginiation

#### Your Environment

* SuiteCRM Version used:
7.9.4
firefox/chrome
php 5.6
debian 7

### LINKED ISSUE #7304: Select All in 7.10.16
#### Issue
You can only select one page of records even if you choose "select all" records.
It doesn't matter if you apply filters or not.
It happens on all modules.

#### Steps to Reproduce
1. In any module's list view ie Contacts

2. try to select all records
Screen Shot 05-21-19 at 04 25 PM
3. You only get one page of records selected
Screen Shot 05-21-19 at 04 25 PM 001
4. If you try to delete those records you'll check that you selected just one page of them

#### Your Environment

* SuiteCRM Version used: 7.10.16
* Any browser
* PHP: 7.0.33 and 7.2.14
* MySQL 5.7.24

lazka:
## Description

The check count isn't valid in the "Select All" case and should be ignored.
On the other hand selectCount is only valid in the "Select All" case and
only exact in case the count query is not disabled in the config.

Fix get_num_selected() to work in all cases, add get_num_selected_string()
which appends a "+" in case count queries are disabled which is used in the
UI to signal the user that there could be more entries selected.

To avoid the UI getting out of sync update all selection count related controls in the
update loop and not just the count label (see https://github.com/salesagility/SuiteCRM/issues/4297)

Fixes #7304

(Also fixes #4297 which I can still reproduce)

## Motivation and Context

The UI should reflect the actual number of selected records.

## How To Test This

* Go to the leads module
* Select all records
* The "Selected: X" label should be correct
* Bulk Action > Delete -> The message should show the correct count

## Types of changes

### Final checklist


pgorod:
@lazka I causally came across this PR and remembered an old unsolved issue I had because I never fully understood how these selections are handled. I was wondering if you could solve it, with what you learned from this PR...

Basically, these two gists are a way to add a custom action to the list views:

https://gist.github.com/pgorod/5de7c6f8d37413654b16e06668d7e1b1
https://gist.github.com/pgorod/c838f258fadf9528b61183e763e978a8

But it's a known issue, that you can't get all the selected rows inside the custom action, you only get the current screen. So if the List view contains 100 rows but is currently showing 20 records (first page), and you hit `Select all`, you only get 20 passed `ids`, when you should get all 100.

Can you understand that and find a solution? Feel free to ignore this request if it is not related to what you did here, or if it doesn't interest you. I just thought I would ask... thanks.

lazka:
> Can you understand that and find a solution? Feel free to ignore this request if it is not related to what you did here, or if it doesn't interest you. I just thought I would ask... thanks.

Yeah, I've hit that too and have disabled all broken actions in my (private) fork when "select all" is active atm. There is currently code in some actions to extract the uids on the php site for all entries mirroring the list state. One solution would be to extract this code into a generic function and use it in all php actions, but I haven't looked further than that.

pgorod:
Hmm interesting. Can you please point me to those places that have code for this?

lazka:
> Hmm interesting. Can you please point me to those places that have code for this?

export_utils.php::export() for example

Mac-Rae:
Assessed :+1:

PR #7541, opened 2019-07-10
ListView: Fix the selection count when executing an action without any selection

lazka:
## Description

When there is a selection from different page but not on the current one and one
selects an action the selection count would increase.

This is because the current selection is appended without checking if there is a selection
to append. This results in a lone "," getting appended which results in an extra empty uid.

Guard against that by checking if there are any uids to append before appending ","

## Motivation and Context

The selection count should match the amount of selected items.

## How To Test This

* Leads List View
* Select a row
* Got to next page
* Bulk actions > delete > cancel
* It should show that only 1 item is selected, not 2

## Types of changes

### Final checklist


Mac-Rae:
Assessed :+1:

PR #7550, opened 2019-07-11
SugarWidgetSubPanelEmailLink: Fix missing opt-in ticks after inline editing

lazka:
## Description

**Note for reviewers: Both commits just remove unneeded code. The indentation changes make it look otherwise..**

The first commit replaces some code with the equivalent getEmailClient().

The second commit:

SugarWidgetSubPanelEmailLink uses a simple mailto link for the non-sugar client
case, but since #6486 EmailUI can create links for external email clients and also include
opt-in related ticks.

Remove the fallback case and let EmailUI handle it.

When inline editing an email field and then pressing enter the then shown link will now
include opt-in ticks when the external email client is used.

## Motivation and Context

Inline editing removes opt-in ticks while it shouldn't.

## How To Test This

* View a lead
* Inline edit the email
* Press enter
* The opt-in ticks should still be there

## Types of changes

### Final checklist

### DIFF
--- a/include/ListView/ListViewDisplay.php
+++ b/include/ListView/ListViewDisplay.php
@@ -488,13 +488,7 @@ class ListViewDisplay
         }
 
 
-        $userPref = $GLOBALS['current_user']->getPreference('email_link_type');
-        $defaultPref = $GLOBALS['sugar_config']['email_default_client'];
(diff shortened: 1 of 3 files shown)


cameronblaikie:
Assessed :+1:

Issue #6967, opened 2019-02-19
Fix CSV export € symbol


holdusback:
#### Issue
€ symbol is badly exported in csv's report. € became ¬. With UTF16-LE.

#### Expected Behavior
A € symbol not ¬

#### Possible Fix
The CRM just doesnt have the option to encode in UTF-8 WITH BOM.
Take the UTF-8 encodage in user advanced settings.

in file: /modules/AOR_Reports/AOR_Report.php
In build_report_csv(), just before print $csv add this:
$csv = chr(239) . chr(187) . chr(191) . $csv;

#### Your Environment
* SuiteCRM Version used: 7.11.1

Dillon-Brown:
Duplicate of #18

Issue #18, opened 2014-03-06
[PATCH] Bad CSV export on Reports module


lsmonki:
The csv export do not escapes quotes ("), do not support new lines on field content, and can't be oppened without problems on excel. Also the http headers do not set the filename ok.

Here is a patch for the `build_report_csv` function on `modules/AOR_Reports/AOR_Report.php`

```
    function build_report_csv(){

        ini_set('zlib.output_compression', 'Off');

        ob_start();
        require_once('include/export_utils.php');

        $delimiter = getDelimiter();
        //text/comma-separated-values

        $user_agent = isset($_SERVER['HTTP_USER_AGENT']) ? $_SERVER['HTTP_USER_AGENT'] : '';
        if ($GLOBALS['locale']->getExportCharset() == 'UTF-8' &&
            ! preg_match('/macintosh|mac os x|mac_powerpc/i', $user_agent)) // Bug 60377 - Mac Excel doesn't support UTF-8
        {
            //Bug 55520 - add BOM to the exporting CSV so any symbols are displayed correctly in Excel
            $BOM = "\xEF\xBB\xBF";
            $csv = $BOM;
        }
        else
        {
            $csv = '';
        }

        $sql = "SELECT id FROM aor_fields WHERE aor_report_id = '".$this->id."' AND deleted = 0 ORDER BY field_order ASC";
        $result = $this->db->query($sql);

        $fields = array();
        $i = 0;
        $first_item = true;
        while ($row = $this->db->fetchByAssoc($result)) {
            $field = new AOR_Field();
            $field->retrieve($row['id']);

            $path = unserialize(base64_decode($field->module_path));

            $field_module = $this->report_module;
            if($path[0] != $this->report_module){
                foreach($path as $rel){
                    $field_module = getRelatedModule($field_module,$rel);
                }
            }
            $label = str_replace(' ','_',$field->label).$i;
            $fields[$label]['field'] = $field->field;
            $fields[$label]['display'] = $field->display;
            $fields[$label]['function'] = $field->field_function;
            $fields[$label]['module'] = $field_module;


            if($field->display){
                if(!$first_item)
                    $csv .= $delimiter;
                else
                    $first_item = false;

                $csv.= '"'.preg_replace("/\"/","\"\"", $field->label).'"';
            }
            ++$i;
        }

        $sql = $this->build_report_query();
        $result = $this->db->query($sql);

        while ($row = $this->db->fetchByAssoc($result)) {
[...]

PR #7580, opened 2019-07-16
export: Update export_excel_compatible to work with all Excel versions

lazka:
## Description

Excel on Windows/macOS when simply opening a CSV file will try to guess the
CSV dialect (encoding, delimiter, etc) and just go with the first guess.
It's still possible to do an explicit import with Excel where you can select the
dialect, but that's not very obvious.

The most common wrong guess is that mostly ASCII utf-8 will get detected as latin-1.

I've tested various encodings with Excel (2019) and codepoints from latin-1:

* utf-8: win/macOS use latin-1
* utf-8+bom: win/macOS correctly use utf-8
* utf-16le: win uses utf-16le, macOS garbled text
* utf-16le+bom: win/macOS fail to detect CSV and load it as one text block

thus utf-8+bom is the preferred mode.

This patch makes the "export_excel_compatible" mode use utf-8+bom for the CSV encoding.

In addition this removes the hardcoded delimiter in export_excel_compatible mode, because excel
uses the Windows user list separator for this and there is no way of knowing it (except in javascript maybe)
and so we better leave this setting up to the user.

## Motivation and Context

I'd like to have exported CSV files work as is in Excel on Windows and macOS (and LibreOffice, but that just works anyway)

## How To Test This

* Set export_excel_compatible to true
* Export a report or a list selection and open it in Excel

## Types of changes

### Final checklist

### DIFF
--- a/export.php
+++ b/export.php
@@ -47,7 +47,6 @@ ini_set('zlib.output_compression', 'Off');
 ob_start();
 require_once('include/export_utils.php');
 global $sugar_config;
-global $locale;
 global $current_user;
 global $app_list_strings;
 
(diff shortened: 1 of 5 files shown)


lazka:
This fixes #6967 and #18, but only when export_excel_compatible is set. But I think it's a good first step.

Mac-Rae:
Hi @lazka,

I'm having some issues with testing as it doesn't appear to function as you would expect.

Below are the four tests I ran;

Test 1 - Without Fix (With Excel Compat Off)
Without(Off)

Test 2 - Without Fix (With Excel Compat On)
Without(On)

Test 3 - With Fix (With Excel Compat Off)
With(Off)

Test 4 - With Fix (With Excel Compat On)
With(On)

As you can see with the fix installed and with the excel compatibility enable the output does not appear to load within excel correctly.

Could you make some updates and I can reassess when you're happy with the PR

lazka:
Thanks for testing. Weird, the non-compat case shouldn't have changed at all.

* Are you on Windows or macOS.
* Did you export from the reports module or the list view?

pgorod:
Note that system locales and system-level delimiter definitions affect the way Excel handles CSV's.
So it would be nice to see what's defined in Windows' Regional settings in these tests.

lazka:
> Note that system locales and system-level delimiter definitions affect the way Excel handles CSV's.

oh, thanks, I didn't know that. I'll try testing with "\t" again, maybe it was used for this reason.
I can't check right know, but I think my format settings are set to german probably.

pgorod:
These are the settings I mean:
image
It's terribly hard to reach this screen in Windows 10... I have to keep clicking Advanced, more settings, more advanced, Additional settings...

Mac-Rae:
Hi @lazka, Sorry i never got back to you, I downloaded the files onto Linux Ubuntu 18.04 and tried the files on Windows 10

lazka:
OK, I've tested now with another Windows user with US default settings and the separator is indeed different (and there is no way to make it work with German and English). There is a special first line `len=,` which tells Excel what the separator is, but that breaks its encoding detection again.

I'll remove the separator bits from this PR because there is no easy fix.

lazka:
> I'll remove the separator bits from this PR because there is no easy fix.

Done. This should be ready now.

Mac-Rae:
Assessed :+1:

PR #7591, opened 2019-07-17
Composerify Smarty

### LINKED ISSUE #6719: Upgrade Smarty and move it to Composer
I saw some other PRs that upgraded dependencies/moved them to composer so I figured this was a relatively simple fix to make :)

The current version of Smarty in SuiteCRM: `2.6.29` (this is weird because the README in the Smarty directory says 2.6.0, but there's a commit mentioning an upgrade to 2.6.29)
Latest version of Smarty: `3.1.33`

Currently the Smarty library is vendored in [`include/Smarty/`](https://github.com/salesagility/SuiteCRM/tree/master/include/Smarty), with version 2.6.29.

We can actually upgrade to the v2.6.31 version on Composer right now with no breaking changes, if we want: https://packagist.org/packages/smarty/smarty#v2.6.31

2.6.31 includes a fix for a PHP 7.2 deprecation. I'm not sure if 3.0 has any major breaking changes that'd be problematic for SuiteCRM, or how the upgrade process would work exactly, but I'd say it's at least worth considering.

EDIT: [Here's the Smarty v2 to v3 upgrade guide](https://raw.githubusercontent.com/smarty-php/smarty/master/SMARTY_2_BC_NOTES.txt)

connorshea:
## Description
Fixes #6719.

This keeps the `Smarty/plugins/` directory for now since I didn't want to move it and cause more breakage.

This removes the security-advisories package because it was preventing me from installing Smarty (unless I upgraded to 3.1, which would almost definitely cause significant breakage). I realize that's not ideal, but we're not changing anything with, if anything we're _improving_ the security of the CRM by upgrading from 2.6.29 to 2.6.31, and moving it to Composer so it's easier to upgrade.

## How To Test This
- Make sure the tests pass.
- Make sure that you can run `composer install`.
- Make sure the CRM's main pages still work and render correctly. 

## Types of changes

This is _technically_ a breaking change if anyone was trying to require a file from `include/Smarty/` directly, but other than that this shouldn't be a problem.

### Final checklist


connorshea:
cc: @Dillon-Brown RE: the security-advisories thing

connorshea:
Ok, I guess that change didn't quite work 🤔

connorshea:
So I diffed the version of the file in the CRM and the version in the original Smarty repo and it looks like we added some custom code to `Smarty_Compiler.class.php` :/

```php
if ( $theme_template == 'true' )
    $include_file = '"'.SugarThemeRegistry::current()->getTemplate(str_replace(array('"',"'"),'',$include_file)).'"';
```

Ours: https://raw.githubusercontent.com/salesagility/SuiteCRM/v7.11.6/include/Smarty/Smarty_Compiler.class.php
Upstream: https://raw.githubusercontent.com/smarty-php/smarty/v2.6.31/libs/Smarty_Compiler.class.php

connorshea:
The only use of `theme_template` seems to be the two in this file: https://github.com/salesagility/SuiteCRM/blob/6208631608d207b85c00186b498a7c2bcf00278d/themes/SuiteP/tpls/header.tpl#L41

So we can probably workaround it without a plugin pretty easily, and then just use vanilla Smarty.

What those two uses of `theme_template` seem to be doing is trying to get the `_head.tpl` and `_headerModuleList.tpl` files from the registered theme directory, but since they're only used in that one file, and none of the themes override it, it should be perfectly safe to just use the relative path directly?

Mac-Rae:
Assessed :+1:

Dillon-Brown:
This should be ok to merge. We'll need to mention the BC changes in the release notes and let third-party plugins know as well. Perhaps a page on the documentation site explaining how to properly require Smarty after this change?

connorshea:
@Dillon-Brown I think it'd fit best in the release notes, so I'll just write an explanation in this comment that can be included in the release notes when next month's release is published.

>Smarty was moved to Composer, as part of this change if you explicitly required a file from `./include/Smarty/` you'll probably need to change that `require`. Composer's autoloader will handle most things for you, simply removing the `require` should be sufficient.
>
>Plugin files are still available at `./include/Smarty/plugins/` and can be `require`d explicitly, and custom plugins should still go in `./custom/include/Smarty/plugins/`.
>
>We will likely upgrade Smarty to 3.x in the future, so any heavily-customized usages of Smarty should keep in mind that their customizations may break in the future.

PR #7609, opened 2019-07-19
Shuffle the codeception tests.

### LINKED ISSUE #7022: Test suite should be run in random order
#### Issue
Essentially, this: https://qualitycoding.org/random-test-order/

#### Expected Behavior
The test suite should be run with a randomly generated seed to guarantee that tests don't rely on each other's state (essentially, if you assume A -> B -> C -> D, and C relies on something B adds to the database – unintentionally or otherwise – C will now fail when run in isolation but not when run as part of the test suite.)

Similarly, it can cause problems when introducing new tests that mess with the existent order and reveal that tests have been unintentionally reliant on state from other tests. It also could potentially mean that a test doesn't _actually_ test the functionality, it's just been reliant on state that was introduced by a previous test (e.g. A creates a User and then doesn't get rid of it. B runs, and the test tries to create a User in a different way that doesn't work, but the database still has the User that A created, so B checks if there's a User in the db and passes despite the code being broken).

#### Actual Behavior
The tests are run without a seed, so they always run in the exact same order every time.

#### Possible Fix

The best fix would probably be to use [Codeception's shuffle feature](https://codeception.com/docs/reference/Configuration) by setting `shuffle: true` in the `codeception.yml` file.

You may want to run the test suite a few times with `shuffle: true` to see if that brings up any existing state problems that would result in random test failures. Ultimately, I'd say this'd be a significant improvement over the current situation even if it may introduce flaky tests occasionally, since it's exposing an anti-pattern.

#### Steps to Reproduce
Look at Travis CI logs and note that the tests are always run in the same order, and that the test runner reports the seed as being blank.

connorshea:
## Description

image

Testing to see if we can fix #7022 right now :) I'm guessing there are tests that are implicitly dependent on one another (I know that's true of PHPUnit, so I haven't enabled it for PHPUnit yet).

## Motivation and Context
This may introduce new flakiness – which is not ideal – but it helps prevent tests that are dependent on state created by other tests, which causes fragile tests. See #7022 for a fuller description of the problems this can solve.

## How To Test This
See if Travis passes!

## Types of changes

### Final checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,7 +76,7 @@ before_script:
 script:
   # Run the wizard installer
   - echo "using install wizard"
-  - ./vendor/bin/codecept run install --env travis-ci-hub -f --ext DotReporter
+  - ./vendor/bin/codecept run install --env travis-ci-hub --ext DotReporter
   - ./build/push_output.sh
 
   # Run the unit tests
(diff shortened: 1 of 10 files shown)


connorshea:
I think the only failure this causes now is the InlineEdit test in AccountsCest. Based on the screenshot that's made when it fails, I think another test is using the search, which causes the backend to save the query, and when it runs before InlineEdit it breaks.

connorshea:
I can't re-run the tests locally with the same order because of the Codeception version ;-;

2.5.0 adds 'seeds' (a random number that acts as the source of Codeception's random ordering) that allow you to reproduce the test order if you want to try debugging a random test failure locally. Unfortunately, with PHP 5.5 we're stuck on 2.4.1 :(

PR #7625, opened 2019-07-24
WIP: Fix the inline edit cancel dialog showing up repeatedly.

### LINKED ISSUE #7432: "Edit not saved" warning when inline-editing in Leads DetailView requires multiple Cancel clicks
#### Issue
When inline-editing a field in the Leads DetailView (and presumably elsewhere in the CRM) and then clicking outside the field, a dialog pops up with a warning about how the edit isn't saved. When you click "Cancel", the dialog will reappear one or more times before it finally disappears.

This is the same issue as #2545, which was supposed to have been fixed back in 2017, but as far as I can tell was not. 

#### Expected Behavior

Hitting Cancel once after the dialog appears would return to the inline edit.

Screen Shot 2019-06-19 at 11 56 22 AM

#### Actual Behavior

The browser alert appears multiple times before being fully dismissed.

#### Possible Fix
There was a supposed fix in 57bf2e3d00bb2ed543c5502505addf40aa6418a4, but either a regression was introduced since then or it never fully fixed the problem.

#### Steps to Reproduce
1. Go to the Demo CRM, or any other CRM you have that's on 7.9.x through 7.11.x, and login
2. Go to a Lead's DetailView
3. Double-click a field to inline-edit it
4. Change the value any way you want, it doesn't matter as long as it differs from the original value
5. Click anywhere outside the field to exit the inline-edit state
6. A popup should appear warning you about abandoning the edit, hit Cancel.
7. The dialog will appear again multiple times, and eventually stop after you hit Cancel enough times.

Screen recording of the bug, I perform it twice here: https://imgur.com/XH6uw9O

#### Context
This isn't a huge deal, and is probably low priority, but still annoying.

#### Your Environment
* SuiteCRM Version used: 7.9.17, 7.10.15, and 7.11.5
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): I've been able to reproduce it in Chrome 75 and Firefox 67
* Environment name and version (e.g. MySQL, PHP 7): MySQL 5.7, PHP 7.0
* Operating System and version (e.g Ubuntu 16.04): Debian Linux (server), macOS Mojave 10.14.5 (client)

connorshea:
## Description
Fix #7432.

This is done by disabling the listener before registering it initially, which removes any listeners that may have existed in a previous page view.

## How To Test This
Try to reproduce this issue, make it sure doesn't occur:

- Go to a Lead detailview page (or any detailview page should work, really).
- Hard reload.
- If you edit a field now and try to click out, the 'cancel' dialog will only show up once.
- Click the next or previous buttons to navigate to a new Lead.
- If you edit a field and try to click out now, and then hit cancel on the dialog it'll show up twice.

With this change it should no longer show up more than once.

## Types of changes

### Final checklist

### DIFF
--- a/include/InlineEditing/inlineEditing.js
+++ b/include/InlineEditing/inlineEditing.js
@@ -273,7 +273,19 @@ function clickedawayclose(field,id,module, type){
     clickListenerActive = true;
 }
 
-$(document).on('click', function (e) {
+/*
+ * Disable any existing click listeners before registering this listener.
+ * This is done to prevent the listener from being registered more than
+ * once across page views.
+ */
+$(document).off().on('click', warnOnLeavingActiveField);
+
+/**
+ * Display a confirmation dialog if the user clicks outside of a field that's actively being edited.
+ * 
+ * @param {Event} e - the click event 
+ */
+function warnOnLeavingActiveField(e) {
     if (clickListenerActive) {
         var field = ie_field;
         var id = ie_id;
@@ -342,7 +354,7 @@ $(document).on('click', function (e) {
             }
         }
     }
-});
+};
 
 /**
  * Depending on what type of field we are editing the parts of the field may differ and need different jquery to pickup the values


connorshea:
Ok this breaks the Actions dropdown, so this isn't a viable solution.

PR #7630, opened 2019-07-25
Avoid adding quotes on already quoted parameters.

### LINKED ISSUE #6996: Escaped strings issue, breaks "My favorites" filters and perhaps other things
### Issue
I ran into an Issue in the Filters on the List View of Cases module. The "My favorites" filter checkbox wasn't working well in 7.11.2.

I found the queries to the database are being sent to the database with double apostrophes like this `favorites.assigned_user_id = ''1''`, and it wouldn't return any results.

The file where that is originally defined is this:
https://github.com/salesagility/SuiteCRM/blob/master/modules/Cases/metadata/SearchFields.php#L155

And the actual replacement takes place here:

https://github.com/salesagility/SuiteCRM/blob/master/include/SearchForm/SearchForm2.php#L1278-L1280

Using this helper function which had recent changes regarding escaping strings:

https://github.com/salesagility/SuiteCRM/blame/master/include/utils.php#L3955

### Possible immediate fix for just this case...
The simple, direct fix is to remove the apostrophes in [SearchFields.php](https://github.com/salesagility/SuiteCRM/blob/master/modules/Cases/metadata/SearchFields.php#L155), make it just 
`and favorites.assigned_user_id = {1}",`

### But this happens in a ton of other places!

```php
modules/Notes/metadata/SearchFields.php:54: and favorites.assigned_user_id = '{1}'",

modules/AOS_Products/metadata/SearchFields.php:91:  and favorites.assigned_user_id = '{1}'",

modules/Accounts/metadata/SearchFields.php:231: and favorites.assigned_user_id = '{1}'",

modules/Calls/metadata/SearchFields.php:180:  and favorites.assigned_user_id = '{1}'",

modules/AOS_Invoices/metadata/SearchFields.php:63:  and favorites.assigned_user_id = '{1}'",

modules/Project/metadata/SearchFields.php:54: and favorites.assigned_user_id = '{1}'",

modules/Leads/metadata/SearchFields.php:95: and favorites.assigned_user_id = '{1}'",

modules/Tasks/metadata/SearchFields.php:65: and favorites.assigned_user_id = '{1}'",

modules/AOS_Contracts/metadata/SearchFields.php:41: and favorites.assigned_user_id = '{1}'",

modules/Spots/metadata/SearchFields.php:56: and favorites.assigned_user_id = '{1}'",

modules/Documents/metadata/SearchFields.php:86: and favorites.assigned_user_id = '{1}'",

modules/Contacts/metadata/SearchFields.php:95:  and favorites.assigned_user_id = '{1}'",

modules/Opportunities/metadata/SearchFields.php:69: and favorites.assigned_user_id = '{1}'",

modules/Meetings/metadata/SearchFields.php:66:  and favorites.assigned_user_id = '{1}'",

modules/AOS_PDF_Templates/metadata/SearchFields.php:42: and favorites.assigned_user_id = '{1}'",

modules/AOS_Quotes/metadata/SearchFields.php:54:  and favorites.assigned_user_id = '{1}'",
```

So I guess it might be better to fix this in the core function, make sure it handles apostrophes well without duplicating them. 

It could also be interesting to **ensure with an automated test** that no pair of `''` goes into a query, unless it's an empty string. So `field=''` is ok, but `field=''some_id''` is not.

### Steps to Reproduce
There are many cases of this problem, but one of them is
1. In Cases module, mark a case as a Favorite. Assign it to yourself.
2. In the List view, hit the `Filter` button
3. Check `My Favorites` and hit `Search`
4. You get an empty list, instead of seeing the Case you created in step 1.

### Context
This would be low priority if we were sure it isn't breaking the app in more places... so maybe it's medium.

#### Your Environment
* SuiteCRM Version used: 7.11.2
* Environment name and version: MySQL, PHP 7

marin-h:
Fix #6996
Avoid adding quotes on already quoted parameters by using already existing `quote` method instead of concatenating quotes with php dot operator.

## Description
Improve the way values are quoted when replacing parameters in `SELECT` clauses, for only adding quotes if needed.
Fixes #6996 

## Motivation and Context
See #6996

## How To Test This
1. Browse ListView of any module that uses "My favorites" filter, for example: Projects.
2. Set "My Favorites" checkbox to checked.
3. Click Search button.
4. The query should be generated properly without throwing errors.
   
## Types of changes

### Final checklist

### DIFF
--- a/include/utils.php
+++ b/include/utils.php
@@ -3989,7 +3989,7 @@ function string_format($format, $args, $escape = true)
             $args[$i] = implode("','", $values);
         }
 
-        $result = str_replace('{'.$i.'}', "'" . $args[$i] . "'", $result);
+        $result = str_replace('{'.$i.'}', $db->quote($args[$i]), $result);
     }
 
     return $result;


Mac-Rae:
Assessed :+1:

PR #5457, opened 2018-03-09
Feature/studio icons

zoltankocsardi:
## Description
Replacing Studio icons (including children)

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [x] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

PR #5273, opened 2018-02-20
Convert SuiteP Icons to font icons

daniel-samson:
## Description

In the navigation bar of the dusk theme.The contrast between the icons and the background is too low. Users are unable to differentiate them from the background.

This is because the color of the icons are difficult to change. So in this pull request it have swapped them for the suitepicon font. Fonts are much easier to style with css. 

Note: There are some icons which I haven't changed like the admin panel icons.

## Motivation and Context

* fixes the icons in the theme

## How To Test This

* Rebuild Sass
* Run repair and rebuild
* Repair JS Minified
* Repair JS Groupings
* Repair JS Languages
* Clear Browser cache
* Test for missing icons
* Test for alignment issues
* To for icons being too small

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [x] My code follows the code style of this project found [here]x(https://suitecrm.com/wiki/index.php/Coding_Standards).
- [ ] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://suitecrm.com/wiki/index.php/Contributing_to_SuiteCRM) guidelines.

PR #7659, opened 2019-07-29
Fix icons not rendering properly in Alerts.

### LINKED ISSUE #6264: The image is broken --> See screenshot
#### Context
This is a bug core on SuiteCRM version 10.4, It related to module Alerts, Please check again.

#### Your Environment

* SuiteCRM Version used: 10.4
 This is a image bug of core --> See screenshot
screenshot_1
- The image is broken --> See screenshot
screenshot_2

* Browser Chrome
* Environment name and version (e.g. MySQL, PHP 7.1):

connorshea:
## Description
Previously, any modules that were plural (most of them) were getting an `s` appended to their names in order to load the image for that module, leading to the CRM trying to load images like `Leadss.gif`, which does not exist. This only applies the `s` when the module name isn't already plural (this is done by just checking if the module name ends with an `s`, which isn't perfect but better than before). This also changes the icons from `.gif` files to html tags that render the icon font, since most other static image icons have been replaced already by #5273 and #5457.

Fixes #6264 (which should not be closed)

Before:

image

Now it looks like this:

Screen Shot 2019-07-30 at 10 54 14 AM

## Motivation and Context
Icons weren't loading.

~~Ideally, this would be fixed by replacing the gif with a `span` using the new icon font, as I mentioned in my comment on #6264, but I'd prefer to just fix this quickly for now and do that at a later time.~~ This now replaces the gif icons with the `span` tag icon font icons. :) 

## How To Test This
1. Log in as someone with notifications, or generate notifications for your own user account somehow.
2. Open the notifications dropdown.
3. See that the icon next to the notification text is visible and loads. Leads and Calls are both good example notifications.

## Types of changes

### Final checklist

### DIFF
--- a/modules/Alerts/templates/default.tpl
+++ b/modules/Alerts/templates/default.tpl
@@ -22,7 +22,8 @@
         <a class="alert-link text-{if $result->type != null}{$result->type}{else}info{/if}" href="index.php?module=Alerts&action=redirect&record={$result->id}">
         {/if}
             {if $result->target_module != null }
-                <img src="index.php?entryPoint=getImage&imageName={$result->target_module}.gif"/>
+                {* Pluralize the module name if necessary. *}
+                <span class="suitepicon suitepicon-module-{$result->target_module|lower|replace:'_':'-'}{if substr($result->target_module, -1) !== 's'}s{/if}"></span>
                 <strong class="text-{if $result->type != null}{$result->type}{else}info{/if}">{$result->target_module}</strong>
             {else}
                 <strong class="text-{if $result->type != null}{$result->type}{else}info{/if}">Alert</strong>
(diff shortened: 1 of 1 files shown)


connorshea:
It seems like this may not work with Calls :/ I need to figure out how to get a bunch of different notification types for my account to test this change.

connorshea:
Err, I think `imageName=Calls.gif` causes it to load an SVG, which isn't valid as the source of an img tag? I guess the CRM is just trying to load whatever it can find regardless of file extension. 😕

gunnicom:
https://github.com/salesagility/SuiteCRM/pull/7245

connorshea:
@gunnicom sorry, I didn't see that PR when searching around about this issue 🙇 

I've now fixed my PR for more cases by replacing the `.gif` with the icon font that was introduced in 7.10 and adding an `s` if there isn't already one at the end of the module name (this isn't ideal, but it does fix the issue for all cases I've been able to create).

connorshea:
This should be ready for review now 👍

Issue #7605, opened 2019-07-19
[Meta] Move remaining vendored libraries to Composer


connorshea:
The goal of this issue is to track a bunch of smaller issues that are dedicated to removing vendored libraries from the repository and replacing them with corresponding Composer packages. This serves a few purposes:

- Easier to upgrade.
- Prevents customization of core code from other libraries (which is bad because it prevents us from upgrading the libraries)
- Easier to track security vulnerabilities in the libraries.
- Fewer files in our repo.
- Easier to track when libraries stop getting maintained.

A lot of packages have already been moved in the last year or two, but there are still quite a few left to move.

This list isn't necessarily comprehensive, it's just all of the vendored libraries that I was able to find.

- [x] Smarty (Issue: #6719, PR: #7591)
- [x] SAML2 (PR: #6850)
- [ ] Twitter OAuth (Issue: #7595)
- [ ] Facebook OAuth 
- [x] HTMLPurifier (Issue: #3100, PR: #5959) - this was actually added to Composer, but never removed from `include/` for 'backwards compatibility'. I'd say we can probably remove it entirely pretty safely. It's already been removed in the `develop` branch. There's also a file for HTMLPurifier at [`jssource/src_files/include/HTMLPurifier/standalone/HTMLPurifier/Printer`](https://github.com/salesagility/SuiteCRM/tree/master/jssource/src_files/include/HTMLPurifier/standalone/HTMLPurifier/Printer).
- [ ] nusoap (Issue: #7590, PR: #7592)
- [ ] HTTP_WebDAV_Server (Issue: #7596)
- [ ] TCPDF (Issue: #6513)
- [ ] [Crypt_Blowfish](https://github.com/salesagility/SuiteCRM/tree/ef79908dad38b294d2f62924f91aeaa805d8f7a3/include/Pear/Crypt_Blowfish)
- [x] [XML_HTMLSax3](https://github.com/salesagility/SuiteCRM/tree/ef79908dad38b294d2f62924f91aeaa805d8f7a3/include/Pear/XML_HTMLSax3) (#8088)
- [x] Pclzip (PR: #7287) - removed in `develop`.
[...]

PR #7672, opened 2019-07-31
Composerify Zend

connorshea:
## Description
Part of #7605. Moves Zend packages to Composer.

I'm opening this to see if CI will pass :)

It only adds zend-oauth since most of the other Zend packages are dependencies of that one, minus Gdata, which I didn't add because it doesn't seem like it's actually used?

## Motivation and Context
Vendored packages are bad, see #7605 for an explanation on that.

## How To Test This
See if the tests pass, make sure OAuth works.

## Types of changes

I guess this is a breaking change?

### Final checklist


connorshea:
okay, so...

For some reason, there's an extra file in our repository called `Zend/Oauth/Provider.php` and it comes from this repository: https://github.com/smalyshev/Zend_OAuth_Provider

And that's why the tests are failing.

cameronblaikie:
Assessed :+1:

Dillon-Brown:
Looks like this was accidentally merged before internal discussion on including a breaking change as part of a minor release or rebasing to develop could take place (We should probably have a label for this... I've updated SuiteBot to handle "Requires Further Review" to prevent us from having to revert stuff in the future).

Ideally, I'd really like this and other composerify PRs to stay in hotfix-7.10.x so I've created a prosal here: #7719 so that we can hopefully get these libs moved over with less fear of breaking customisations, etc.

Thanks!

PR #7719, opened 2019-08-15
Fix/backwards compatiblity

Dillon-Brown:
## Description

Adds a new way to alias class names when moving a dependency to composer by autoloading the deprecated.php file. I've also added in some missing packages from Zend and re-added previously removed Zend classes as empty PHP files to ensure that existing customisations that use require_once(Some Zend file) won't fatal.

## Motivation and Context

This has a few benefits over something like using class_alias in the old file which then autoloads the new file due to type hints not triggering the autoloader. By using a class alias, both class names are treated as the same exact class. If the deprecated class extends the original class, a type hint against the deprecated class will throw an error if an object of the new class is passed. Also, it doesn't work for abstract classes, interfaces, etc.

By putting everything in ```deprecated.php``` we can also have a clear overview of all the deprecated classes in one place. This also allows us to get ```@depreacted``` warnings in the IDE.

## Types of changes

### Final checklist

### DIFF
--- a/Zend/Exception.php
+++ b/Zend/Exception.php
@@ -1,0 +1,5 @@
+<?php
+/**
+ * Class for backwards compatibility only
+ */
+
--- a/Zend/Gdata.php
+++ b/Zend/Gdata.php
@@ -1,0 +1,5 @@
+<?php
+/**
+ * Class for backwards compatibility only
+ */
+
--- a/Zend/Gdata/App.php
+++ b/Zend/Gdata/App.php
@@ -1,0 +1,5 @@
+<?php
+/**
+ * Class for backwards compatibility only
+ */
+
--- a/Zend/Gdata/App/AuthException.php
+++ b/Zend/Gdata/App/AuthException.php
@@ -1,0 +1,5 @@
+<?php
+/**
+ * Class for backwards compatibility only
+ */
+
--- a/Zend/Gdata/App/BadMethodCallException.php
+++ b/Zend/Gdata/App/BadMethodCallException.php
@@ -1,0 +1,5 @@
+<?php
+/**
+ * Class for backwards compatibility only
+ */
+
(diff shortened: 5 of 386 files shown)


connorshea:
In this case, I kind of feel like it'd be better to rip off the bandaid now (since all the same classes are still there, it'd just require removing the `require`s if anyone is using them) rather than waiting to rip it off in the next major release.

I'm all for following 'SemVer' or whatever we want to call the system we're using, but if we need to keep this level of compatibility we're going to be toiling away with legacy code for a very very long time. There's something to be said for smaller, slightly painful changes over time, compared to a bunch of changes – some of them really painful – occurring all at once, in my opinion.

Though admittedly, part of this is my own selfish desire to make improvements to the codebase without too much hassle on my end :P

connorshea:
Oh, also this adds Zend GData (and some other packages) to Composer even though we don't use it at all. We should probably make note of removing that in the future somewhere.

Dillon-Brown:
@connorshe I can't say I disagree :). Although I think as developers we're probably biased towards refactoring legacy code even if it breaks backwards compatibility just because working with a large legacy codebase can be a pain. 

With this PR I'm mostly looking to strike a balance between keeping things backwards compatible while at the same time making 7.10.x easier to deal with (At least this way we can have a central place to see all the deprecated libs + IDEs can pick up the ```@deprecated``` tag).

Oh and just to note for anyone else looking at moving libs to composer (Assuming this or a similar solution gets merged), I made these commented PHP files with a simple bash script:

```
#!/usr/bin/env bash

for i in `cat file_list.txt | sed '/^$/d;s/^ *//;s/ *$//'`; 
  do mkdir -p `dirname $i`; 
     touch `echo $(dirname $i)$(echo "/")$(basename $i)`;
     cat content.txt >> `echo $(dirname $i)$(echo "/")$(basename $i)`;
  done
```

Updating ```deprecated.php`` should be pretty self-explanatory but you can also just tag me and I can assist with making existing PRs backwards compatible where possible. Thanks!

PR #7732, opened 2019-08-21
Extend Core Modules Without Breaking Everything

Jason-Dang:
##### Done:
- Add new global variables to store custom meta for core modules.
- Change get bean method to check for custom bean classes.
- Add method to dynamically load core and custom bean class files.
- Add method to get custom/core bean class.
- Add method to get custom object name.
- Add method to get core bean file.
- Add method to get custom bean file.
- Add method to get bean meta info as array using above methods.
- Cleanup Bean Factory code.
- Use Script to iterate through entire bean list and replace all occurrences of explicit instantiations of bean class with bean factory usage and then incrementally commit per file.

##### TODO:
- Replace dynamic instantiations of core modules' classes with Bean Factory usage.
   - Occurrences of `new $module();` etc. need to be replaced accordingly.
- Replace usages of the lower-cased module name as the module table name in some queries.
   - Occurrences where the module name is retrieved from `beanList` key(s), lower-cased and used as the table name, replace with `$bean->getTableName();` (may need to instantiate a bean in some cases or better yet add a static property and method for static info like table names etc.).

**Note:** *For the above the module directory or plural name needs to be used instead of the class/object name such as `Accounts` instead of `Account`.*

## Motivation and Context

Previously unable to extend almost all core module classes without breaking system in various areas.

## How To Test This

Smoke test the entire CRM.

## Types of changes

### Final checklist


Jason-Dang:
Close in favour of #7733.

PR #7733, opened 2019-08-21
BeanFactory Extend Core Modules

Jason-Dang:
## Description

Add the ability to define custom core module classes in a non-breaking way. This PR can be merged as is without implementing the next steps as it will not impact any existing functionality. The next steps will provide full coverage so that every part of the CRM respects custom core module classes.

##### Changes:
- Add new global variables to store custom meta for core modules.
- Change get bean method to check for custom bean classes.
- Add method to dynamically load core and custom bean class files.
- Add method to get custom/core bean class.
- Add method to get custom object name.
- Add method to get core bean file.
- Add method to get custom bean file.
- Add method to get bean meta info as array using above methods.
- Cleanup Bean Factory code.
- Use Script to iterate through entire bean list and replace all occurrences of explicit instantiations of bean class with bean factory usage and then incrementally commit per file.

##### Next Steps:
- Replace dynamic instantiations of core modules' classes with Bean Factory usage.
   - Occurrences of `new $module();` etc. need to be replaced accordingly.
- Replace usages of the lower-cased module name as the module table name in some queries.
   - Occurrences where the module name is retrieved from `beanList` key(s), lower-cased and used as the table name, replace with `$bean->getTableName();` (may need to instantiate a bean in some cases or better yet add a static property and method for static info like table names etc.).

**Note:** *For the above the module directory or plural name needs to be used instead of the class/object name such as `Accounts` instead of `Account`.*

## Motivation and Context

Previously unable to extend almost all core module classes without breaking system in various areas. Those that could be extended, don't have their custom classes respected everywhere in the CRM.

## How To Test This

Smoke test the entire CRM with and without extending core module classes.

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOR_Conditions/AOR_Condition.php
+++ b/modules/AOR_Conditions/AOR_Condition.php
@@ -109,7 +109,7 @@ class AOR_Condition extends Basic
             if (isset($post_data[$key . 'deleted'][$i]) && $post_data[$key . 'deleted'][$i] == 1) {
                 $this->mark_deleted($post_data[$key . 'id'][$i]);
             } else {
-                $condition = new AOR_Condition();
+                $condition = BeanFactory::newBean('AOR_Conditions');
                 foreach ($this->field_defs as $field_def) {
                     $field_name = $field_def['name'];
                     if (isset($post_data[$key . $field_name][$i])) {
--- a/modules/AOR_Conditions/conditionLines.php
+++ b/modules/AOR_Conditions/conditionLines.php
@@ -54,7 +54,7 @@ function display_condition_lines($focus, $field, $value, $view)
                 $result = $focus->db->query($sql);
 
                 while ($row = $focus->db->fetchByAssoc($result)) {
-                    $condition_name = new AOR_Condition();
+                    $condition_name = BeanFactory::newBean('AOR_Conditions');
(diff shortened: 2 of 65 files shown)


connorshea:
Some feedback after looking through this for a little while:

- I think it'd be better if you had commits of 5 or 10 files, rather than one commit for every single file. It would still be pretty easy to go back and `git bisect` if any of these changes cause bugs, without adding so many commits to the commit history.
- Similarly, I think it might be easier to review if we split it into 5 (or more) PRs instead of 1 PR with >500 files changed. I suppose that'd be more up to the SalesAgility team, just my opinion :) I definitely prefer smaller PRs if at all possible.
- The commit messages all being `PHP Commit - BeanFactory Extend Core Modules` is pretty odd, I know these were generated by a script but I think it'd be better to at least differentiate them, e.g. like `BeanFactory Extend Core Modules 1/100` or some such.

I'm glad GitHub added the "Viewed file" checkboxes recently, otherwise this'd be a real monster to review :P

Jason-Dang:
Closed in favour of https://github.com/salesagility/SuiteCRM/pull/7795 & https://github.com/salesagility/SuiteCRM/pull/7806 to address feedback above :)

PR #5640, opened 2018-04-02
Fix/5639 subpanel compose email

PedroErnst:
Fixes a few bug related to the compose email button in a records' subpanel.

## Description

1. Fixes an issue with compose buttons not being displayed at all due to the opt in span throwing off a jquery find.
2. Correctly aligns the compose button with the other buttons in the subpanel dropdown.
3. Fixes an issue with Cases compose modal not showing the related record and not picking up any of the information from it (#5639 ).

## Motivation and Context

Fix bugs

## How To Test This

1. Enter a contact/account record detail view
2. See the subpanel compose email button shows
3. See the correct email is displayed as the button caption
4. See the buttton is correctly aligned
5. Enter a case record
6. Press compose email button in subpanel
7. See the compose modal is correctly filled with data

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [x] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

PR #7760, opened 2019-08-27
SugarEmail: Fix "to" field not being filled when the last record doesn't have an email

lazka:
## Description

In the list view when selecting multiple records or a page and the last element doesn't have
and email then nothing gets listed in the "to" field of the email widget.
Similarly when multiple are selected and some don't have an email then invalid addresses (<>)
are listed in the "to" field.

This was introduced in #5640 (likely by accident, see discussion there) by checking the last
email instead of during looping over them when deciding when to add one.

This moves the empty email check into the loop.

## Motivation and Context

We have some records without emails and in case they end up in a selection then emails are broken.

## How To Test This

* Go to a list view
* Select an entry with an email
* Select an entry below without an email
* Action menu > Email
* The email of the first entry should be in the "to" field

## Types of changes

### Final checklist


Mac-Rae:
Hi @lazka, I havent been able to replicate this working locally however I will get @cameronblaikie to test this tomorrow to confirm.

lazka:
(I always have to do a "javascript repair" after changing things there to avoid browser caches, maybe that helps)

cameronblaikie:
Assessed :+1:

PR #7771, opened 2019-08-27
Add dotenv support for the test environment

### LINKED ISSUE #7620: Introduce dotenv files
#### Issue
This is part of #7344, but I'm splitting it out into a separate issue.

Introduce `.env` files for use with testing, either [phpdotenv](https://github.com/vlucas/phpdotenv) or [Symfony Dotenv](https://github.com/symfony/dotenv).

This is mostly useful for running multiple CRMs side-by-side on the same computer, and also because it makes it easier to figure out what the local environment variables are.

For example, I have a CRM for contributing to the upstream and a CRM for our customized version of the CRM. I'd like to have a dotenv file in each so that I can set environment variables for the test suite that differ between the two sites.

CRM 1 is at http://localhost and CRM 2 is at http://localhost:8080, but the environment variable I have set on my Mac is set to `INSTANCE_URL=http://localhost`, which means that when I run the tests on CRM 2, I end up running them on the CRM 1 site. The only way to fix this is to change the environment variables every time I switch between sites, which is _really_ annoying.

If we introduced dotenv files, we could have the variables set in the root of the site, e.g. `CRM1/.env` and `CRM2/.env`.

CRM1 `.env`:
```
INSTANCE_URL="https://localhost"
DATABASE_NAME="automated_tests"
```

CRM2 `.env`: 
```
INSTANCE_URL="https://localhost:8080"
DATABASE_NAME="automated_tests2"
```

We could also have `.env.test` to separate environment variables between 'development' and 'test', e.g. when running the tests we don't want to mess with our database we use for development.

TL;DR the main benefits are:

- Easier to configure the environment variables for a site
- Make it possible to switch between two separate instances to run tests
- Everything is configured in the instance's directory, rather than some configuration being externally stored as environment variables.

connorshea:
## Description
Fixes #7620. This introduces support for loading environment variables via [phpdotenv](https://github.com/vlucas/phpdotenv).

I chose phpdotenv over Symfony dotenv because Symfony Dotenv doesn't support overloading existing environment variables.

## Motivation and Context

If you wanted to have multiple CRMs locally, or a separate configuration when running the test suite vs. developing the CRM, that wasn't possible without manually updating your `~/.bash_profile` or updating your `config.php` file.

That makes running the tests or contributing upstream unnecessarily difficult. With a dotenv file, we can fix both these problems :)

Simply copy the `.env.dist` file to `.env.test` in the root of your project and modify each variable as necessary to set up your project. The environment variables defined in `.env.test` only apply to the test suites, and will override any environment variables set elsewhere (e.g. in `~/.bash_profile`). If the `.env.test` file doesn't exist, the CRM works the same as before.

## How To Test This
- Copy the `.env.dist` file to `.env.test`, modify it to fit your existing environment if necessary.
- Run the tests (e.g. `./vendor/bin/robo test:acceptance`).

You can modify the `tests/_bootstrap.php` file to include something like `echo getenv('INSTANCE_URL');` after the dotenv file is loaded, and then see if the tests actually output the instance URL set in your `.env.test` file. If they do, it worked :D

You can also make sure the tests continue to work if the `.env.test` doesn't exist.

I'm not 100% sure whether these environment variables are loaded with PHPUnit tests. I need to look into that a bit more.

## Types of changes

### Final checklist

I should probably update the automated testing docs to mention this improvement.


connorshea:
@Dillon-Brown *poke*

cameronblaikie:
Looks good @connorshea :smile: 
Assessed :+1:

PR #7797, opened 2019-09-02
Fixed #7779 - PHP Fatal error in modules/Connectors

### LINKED ISSUE #7779: PHP Fatal error in modules/Connectors/connectors/sources/ext/rest/twitter/config.php PHP version 7.1.31
Code in _modules/Connectors/connectors/sources/ext/rest/twitter/config.php_ produces _PHP Fatal error:  Uncaught Error: Cannot use string offset as an array_ 

#### Issue
Using SuiteCRM v7.10.15 with PHP 7.1.31, code in _modules/Connectors/connectors/sources/ext/rest/twitter/config.php_ produces _PHP Fatal error:  Uncaught Error: Cannot use string offset as an array_ when hitting CRM home.

Dashlets and rest of the page are not displayed.

#### Expected Behavior
Expected to see dashlets and rest of the page

#### Actual Behavior
Home page is not properly displayed and apache error.log shows _PHP Fatal error:  Uncaught Error: Cannot use string offset as an array in .../modules/Connectors/connectors/sources/ext/rest/twitter/config.php:4_

#### Possible Fix
Problem can be fixed by putting _$config = array();_ before the rest of the code in the file (thus initializing $config variable before use). $config is NULL at this point of execution.

Alternatively, problem can be solved by changing PHP version to 7.3.XX.

#### Steps to Reproduce
Using SuiteCRM 7.10.15 with PHP 7.1.31
1. Hit the CRM home (after logging in)

#### Context
Problem occurs from the fact that variable $config used in mentioned file is not initialized before use. PHP then assumes it to be a string (while it should be array).

While this behavior can be blamed on PHP by initializing (implicitly) variable to string type, initializing variables (explicitly in code) is good practice and can prevent this types of problems (and many other), so I feel it should be fixed in SuiteCRM (_modules/Connectors/connectors/sources/ext/rest/twitter/config.php_)

#### Your Environment

* SuiteCRM Version used: 7.10.15
* Environment name and version (e.g. MySQL, PHP 7): PHP 7.1.31

Dillon-Brown:
## Description

modules/Connectors/connectors/sources/ext/rest/twitter/config.php produces PHP Fatal error: 
```Uncaught Error: Cannot use string offset as an array when hitting CRM home.```

## Motivation and Context

Simply initialise the config array before assignments.

## How To Test This

Check for PHP fatals on the homepage.

## Types of changes

### Final checklist

### DIFF
--- a/modules/Connectors/connectors/sources/ext/rest/twitter/config.php
+++ b/modules/Connectors/connectors/sources/ext/rest/twitter/config.php
@@ -1,5 +1,6 @@
 <?php
 /***CONNECTOR SOURCE***/
+$config = [];
 $config['name'] = 'Twitter';
 $config['properties']['consumer_key'] = '';
 $config['properties']['consumer_secret'] = '';


Mac-Rae:
Assessed :+1:

PR #7803, opened 2019-09-03
ShowDuplicates: Only show accessible duplicates

lazka:
## Description

In case a user tries to create a new lead/account etc it shows duplicates
from other security groups to which the user doesn't have access to.

This adds checks to the various checkForDuplicates() implementations to
only return duplicates which the user can view.

## Motivation and Context

Users shouldn't see things to which they don't have the rights.

## How To Test This

* Create a lead as user A (security group X)
* Create a lead with the same name as user B (security group Y)
* There should be a duplicate page when creating the second lead.

## Types of changes

### Final checklist

### DIFF
--- a/include/SugarObjects/forms/PersonFormBase.php
+++ b/include/SugarObjects/forms/PersonFormBase.php
@@ -240,7 +240,15 @@ abstract class PersonFormBase extends FormBase
             }
         } //if
 
-        return !empty($rows) ? $rows : null;
+        $can_view = [];
+        foreach ($rows as $row) {
+            $bean = BeanFactory::getBean($this->moduleName, $row['id']);
+            if ($bean->ACLAccess('view')) {
+                $can_view[] = $row;
+            }
+        }
+
+        return !empty($can_view) ? $can_view : null;
     }
 
 
--- a/modules/Accounts/AccountFormBase.php
+++ b/modules/Accounts/AccountFormBase.php
@@ -86,22 +86,24 @@ class AccountFormBase
             $query .=   ' ('. $temp_query . ' ) ';
         }
 
+        $rows = array();
         if (!empty($query)) {
-            $rows = array();
             $db = DBManagerFactory::getInstance();
             $result = $db->query($query);
-            $i=-1;
             while (($row=$db->fetchByAssoc($result)) != null) {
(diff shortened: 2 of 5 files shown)


pgorod:
@eggsurplus, this should interest you...

eggsurplus:
@pgorod Thanks. This isn't going to be desired by many. While duplicates are a gray area, most companies will not want duplicates to be entered.

Yes, seeing duplicates that they don't have access to gives them high level visibility that the record exists, but they will not have access to that duplicate record to drill in further. This limits the access that the user has.

As far as whether they should see the duplicate at all, since they are entering the record they obviously must have a need for access. Said another way, if that duplicate didn't exist then they would be able to create that record and have access to it. The current implementation lets them know that the record exists which will lead them to contact the appropriate person to get corrected permissions to that record. If that person responsible deems that there isn't anything particularly sensitive about that record they can then enable access as required.

That's a wordy way of saying that we should leave the current way duplicates are handled the same.

pgorod:
Thanks, that's certainly food for thought.

What do you think of this @lazka?

lazka:
To give some context, we have multiple independent groups and we need duplicate records for the same persons because of privacy laws. Different groups get different types of consent for what data they are allowed to save and what they can do with it.

Even if we would try to avoid duplicates there is no way of knowing who to contact since the records can't be viewed and clicking them just results in an error.

But, I can see how for other organizations the current behavior might be preferred.

eggsurplus:
@lazka makes sense. Process-wise, when they get a no access error it could be a set process where they then access a defined data admin who then polices such cases as necessary. Either they provide access or they force a duplicate to be created and give the person access to that new duplicate.

I don't believe software needs to be used here. Just a defined internal process.

Mac-Rae:
Assessed :+1:

eggsurplus:
Unless this is a configurable behavior this shouldn't be pulled in. There are good reasons for wanting either behavior. Thus it should stay as is to match current expectation or have this behavior be configurable.

Mac-Rae:
Yes @eggsurplus, I agree this is being left here for now and won't be merged pending review and update from @samus-aran &/ @Dillon-Brown

PR #7807, opened 2019-09-04
Fixed #7740 - Discussion on replacing the StateChecker with database truncation in tests

### LINKED ISSUE #7740: Discussion on replacing the StateChecker with database truncation in tests
I was hoping to start a discussion on the current StateChecker used for PHPUnit tests and whether we can replace it with a different solution.

I realize there's been a decent investment of time and energy into the feature, but I think it'd be best if we were to move from the StateChecker to truncating the database instead.

There are a few reasons I think it'd be beneficial to do this:

- It's quite a complex solution to the problem, I've spent a while reading the docs and trying to understand it, but I'm still honestly not 100% sure how it works/how to use it.
- Because of the complexity, it's harder for new contributors to write tests.
- The StateChecker can cause tests to pass inconsistently, depending on whether it's enabled or not.
- When you run a test and it fails to clean up after itself, it can be really hard to get back to the state where things started. This makes developing tests that mess with database records or files pretty difficult and error-prone.
- The StateChecker causes the tests to be much slower than they'd otherwise be, so people rarely ever actually run the tests themselves.
- Truncating the database leads to more consistent and clearer tests, because all dependencies and records that need to be created are done so explicitly, and it's generally what's done by most other test suites I've worked with ([Laravel does this](https://laravel.com/docs/5.8/database-testing#resetting-the-database-after-each-test), for example).

The biggest problems I foresee from making this change would be:

- The time investment for moving to database truncation.
- The existing tests being reliant on certain records from other tests being existent.
- Potentially slower tests, though I'm skeptical as to whether this would actually be slower than the existing system.
- Files and global variables - and potentially other state I'm not thinking of - wouldn't be handled by the database truncation. We'd need to replace the file state checking with something else, or I guess leave part of the statechecker intact to handle that aspect. For global variables at least, [this can be handled by PHPUnit itself](https://thephp.cc/news/2015/04/phpunit-4-6-and-global-state).
- Truncating the database can cause problems due to the fact that the CRM needs to be installed before it can be used for most of these tests. However, it _should_ be possible to use Fixtures to get to a 'known-good' starting state between each the test.

There may be things I don't know about the statechecker that are causing me to misunderstand things, and I'd be happy to be corrected if anything I said here was incorrect, but I hope this at least creates an interesting discussion to help us improve the test suite going forward.

Thanks :)

Dillon-Brown:
## Description

WIP PR for discussing/investigating alternatives to StateChecker/StateSaver. The goal of this change is to reduce instances of "flaky tests" while also reducing the complexity involved when writing unit tests (See #7740 for details). 

I've so far implemented two traits for truncating DB tables and for SQL transactions and rollbacks that will run at the start of every test. From my testing, it seems that truncating the DB would significantly increase the test time which may be a good solution for smaller projects but I don't think it would be viable here. I would instead propose we use transactions/rollbacks for DB operations while global variables can be handled by [PHPUnit](https://thephp.cc/news/2015/04/phpunit-4-6-and-global-state). We'll potentially need to implement a custom handler for file operations. 

## Motivation and Context

Issue reference: #7740 

## How To Test This

1. Check Travis
2. Run unit tests locally: ```./vendor/bin/phpunit --stop-on-failure --stop-on-error --colors --configuration ./tests/phpunit.xml.dist ./tests/unit/phpunit```

## Types of changes

### Final checklist

- [X] My code follows the code style of this project found [here](htps://docs.suitecrm.com/community/contributing-code/coding-standards/).


tests/unit/phpunit/data/SugarBeanTest.php
```diff
@@ -1856,6 +1844,9 @@ public function testUnPopulateDefaultValues()
         // test
         $GLOBALS['log']->reset();
         $GLOBALS['log']->fatal('test');
+        $this->db->query(/** @lang sql */
+            "INSERT INTO contacts (id, date_entered, date_modified, modified_user_id, created_by, description, deleted, assigned_user_id, salutation, first_name, last_name, title, photo, department, do_not_call, phone_home, phone_mobile, phone_work, phone_other, phone_fax, primary_address_street, primary_address_city, primary_address_state, primary_address_postalcode, primary_address_country, alt_address_street, alt_address_city, alt_address_state, alt_address_postalcode, alt_address_country, assistant, assistant_phone, lead_source, reports_to_id, birthdate, campaign_id, joomla_account_id, portal_account_disabled, portal_user_type) VALUES ('test_parent_contact_1', '2017-08-04 00:00:11', '2017-08-11 00:00:22', 'aaa', 'bbb', 'ccc', '0', 'eee', 'fff', 'ggg', 'hhh', 'jjj', 'kkk', 'lll', '1', 'mmm', 'nnn', 'ooo', 'ppp', NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, 'Single');"
```
connorshea: Could we improve the formatting on this query?

tests/unit/phpunit/include/MVC/Controller/SugarControllerTest.php
```diff
@@ -247,6 +221,6 @@ public function testcheckEntryPointRequiresAuth()
         $this->assertFalse($result);
         
         // clean up
```
connorshea: Remove this comment

tests/unit/phpunit/include/MVC/View/ViewFactoryTest.php
```diff
@@ -1,11 +1,13 @@
...
 
-class ViewFactoryTest extends SuiteCRM\StateCheckerPHPUnitTestCaseAbstract
+use SuiteCRM\Test\SuitePHPUnitFrameworkTestCase;
+
+class ViewFactoryTest extends SuitePHPUnitFrameworkTestCase
 {
     public function testloadView()
     {
-        $state = new SuiteCRM\StateSaver();
+
```
connorshea: Remove all this extra space.

connorshea: Actually, let's just make a note to clean up the `tests/unit/phpunit/include/` directory in a PR later.

tests/unit/phpunit/modules/AOR_Reports/AOR_ReportTest.php
```diff
@@ -30,31 +32,6 @@ public function testbean_implements()
...
-        $state->pushTable('aod_indexevent');
-        $state->pushTable('tracker');
-        $state->pushTable('aod_index');
-        $state->pushTable('aor_charts');
-        $state->pushTable('aor_fields');
-        $state->pushTable('aor_reports');
-        $state->pushFile('modules/AOD_Index/Index/Index/read.lock.file');
-        $state->pushFile('modules/AOD_Index/Index/Index/segments_3n');
-        $state->pushFile('modules/AOD_Index/Index/Index/read-lock-processing.lock.file');
-        $state->pushFile('modules/AOD_Index/Index/Index/write.lock.file');
```
connorshea: Worth noting the `pop`/`pushFiles` that this PR removes so we can go back and add whatever new solution we come up with.

Given that this means the tests now leave behind files, have you tried running the test suite more than once locally to see if it still works?

tests/unit/phpunit/modules/SchedulersJobs/SchedulersJobTest.php
```diff
@@ -267,18 +248,13 @@ public function testpostponeJobAndMark_deleted()
         $this->assertEquals(null, $result);
         
         // clean up
```
connorshea: A bunch of 'clean up' comments left in this file, pls fix :)

connorshea:
Oh also:

@Dillon-Brown

image

connorshea:
Should the `tests:unit` task and the PHPUnit command in CI be updated to use `--strict-global-state`? How much of a difference in test duration does that make?

connorshea:
It's weird that the PHPUnit tests took 6 and 7 minutes on PHP 5.5/5.6 :?

Is that because 7.3 is just faster or because there's a memory leak somewhere 🤔 The latest commit from hotfix-7.10.x only used 120MB of memory running the tests, but PHP 7.3 on this PR took 1300MB.

modules/MailMerge/Save.php
```diff
@@ -69,10 +69,9 @@
     $items[] = $seed;
 }
 
-$state = new \SuiteCRM\StateSaver();
-$state->pushPHPConfigOptions();
+$maxExecutionTime = ini_get('max_execution_time');
 
-ini_set('max_execution_time', 600);
+set_time_limit( 600);
```
connorshea:
```suggestion
set_time_limit(600);
```

include/Sugar_Smarty.php
```diff
@@ -147,22 +147,23 @@ public function fetch($resource_name, $cache_id = null, $compile_id = null, $dis
...
 
         if (!(isset($sugar_config['developerMode']) && $sugar_config['developerMode'])) {
             $level = isset($sugar_config['smarty_error_level']) ? $sugar_config['smarty_error_level'] : 0;
+            $errorLevelStored = error_reporting();
             error_reporting($level);
         }
         $fetch = parent::fetch(get_custom_file_if_exists($resource_name), $cache_id, $compile_id, $display);
-        $state->popErrorLevel('sugar_smarty_errors', 'error_reporting', false);
+
+        if (!(isset($sugar_config['developerMode']) && $sugar_config['developerMode'])) {
```
connorshea: err, does this ever actually return true?

tests/unit/phpunit/modules/AOR_Reports/AOR_ReportTest.php
Dillon-Brown: Ran a few dozens times without any issues (Although I imagine there might be some I just got lucky enough to miss).

Dillon-Brown:
Starting to run into some of those PHP 5 issues. I would have liked to get the tests running in a random order and in parallel if possible but it looks like this isn't availible with PHPUnit 4.8.....

Dillon-Brown:
> Should the `tests:unit` task and the PHPUnit command in CI be updated to use `--strict-global-state`? How much of a difference in test duration does that make?

Tried this out but looks like it always dies when trying to run a second test: https://travis-ci.org/Dillon-Brown/SuiteCRM/jobs/581197078

connorshea:
Needs a rebase. I think we should merge this ASAP if we can, it'd be a really nice improvement.

connorshea:
@Dillon-Brown this needs a rebase please :)

include/TestCaseAbstract.php
```diff
@@ -38,51 +38,58 @@
...
         }
-        
-        $this->beforeStateCheck();
+
+        if (self::$cleanupStrategy === 'transaction') {
+            $this->startDBTransaction();
+        } elseif (self::$cleanupStrategy === 'refresh') {
+            $this->refreshDatabase();
+        } else {
+            throw new Exception('Failed to truncate database');
```
connorshea:
```suggestion
            throw new Exception('Failed to cleanup database, invalid cleanup strategy specified.');
```

include/Imap.php
```diff
@@ -74,11 +74,8 @@ class Imap implements ImapInterface
      */
     public function open($mailbox, $username, $password, $options = 0, $n_retries = 0, array $params = null)
     {
-        $state = new StateSaver();
-        $state->pushErrorLevel();
-        error_reporting(0);
```
connorshea: 🤔 Should we keep this error reporting stuff?

tests/unit/phpunit/include/MVC/View/views/view.listTest.php
```diff
@@ -52,7 +46,7 @@ public function testlistViewProcess()
         
         
         
-        // clean up
+
         DBManagerFactory::getInstance()->query("DELETE FROM email_addresses");
```
connorshea: 🤔 should we remove this stuff? Is it just old stuff that was never removed when we added the state checker?

Dillon-Brown: Looks like a lot of the DB 
```sql
DELETE FROM
```
stuff was added after statechecker so ¯\\_(ツ)_/¯. We should probably try removing to see what happens, ideally shouldn't be necessary anymore.

include/Imap.php
Dillon-Brown: iirc by default the ```pushErrorLevel()``` function actually immediately logs not to use it and to remove ```error_reporting``` from your code. Imo we should remove it if we can.

connorshea:
So, I rebased this and managed to get it working :D

Unfortunately, it still took quite a bit of work and it's still pretty broken (lots of failures on the unit tests, I had to dump my dev database into the test table for it to not fail with errors about missing tables). But that's not really much different from how it was before.

```
Tests: 1388, Assertions: 1721, Errors: 838, Failures: 13, Skipped: 15, Incomplete: 128.
```

Then I realized I didn't do the `tests/config.test.php` correctly, it needs a _full_ configuration in it or it won't work right. Also, it seems to replace the original config.php when you run the tests. So that's not ideal. We should maybe fix that.

Now it's down to 47 errors/3 failures! :D

```
Tests: 1388, Assertions: 4088, Errors: 47, Failures: 3, Skipped: 16, Incomplete: 147.
```

A few of which are related to forgetting to remove the StateChecker from new tests, and a few others which are related to me probably screwing up the rebase on the InboundEmailTest file.

Branch is here if it's of any interest :P

https://github.com/connorshea/SuiteCRM/tree/connors-terrible-state-checker-rebase

connorshea:
@Dillon-Brown the composer.lock needs an update :)

cameronblaikie:
Assessed :+1:

tests/unit/phpunit/modules/ACLActions/ACLActionTest.php
```diff
@@ -117,37 +103,27 @@ public function testhasAccess()
...
         self::markTestIncomplete('Need to fix checking user access. Hint: session is a system state perhaps its failing because the user session');
-        
-        $state = new SuiteCRM\StateSaver();
-        $state->pushGlobals();
-        
+
         $this->assertFalse(ACLAction::userHasAccess('', '', '')); //test with empty module and action
-        $this->assertTrue(ACLAction::userHasAccess('', 'Accounts', 'list')); //test with e,pty user and valid module and action
+        $this->assertTrue(ACLAction::userHasAccess('', 'Accounts',
+            'list')); //test with e,pty user and valid module and action
```
craigpanton: Small typo in comment

Dillon-Brown: Fixed :D.

connorshea:
LGTM 👍

connorshea:
Okay, one problem I've noticed from this PR: The `config.php` gets overridden with data from the `config.test.php` file. This doesn't happen for the `hotfix-7.10.x` branch. I've narrowed it down to being caused by something in the `InboundEmailTest.php` file.

Though, it only seems to copy it partially? Not really sure what's up with that.

tests/unit/phpunit/modules/InboundEmail/InboundEmailTest.php
```diff
@@ -22,51 +24,15 @@ function tempFileWithMode($mode)
...
-        $state = new SuiteCRM\StateSaver();
-        $state->pushTable('inbound_email_cache_ts');
-        $state->pushTable('inbound_email_autoreply');
-        $state->pushTable('inbound_email');
-        $state->pushTable('aod_index');
-        $state->pushTable('folders');
-        $state->pushTable('folders_subscriptions');
-        $state->pushTable('config');
-        $state->pushTable('email_cache');
-        $state->pushFile('config.php');
```
connorshea: This is probably the source of the problem, InboundEmailTest.php messes with the config.php.

Can we work around this somehow? Ideally, it wouldn't mess with the `config.php` at all because it would also screw it up when it fails.

We can fix this after merging this PR if we want to, I don't think it should be a blocker.

connorshea: I'm fairly certain it's the `saveInboundEmailSystemSettings` test specifically.
https://github.com/salesagility/SuiteCRM/blob/c9d2132df8db5955a0809301dacb9fb5db872b6c/modules/InboundEmail/InboundEmail.php#L5969-L5986

connorshea: Looking at the other uses of pushFile/popFile – besides the AOD Index which I'm not sure about – I don't think they actually need to be kept. TaskTest pops/pushes the config.php for no obvious reason, and CurrentLanguageTest does the same thing with the config_override.php, again for no apparent reason. So this is the only one that's really a problem.

connorshea:
👀

Mac-Rae:
Assessed :+1:

connorshea:
@Dillon-Brown @Mac-Rae don't forget to merge the Docs PR :) https://github.com/salesagility/SuiteDocs/pull/346

PR #7808, opened 2019-09-04
Fixed #7621 - Add support for config_override.test.php

### LINKED ISSUE #7621: Add support for config_override.test.php
#### Issue
This is part of #7344, but I'm splitting it out into a separate issue.

Add support for a `config_override.test.php` file, which would be used to define configuration values for the test environment specifically (e.g. so you can run the tests on a separate database from the development site).

There are certain config options you'd want to use specifically for the test environment, like `imap_test` or statechecker-related options. It'd be nice to have a `config_override.test.php` which would only be used when the tests are being run, to allow separating the development and test environments.

I'm not sure if the `config_override.php` should still be loaded in the test environment, I guess it should be? It'd load `config.php`, then `config_override.php` which would override any settings in `config.php`, then `config_override.test.php` which would override any settings in `config_override.php`. When not running the tests, the `config_override.test.php` would be ignored.

[Rails comes with separate environments for development, test, and production](https://guides.rubyonrails.org/configuring.html#creating-rails-environments), for example. Laravel works the same way. I think it's a pretty smart pattern, and makes it easier to write code and run tests at the same time, without one messing with the other.

Dillon-Brown:
## Motivation and Context

Issue reference: #7621 

## How To Test This

1. Check Travis
2. Add a ```tests/config.test.php``` file and then give it a custom value like so: 
```
$testConfig = [
    'imap_test' => true,
];
```
3. Check that the custom config only runs during the test suite execution and overrides the ```config.php``` and ```config_override.php``` values.

## Types of changes

### Final checklist

### DIFF
--- a/lib/Robo/Traits/CliRunnerTrait.php
+++ b/lib/Robo/Traits/CliRunnerTrait.php
@@ -65,12 +65,26 @@ trait CliRunnerTrait
          * since they have side effects and can make other
          * Robo Tasks fail (i.e. failed database connection).
          */
-
         $root = __DIR__ . '/../../../';
 
         require $root . 'config.php';
         require $root . 'config_override.php';
         require_once $root . 'include/entryPoint.php';
+
+        // Load up the config.test.php file. This is used to define configuration values for the test environment.
+        $testConfig = [];
+
+        if (is_file($root . 'tests/config.test.php')) {
+            require_once $root . 'tests/config.test.php';
+        }
+
+        foreach (array_keys($testConfig) as $key) {
+            if (isset($sugar_config[$key])) {
+                $sugar_config[$key] = $testConfig[$key];
+            } else {
+                $sugar_config[] = $testConfig[$key];
+            }
+        }
 
         $current_language = 'en_us';
         $app_list_strings = return_app_list_strings_language($current_language);
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ install/status.json
 /config.php
 /config_override.php
 /config_si.php
+tests/config.test.php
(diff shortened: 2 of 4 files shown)


connorshea:
We should probably have a `config.test.example.php` or `config.test.dist.php` file in the repo for users to copy from. :)

Just this should be fine? Maybe a comment explaining what it's for.

```php
<?php
$testConfig = [
    'imap_test' => true,
];
```

This'll be great, thank you! ❤️

cameronblaikie:
Assessed :+1:

cameronblaikie:
Needs input from @Dillon-Brown before merging :D.

cameronblaikie:
Assessed :+1:

lib/Robo/Traits/CliRunnerTrait.php
```diff
@@ -65,13 +65,27 @@ protected function bootstrap()
...
         $root = __DIR__ . '/../../../';
 
         require $root . 'config.php';
         require $root . 'config_override.php';
         require_once $root . 'include/entryPoint.php';
 
+        // Load up the config.test.php file. This is used to define configuration values for the test environment.
+        $testConfig = [];
+
+        if (is_file($root . 'tests/config.test.php')) {
```
connorshea: it occurs to me now, wouldn't this effect all the Robo tasks that use CliRunnerTrait?

connorshea: Also: when I run `./vendor/bin/robo tests:unit`, at least on the StateChecker-removal branch, the `config.php` file is overwritten by the contents of the `config.test.php`.

PR #7834, opened 2019-09-09
Workflow: Properly delete records which are marked as deleted

lazka:
## Description

There is currently no action for deleting a record, only for modifying one.

The only way atm is to set the "deleted" field during the workflow action which
seems to work but leaves tangling relations in the DB and not all code handles that
properly, leading to weird errors/states and a bloated DB

Improves the current way by checking if the delete field was set during the action
and properly delete the record after the action is done, including relations etc.

## Motivation and Context

Searching for how to delete with a workflow leads to descriptions pointing to set the deleted field, but that's not working right currently.

## How To Test This

* Delete a record with a workflow, all relations should be marked as deleted as well.

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOW_Actions/actions/actionCreateRecord.php
+++ b/modules/AOW_Actions/actions/actionCreateRecord.php
@@ -380,7 +380,20 @@ class actionCreateRecord extends actionBase
         $record->process_save_dates =false;
         $record->new_with_id = false;
 
+        /* Since we only work on non-deleted records this means the delete field
+         * was set during this action.
+         * Complete the deletion process by calling mark_deleted() after save() */
+        $was_deleted = false;
+        if ($record->deleted) {
+            $record->deleted = 0;
+            $was_deleted = true;
+        }
+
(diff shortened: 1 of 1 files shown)


lazka:
(@holdusback ran into this problem and mentioned it on gitter)

Mac-Rae:
Could you link us to the issue on the forums as I was able to replicate this working fine?

pgorod:
@Mac-Rae I think we discussed this on Gitter, Sep 5.

But if I understand correctly, the point is: the record gets deleted ok, you can only see a problem if you go in the database and look for orphaned relationships and orphaned custom table entries.

With this PR these deletes are handled more completely.

holdusback:
Had this issue cause of one workflow :
On every saved lead, check if the statut equal to converted
If converted -> Delete the Lead to prevent duplicate.
image


@lazka told me that this way doesn't actually delete everything. Realtion in the DB are not deleted (wich is not really healthy.)

PR #7855, opened 2019-09-12
FIXED #7613 - Status/State usage causing translation errors (State/Region)

### LINKED ISSUE #7613: Status/State usage causing translation errors
#### Issue

SuiteCRM language files uses "State" as a constituent political entity (in the US there 50 "States" divided into counties or county-equivalents). But STATE is also used as situation/condition State. 
On top of that "Status" is also used as condition. 
But for many languages all 3 could be translated with the same word!
This causes issues when in single word language values, as there is no context, making it impossible for translators to make a proper job. But it also may cause issues for devs using STATE strings in future.
Note: I'm not talking on the strings that uses "Address State" as this shows an obvious use, just the single words as they do not give any context.

#### Actual Behavior

Values used as a single word:

- 6 strings with "State" used for situation, condition. Examples:  `'LBL_STATE' => 'State:',` or `'LBL_LIST_STATE' => 'State'`, or `'LBL_AN_SERVICE_STATE' => 'State',`
- 13 strings for "State" used in address fields for a Region. Examples: `'LBL_STATE' => 'State:',` or `'LBL_LIST_STATE' => 'State'`,
- 73 strings for "Status". Examples: `'LBL_STATUS' => 'Status:'` or `'LBL_LIST_STATUS' => 'State'`,

#### Possible Fix

"Easy" fix: to change "State" (when political entity) values in language files into a new value. 
As Regional political entity/divisions are different in each countries (some don't even have any) no single word seems a proper one to also be used in english.
This "State" should be considered as the main regional administrative division. It could be regions, districts or provinces, and can be different in different countries. https://en.wikipedia.org/wiki/District

So my proposal is to use a value `State or Region`. Example:
`    'LBL_STATE' => 'State or Region:', //For address fields`

Then each translator can pick a word suitable for their own country/language equivalent political entity.
A proper fix would also require to also change the key part of the language string so there would not be errors from the devs when using/creating a string for STATE.

#### Context

In Crowdin I already changed this a long time ago but this requires me to keep a local non synchronized copy from Github to Crowdin with my local edits and this a time consuming when updating source files.

### Additional note: 
You can also find "Status" and "State" in the same screen, and none are used as a political entity, so its not possible to change all non political "State" for "Status" occurrences:
state_status

State as address
address

cameronblaikie:
using "State/Region" for address fields
## Motivation and Context

https://github.com/salesagility/SuiteCRM/issues/7613

## How To Test This

Go to any form that has a input box for adding which State your address is and it should say "State/Region" now.
## Types of changes

### Final checklist


Mac-Rae:
Assessed :+1:

horus68:
@cameronblaikie  So no `LBL_LIST_STATE` changed as requested in the my initial issue?
They are required too!

PR #5425, opened 2018-03-07
refactoring email template parser

zoltankocsardi:
## Description
Small refactoring for parsing email template

## How To Test This
- create a survey campaign and see, if the variables parsed

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [x] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

PR #5337, opened 2018-02-26
fix survey variables with new parser

zoltankocsardi:
## Description
Survey variables were not parsed properly.

## How To Test This
- create a new survey campaign
- add some survey variables to the template
- send test email
- see if the variables replaced

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [x] My code follows the code style of this project found [here](https://suitecrm.com/wiki/index.php/Coding_Standards).
- [ ] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://suitecrm.com/wiki/index.php/Contributing_to_SuiteCRM) guidelines.

PR #7866, opened 2019-09-16
Fixed #6535 - Replace contact_xxx in templates also for leads/prospects/users

### LINKED ISSUE #6535: Email Template Not Replacing Contact/Lead/Target Name Variable in Leads
#### Issue

I do see there are other issues with emails already listed #3829 #3321 #3310 #6462 but don't seem to be exactly the same but may be related.

When creating an email template and then inserting Contact/Lead/Target variable for $contact_first_name, for example if you send it to a lead it does not work (it worked in previous versions) it just leaves the variable there.  However, this works fine if you send it to a contact and replaces the variable with the first name no problem.  If I use the variable $lead_first_name and send it to a lead it work no problem.  However, using the Contact/Lead/Target variable is supposed to grab the right first name regardless of the record.

I should mention, it works as part of a workflow and subs in the right variable, just when you select a single lead to send an email to, it doesn't populate the value.

#### Expected Behavior

Using the Contact/Lead/Target variable is supposed to grab the right first name regardless of the record when used in a template
#### Actual Behavior

When creating an email template and then inserting Contact/Lead/Target variable for $contact_first_name, for example if you send it to a lead it does not work (it worked in previous versions) it just leaves the variable there as $contact_first_name when sending to the client.

I've checked error.log and there are no errors there to report.

#### Steps to Reproduce

1. Create an email template 
2. Add field Contact/Lead/Target (any field) $contact_first_name for example
3. Send the email to a Lead (and relate the email to the lead before sending)
4. Doesn't populate the lead's first name in email template

#### Context

This should be a high priority.  Sending out emails to people that you can't personalize kind of defeats the purpose of a CRM!  Looks like there are lots of bugs around sending emails and variables.  Would be great to get this fixed.

#### Your Environment

* SuiteCRM Version used:  7.10.10
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Chrome 70.0.3538.102
* Environment name and version (e.g. MySQL, PHP 7):  PHP 7.1
* Operating System and version (e.g Ubuntu 16.04): Apache

lazka:
## Description

This was removed in #5337, reverted in #5397 and then re-removed in #5425.

The original template code does more things, but this should at least make
the basics work again and adds some tests.

Fixes #6535

## Motivation and Context

Sending emails with templates to leads

## How To Test This

* Create a campaign
* Add a lead to the test list
* Include a lead variable (like name)
* Send a test email
* The variable should be replaced

## Types of changes

### Final checklist

### DIFF
--- a/modules/EmailTemplates/EmailTemplateParser.php
+++ b/modules/EmailTemplates/EmailTemplateParser.php
@@ -174,6 +174,18 @@ class EmailTemplateParser
 
         $parts = explode($charUnderscore, ltrim($variable, $charVariable));
         list($moduleName, $attribute) = [array_shift($parts), join($charUnderscore, $parts)];
+
+        /* Leads/Prospects/Users have a special variable naming scheme.
+        $contact_xxx for leads/prospects and $contact_user_xxx for users */
+        if (strtolower($moduleName) === 'contact') {
+            if (in_array($this->module->object_name, ['Lead', 'Prospect'], true)) {
+                $moduleName = strtolower($this->module->object_name);
+            } else if ($this->module->object_name == 'User' && str_begin(strtolower($attribute), 'user_')) {
+                $attribute = explode('_', $attribute, 2)[1];
+                $moduleName = 'user';
+            }
+        }
+
         if (in_array($attribute, static::$allowedVariables, true)) {
             return $this->getNonDBVariableValue($attribute);
         }
--- a/tests/unit/phpunit/modules/EmailTemplates/EmailTemplateTest.php
+++ b/tests/unit/phpunit/modules/EmailTemplates/EmailTemplateTest.php
@@ -1,5 +1,7 @@
 <?php
 
+require_once 'modules/EmailTemplates/EmailTemplateParser.php';
+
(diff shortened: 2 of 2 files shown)


modules/EmailTemplates/EmailTemplateParser.php
```diff
@@ -174,6 +174,18 @@ private function getValueFromBean($variable)
 
         $parts = explode($charUnderscore, ltrim($variable, $charVariable));
         list($moduleName, $attribute) = [array_shift($parts), join($charUnderscore, $parts)];
+
+        /* Leads/Prospects/Users have a special variable naming scheme.
```
Dillon-Brown: Removed, reverted, re-removed and now re-added again :laughing:. I'd potentially want to investigate this further to see exactly why this was done and what the differences are between this and the old parser but this looks good for now since you've added in unit tests.

lazka: I don't understand it either why a new parser was added instead of fixing the old one. There could have been a good reason for it of course, not sure :)

Issue #7849, opened 2019-09-11
Integration with Sentry error tracking


connorshea:
I wanted to propose integrating the CRM with support for reporting errors to a [Sentry](https://sentry.io/) instance.

I use Sentry for a personal project, and we used it for GitLab pretty extensively. GitLab comes with error reporting to Sentry included by default (though not enabled by default, obviously). Sentry's an open source error tracking product that you can either self-host or use as 'SaaS', it's written in Python and the devs are quite good from my experience interacting with them.

Essentially, what I'm proposing is having an "Error Tracking" page in the Admin section, that would let you activate error reporting to Sentry by providing the "Error Tracking" page with a Sentry URL, Sentry DSN, and Sentry Key. The CRM would ship with the Sentry PHP package, although it'd be disabled by default, and it would only report errors to a Sentry instance once it's been configured in the Admin panel. (or maybe the configuration should be done in `config.php`?)

The reason I want this is to enable us to track errors more easily, because right now we have to go through the php error logs manually to look for any errors that seem like they could be problematic. And once we do find an error in the logs, there's not much information to debug it with, and it lacks a lot of potentially important context (e.g. the page the user was on, the user that ran into the error, the # of users who have hit the error, the # of times the error has occurred, etc.).

Unfortunately, the latest iteration of the Sentry PHP client requires PHP 7.1 and higher, so we would have to use the 1.x client: https://docs.sentry.io/clients/php/

I don't know how long they intend to support the 1.x client, but hopefully it'll be a while. Eventually, we should obviously upgrade to 2.x.

cc: @Abuelodelanada I know you mentioned trying to integrate with Sentry before.

Abuelodelanada:
Hi @connorshea 

I did it about 7/8 years ago. I think Sentry it's different now. Anyway it is a wonderful tool!
Having an Admin section to setup will be great. All the things you setup in Admin will be written in `config_override.php` so it is almost the same than using `config.php`

Dillon-Brown:
Very cool tool, I definitely want to play around with this some more.


 ![sentry](https://user-images.githubusercontent.com/26431166/64993317-a6369100-d8cd-11e9-9229-da11f6c82fad.png)

PR #7873, opened 2019-09-17
Feature/sentry

Dillon-Brown:
## Description

Allows SuiteCRM exception handling to utilize [Sentry](https://sentry.io).

## Motivation and Context

Issue reference: #7849 

## How To Test This

1. Register a new application on https://sentry.io
2. Set the following config values:

```
    'sentry' => [
        'enabled' => true,
        'dsn' => '*Get this from sentry',
    ],
```
3. Check that the SuiteCRM exceptions and error logging goes to sentry.

## Types of changes

### Final checklist


Sentry.php
```diff
@@ -0,0 +1,134 @@
...
+        return self::$instance = new self();
+    }
+
+    public function __construct()
+    {
+        global $sugar_config;
+
+        try {
+            if (!isset($sugar_config['sentry']['dsn'])) {
+                throw new \SuiteCRM\Exception\Exception('You must configure Sentry DSN.');
```
connorshea:
```suggestion
                throw new \SuiteCRM\Exception\Exception('You must configure a Sentry DSN in config.php.');
```

Sentry.php
```diff
@@ -0,0 +1,134 @@
...
+        return new Raven_Client(self::$dsn);
+    }
+
+    protected function init()
+    {
+        global $current_user;
+
+        if (self::$enabled) {
+            $user_context = $current_user->id;
+            self::$client->user_context($user_context);
```
connorshea: If there's no user logged in, what does `$current_user->id` return? Just `null`?

connorshea: Can we provide the user's name and/or email as well? Just having the ID doesn't tell us that much.

connorshea: And also the PHP Version and SuiteCRM version, if possible 😇

Sentry.php
```diff
@@ -0,0 +1,134 @@
...
+
+        self::$enabled = $sugar_config['sentry']['enabled'];
+        self::$dsn = $sugar_config['sentry']['dsn'];
+        self::$client = $this->buildClient();
+        $this->init();
+    }
+
+    protected function buildClient()
+    {
+        return new Raven_Client(self::$dsn);
```
connorshea: If you only pass the DSN does it automatically know what project it's going to be sent to?

The docs say to use a format like `new Raven_Client('https://<key>@sentry.io/<project>');`

I think it'd be ideal if we could also configure it to use a custom domain, for people who want to self-host Sentry.

Sentry.php
```diff
@@ -0,0 +1,134 @@
...
+
+        return self::$instance = new self();
+    }
+
+    public function __construct()
+    {
+        global $sugar_config;
+
+        try {
+            if (!isset($sugar_config['sentry']['dsn'])) {
```
connorshea: Throwing an error here is probably unnecessary if it's not enabled? If it's not enabled, it should just be silent.

include/MVC/Controller/SugarController.php
```diff
@@ -397,15 +398,22 @@ protected function showException(Exception $e)
...
         if (isset($this->bean)) {
             $logicHook->setBean($this->bean);
-            $logicHook->call_custom_logic($this->bean->module_dir, "handle_exception", $e);
+            $logicHook->call_custom_logic($this->bean->module_dir, 'handle_exception', $e);
         } else {
-            $logicHook->call_custom_logic('', "handle_exception", $e);
+            $logicHook->call_custom_logic('', 'handle_exception', $e);
+        }
+
+        if (!empty($sugar_config['sentry']['enabled'])) {
```
connorshea: Does this cause a problem if there's no `'sentry'` key in the config?

include/SugarLogger/LoggerManager.php
```diff
@@ -119,6 +121,12 @@ public function __call(
             }
             //tell the logger to log the message
             self::$_loggers[$logger]->log($method, $message);
+
+            // Log to sentry if enabled
+            if (!empty($sugar_config['sentry']['enabled'])) {
```
connorshea: It might be useful to pass the log message's `level` along here, it seems to be supported by the client: https://docs.sentry.io/clients/php/usage/

Sentry.php
connorshea: As an escape hatch in case the user wants to customize something without waiting for a SuiteCRM update to add it, we should probably have a `$sugar_config['sentry']['extra_config']` array that'd be passed as the second argument to Raven_Client.

https://docs.sentry.io/clients/php/config/

lazka: The dsn includes the host it sends to, so on your own sentry it would be `https://<key>@sentry.yourcompany.com/<project>`

Abuelodelanada:
@Dillon-Brown Having integration between SuiteCRM and Sentry is awesome! Great PR!

Sentry.php
connorshea: We should make sure to mention that in the docs :)

Sentry.php
Dillon-Brown: I was thinking of a case where you had ```['sentry']['enabled']``` set to true but hadn't set up the ```[dsn]```

Sentry.php
connorshea: Okay, yeah this is a problem because `$current_user->id` can return an error if `$current_user` is null.

```
ERROR: Trying to get property 'id' of non-object 
in /Users/connorshea/code/sites/suitecrm-upstream/Sentry.php:85
```

include/SugarLogger/LoggerManager.php
connorshea: It does actually already know the error level, kind of. It doesn't quite match the configured log levels from the CRM, but 🤷‍♂

connorshea:
If you log something like `$GLOBALS['log']->fatal("Test");` it reports it to Sentry as an error >.< I'm not sure if that's intentional?

Sentry.php
connorshea: Does the constructor only get called if `enabled` is true?

Dillon-Brown:
Thanks for the code-review @connorshea! took a quick look and I agree with pretty much everything you've pointed out. While this won't make the next minor release, I'd ideally aim to have it merged in right after.

connorshea:
hmm, I get `Trying to get property 'id' of non-object` whenever I see an error even though I'm logged in. Are we sure `$current_user` will always exist?

Sentry.php
connorshea: I think we should actually be handling this in the `handleException` and `captureMessage` functions, rather than in the install step, since it seems like the install hook doesn't actually have a current_user.

Sentry.php
```diff
@@ -0,0 +1,134 @@
...
+
+    private function __construct()
+    {
+        global $sugar_config;
+
+        try {
+            if (!isset($sugar_config['sentry']['dsn'])) {
+                throw new \SuiteCRM\Exception\Exception('You must configure a Sentry DSN in config.php.');
+            }
+        } catch (\SuiteCRM\Exception\Exception$exception) {
```
connorshea:
```suggestion
        } catch (\SuiteCRM\Exception\Exception $exception) {
```

connorshea:
I've opened https://github.com/Dillon-Brown/SuiteCRM/pull/143 to resolve my `extra_config` comment (it's called `raven_client_config` instead) and also add the SuiteCRM version to the error context.

So I have this in my config:
```
$sugar_config['sentry']['raven_client_config'] = array(
  'release' => '1234abc',
  'environment' => 'development'
);
```

and this is what it looks like in Sentry:

Screen Shot 2019-09-30 at 12 37 04 PM

I imagine you could automatically set the release SHA in your `config_override.php` every time you deploy a new version of the CRM, that's what I intend to do at least.

There are a bunch of things you can do with the Raven Client's config, e.g. set extra tags or exclude certain exception types: https://docs.sentry.io/clients/php/config/#available-settings

connorshea:
Screen Shot 2019-09-30 at 12 47 29 PM

The Sentry section in System Settings looks a bit wonky, we probably need to change the HTML a bit? (or I just haven't run the CSS repair recently).

modules/Configurator/tpls/EditView.tpl
```diff
@@ -372,15 +372,29 @@
...
+				{assign var='sentry_checked' value=''}
+			{/if}
 
-<table  width="100%" border="0" cellspacing="1" cellpadding="0" class="edit view">
-{if $logger_visible}
-<tr>
-<th align="left" scope="row" colspan="6"><h4>{$MOD.LBL_LOGGER}</h4></th>
-</tr>
-	<tr>
+		<td><input type='hidden' name='sentry_enabled' value='false'><input name='sentry_enabled' type="checkbox"
```
connorshea: I think the formatting here got screwed up?

Dillon-Brown: Yeah, didn't actually check what this looked like or even if it worked, just threw it into the draft so I wouldn't forget to make this configurable somewhere in the front-end :P.

include/entryPoint.php
```diff
@@ -84,12 +84,18 @@
...
 $GLOBALS['sugar_config'] = $sugar_config;
 require_once 'include/SugarObjects/SugarConfig.php';
+require_once 'include/SugarLogger/LoggerManager.php';
 
 if (!empty($sugar_config['xhprof_config'])) {
     require_once 'include/SugarXHprof/SugarXHprof.php';
     SugarXHprof::getInstance()->start();
 }
 
+if (!empty($sugar_config['sentry']['enabled'])) {
```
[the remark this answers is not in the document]
Dillon-Brown: Agreed

Sentry.php
Dillon-Brown: It should, although I think I had set it just to check if ```['sentry']['enabled']``` exists which would need to be changed.

pgorod:
Some work will probably be required to integrate this PR with #8640.

pgorod:
@Dillon-Brown I hope everything is well with you! Congrats on the Suite8 Beta release!

I'd like to ask you what is the meaning of all these closed PR's? It's not very obvious from the Github UI what happens with the force-push. Are these merged and part of the next 7.x release?

Dillon-Brown:
Hey @pgorod, the couple of PRs I've closed recently are ones that I think are unlikely to be included anytime soon, need significant work, or are invalid in some way. This one I closed because it would need to be looked at as a develop branch feature. I'm still keeping track of them and may go back and re-create them if I think it will bring value in the future. At the moment @jack7anderson7 has been looking through the list of PRs requiring assessment so I thought it'd be easier if I cleaned up some of my own.

In terms of force pushing, everything should be visible as a PR with "merged" status. Nothing should get force pushed into the branch. What we do for releases is merge hotfix-7.10.x -> hotfix and then hotfix -> hotfix-7.10.x and finally hotfix -> develop. This ensures that all fixes end up in all branches and you don't need to have duplicate PRs.

For Suite-8, the current ```next``` branch you can see on the SuiteCRM repo is the version of Suite-7 that Suite-8 is going to use. How this works is that we do Suite-7 commits to our internal Suite-8 repo and those get pushed out to the Suite-7 ```next``` branch which we also rebase onto the latest Suite-7 ```develop``` to ensure any fixes done are brought across.

PR #3158, opened 2017-02-19
Fix #2187 Update click-to-call to work from current browsers.

chris001:
Fix for #2187  Replaced `callto:` with `tel:` in compliance with the RFC.

## Motivation and Context

Skype started the `callto:` standard for click to call.  That's now outdated.  New standard is `tel:`.

## How To Test This

1. Open mobile smart phone or tablet, and open the browser app.
2. Log into SuiteCRM.
3. Browse to a Contact.
5. Tap the phone number for the Contact.
6. Your mobile device dials the Contact's phone number with the built in dialer app.
Previously, the application was generating an old style `callto:` link which when clicked fails to dial the Contact's phone number on current browsers.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [x] My code follows the code style of this project found [here](https://suitecrm.com/wiki/index.php/Coding_Standards).
- [ ] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://suitecrm.com/wiki/index.php/Contributing_to_SuiteCRM) guidelines.

Issue #2187, opened 2016-09-09
Feature Request - "tel:" encoding for click-to-call on mobile


sieberta:
#### Suggestion

It would be nice if you could click a number to call it.
#### Expected Behavior

Phone numbers should be formatted as "tel:" links.
#### Actual Behavior

They aren't.
#### Possible Fix

[https://suitecrm.com/forum/developer-help/6536-skypeout-callto-tel-code-location](https://suitecrm.com/forum/developer-help/6536-skypeout-callto-tel-code-location)
#### Context

It doesn't effect me at all, because I've already deployed this personally.

pgorod:
This is not really a suggestion for improvement, but rather a bug, since `callto:` has been discarded by the [RFC 3966](https://tools.ietf.org/html/rfc3966) (from 2004) explaining that 

> URI schemes locate a resource and do not specify an action to be taken

So a change to `tel:` seems to have only positive consequences. The most significant being that your cell phone can call them when you click, something quite relevant in 2017.

Skypeout can always pick up those links also, if you tell your computer to register that application as responsible for handling `tel:` URIs.

This would be a nice time to drop the `Skypeout` name from the `Admin-->System Settings` screen. It is just one of the many programs that can handle phone calls. I suggest to change that setting into something more generic like `Allow clicks on phone numbers to start calls`.

PR #7879, opened 2019-09-17
Render phone fields as links

### LINKED ISSUE #7876: Generate tel: links for phone numbers
#### Issue
Phone fields in the CRM aren't rendered as `tel:` links, which they should be to allow users to easily call leads, accounts, etc.

Screen Shot 2019-09-17 at 1 01 26 PM

#### Expected Behavior
The phone fields should be clickable `tel:` links.

#### Actual Behavior
The phone fields do not render as links.

#### Possible Fix
Fix the phone fields returning without links.

#### Steps to Reproduce
Note: This is reproducible on the demo instance
1. Go to a Lead, Account, or Contact (other modules probably have the same problem as well) that has a phone number.
2. Try to click the phone number to call it.
3. It does not work, because the phone number is not a `tel:` link, or a link at all.

#### Context
We have to have users install a 'telephone linker' extension for Chrome in order to make calls easily, which is pretty annoying. This should work out-of-the-box.

It seems like this commit was supposed to fix it, but I guess the sugar_phone function isn't used or something? https://github.com/salesagility/SuiteCRM/commit/02b7f9099b0ac8298067e5dfdb77ac5a90924043

#### Your Environment
* SuiteCRM Version used: 7.10.19, 7.11.8
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Firefox 69
* Operating System and version (e.g Ubuntu 16.04): macOS Mojave 10.14.6

connorshea:
## Description
Fixes #7876. This renders phone fields as links.

Before:

Screen Shot 2019-09-17 at 2 46 23 PM

After:
Screen Shot 2019-09-17 at 2 40 49 PM

This also includes a minor change to include `phone` classes in phone fields, since the code for doing that existed but wasn't actually functional. I'm honestly not sure why it was done that way, with a special case only for phone fields, but 🤷‍♂ 

## Motivation and Context
Being able to open phone numbers as links is really useful for easily making calls, etc.

## How To Test This
1. Go to a Lead, Contact, Account, etc, make sure they have a phone number associated, or add one yourself.
2. See that the phone number field is now a link.
3. Make sure an empty phone field doesn't have a `tel:` link.
4. Make sure other fields haven't become phone links. 

## Types of changes

### Final checklist


pgorod:
This is interesting, thanks. We do already have `tel:` links somewhere, I remember a discussion about that years ago. Isn't it applying everywhere?

An issue I have with this is that if it depends too much on phone formats, it won't work for most people. Not every country writes phone numbers in American format. Does this require a format to work? For example, does it require parenthesis?

pgorod:
Here it is: 

Issue #2187
PR #3158

Mac-Rae:
Assessed :+1:

pgorod:
Hmm I'm not so sure about this, no... at least I'd like to hear more from Connor on this one.

The previous code added the `tel` link when all these conditions were true:
1. In Admin / System setting this feature is turned on
1. There is a value present to display
1. That value has a format such that it can be used to start a phone call
Then the created link would go through another function in order to format it correctly: 

https://github.com/salesagility/SuiteCRM/blob/master/include/utils.php#L3022-L3025

Somebody, at some point, must have thought that was useful ;-)

I do think that there's probably a bug in there somewhere, I've seen people on the Forums complaining that this isn't working.

Now what we're doing with this PR is: 

- if we have a value, whatever the format, show it, just as it is

**So my questions are:**

Do we really want to ignore the System setting? Why?

What is the rationale behind dropping the `skype_formatted` and `format_skype` functions?

(Note that this is not about Skype as a specific product; we removed that from the UI, but it lives on in the code. It is meant to be generic for any phone app.)

connorshea:
@pgorod I don't see a good reason not to ignore the system setting, just because "someone implemented it like that at one point" doesn't mean we should just go along with it forever :) I'd like to know the logic behind not providing `tel:` links by default in the first place, but I don't think that we can find that information easily (at least, I wasn't able to when working on this) and I don't think that should be an excuse not to make this improvement.

As for why this drops the 'skype formatting', for one it doesn't drop it, it just doesn't use it if the setting isn't enabled. For two, I just don't fully understand the code for the skype-related stuff. I _think_ it only tries to run `skype_format` when the number is in 'skype format' (starts with a `+`, `00`, or `011`, I think is what it's checking for?), but that seems really arbitrary to me.

The previous code, as far as I can tell, wouldn't render perfectly valid phone numbers as links (e.g. `720-123-1234` wouldn't get caught as a phone number, which I guess it isn't unless you assume a default of American phone numbers, but that's what most Americans do).

I don't see a particularly good reason not to have the phone field render as a link by default, honestly. Even if you click it accidentally, modern browsers tend to prompt you to make sure you're wanting to make a call, and ask what application to do it in.

pgorod:
Ok, I am just asking, I actually lean to agreeing with you, we could just "liberalize" this and nobody would care.

I guess the only reason to leave the option would be if in some company, having a link that doesn't work (because no phone app is installed) would be deemed too confusing or inadequate for users. But this should be pretty rare, I guess.

Maybe we can have more opinions here (@samus-aran ?)

If we do make the change, then we should probably make a cleaner cut - remove the unnecessary functions, and remove the System setting altogether.

I like the fact that more formats are allowed (as I said, the american format restriction didn't work well in my country). But I wonder if for some apps we wouldn't need to prepare the numbers before sending them to the phone app.

Is that custom smarty function customizable? I mean, can we override it in `custom` dir?

Dillon-Brown:
Code looks good, will mark this as an enhancement for now.

connorshea:
@pgorod it can indeed be customized by creating a file at `custom/include/Smarty/plugins/function.sugar_phone.php`. At least, it can on 7.10.19. I'm pretty sure the same is true of 7.10.20.

connorshea:
The `skype_formatted` and `format_skype` methods in `utils.php` are no longer used, but I guess removing them would be a breaking change? Should I mark them as deprecated?

connorshea:
I've just marked the functions as `@deprecated` in 1760f0d, but I think #7967 would probably be a better long-term solution :)

connorshea:
@pgorod Does this look good to you?

pgorod:
@connorshea LKOGETM (looks kind of good enough to me)

but since this is marked as Enhancement I think it requires further approvals, not just from me.

HVStechnik:
👍 looking forward to this enhancement

connorshea:
@Dillon-Brown can this be merged? :)

Mac-Rae:
Assessed :+1:

PR #7894, opened 2019-09-20
add setting for an large database, e.g. 2000000 contacts, for an faster list queue query

andre-karrlein:
## Description
We had a problem with our large database, that the contact list view load over 1 minute.
With this change it loads in several seconds.

## Motivation and Context
We had a problem with our large database, that the contact list view load over 1 minute.
The db query needs more than 40 seconds.

## How To Test This
Open the contact list view with over 2000000 contacts. Enable this feature via the config_override.php setting $sugar_config['large_table']['contacts'].

## Types of changes

### Final checklist

### DIFF
--- a/data/SugarBean.php
+++ b/data/SugarBean.php
@@ -3465,7 +3465,9 @@ class SugarBean
             '',
             false,
             null,
-            $singleSelect
+            $singleSelect,
+            false,
+            'LIMIT ' . $row_offset . ', ' . $limit
         );
         return $this->process_list_query($query, $row_offset, $limit, $max, $where);
     }
@@ -3503,6 +3505,7 @@ class SugarBean
      * @param object $parentbean creating a subquery for this bean.
      * @param bool $singleSelect Optional, default false.
      * @param bool $ifListForExport
+     * @param string $limit Optional, default empty
      * @return String select query string, optionally an array value will be returned if $return_array= true.
      */
     public function create_new_list_query(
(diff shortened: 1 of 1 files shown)


Abuelodelanada:
Hi @andre-karrlein !

I am not a SQL Server Fan ;-) but as far as I understand if you use `LIMIT` in SugarBean.php the SQL Server compatibility will be broken, because there is no `LIMIT` statement in [that engine](https://stackoverflow.com/a/603735).

There is a method called `limitQuery()` for [MySQL](https://github.com/salesagility/SuiteCRM/blob/master/include/database/MysqlManager.php#L266) and for [SQLServer](https://github.com/salesagility/SuiteCRM/blob/master/include/database/MssqlManager.php#L433)

Check that the way every engine writes a 'LIMIT' is different.


Most of the performance problems with database are related with missing indices, bad configuration, low resources, etc.

May you capture the query for that ListView and execute:

```sql
EXPLAIN SELECT ...... FROM ....
```

And paste the result?

May be your problem can be solved in an easier way.

data/SugarBean.php
```diff
@@ -3465,7 +3465,9 @@ public function get_list(
             '',
             false,
             null,
-            $singleSelect
+            $singleSelect,
+            false,
+            'LIMIT ' . $row_offset . ', ' . $limit
```
cppmonkey: Use of the comment syntax may help with comparability
https://dev.mysql.com/doc/refman/8.0/en/comments.html

`'/*! LIMIT ' . $row_offset . ', ' . $limit . ' */' `

PR #7905, opened 2019-09-22
Make internal errors with 500 HTTP header

oc666:
When there are internal crashes the application exit with 200 HTTP header, which is not correct, for example, on db failure.
This PR will change the HTTP header to be 500.

## Description
Today, the HTTP header can be confused monitor systems, when the SuiteCRM application return 200 when it's not correct.

## Motivation and Context
Allow monitor systems to monitor SuiteCRM application.

## How To Test This
Before the change: turn off DB and try to use the application. You'll get 200 HTTP header.
After the change: turn off DB and try to use the application. You'll get 200 HTTP header.

## Types of changes

### Final checklist

### DIFF
--- a/include/utils.php
+++ b/include/utils.php
@@ -1790,6 +1790,7 @@ function sugar_die($error_message, $exit_code = 1)
 {
     global $focus;
     sugar_cleanup();
+    http_response_code(500);
     echo $error_message;
     throw new \Exception($error_message, $exit_code);
 }


pgorod:
This makes some sense, but I wonder if `sugar_die` is only called for application crashes, or if people are relying on it just to exit even when there is no special error:

https://github.com/salesagility/SuiteCRM/search?q=sugar_die&unscoped_q=sugar_die

My point is that perhaps the weight of the past makes this change have the opposite effect of what you intended. You can confuse monitoring software with "false positives" of `HTTP 500` errors that aren't really problematic.

I am just raising the concern for discussion - I don't know much about this and I haven't done any extensive analysis.

connorshea:
@pgorod if people are using it that way, they probably shouldn't be :) But I think that's probably a valid concern, maybe we should make it configurable. with the default set to returning a 500 error?

Mac-Rae:
Hi @oc666 ,

Please rebase this pull request onto hotfix-7.10.x as pull requests directly to master are no longer accepted.
For more information please take a look at our [contribution guidelines](https://docs.suitecrm.com/community/contributing-code/).

If you would like assistance please ensure you select "Allow edits from maintainers" checkbox and leave us a message @tagging me! 

Thanks :+1:

oc666:
Create new [#8471 PR for this](https://github.com/salesagility/SuiteCRM/pull/8471).

PR #7909, opened 2019-09-23
Fix/htaccess

Dillon-Brown:
## Description

- Updated .htaccess rules to be consistent when upgrading vs installing.
- Disabled ETags.
- Disabled MultiViews.
- Added nosniff content rule.
- Removed "X-Powered-By" from response headers.
- Updated many ExpiresByType rules.

## How To Test This

1. Install a fresh instance of SuiteCRM using this branch.
2. Check that the htaccess is using the new rules.
3. Check that repair and rebuild correctly generates the .htaccess.
4. Check that using an upgrade pack correctly generates the .htaccess.

## Types of changes

### Final checklist

### DIFF
--- a/install/install_utils.php
+++ b/install/install_utils.php
@@ -984,7 +984,6 @@ EOQ;
     $restrict_str .= <<<EOQ
 RedirectMatch 403 {$ignoreCase}.*\.log$
 RedirectMatch 403 {$ignoreCase}/+not_imported_.*\.txt
-RedirectMatch 403 {$ignoreCase}/+(custom/soap|custom/cache|custom/xtemplate|custom/data|custom/examples|custom/include|custom/log4php|custom/metadata|custom/modules|custom/vendor)/+.*\.(php|tpl)
 RedirectMatch 403 {$ignoreCase}/+(soap|cache|xtemplate|data|examples|include|log4php|metadata|modules|vendor)/+.*\.(php|tpl)
 RedirectMatch 403 {$ignoreCase}/+emailmandelivery\.php
 RedirectMatch 403 {$ignoreCase}/+.git
--- a/install/install_utils.php
+++ b/install/install_utils.php
@@ -1020,6 +1020,10 @@ EOQ;
     RewriteRule ^Api/access_token$ Api/index.php/access_token [L]
     RewriteRule ^Api/V8/(.*?)$ Api/index.php/V8/$1 [L]
 </IfModule>
+<IfModule mod_headers.c>
+    Header unset ETag
+    FileETag None
+</IfModule>
 <IfModule mod_expires.c>
  ExpiresActive on
  ExpiresDefault "access plus 1 month"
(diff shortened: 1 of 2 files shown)


install/install_utils.php
```diff
@@ -982,23 +984,34 @@ function handleHtaccess()
     $restrict_str .= <<<EOQ
 RedirectMatch 403 {$ignoreCase}.*\.log$
 RedirectMatch 403 {$ignoreCase}/+not_imported_.*\.txt
-RedirectMatch 403 {$ignoreCase}/+(soap|cache|xtemplate|data|examples|include|log4php|metadata|modules)/+.*\.(php|tpl)
+RedirectMatch 403 {$ignoreCase}/+(custom/soap|custom/cache|custom/xtemplate|custom/data|custom/examples|custom/include|custom/log4php|custom/metadata|custom/modules|custom/vendor)/+.*\.(php|tpl)
```
mattlorimer: "custom/xtemplate custom/data custom/examples custom/log4php" etc, should it not be custom/application,custom/Extension etc, plus these are already covered by the next line anyway
is there maybe a regex  solotion to just inlucde the directories we want

Dillon-Brown: Yes, you're right, it's already covered by the next line. For the ```custom/application``` and ```custom/Extension``` I would think that those shouldn't be restricted.

install/install_utils.php
```diff
@@ -982,23 +984,33 @@ function handleHtaccess()
...
 
 <IfModule mod_rewrite.c>
     Options +SymLinksIfOwnerMatch
+    Options -Indexes
+    Options -MultiViews
     RewriteEngine On
     RewriteBase {$basePath}
     RewriteRule ^cache/jsLanguage/(.._..).js$ index.php?entryPoint=jslang&modulename=app_strings&lang=$1 [L,QSA]
     RewriteRule ^cache/jsLanguage/(\w*)/(.._..).js$ index.php?entryPoint=jslang&modulename=$1&lang=$2 [L,QSA]
+    RewriteRule ^cache/jsLanguage/(.._..).js$ index.php?entryPoint=jslang&module=app_strings&lang=$1 [L,QSA]
```
mattlorimer: where is module used, looking at the entry point it uses modulename, so I can see a reason for having both?

install/install_utils.php
mattlorimer: why not?

install/install_utils.php
Dillon-Brown: These two lines were pulled from ```modules/Administration/UpgradeAccess.php```. I also don't see any obvious reason for this but I kept it in just in case there was some legacy backwards compatibility reason this was done.

install/install_utils.php
Dillon-Brown: My thinking was that we wouldn't want to restrict what you can do in custom beyond the default core restrictions. I wouldn't want to break someone's customisations in a minor release unnecessarily (Even if they are doing it the wrong way).

install/install_utils.php
mattlorimer: ok, but makes no sense it can't do both, I assume it just defaults to the first one so the second is irrelevant anyway

Dillon-Brown: Agreed, will remove.

PR #7948, opened 2019-10-01
Fix/backport mpdf

Dillon-Brown:
## Description

This is a duplicate of #7118 for the hotfix-7.10.x branch.

## Motivation and Context

This allows us to composerify mPDF while still maintaining backwards compatibility by including empty PHP files and aliasing the changed class names.

## How To Test This

1. Check that mPDF works as expected.

## Types of changes

### Final checklist


modules/AOS_PDF_Templates/PDF_Lib/classes/meter.php
```diff
@@ -1,228 +1,4 @@
...
-
-                $svg .= '</g></svg>';
-            }
-        }
-
-        return $svg;
-    }
-}	// end of class
+/**
+ * Class for backwards compatibility only
```
connorshea: We shouldn't bother with it right now, but I wonder if we should print deprecation notices if these files are included? Is there a way to do that, or would Composer autoload these files and cause false positives?

connorshea:
`Warning: Class 'MpdfException' not found in /home/travis/build/salesagility/SuiteCRM/deprecated.php on line 66` 🤔

deprecated.php
```diff
@@ -58,3 +58,14 @@ class_alias('SuiteCRM\\Zend_Oauth_Provider', 'Zend\\Oauth\\Provider');
     /** @deprecated this is an alias for Zend_Oauth_Provider */
     class Provider {}
 }
+
+
+/**
+ * Backwards compatibility for mPDF
+ */
+class_alias('MpdfException', 'mPDF_exception');
```
connorshea: I think you got this backward

connorshea: Or maybe not, why is this erroring 🤔

connorshea:
@Dillon-Brown FWIW, we finally got the backport merged in our 7.10 fork a while ago, and deployed it to production on Monday. It required one change to a PDF Template like I mentioned in my comment on #7118, but other than that we haven't had any problems reported yet.

I'm not sure if we should merge this since it's _technically_ a breaking change, but it also doesn't seem like a good idea to just continue without the fix for all these errors.

mattlorimer:
no longer required

PR #7954, opened 2019-10-01
Fixed #7953 - Elasticsearch default size setting

### LINKED ISSUE #7953: Elasticsearch default search count does not reflect settings
#### Issue

Suite provides settings that influence the dropdown to select the number of results. These are stored in the `search.pagination.min/max/step` settings. 
The default number of search results is hardcoded and does not consider the pagination settings. 
This can lead to an odd user experience. If, for example, the Minimum pagination setting is 25, then it's odd to only receive 10 results on the initial search. 

#### Expected Behavior

The default search should be _at least_ the minimum value available in the dropdown box.

#### Actual Behavior

The default search is hardcoded to 10. 

#### Possible Fix

See PR
#### Steps to Reproduce

1. Enable Elasticsearch and Index your records
2. Set the `$config['search']['pagination']['min'] value to a number other than 10
3. Perform a generic search (from the navbar) that will return many results
4. You will only see 10 records no matter what your config settings are

#### Context

My customers would like to see more results out of the gate. This is especially relevant when something like Notes get ranked higher than, say, an Account - Users don't want to have to go to page 2 and currently it's too difficult to filter by module. 

#### Your Environment

* SuiteCRM Version used:7.11.8
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)):Chrome
* Environment name and version (e.g. MySQL, PHP 7):LAMP
* Operating System and version (e.g Ubuntu 16.04):LAMP

tsummerer:
## Description

This PR addresses #7953 
It changes the SearchQuery class to check the config before falling back to a default search size. 

This is important for situations where a CRM has a minimum pagination size greater or less than 10. 

I also took the liberty of adding a dedicated setting for the default query size. 

## Motivation and Context

This PR gives developers the ability to set and potentially raise the default Elasticsearch query size. This can be especially helpful if there are 'cluttered' results (like if you're searching for a specfic Account and for whatever reason a bunch of Notes rank higher). 

## How To Test This

1. Enable Elasticsearch and Index your records
2. Perform a generic search from the Navbar - you should get 10 results
3. Add the following to your config: `$sugar_config['search']['pagination']['min'] = 5`
4. Repeat the search - you should get 5 results.
5. Add the following to your config: `$sugar_config['search']['query_size'] = 11`
6. Repeat the search - you should get 11 results

## Types of changes

### Final checklist

- [x] (maybe) My change requires a change to the documentation.

### DIFF
--- a/lib/Search/SearchQuery.php
+++ b/lib/Search/SearchQuery.php
@@ -28,6 +28,8 @@ if (!defined('sugarEntry') || !sugarEntry) {
  */
 class SearchQuery implements \JsonSerializable
 {
+    const DEFAULT_SEARCH_SIZE = 10;
+
     /** @var string Search query string */
     private $query;
     /** @var int The number of results per page */
@@ -49,13 +51,10 @@ class SearchQuery implements \JsonSerializable
      * @param int         $from         Offset of the search. Used for pagination
      * @param array       $options      [optional] used for additional options by SearchEngines.
      */
-    private function __construct($searchString, $engine = null, $size = 10, $from = 0, array $options = [])
+    private function __construct($searchString, $engine = null, $size = null, $from = 0, array $options = [])
     {
         $this->query = strval($searchString);
-        $this->size = intval($size);
-        if ($this->size < 0) {
-            $this->size = 1;
-        }
+        $this->size = $size ? intval($size) : $this->getDefaultSearchSize();
         $this->from = intval($from);
         $this->options = $options;
         $this->engine = $engine !== null ? strval($engine) : null;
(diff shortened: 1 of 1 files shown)


lib/Search/SearchQuery.php
```diff
@@ -95,7 +95,7 @@ public static function fromRequestArray(array $request)
     {
         $searchQuery = self::filterArray($request, 'search-query-string', '', FILTER_SANITIZE_STRING);
         $searchQueryAlt = self::filterArray($request, 'query_string', '', FILTER_SANITIZE_STRING);
-        $searchSize = self::filterArray($request, 'search-query-size', 10, FILTER_SANITIZE_NUMBER_INT);
+        $searchSize = self::filterArray($request, 'search-query-size', static::getDefaultSearchSize(), FILTER_SANITIZE_NUMBER_INT);
```
Dillon-Brown: Taking a quick look, would this not be better as a constant? that way we would avoid having a magic number while also allowing us to remove the: 
```php
public static function getDefaultSearchSize();
```
function in favour of using the class constructor since we would already have the ```private $size``` var and the getter function:
```php
    public function getSize()
    {
        if ((int)$this->size < 0) {
            $this->size = 1;
        }
        return (int)$this->size;
    }
```

tsummerer: Yeah - a constant is more clear. 
I refactored to remove the static function. Initially, I hadn't realized that I could pass a null to the `filterArray` function. 

So I'm setting the default value in the constructor now. I still kept the code in it's own function in order to keep things short and specific.

Issue #2810, opened 2016-12-20
Should dir_mode of default_permissions (from config.php) override sugar_mkdir calls?


pgorod:
This is another possible issue from my analysis of permissions problems in SuiteCRM. I'm not _sure_ it's an issue, so I phrased it as question; but it's worth considering it carefully - and I _think_ it's a bug.

In calls to `sugar_mkdir`, an optional parameter is available to specify `dir_mode` permissions of the new directory. Here are the calls throughout our code:
```
ModuleInstall/ModuleInstaller.php:2052:        sugar_mkdir( $dest );

ModuleInstall/ModuleScanner.php:212:      'sugar_mkdir',

data/Relationships/RelationshipFactory.php:202:        sugar_mkdir(dirname($this->getCacheFile()), null, true);

install/installSystemCheck.php:244:               if ((is_dir($dirname)) || @sugar_mkdir($dirname,0755, true)) // set permissions to restrictive - use make_writable to change in a standard way to the required permissions

install/performSetup.php:487:    sugar_mkdir('custom/modules/Home', 0775);
install/performSetup.php:491:    sugar_mkdir('custom/include', 0775);

install/install_utils.php:2141:    if ((is_dir($dirname)) || @sugar_mkdir($dirname,0755)) {

modules/ModuleBuilder/MB/MBPackage.php:596:       sugar_mkdir($path . $trimmedPath, NULL, true);

modules/MailMerge/Merge.php:155:        sugar_mkdir($dataDir);

modules/MailMerge/MailMerge.php:153:              sugar_mkdir($this->mm_data_dir);
modules/MailMerge/MailMerge.php:154:              sugar_mkdir($this->mm_data_dir.'/Temp/');

modules/Configurator/controller.php:183:                sugar_mkdir('custom/modules/Home', 0775);
modules/Configurator/controller.php:187:                sugar_mkdir('custom/include', 0775);

modules/Administration/Common.php:55:    return sugar_mkdir("custom/include/language", null, true);
modules/Administration/Common.php:69:   return sugar_mkdir("custom/modules/$module/language", null, true);

modules/EmailTemplates/EmailTemplateFormBase.php:206:       sugar_mkdir('public', 0777);

modules/EmailTemplates/EmailTemplate.php:897:      sugar_mkdir('public', 0777);

modules/AOS_PDF_Templates/templateParser.php:75:        sugar_mkdir('public', 0777);

[...]

PR #7957, opened 2019-10-02
Fix/silent upgrade

Dillon-Brown:
## Description

Fixed an issue with the file pathing in step 2 of the silent upgrader and added R&R to the final step. This issue was preventing users from upgrading and even with a workaround would prevent them from getting to the admin panel to run a repair and rebuild.

## How To Test This

1. Unzip an upgrade pack and copy the updated ```silentUpgrade_step2.php``` file in this PR to it before recompressing.
2. Move the upgrade pack into the CRM and run the following command found here: https://docs.suitecrm.com/admin/installation-guide/upgrading/

## Types of changes

### Final checklist

### DIFF
--- a/modules/UpgradeWizard/silentUpgrade_step2.php
+++ b/modules/UpgradeWizard/silentUpgrade_step2.php
@@ -475,6 +475,10 @@ if (version_compare($sugar_version, '6.5.0', '<') && function_exists('repairUpgr
     repairUpgradeHistoryTable();
 }
 
+// Run quick repair and rebuild
+$repair = new RepairAndClear();
+$repair->repairAndClearAll(['clearAll'], ['All Modules']);
+
 //TAKE OUT TRASH
 if (empty($errors)) {
     set_upgrade_progress('end', 'in_progress', 'unlinkingfiles', 'in_progress');
--- a/modules/UpgradeWizard/silentUpgrade_step2.php
+++ b/modules/UpgradeWizard/silentUpgrade_step2.php
@@ -308,17 +308,7 @@ if (isset($argv[4])) {
     exit(1);
 }
 
-$unzip_dir = sugar_cached("upgrades/temp");
-
-include "$unzip_dir/manifest.php";
-if (is_file("$unzip_dir/manifest.php")) {
-    include "{$argv[1]}/manifest.php";
-}
-
-if (isset($manifest['copy_files']['from_dir']) && $manifest['copy_files']['from_dir'] != "") {
-    $zip_from_dir = $manifest['copy_files']['from_dir'];
-}
-
+$unzip_dir = sugar_cached('upgrades/temp');
 
 $install_file = $sugar_config['upload_dir'] . '/upgrades/patch/' . basename($argv[1]);
 sugar_mkdir($sugar_config['upload_dir'] . '/upgrades/patch', 0775, true);


modules/UpgradeWizard/silentUpgrade_step2.php
```diff
@@ -308,17 +308,7 @@ function rebuildRelations($pre_path = '')
     exit(1);
 }
 
-$unzip_dir = sugar_cached("upgrades/temp");
-
-include "$unzip_dir/manifest.php";
-if (is_file("$unzip_dir/manifest.php")) {
```
mattlorimer: perhaps this was meant to be !is_file ??

anyway is it the same issue in modules/UpgradeWizard/silentUpgrade_step1.php:545

willrennie:
I have assessed this running silent upgrades from 7.11.3 and 7.11.5 to 7.11.8. I receive an 'Upgrade Complete!' message within the terminal but it appears I have a myriad of issues. Mainly, labels across the CRM appear broken until I run permissions. It also seems that the upgraded files are not applied as when I check the about page, my CRM is still on the pre-upgrade version and the upgrade does not appear to have been successfully implemented.

mattlorimer:
@willrennie that is bug with 7.11.8 specifically not related to this

modules/UpgradeWizard/silentUpgrade_step1.php
```diff
@@ -541,10 +541,8 @@ function sugar_cached($file)
 
     $unzip_dir = sugar_cached('upgrades/temp');
 
-    include "$unzip_dir/manifest.php";
-    if (is_file("$unzip_dir/manifest.php")) {
-        include "{$argv[1]}/manifest.php";
-    }
+    $install_file = $sugar_config['upload_dir'] . '/upgrades/patch/' . basename($argv[1]);
+    sugar_mkdir($sugar_config['upload_dir'] . '/upgrades/patch', 0775, true);
```
pgorod: We shouldn't be setting permissions like this. I suggest using `make_writable` function which can also be improved, but still is better than using fixed permissions values directly https://github.com/salesagility/SuiteCRM/blob/master/install/install_utils.php#L1308

If an admin doesn't want to add `world` permissions, which is a good idea, and even takes the trouble of specifying this in `default_permissions` in `config.php`, who are we to then go open up his system like this?

Dillon-Brown: @pgorod Totally agreed, this is a hacky fix ontop of a hacky fix intended to just "force" the upgrade to work. We actually do an absolute ton of other things forcefully and without warning regarding permissions and files during the upgrade process (E.g. just uploading a pack, even the wrong one, will steamroll force replace all the DBManagers and a couple dozen other files).

I'm rather reluctant to change anything here since it requires so much testing upgrading legacy instances with customisations.

pgorod: Anyway, it seems `sugar_mkdir` will happily ignore whatever you pass it there and just use the value from `config.php`.
See #2810

Go ahead with whatever you think is best at this point. I agree that a proper fix needs to go deeper than this, and involves a lot more changes. Sigh...

Dillon-Brown: > Go ahead with whatever you think is best at this point. I agree that a proper fix needs to go deeper than this, and involves a lot more changes. Sigh...

Just to note, I'll be making another PR (Will likely need to go to develop though) after the next release that will fix all of these problems and bring the upgrade time down to 30 or so seconds :).

pgorod: In that case, to add to your work, I have a couple more requests

- a check that warns users that they are trying to use a full installation pack, or an add-on, in the upgrade wizard (instead of just saying it's invalid).

- a button to upgrade directly from the Internet, avoiding the download-then-upload loop. Just "Upgrade available, click here to apply it".

Of course you won't have time for all this but I thought I might as well ask for it!

Dillon-Brown: Done and done :D. The upgrade from the internet I was going to split into a third PR though.

pgorod: That's I call _agile_ 😱

cameronblaikie:
Okay looks good to me :+1: Assessed

craigpanton:
LGTM :+1:

Mac-Rae:
Assessed :+1:

Issue #7727, opened 2019-08-21
Composer install fails 


Dillon-Brown:
#### Issue

When running ```composer install``` on a windows based setup, you will run into this issue:

```'rm' is not recognized as an internal or external command```

This is due to the post-install-cmd script containing a Linux based command.

#### Expected Behavior

Composer should pull down packages listed in composer.json regardless of OS.

#### Actual Behavior

Composer fails on non linux-based systems.

#### Possible Fix

I'd suggest updating this to simply call a PHP file and do the required operation there.

#### Steps to Reproduce

1. Run ```composer install```.

#### Context

This prevents someone who downloads from the repo from getting SuiteCRM running. Although a simple workaround would be to run ```composer update``` instead (Not ideal in a prod environment).

#### Your Environment

* SuiteCRM Version used: 7.11.7
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Chrome
* Environment name and version (e.g. MySQL, PHP 7): PHP 5.6
* Operating System and version (e.g Ubuntu 16.04): W10 + Ampps

PR #7980, opened 2019-10-05
Fix post-install-cmd to work on Windows

brunoti:
Fixes: #7727

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -59,7 +59,11 @@
     "mockery/mockery": "^0.9.9",
     "symfony/yaml": "^3.4"
   },
-  "scripts": {},
+  "scripts": {
+    "post-install-cmd": [
+      "php ./install/composer_post_install_cmd.php"
+    ]
+  },
   "prefer-stable": true,
   "autoload": {
     "files": [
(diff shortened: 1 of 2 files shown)


pgorod:
Hi, thanks for this!

But you need to rebase to your `hotfix-7.10.x` branch so that it is up-to-date with our `hotfix-7.10.x`, otherwise more than a thousand commits are showing as part of your PR...

Another thing: I would personally prefer if the new `php composer_post_install_cmd.php` could be placed in a subdirectory, not in root directory, just to keep things cleaner. I suggest putting it in `install` directory. Thanks!

brunoti:
> Hi, thanks for this!
> 
> But you need to rebase the PR onto an updated `hotfix-7.10.x` branch in your repo so that it is up-to-date with our `hotfix-7.10.x`, otherwise more than a thousand commits are showing as part of your PR...
> 
> Another thing: I would personally prefer if the new `php composer_post_install_cmd.php` could be placed in a subdirectory, not in root directory, just to keep things cleaner. I suggest putting it in `install` directory. Thanks!

Ok! I will fix this later

Mac-Rae:
Hi @brunoti,

Could you sign the clahub?

PR #7989, opened 2019-10-08
BeanFactory: Don't return deleted beans from the cache

lazka:
## Description

In case a bean is retrieved, deleted and then fetched again the bean factory
would return it despite the bean being deleted. This means the bahviour is different
depending on whether the bean is cached or not.

Avoid this by not returning a cached bean if it is marked deleted and the delete
filter isn't disabled by the caller.

## Motivation and Context

I was writing a test and was confused why deleted beans got still returned.

## How To Test This

See the unit test.

## Types of changes

### Final checklist


cameronblaikie:
Assessed :+1:

PR #8003, opened 2019-10-08
Fixed #6046 for 7.10.x - "rank" is a SQL reserved word, can't be used as a field

### LINKED ISSUE #6046: SQL "rank" is a reserved word, can't be used as a field name in MySQL 8.0.2 and later

Abuelodelanada:
## Description

It seems that rank is an SQL reserved word since MySQL 8.0.2:

## Motivation and Context

Issue reference: #6046

## How To Test This

Install SuiteCRM 7.10.x using MySQL 8.0.2

## Types of changes

### Final checklist

### DIFF
--- a/metadata/user_feedsMetaData.php
+++ b/metadata/user_feedsMetaData.php
@@ -8,7 +8,7 @@ if (!defined('sugarEntry') || !sugarEntry) {
  * SugarCRM, Inc. Copyright (C) 2004-2013 SugarCRM Inc.
  *
  * SuiteCRM is an extension to SugarCRM Community Edition developed by SalesAgility Ltd.
- * Copyright (C) 2011 - 2018 SalesAgility Ltd.
+ * Copyright (C) 2011 - 2019 SalesAgility Ltd.
  *
  * This program is free software; you can redistribute it and/or modify it under
  * the terms of the GNU Affero General Public License version 3 as published by the
--- a/metadata/user_feedsMetaData.php
+++ b/metadata/user_feedsMetaData.php
@@ -41,17 +41,41 @@ if (!defined('sugarEntry') || !sugarEntry) {
  * display the words "Powered by SugarCRM" and "Supercharged by SuiteCRM".
  */
 
-$dictionary['users_feeds'] = array( 'table' => 'users_feeds'
-                                  , 'fields' => array(
-    
-       array('name' =>'user_id', 'type' =>'varchar', 'len'=>'36', )
-      , array('name' =>'feed_id', 'type' =>'varchar', 'len'=>'36', )
-      , array('name' =>'suite_rank', 'type' =>'int', 'required' => false)
-      , array('name' => 'date_modified','type' => 'datetime')
-      , array('name' =>'deleted', 'type' =>'bool', 'len'=>'', 'default'=>'0', 'required' => false)
-                                                      )
-                                 , 'indices' => array(
-  
-       array('name' =>'idx_ud_user_id', 'type' =>'index', 'fields'=>array('user_id', 'feed_id'))
-                                                      )
-                                  );
+$dictionary['users_feeds'] = [
+    'table' => 'users_feeds',
+    'fields' => [
+        [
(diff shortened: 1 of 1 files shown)


metadata/user_feedsMetaData.php
```diff
@@ -46,7 +46,7 @@
     
        array('name' =>'user_id', 'type' =>'varchar', 'len'=>'36', )
       , array('name' =>'feed_id', 'type' =>'varchar', 'len'=>'36', )
-      , array('name' =>'rank', 'type' =>'int', 'required' => false)
+      , array('name' =>'suite_rank', 'type' =>'int', 'required' => false)
```
Dillon-Brown: While this change did go into ```7.8.x```, I'm thinking it probably shouldn't have for backwards compatibility reasons (Assuming we can create a workaround). I haven't tried this out but would adding an alias here work?

*Not super necessary but it would also be nice to clean up some of this file while we're here. Attached a .txt to show what I mean (Can't wait for GitHub multi-line code suggestions).

[user_feedsMetaData.txt](https://github.com/salesagility/SuiteCRM/files/3704618/user_feedsMetaData.txt)

Abuelodelanada: I don't think alias will solve the bug. Let's think we define the field like you wrote in .txt:

```php
        [
            'name' => 'rank',
            'type' => 'int',
            'alias' => 'suite_rank',
            'required' => false
        ],
```

This will be [translated](https://github.com/salesagility/SuiteCRM/blob/master/data/SugarBean.php#L3656) into SQL:

```SQL
SELECT rank as suite_rank
```

But the problem is the name `rank` when **creating** the table:

```sql
Error creating table: users_feeds Query Failed: CREATE TABLE users_feeds (user_id varchar(36)  NULL ,feed_id varchar(36)  NULL ,rank int  NULL ,date_modified datetime  NULL ,deleted bool  DEFAULT '0' NULL  , KEY idx_ud_user_id (user_id, feed_id)) CHARACTER SET utf8 COLLATE utf8_general_ci: MySQL error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'rank int  NULL ,date_modified datetime  NULL ,deleted bool  DEFAULT '0' NULL  , ' at line 1
```

I think we face these situations:

**If we do not fix in 7.10:**

1. SuiteCRM 7.10 installation with MySQL 8.0.x: The installation will fail.
2. SuiteCRM 7.10 installed with MySQL 5.7 and upgraded to MySQL 8.0.x: I think the upgrade process of MySQL will fail or at least will show a warning.

**If we fix in 7.10:**

1. SuiteCRM 7.10 installation with MySQL 8.0.x: The installation will work ok.

2. SuiteCRM 7.10.x (pre-fix) installed with MySQL 5.7.x and upgraded to 7.10.x (post-fix): The QR&R will perform an ```ALTER TABLE users_feeds...``` to change the name of the field `rank` to `suite_rank`

3. SuiteCRM 7.10 (post-fix) installed with MySQL 5.x and upgraded to MySQL 8.0: Will work ok.


I agree with you, we can clean up this file.

metadata/user_feedsMetaData.php
```diff
@@ -41,17 +41,41 @@
...
+            'type' =>'varchar',
+            'len'=>'36',
+        ],
+        [
+            'name' =>'feed_id',
+            'type' =>'varchar',
+            'len'=>'36',
+        ],
+        [
+            'name' => 'suite_rank',
```
mattlorimer: 'ranking' would maybe be a better option over 'suite_rank', overall this is a breaking change

Abuelodelanada:
I close this PR because PR #8705 it's much better!
@Dillon-Brown @Mac-Rae Can you take a look at  #8705 please?

Abuelodelanada:
I close this PR because PR #8705 it's much better!
@Dillon-Brown @Mac-Rae Can you take a look at  #8705 please?

PR #8028, opened 2019-10-10
Fixed #7874 - removes ability to use custom _head.tpl file

### LINKED ISSUE #7874: 7.11.8 removes ability to use custom _head.tpl file
#### Issue

Prior to 7.11.8 you could place a custom file `custom/themes/SuiteP/tpls/_head.tpl` and it would be used by the application. 

Since the update, the path is hardcoded and does not check the custom directory. 

This file controls the `<head>` section of the page. It was a nice place to add global javascript/css dependencies. (ie. bring in a third party library, etc). 

Many platforms (ecommerce, website, etc) give folks a way to add scripts to the head - it's a common use case. So it's disappointing to have the feature removed.

#### Expected Behavior

A file placed in `custom/themes/SuiteP/tpls/_head.tpl` will be loaded by the app.

#### Actual Behavior

The path to the `_head.tpl` file is hardcoded in `header.tpl` and cannot be modified without overwriting core. 

See #7591
#### Possible Fix

@connorshea  suggested a plugin [here](https://github.com/salesagility/SuiteCRM/pull/7591#issuecomment-512596068)

#### Steps to Reproduce

1. Add a file to the custom directory : `custom/themes/SuiteP/tpls/_head.tpl`. Let this file contain some custom javascript or something
2. QRR and see if your file gets loaded

#### Context

I was using `_head.tpl` to load Global javascript libraries from a cdn. 

#### Your Environment

* SuiteCRM Version used: 7.11.8
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)):Chrome
* Environment name and version (e.g. MySQL, PHP 7):LAMP
* Operating System and version (e.g Ubuntu 16.04):LAMP

604media:
Fixed #7874 - removes ability to use custom _head.tpl file


cameronblaikie:
Assessed :+1:

connorshea:
Should maybe rename this PR, right now it can be read as you're removing the ability to do this :)

Dillon-Brown:
> Should maybe rename this PR, right now it can be read as you're removing the ability to do this :)

Lol good point

PR #8031, opened 2019-10-10
Fixed #7758 - Disable Action menu has no effect on menus in subpanel

### LINKED ISSUE #7758: Disable Action menu has no effect on menus in subpanel
Although you disable "Show actions in menu" ( enable_action_menu ) actions will still be shown as menu in subpanel action menu.
The setting only affects the top menu, but it did affect the subpanel menu, too in 7.8.x

#### Expected Behavior
Subpanel action menu is shown as seperate button

#### Actual Behavior
Subpanel menu is shown as one dropdown action menu.

#### Steps to Reproduce
1. Go to "System Settings->Display actions within menus 
2. uncheck
3. Go to a module with an activity subpanel
4. See the action menu is a menu not separate buttons

#### Context
As our sales most time uses "New meeting" the action menu with "New Task" as default in the first place is not good. So the option or possibility to reorder would fix this.
Had no Problem with separating the buttons before updating to 7.11.x from 7.8.x

#### Your Environment

* SuiteCRM Version used: 7.11.7
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Firefox or Chrome
* Environment name and version (e.g. MySQL, PHP 7): PHP 7.3, Apache
* Operating System and version (e.g Ubuntu 16.04): Ubuntu 18.04

604media:
Fixed #7758 - Disable Action menu has no effect on menus in subpanel


include/SubPanel/SubPanelTiles.php
```diff
@@ -443,7 +443,6 @@ public function get_buttons($thisPanel, $panel_query=null)
         $widget_contents = smarty_function_sugar_action_menu(
             [
                 'buttons' => $buttons,
-                'flat' => !empty($thisPanel->_instance_properties['flat']),
```
marin-h: This was meant for displaying action menu with flat buttons instead of dropdown options (see screenshot)... it was a useful feature, why was this removed?

This was possible adding `'flat' => true,` in subpanel's definition.

imagen

marin-h: @Dillon-Brown @cameronblaikie when you have the time can you please check this out, we were using this feature in production. Thanks!

PR #8038, opened 2019-10-11
Update LICENSE.txt

NeelPrime:
#8036 

I checked also grammar.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -101,7 +101,7 @@ menu, a prominent item in the list meets this criterion.
 
   The "source code" for a work means the preferred form of the work
 for making modifications to it.  "Object code" means any non-source
-form of a work.
+form of work.
 
   A "Standard Interface" means an interface that either is an official
 standard defined by a recognized standards body, or, in the case of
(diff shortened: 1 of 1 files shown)


pgorod:
`other than` is correct, `other then` is not, you're changing a correct expression to an incorrect one...

I don't think we should change anything in this file, the changes proposed here are not relevant, and mainly, this is a legal text, probably copied from somewhere, we should leave this to the lawyers, IMHO.

horus68:
This is a legal docuemnto and should be synched with the source only... the GNU AFFERO GENERAL PUBLIC LICENSE - Version 3

but, there is an issue...

- The file used by SuiteCRM is **Version 3, 19 November 2007**
https://github.com/salesagility/SuiteCRM/blob/038cd0ddab338275b08b63cff823ddd211f4ed16/LICENSE.txt

- The actual file in use by gnu.org is **Version 3, 29 June 2007**
https://www.gnu.org/licenses/gpl-3.0.html

I rest my case and leave it to someone expert on this!

Dillon-Brown:
Closing this as we shouldn't be changing the license file.

PR #8040, opened 2019-10-11
Remove dead code

connorshea:
## Description
This removes dead code for the Activities SubPanel. I couldn't figure out how to cause anything involving this code to occur, the code isn't covered at all in the test suite (suggesting to me that it's at least _not_ ever used in the test suite, not that that means a lot with only 10% coverage), and as far as I can tell the subpanels are handled in a generic way in the current iteration of SuiteCRM.

## Motivation and Context
I originally started this by trying to turn it into a Smarty template, but I couldn't figure out where the code was used.

## How To Test This

Mess with a bunch of views and see if anything looks broken or if there are errors in `php_errors.log`.

Also make sure the test suite passes, obviously.

## Types of changes
Technical debt :)

### Final checklist


Dillon-Brown:
Agreed this doesn't actually get used anywhere. I think we'd need this to go to develop anyway though as someone might have a ```require_once``` pointing to these files.

PR #8042, opened 2019-10-11
Ignore vendored library files in .codecov.yml

connorshea:
This now matches the coverage tracking in hotfix.

Make sure this passes CI, obviously, and make sure it's valid and makes sense.

Here's the file on hotfix for comparison: https://github.com/salesagility/SuiteCRM/blob/hotfix/.codecov.yml


cameronblaikie:
Assessed :+1:

PR #8050, opened 2019-10-14
Fix #8001 Non-distinct person entries for each meeting/call invited to

### LINKED ISSUE #8001: API returns non-distinct values in Person modules
#### Issue

(EDIT: The following applies to Contacts and Leads as well)

The `/module/Users` endpoint returns nonsense since 7.11.4.

Instead of returning one record for each User, it returns one record for each user's calls and meetings. 
So if admin has 10 calls and 10 meetings, there are at least 20 admin user records in the results. 

This makes it nearly impossible to get a list of CRM Users. 

In a vanilla instance with demo data, there are 2,125 User records returned even though there are only 7 users. 

The problem was introduced by #7147

The root cause seems to be the `c_accept_status_fields` and `m_accept_status_fields` fields. These cause a `LEFT JOIN` to be added to the query which sends everything into chaos. 

#### Expected Behavior

The Users api endpoint returns one record for each user

#### Actual Behavior

The Users api endpoint returns multiple copies of each user (if they are invitees on any Meetings/Calls). The pagination is completely off and it's nearly impossible to get a list of active users on an instance. 
#### Possible Fix

I'm not completely sure. A quick bandaid would be to figure out how to turn the `singleSelect` off for the Users module. 

##### Scenario 1: Fresh Instance
(Assuming less than 100 users)

1. Spin up a fresh instance (no dummy data) and configure the API. 
2. Do a `GET` to `/Api/V8/module/Users?page[size]=100`
3. You should receive a record for each User and the `meta.records-on-this-page` should equal your user count (assuming you have less than 100 users)
4. Log in to the UI and create **2** Meetings. Your user should be one of the invitees on each.
5. Do a `GET` to `/Api/V8/module/Users?page[size]=100` 

Expected : same results as step 3.
Actual : There will be n+1 results. The User with 2 meetings will be returned twice.

##### Scenario 2: Instance with Dummy Data

1. Set up the instance and configure API
2. Do a `GET` to `/Api/V8/module/Users?page[size]=100`

Expected : 7 results
Actual : 100 results - mostly duplicates. 

#### Context

I am trying to get a list of Active Users.

i think that this is high priority because this is an important and common endpoint. 

#### Your Environment

* SuiteCRM Version used: 7.11.8
* Environment name and version (e.g. MySQL, PHP 7): LAMP
* Operating System and version (e.g Ubuntu 16.04): LAMP

Jason-Dang:
## Description

- Set default field list from field definitions as is done in core when no fields given.
- Filter relationship info fields from default field list to prevent extra joins for accept status fields.

## Motivation and Context

Fixes issue https://github.com/salesagility/SuiteCRM/issues/8001 where duplicate entries are returned for every meeting/call the person records is an invitee of.

## How To Test This

- Create Contact, User, Lead.
- Create any number of meetings/calls (at least 2) and add the person records as invitees.
- Make a get list call for each person module and see that duplicates are returned corresponding to the number of meetings/calls invited to.

## Types of changes

### Final checklist

### DIFF
--- a/Api/V8/Service/ModuleService.php
+++ b/Api/V8/Service/ModuleService.php
@@ -113,6 +113,23 @@ class ModuleService
         $limit = $size === BeanManager::DEFAULT_ALL_RECORDS ? BeanManager::DEFAULT_LIMIT : $size;
         $deleted = $params->getDeleted();
 
+        if (empty($fields)) {
+            $fields = array_column($bean->field_defs, 'name');
+        }
+
+        $filteredFields = array_filter(
+            $fields,
+            function ($field) use ($bean) {
+                return (
+                    in_array($field, array_column($bean->field_defs, 'name'), true)
+                    && (
+                        empty($bean->field_defs[$field]['link_type'])
+                        || $bean->field_defs[$field]['link_type'] !== 'relationship_info'
+                    )
+                );
+            }
+        );
+
         $beanListResponse = $this->beanManager->getList($module)
             ->orderBy($orderBy)
             ->where($where)
@@ -120,6 +137,7 @@ class ModuleService
             ->limit($limit)
             ->max($size)
             ->deleted($deleted)
+            ->fields($filteredFields)
             ->fetch();
 
         $data = [];


Api/V8/Service/ModuleService.php
```diff
@@ -113,13 +113,31 @@ public function getRecords(GetModulesParams $params, Request $request)
         $limit = $size === BeanManager::DEFAULT_ALL_RECORDS ? BeanManager::DEFAULT_LIMIT : $size;
         $deleted = $params->getDeleted();
 
+        if (empty($fields)) {
+            $fields = array_column($bean->field_defs, 'name');
+        }
+
+        $filteredFields = array_filter(
+            $fields,
+            function ($field) use ($bean) {
```
Dillon-Brown:
```suggestion
            static function ($field) use ($bean) {
```

Api/V8/Service/ModuleService.php
```diff
@@ -113,13 +113,31 @@ public function getRecords(GetModulesParams $params, Request $request)
...
+                return (
+                    in_array($field, array_column($bean->field_defs, 'name'), true)
+                    && (
+                        empty($bean->field_defs[$field]['link_type'])
+                        || $bean->field_defs[$field]['link_type'] !== 'relationship_info'
+                    )
+                );
+            }
+        );
+
```
PedroErnst: Potentially worth extracting this bit into its own function. The `getRecords` function is already very long.

Dillon-Brown: Even better would be to split it out the entire function into separated classes imo. Obviously not a required change though :).

Jason-Dang: Hey hey hey, if you want me to re-write  everything just say ;) I've extracted the new logic to BeanManager class after debating a few places to put it.

willrennie:
Assessed. Fix working as documented :+1:

Issue #8057, opened 2019-10-14
PHP 7.4 Deprecations/Changes


connorshea:
This issue is meant to track all the deprecation warnings in SuiteCRM when running on PHP 7.4.

Notable deprecations:

- `implode()` now emits a deprecation warning when using it with the arguments in reverse order. use `implode($glue, $parts)` instead of `implode($parts, $glue)`.
- Accessing array values using curly braces is deprecated. e.g. `$array[0]` is good and `$array{0}` is bad.
   - I think all of these have been fixed in hotfix-7.10.x now, except for the uses that are in vendored libraries. We can fix those too if we want, but I didn't because I don't like messing with vendored code.
- The `get_magic_quotes_gpc()` and `get_magic_quotes_runtime()` functions are deprecated. They always return `FALSE` (and have since 5.3, I think? We can pretty remove any uses of these two in conditionals since they always return false).
- Nested ternary operators without explicit parentheses
  - [Explanation](https://www.php.net/manual/en/migration74.deprecated.php#migration74.deprecated.core.nested-ternary)

See these PRs for more info: #7987, #7976, and #8000.

A lot of the deprecated code comes from vendored dependencies, which is a problem since they're not trivial to upgrade. Thankfully, a lot of them have been moved to Composer in `develop`. We might want to backport those changes.

Some useful resources:

- https://stitcher.io/blog/new-in-php-74
- https://www.php.net/manual/en/migration74.php
- https://www.php.net/manual/en/migration74.deprecated.php

connorshea:
Also I guess it's worth mentioning these two uses of `each()`, which was deprecated in 7.2:

- https://github.com/salesagility/SuiteCRM/blob/d9f5d17197c91de9ac37c247180888fde1f0b75c/modules/ModuleBuilder/views/view.relationship.php#L143
- https://github.com/salesagility/SuiteCRM/blob/d9f5d17197c91de9ac37c247180888fde1f0b75c/modules/ModuleBuilder/parsers/relationships/AbstractRelationships.php#L440

connorshea:
Here are some other deprecation notices from #7987:

```
Deprecated: The behavior of unparenthesized expressions containing both '.' and '+'/'-' will change in PHP 8: '+'/'-' will take a higher precedence in /home/travis/build/salesagility/SuiteCRM/modules/InboundEmail/InboundEmail.php on line 1817

[...]

PR #8058, opened 2019-10-14
Fix deprecated usages of implode

connorshea:
Part of #8057. This fixes usages of `implode()` where the arguments are reversed. PHP 7.4 deprecates the ability to use `implode($parts, $glue)` in favor of only supporting `implode($glue, $parts)`.

This PR fixes a few of these deprecates uses.


cameronblaikie:
Assessed 👍

PR #8073, opened 2019-10-16
Remove all uses of get_magic_quotes_gpc.

connorshea:
This doesn't touch the vendored HTMLPurifier, which does still use this function.

[It always returns `FALSE` since PHP 5.4](https://www.php.net/manual/en/function.get-magic-quotes-gpc.php), and 7.4 deprecates it entirely. This PR removes those branches of code that never get touched.

Part of #8057.

`get_magic_quotes_runtime` is also deprecated, but it's only used in vendored libraries, so I haven't touched them.

### DIFF
--- a/include/utils.php
+++ b/include/utils.php
@@ -2294,22 +2294,11 @@ function clean_incoming_data()
     global $sugar_config;
     global $RAW_REQUEST;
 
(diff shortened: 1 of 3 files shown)


Mac-Rae:
Assessed :+1:

PR #8075, opened 2019-10-16
Fix misc PHP 7.4 deprecations

connorshea:
## Description
This fixes these three deprecation warnings from PHP 7.4:

```
Deprecated: The behavior of unparenthesized expressions containing both '.' and '+'/'-' will change in PHP 8: '+'/'-' will take a higher precedence in /home/travis/build/salesagility/SuiteCRM/modules/InboundEmail/InboundEmail.php on line 1817

Deprecated: Unparenthesized `a ? b : c ? d : e` is deprecated. Use either `(a ? b : c) ? d : e` or `a ? b : (c ? d : e)` in /home/travis/build/salesagility/SuiteCRM/modules/InboundEmail/InboundEmail.php on line 4366

Deprecated: The behavior of unparenthesized expressions containing both '.' and '+'/'-' will change in PHP 8: '+'/'-' will take a higher precedence in /home/travis/build/salesagility/SuiteCRM/modules/Users/authentication/SugarAuthenticate/SugarAuthenticate.php on line 404
```

Make sure the tests pass.

I'm not 100% sure the fix for InboundEmail's ternary expression is the correct fix, so please check that especially :)

Part of #8057.

### DIFF
--- a/modules/InboundEmail/InboundEmail.php
+++ b/modules/InboundEmail/InboundEmail.php
@@ -1814,7 +1814,7 @@ class InboundEmail extends SugarBean
                 'mbox' => $mailbox,
                 'totalcount' => $total
             );
-            $GLOBALS['log']->info("INBOUNDEMAIL: $status : Downloaded " . $start + sizeof($searchResults) . "messages of $total");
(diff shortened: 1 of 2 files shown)


Mac-Rae:
Assessed :+1:

PR #7196, opened 2019-04-17
*WIP Fix/inspections language migration

Dillon-Brown:
## Description

- Use PHP_SAPI, PHP_VERSION and M_PI constant instead of function calls.
- Removed alias functions and instead call origin.
- Replace ```dirname(__FILE__)``` with ```__DIR__```.
- Replace ternary operator with ["Elvis operator"](https://en.wikipedia.org/wiki/Elvis_operator).
- ```instanceof``` can be used.
- Replace ```is_null``` with ```$var === null```.

## Motivation and Context

1. Constants for these functions have been available since [PHP 4](https://www.php.net/manual/en/function.phpversion.php).
2. Alias functions simply reference other functions, this can cause surprises like ```magic_quotes_runtime``` being dropped in PHP 7.0.
3. ```__DIR__``` is equal to ```dirname(__FILE__)```, see [docs](https://www.php.net/manual/en/language.constants.predefined.php).
4. Simplify ternary operator.
5. ```instanceof``` is a language construct and should be preferred.
6. ```$var === null``` should be preferred over function call ```is_null```.

## How To Test This

See Travis, this should not change existing functionality.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [X] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change to the documentation.
- [X] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

PR #8076, opened 2019-10-16
Fix a deprecated usage of join.

connorshea:
join is an alias of implode. PHP 7.4 deprecates using the delimiter as the second argument.

Part of #8057. See also #8058.

### DIFF
--- a/include/SubPanel/SubPanelTiles.php
+++ b/include/SubPanel/SubPanelTiles.php
@@ -381,7 +381,7 @@ class SubPanelTiles
             array_push($tab_names, $tab);
         }
 
-        $tab_names = '["' . join($tab_names, '","') . '"]';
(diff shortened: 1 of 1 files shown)


gunnicom:
I think we should use not both join and implode. It would be better to use one or the other.
Quick search fund 528 implode( in my instance and 207 join(
I prefer implode as it is the opposite of explode, and join is "only" an alias.

Dillon-Brown:
@gunnicom Agreed, we should remove alias functions. I had a PR for this here: https://github.com/salesagility/SuiteCRM/pull/7196 but it will need rebased. Should be good to just merge this one in now and fix all the alias stuff in another PR.

cameronblaikie:
Assessed :+1:

PR #8080, opened 2019-10-17
Replaced alias functions

Dillon-Brown:
## Description

There's no reason to use aliases when we can just call the origin.

## Motivation and Context

https://github.com/salesagility/SuiteCRM/pull/8076#issuecomment-543129911

## How To Test This

There's really no differences to test here, I would just look at Travis which will lint all the modified files.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [X] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change to the documentation.
- [X] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

PR #8085, opened 2019-10-17
Fix a deprecated string concatenation.

connorshea:
This was originally done in #8075 but was overridden by #8080.

@Dillon-Brown 👀


Dillon-Brown:
Whoops, not sure how I missed that one..... Will get this merged soon as Travis passes.

PR #8074, opened 2019-10-16
Remove deprecated helper files

connorshea:
## Description
This code hasn't really been touched in years (other than with formatting commits that changed pretty much every file in the repo) and were marked as deprecated before SuiteCRM was even forked. So I've removed them.

## How To Test This
Make sure the test suite passes, I guess?

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist
- [x] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

PR #8089, opened 2019-10-17
Remove support for FreeTDS.

connorshea:
## Description
FreeTDS is an offshoot of MS SQL, but we don't test it, haven't updated anything related to it in years, and don't seem to mention support for it anywhere outside a few places in the code. The docs do not mention it anywhere.

This PR removes it entirely.

## Motivation and Context
FreeTDS support is presumably not working anymore (or if it works, it's a miracle). There's not a single commit, issue, or PR in all of SuiteCRM that refers FreeTDS, besides the one I opened yesterday that removes the FreeTDSHelper (#8074) and another where some formatting of a language file got messed up and happened to modify the line for FreeTDS's language string.

Essentially, this code is unused, unsupported, and unadvertised.

## How To Test This
🤷‍♂ Just make sure the tests pass and MySQL/MS SQL continues to work fine.

## Types of changes

### Final checklist


Mac-Rae:
I've tested this and it looks to be working fine, however, I have a few issues that I believe may be unrelated or an issue with my testing environment. Could @cameronblaikie take a second look at this?

What needs to be tested;
Case module (User communication via update field)
Inline Editing
Run an R&R and check for output

willrennie:
I am aware of some projects/people using the FreeTDS service within SuiteCRM so @connorshea I don't think a straight removal would be the best solution here. Just wondering if you think it is worth us looking at updating and putting some work into publicising the compatibility...

horus68:
self ping on language changes

connorshea:
@willrennie I'm generally of the opinion that supporting fewer configurations is usually better, especially when it comes to something core to the functionality of the CRM like a database.

But if there are users that are actually using SuiteCRM with FreeTDS, then I guess we shouldn't remove it. That said, I don't think it's worth updating and I definitely wouldn't encourage its usage.

PR #8093, opened 2019-10-18
Remove all deprecated Smarty files.

connorshea:
Smarty is included via Composer now (#7591, #7729), so these files can be removed. If anyone is requiring these files in their custom code, they should make sure to remove those requires.

All Smarty plugins remain as-is.


Mac-Rae:
Assessed :+1:

PR #8094, opened 2019-10-18
Remove all deprecated Zend files

connorshea:
This was deprecated in 7.10.x (#7672, #7719) and can safely be removed entirely. These libraries are loaded by Composer now.

Anyone requiring files in the `Zend/` directory should remove those requires. All relevant classes are loaded via Composer's autoload functionality.


Dillon-Brown (approved):

Mac-Rae:
Wanting to test AOD before merge however develop has a bug that prevents AOD from working so this is on pause until we get the change to resolve that issue. Apologies for the inconvenience

connorshea:
What's the status on this? :)

PR #8095, opened 2019-10-18
Remove SugarXHProf

connorshea:
## Description

The [XHProf project](https://github.com/phacility/xhprof) has been abandoned for over 4 years, and the version vendored here is much older than even that. It's undocumented, unmaintained, and it doesn't seem like it even works on PHP 7.x [if this comment is to be believed](https://www.php.net/manual/en/book.xhprof.php#123403).

This also removes the sugar config option 'xhprof_config', which doesn't seem to have been documented anywhere.

> The best way to choose what to keep and what to throw away is to take each item in one’s hand and ask: 'Does this spark joy?' If it does, keep it. If not, dispose of it.
> \- Marie Kondō

😉 

## Motivation and Context
We want to remove as much vendored third party code as possible, especially if it's not being used at all.

Part of #7605.

## How To Test This
Search around for references to xhprof, make sure it isn't lingering anywhere. As far as I can tell, it's not used in the Diagnostic Tool or anything, but feel free to try that out in the Admin panel just in case.

## Types of changes

### Final checklist


Mac-Rae:
Assessed :+1:

Mac-Rae:
Decent Quote :rofl:

PR #8108, opened 2019-10-22
Fix multiple date parameters in reports

JimMackin:
## Description

Fix  #8107 by using the fix suggested there - replace the `$dateCount` keys with the use of `$key`. This ensures that the date parameters work correctly

## Motivation and Context

Needed to fix #8107 - otherwise date parameters can't be updated in some situations.

## How To Test This

    Create a new calls report
    Create a parameter for status set to anything
    Create two date parameters on start date
    Be sure to use the "Date" type with, for example "Now + 12 months"
    Set the second date parameter to something similar but distinct
    Save the report and click update on the parameters
    Examine the updated date parameters - the first will be empty and the second will have taken the firsts settings before this fix and will display correctly afterwards.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [x] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

PR #8110, opened 2019-10-22
Fix #8109 - Reports Incorrect Module path

### LINKED ISSUE #8109: Reports Incorrect Module path
#### Issue

In Reports, when creating a report, the module path can be incorrect when selecting a field to display or a condition. This seems to occur when you use the same module but from 2 different sources e.g using accounts, the the related member of, which is also Accounts. This would use the module path of Accounts for both where it should be  Accounts:member of for the second field.

#### Expected Behaviour
The correct Module path for each field

#### Actual Behaviour
incorrect Module path when using the same module, with 2 different paths

#### Possible Fix

generate the field data based on module path not the module

#### Steps to Reproduce

1. Create a report and select Accounts as the module
2. select the name field from the Account module fields, add it to the report
3. Now select the member of relationship,  select the name field from the member of relationships fields, add it to the report
4.see the module path is the same for both field where it should be different 

#### Context

Report don't work so Medium prority

#### Your Environment

* SuiteCRM Version used:7.11.8
* Environment name and version (e.g. MySQL, PHP 7): PHP 7.1
* Operating System and version (e.g Ubuntu 16.04):18:04

mattlorimer:
Fix issue #8108 

## Description

See issue #8108 

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOR_Reports/AOR_Report_Before.js
+++ b/modules/AOR_Reports/AOR_Report_Before.js
@@ -115,12 +115,12 @@ $(document).ready(function(){
       });
     };
 
-    function loadTreeLeafData(node){
+    function loadTreeLeafData(node) {
       var module = node.module;
       var module_name = node.name;
       var module_path_display = node.module_path_display;
 
-      if(!treeDataLeafs[module]) {
+      if (!treeDataLeafs[module_path_display]) {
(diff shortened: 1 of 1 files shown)


Mac-Rae:
Assessed :+1:

PR #8123, opened 2019-10-23
Add parentheses to fix a deprecated string concatenation

connorshea:
PHP 7.4 warns about upcoming changes to the order of operations when using +/- in a string concatenation, this adds explicit parentheses to fix the deprecation warning.

```
Deprecated: The behavior of unparenthesized expressions containing both '.' and '+'/'-' will change in PHP 8: '+'/'-' will take a higher precedence.
```

Part of #8057.


Mac-Rae:
Assessed :+1:

PR #8154, opened 2019-10-30
Fix a SQL query in the ACLAction code.

### LINKED ISSUE #8153: PHPUnit tests not completing in CI on hotfix-7.10.x
#### Issue
PHPUnit is acting strangely in CI since at least a week ago, specifically on hotfix-7.10.x (I haven't checked other branches yet).

It just suddenly stops after the 64th test is run:

```
.   61 / 1384 (  4%)
	SugarControllerTest::testloadBean ............................. [done]
.	SugarControllerTest::testexecute .............................. [done]
.	SugarControllerTest::testprocess .............................. [done]
The command "./vendor/bin/phpunit --stop-on-failure --stop-on-error --colors --configuration ./tests/phpunit.xml.dist ./tests/unit/phpunit" exited with 0.
```

#### Expected Behavior
The tests should all be run.

#### Actual Behavior
The tests stop after around 64 tests have run. It's almost definitely not a problem with the logging because the tests are taking 2-3 minutes less than they normally would, so the other >1000 unit tests aren't being run.

#### Possible Fix
Not really sure?

You can see the normal behavior, for [this commit](https://github.com/salesagility/SuiteCRM/commit/891ed62f7895c1c80bb76c756a7bce60d7e52ff3), in [this CI job](https://travis-ci.org/salesagility/SuiteCRM/jobs/601242180?utm_medium=notification&utm_source=github_status)

You can see the bad behavior, for [this commit](https://github.com/salesagility/SuiteCRM/commit/0262a79171a3d3e75ef06d3760be4ccb60fd3a7a), in [this CI job](https://travis-ci.org/salesagility/SuiteCRM/jobs/602261388?utm_medium=notification&utm_source=github_status).

#### Steps to Reproduce
1. Look at a recent commit to hotfix-7.10.x
2. Look at the CI jobs for it
3. View the log for the PHP 5.6 or 7.3 jobs
4. See that PHPUnit stops after around 64 tests, but still has an exit code of 0?

#### Context
It means that we could be merging bad code without knowing it.

connorshea:
## Description

The SQL query was causing a bunch of errors in CI and potentially was the cause of the unit tests stopping early in CI.

[`suitecrm.log`](https://www.simpleupload.co.uk/uploads/xpkv7Zsuitecrm.log) was filled with messages like this:

```
Thu Oct 24 16:04:53 2019 [7958][-none-][FATAL]  Query Failed: SELECT * FROM 'acl_actions' WHERE name = 'export' AND category = 'Leads' AND acltype = 'module' AND deleted = 0: MySQL error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ''acl_actions' WHERE name = 'export' AND category = 'Leads' AND acltype = 'module' at line 1
```

This potentially fixes #8153.

## Motivation and Context

This problem was introduced with fa1dbf3fb22666f5784e8c38286ac03b376bf66c, as far as  I can tell.

## How To Test This

Make sure CI passes. Make sure all the acl_actions SQL query errors go away.

## Types of changes

### Final checklist

### DIFF
--- a/modules/ACLActions/ACLAction.php
+++ b/modules/ACLActions/ACLAction.php
@@ -91,12 +91,12 @@ class ACLAction extends SugarBean
             foreach ($ACLActions[$type]['actions'] as $action_name => $action_def) {
                 $action = new ACLAction();
 
-                $tableNameQuoted = $db->quoted($action->table_name);
+                $tableName = $action->table_name;
                 $actionNameQuoted = $db->quoted($action_name);
                 $categoryQuoted = $db->quoted($category);
                 $typeQuoted = $db->quoted($type);
 
-                $query = "SELECT * FROM " . $tableNameQuoted .
+                $query = "SELECT * FROM " . $tableName .
                     " WHERE name = " . $actionNameQuoted .
                     " AND category = " . $categoryQuoted .
                     " AND acltype = " . $typeQuoted .
@@ -138,13 +138,13 @@ class ACLAction extends SugarBean
             foreach ($ACLActions[$type]['actions'] as $action_name => $action_def) {
                 $action = new ACLAction();
 
-                $tableNameQuoted = $db->quoted($action->table_name);
+                $tableName = $action->table_name;
                 $actionNameQuoted = $db->quoted($action_name);
                 $categoryQuoted = $db->quoted($category);
                 $typeQuoted = $db->quoted($type);
 
-                $query = "SELECT * FROM " . $tableNameQuoted .
-                    " WHERE name=" . $actionNameQuoted .
+                $query = "SELECT * FROM " . $tableName .
+                    " WHERE name = " . $actionNameQuoted .
                     " AND category = " . $categoryQuoted .
                     " AND acltype = " . $typeQuoted .
                     " AND deleted = 0";
--- a/modules/ACLActions/ACLAction.php
+++ b/modules/ACLActions/ACLAction.php
@@ -106,7 +106,7 @@ class ACLAction extends SugarBean
 
                 // Only add if an action with that name and category don't exist
                 $row = $db->fetchByAssoc($result);
-                if ($row === null) {
+                if ($row === false) {
                     $action->name = $action_name;
                     $action->category = $category;
                     $action->aclaccess = $action_def['default'];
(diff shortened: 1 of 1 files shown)


connorshea:
Okay, that didn't actually solve the PHPUnit-stopping-early issue :/ I'm not sure what's causing that. It did fix all the bad queries, though.

lazka:
I bisected it:

Install fails for those commits, so I can't test with them.

> $ git bisect skip
There are only 'skip'ped commits left to test.
The first bad commit could be any of:
fa1dbf3fb22666f5784e8c38286ac03b376bf66c
3e102d825b023f5162b87b8190bdb619c7742e18
We cannot bisect more!

lazka:
Reverting both commits fixes things

lazka:
Here is the fix (on top of yours):

```
diff --git a/modules/ACLActions/ACLAction.php b/modules/ACLActions/ACLAction.php
index 555bfd929..b1f0becee 100755
--- a/modules/ACLActions/ACLAction.php
+++ b/modules/ACLActions/ACLAction.php
@@ -106,7 +106,7 @@ class ACLAction extends SugarBean
 
                 // Only add if an action with that name and category don't exist
                 $row = $db->fetchByAssoc($result);
-                if ($row === null) {
+                if ($row === false) {
                     $action->name = $action_name;
                     $action->category = $category;
                     $action->aclaccess = $action_def['default'];
@@ -153,7 +153,7 @@ class ACLAction extends SugarBean
 
                 // Only add if an action with that name and category don't exist
                 $row = $db->fetchByAssoc($result);
-                if ($row !== null) {
+                if ($row !== false) {
                     $action->mark_deleted($row['id']);
                 }
             }
```

lazka:
Here my suggestion on how to avoid this in the future: #8156

connorshea:
The test suite is failing now because of this change: https://github.com/salesagility/SuiteCRM/commit/dd163c5dcae68f9093e7087b2a65ce1002ec20fd#diff-59e5774d555221fc4b93fa67e023a628

I'm not sure how best to fix this since I'm not sure what the test is supposed to be testing.

connorshea:
This was merged as deeae6b74a5bcd606c6496a207b095984b836b19, closing.

PR #8156, opened 2019-10-31
tests: throw an error in case exit() is called during testing

lazka:
## Description

When exit() is called during tests the test suite will stop without returning an error.

To avoid that add our own exit() wrapper which throws an exception instead when
tests are running. There are many more calls to exit() in the code base, this just changes the
common path through sugar_cleanup()

## Motivation and Context

Make tests fail when they fail.

## Types of changes

### Final checklist

### DIFF
--- a/include/utils.php
+++ b/include/utils.php
@@ -3292,6 +3292,22 @@ function pre_login_check()
     }
 }
 
+/**
+ * Like exit() but will throw an exception if called during tests.
+ *
+ * This is to avoid exit() stopping the test suite without us noticing.
+ *
+ * @param int|string $status
+ * @throws Exception
+ */
+function suite_exit($status = 0)
+{
+    if (defined('SUITE_PHPUNIT_RUNNER'))
+        throw new Exception("exit() called during tests with status: $status");
+    else
+        exit($status);
+}
+
 function sugar_cleanup($exit = false)
 {
     static $called = false;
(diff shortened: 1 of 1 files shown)


include/utils.php
```diff
@@ -3292,6 +3292,22 @@ function set_focus() {}
...
 
+/**
+ * Like exit() but will throw an exception if called during tests.
+ *
+ * This is to avoid exit() stopping the test suite without us noticing.
+ *
+ * @param int $status
+ * @throws Exception
+ */
+function sugar_exit($status = 0)
```
Dillon-Brown: Actually, we should probably drop the sugar here since it's a new function.

```suggestion
function suite_exit($status = 0)
```

include/utils.php
```diff
@@ -3309,7 +3325,7 @@ function sugar_cleanup($exit = false)
     //added this check to avoid errors during install.
     if (empty($sugar_config['dbconfig'])) {
         if ($exit) {
-            exit;
+            sugar_exit();
```
Dillon-Brown:
```suggestion
            suite_exit();
```

include/utils.php
```diff
@@ -3344,7 +3360,7 @@ function sugar_cleanup($exit = false)
         $db = DBManagerFactory::getInstance();
         $db->disconnect();
         if ($exit) {
-            exit;
+            sugar_exit();
```
Dillon-Brown:
```suggestion
            suite_exit();
```

lazka:
thanks, done

connorshea:
I think this needs a rebase so it'll pass.

lazka:
Thanks, done.

Mac-Rae:
Assessed :+1:

PR #8163, opened 2019-10-31
Punctuation errors

trishitapingolia:
Fixed punctuation errors.


serhiisamko091184:
Hello @trishitapingolia ,

one more time thank you for the contribution, 
we can accept your contribution only after signing CLA.
We appreciate if it would be possible to recreate this PR with signed CLA in the future.

Many thanks! 

Regards,
Serhii

PR #7987, opened 2019-10-07
Add PHP 7.4 to Travis CI.

connorshea:
## Description
This is essentially the same as #7974, but built on hotfix-7.10.x and with the fix from #7976. I wanted to see if other issues crop up now that that one is fixed.

## Motivation and Context
PHP 7.4 is coming out at the end of November and we should try to be ready for that.

## How To Test This
See if CI passes.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist
- [x] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

PR #8170, opened 2019-11-04
Test PHP 7.4 with no state checker

connorshea:
This is just a test of the changes from #7987 with the changes from #7807. I think a lot of the unit test failures are caused by the state checker, so I wanted to verify that.

I should really set up Travis CI on my fork so I don't need to open PRs like this, but I haven't yet.

Don't merge this.


connorshea:
Yup, from 78 errors with the state checker to 42 without it.

PR #8201, opened 2019-11-06
Fixed #8178 - 7.11.9 Employee role control is lost, invalid

### LINKED ISSUE #8178: 7.11.9 Employee role control is lost, invalid
Employee role permissions are lost and role control does not work.

In 7.11.8 I set up different employee roles and it worked fine. However, after upgrading to 7.11.9, the role is invalid and all employees have the same default permissions.

Dillon-Brown:
## Description

Fixes an issue with ACL roles allowing users to see records they don't have ownership of.

## Motivation and Context

Issue reference: #8178 

## How To Test This

1. Setup your ACL role the same as the screenshot in #8178.
2. Check that your user has the correct level of access.

## Types of changes

### Final checklist

### DIFF
--- a/modules/ACLActions/ACLAction.php
+++ b/modules/ACLActions/ACLAction.php
@@ -359,20 +359,20 @@ class ACLAction extends SugarBean
         /* END - SECURITY GROUPS */
         while ($row = $db->fetchByAssoc($result, false)) {
             /* BEGIN - SECURITY GROUPS */
-            if ($has_user_role === false && $row['user_role'] === 1) {
+            if ($has_user_role == false && $row['user_role'] == 1) {
                 $has_user_role = true;
             }
-            if ($has_role === false && ($row['user_role'] === 1 || $row['user_role'] === 0)) {
+            if ($has_role == false && ($row['user_role'] == 1 || $row['user_role'] ==0)) {
                 $has_role = true;
             }
             //if user roles should take precedence over group roles and we have a user role
             //break when we get to processing the group roles
-            if ($has_user_role === true && $row['user_role'] === 0
+            if ($has_user_role == true && $row['user_role'] == 0
                 && isset($sugar_config['securitysuite_user_role_precedence'])
-                && $sugar_config['securitysuite_user_role_precedence'] === true) {
+                && $sugar_config['securitysuite_user_role_precedence'] == true) {
                 break;
             }
-            if ($row['user_role'] === -1 && $has_role === true) {
+            if ($row['user_role'] == -1 && $has_role == true) {
                 break; //no need for default actions when a role is assigned to the user or user's group already
             }
             /* END - SECURITY GROUPS */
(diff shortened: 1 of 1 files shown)


AdminPL:
After applying this patch users still see records in report module which they are not supposed to see on list. ( thankfully cannot edit or view the record ).

Same thing happens with  moudles created by module builder.
Users in security group set to to only  list and view  of their own records  see the list of all records, can't view or edit.

Dillon-Brown:
Hi @AdminPL, I've pushed an extra fix to this PR but so far I haven't been able to replicate your issue. Perhaps you also need the fix here? #8188

AdminPL:
@Dillon-Brown When i apply both fixes:

Modules:
Quotes and Invoices works fine. I have3 type ofsecurity users and all ACLS works as intended but on custom module problem remains - user can see list view.

When i apply on your patch custom modules works fine, but in "normal" ACLS don't work for list view.

Dillon-Brown:
@AdminPL, hmm interesting. This is what I'm seeing:

Role:
role

With both fixes:
withfix

Without fixes:
withoutfix

AdminPL:
I made quick rebuild and right now ( both fiex apllied )

This the invoice module.
image


On the left site i am logged in as administrator and on the left side i have incognito window logged in as user. 


In custom module i had to hide personal data, but no rules are applied. User can see and access all data.
image

Reports module somehow works fine ?

I had no errors while update and no errors related with this specific action in any logs

Dillon-Brown:
@AdminPL Perfect, I can replicate. Thanks!

Dillon-Brown:
@AdminPL Pushed up a fix to this PR that should make ACLAction work the same as it did in 7.10.20.

AdminPL:
@Dillon-Brown Seems everything is fine right now. I'll do some extended testing before moving production to 7.11.9 and give you more feeedback if I find anything.

Dillon-Brown:
@AdminPL :+1: Just so you're aware we'll be looking at putting out a 7.11.10 and 7.10.22 release to resolve this and a few other issues either today or tomorrow.

AdminPL:
There is one more bug to be done ( it is present also in 7.11.8 ) When you set "view" and "list" to group. Records of the owner works fine ( user can enter ) , but records which are in group but user is not the owner are only listed but can't be accessed.

Dillon-Brown:
@AdminPL Seems to be working for me after I ran repair roles.

Role:
Role

Contact module with a single record assigned to "max" and in a security group with the user "jim":
Contacts

AdminPL:
Repair roles doesn't work for me.
image

There is no trace in logs

samus-aran:
> Repair roles doesn't work for me.
> image
> 
> There is no trace in logs

Hey @AdminPL We haven't been able to replicate this issue. Could we open a new ticket specific for that issue? We are merging this it helps resolves other issues.

AdminPL:
@samus-aran Should i create new ticket ?

PR #8257, opened 2019-11-14
Fix #8261, Fix #7242 - Handling of temp files during Upgrades

### LINKED ISSUE #7242: Upgrade stuck on "Upgrade Scripts in Progress" till timeout
#### Issue
Upgrade of SuiteCRM 7.11.3 to 7.11.4 stuck on "Upgrade Scripts in Progress" till browser error "The connection was reset" (probably because of some timeout).

#### Expected Behavior
The last steps of upgrade should be completed successfully.

#### Actual Behavior
Browser error "The connection was reset".

upgradeWizard.log last rows:
> Wed, 01 May 2019 09:48:29 +0300 [UpgradeWizard] - Cleaning up the session.  Goodbye.
> Wed, 01 May 2019 09:48:29 +0300 [UpgradeWizard] - at unlinkUWTempFiles()

#### Possible Fix
Restarting Apache after browser error.

#### Steps to Reproduce
1. Admin -> Upgrade Wizard
2. Uploading 7.11.X-TO-7.11.4 file
3. Completing (successfully) all the steps till "Upgrade Scripts in Progress"

#### Context
1. Although the error, the upgrade looks like been applied successfully, at least per About menu and working SuiteCRM (after Apache restart).
2. This issue observed on each 7.11 point release.
3. Next upgrade not possible till all cache/upgrades/temp directory are cleaned.
4. If relevant, Apache:
> Max Requests -> Per Child: 0 - Keep Alive: on - Max Per Connection: 100
> Timeouts -> Connection: 300 - Keep-Alive: 5

#### Your Environment
* SuiteCRM Version used: 7.11.3 (before upgrade)
* Browser name and version: Firefox 66.0.3 (64-bit)
* Environment name and version: MySQL 5.7.26-0ubuntu0.18.04.1, PHP 7 7.2.17-0ubuntu0.18.04.1, Apache 2.4.29, localhost
* Operating System and version: Kubuntu 18.04.2 (64-bit)

### LINKED ISSUE #8261: Interrupted / Hanging Upgrades
#### Issue
Several manifestations of Upgrade problems. The most frequent are the upgrade hanging at a certain step, or at the very end (in which case the upgrade actually succeeds in taking you to the next version). Sometimes there is a HTTP 500 error sent to the client.

Sometimes there is a `php zend_mm_heap corrupted` in the Apache log.

The problem occurs due to long delays in cleaning up the temp directory, which crash PHP in a way that doesn't leave any traces in the logs most of the times... the problem tends to aggravate significantly after occurring more than once, since the files accumulate in this directory.

#### Quick Workaround
If you just want to get your upgrade done, often all that is needed is to clear the entire contents of `cache/upgrades/temp` and start again.

#### Actual Behavior
The [function `unlinkUWTempFiles`](https://github.com/salesagility/SuiteCRM/blob/v7.11.10/modules/UpgradeWizard/uw_utils.php#L2258) calls a [function `findAllFiles`](https://github.com/salesagility/SuiteCRM/blob/v7.11.10/include/dir_inc.php#L206) which is _very_ inefficient for large directories. It recursively goes through every file in the directory tree, adding each element to an array, then sorting it, then sorting it again, and then deleting each file individually.

So it's a long period of disk read activity, starving the process memory, followed by a long period of CPU intensive operations, followed by a long period of disk write activity.

#### Possible Fix
This sort of directory iteration should be done using a PHP language construct called `Iterator`. Instead of making a huge array of files first, and then looping through each item to perform some function, it should just get the necessary item at the moment it needs to act on it. No significant memory consumption, and much simpler code. I will provide a PR.

#### Steps to Reproduce
1. Run an upgrade to one of the newest versions, like 7.10.9 or 7.10.10.
2. Keep an eye on `cache/upgrads/temp` to see how many copies of the unzipped package you have there. They add up after each time the process hangs.
3. If necessary to help reproduce, make these copies yourself. Fill that directory with tens of thousands of files. Each package has around 22k. I had a real-life situation with over 90k files.
4. After the upgrades conclude, this directory should be completely empty. If not, then this issue occurred.

#### Additional tools that might help (Linux, needs `sudo`)

* `htop` - a fancier `top`. Shows CPU usage and other details. when you see an Apache process that is taking up 100% of the CPU, note its `PID`.
* `iotop` see aggregated disk activity, by process. Similar to `top` or `htop`, but for disk access.
* `lsof -n -p 1234 | grep -v mem` - get a momentary glimpse of what a process with PID `1234` is doing in terms of file accesses, file by file. You can repeat this command and see that it's iterating the directory, and check if it's reading or writing.
* `watch` - repeats a command every `n` seconds. Useful to leave separate terminals showing you regular updates of the above tools' output. I used these 
```bash
Observe file accesses of process with PID 1234:
watch -n 2 'lsof -n -p 1234 | grep -v mem'

Track directory size dynamically:
watch -n 2 'du -h --max-depth=1 cache/upgrades/temp'
```

#### Environment
* SuiteCRM Version used: problems started around version 7.10.7 up to 7.10.10

pgorod:
Fixes many Upgrader problems caused by the increase of our upgrade packages size, and aggravated each time the upgrade didn't finish correctly.

## Description
See the Issue #8261 for symptoms and diagnosis tools. Also #7242.

## Future work

The functions [`uwfindAllFiles`](https://github.com/salesagility/SuiteCRM/search?q=uwFindAllFiles&unscoped_q=uwFindAllFiles) and [`findAllFiles`](https://github.com/salesagility/SuiteCRM/search?q=findAllFiles&unscoped_q=findAllFiles) should be dropped entirely, and the places that use them should be refactored so that they use Iterators and don't build huge arrays in memory before acting on the files.

Exceptions to this can be made if you're sure the place where the function is used will never have to iterate over a large set of files.

## How To Test This

**Warning**: if you just pull these changes from GitHub and then run an upgrade, the files in the package will replace these and you won't test anything. Make sure you add these files _into_ the package before starting the test upgrade. 

See the Issue's "steps to reproduce" for steps to test.

## Types of changes

### Final checklist

### DIFF
--- a/modules/UpgradeWizard/systemCheck.php
+++ b/modules/UpgradeWizard/systemCheck.php
@@ -58,7 +58,13 @@ $skipDirs = array(
     '.svn',
     '.git',
 );
-$files = uwFindAllFiles(getcwd(), array(), true, $skipDirs);
+
+$files = new RecursiveIteratorIterator(
+    new RecursiveDirectoryIterator(
+        getcwd(),
+        RecursiveDirectoryIterator::SKIP_DOTS | RecursiveIteratorIterator::SELF_FIRST
+    )
+);
 
 $i=0;
 $filesOut = "
@@ -74,6 +80,10 @@ $filesOut = "
 
 $isWindows = is_windows();
 foreach ($files as $file) {
+    if (dirInArray($file, $skipDirs)) {
+        continue;
+    }
+
     if ($isWindows) {
         if (!is_writable_windows($file) && file_exists($file)) {
             logThis('WINDOWS: File ['.$file.'] not readable - saving for display');
--- a/modules/UpgradeWizard/uw_utils.php
+++ b/modules/UpgradeWizard/uw_utils.php
@@ -118,8 +118,13 @@ function commitMakeBackupFiles($rest_dir, $install_file, $unzip_dir, $zip_from_d
 
     if (file_exists($rest_dir) && is_dir($rest_dir)) {
         logThis('backing up files to be overwritten...', $path);
-        $newFiles = findAllFiles(clean_path($unzip_dir . '/' . $zip_from_dir), array());
-
+        $newFiles = new RecursiveIteratorIterator(
+            new RecursiveDirectoryIterator(
+                clean_path($unzip_dir . '/' . $zip_from_dir),
+                RecursiveDirectoryIterator::SKIP_DOTS | RecursiveIteratorIterator::SELF_FIRST
+            )
+        );
+	    
         // keep this around for canceling
         $_SESSION['uw_restore_dir'] = getUploadRelativeName($rest_dir);
 
@@ -187,7 +192,12 @@ function commitCopyNewFiles($unzip_dir, $zip_from_dir, $path='')
         }
     }
 
-    $newFiles = findAllFiles(clean_path($unzip_dir . '/' . $zip_from_dir), array());
+    $newFiles = new RecursiveIteratorIterator(
+        new RecursiveDirectoryIterator(
+            clean_path($unzip_dir . '/' . $zip_from_dir),
+            RecursiveDirectoryIterator::SKIP_DOTS | RecursiveIteratorIterator::SELF_FIRST
+        )
+    );
     $zipPath = clean_path($unzip_dir . '/' . $zip_from_dir);
 
     // handle special do-not-overwrite conditions
@@ -525,7 +535,7 @@ function getAllCustomizedModules()
 }
 
 /**
- * Array of all Modules in the version bein upgraded
+ * Array of all Modules in the version being upgraded
  * This method returns an Array of all modules
  * @return $modules Array of modules.
  */
@@ -713,7 +723,10 @@ function deleteChance()
  */
 function upgradeUWFiles($file)
 {
+    global $base_tmp_upgrade_dir;
+
     $cacheUploadUpgradesTemp = mk_temp_dir(sugar_cached("upgrades/temp"));
+    $_SESSION['unzip_dir'] = strstr($cacheUploadUpgradesTemp, $base_tmp_upgrade_dir);
 
     unzip($file, $cacheUploadUpgradesTemp);
 
@@ -733,29 +746,26 @@ function upgradeUWFiles($file)
     }
     // upgradeWizard
     if (file_exists("$from_dir/modules/UpgradeWizard")) {
-        $allFiles[] = findAllFiles("$from_dir/modules/UpgradeWizard", $allFiles);
+        $allFiles[] = findAllFiles("$from_dir/modules/UpgradeWizard", []);
     }
     // moduleInstaller
     if (file_exists("$from_dir/ModuleInstall")) {
-        $allFiles[] = findAllFiles("$from_dir/ModuleInstall", $allFiles);
+        $allFiles[] = findAllFiles("$from_dir/ModuleInstall", []);
     }
     if (file_exists("$from_dir/include/javascript/yui")) {
-        $allFiles[] = findAllFiles("$from_dir/include/javascript/yui", $allFiles);
+        $allFiles[] = findAllFiles("$from_dir/include/javascript/yui", []);
     }
     if (file_exists("$from_dir/HandleAjaxCall.php")) {
         $allFiles[] = "$from_dir/HandleAjaxCall.php";
     }
     if (file_exists("$from_dir/include/SugarTheme")) {
-        $allFiles[] = findAllFiles("$from_dir/include/SugarTheme", $allFiles);
+        $allFiles[] = findAllFiles("$from_dir/include/SugarTheme", []);
     }
     if (file_exists("$from_dir/include/SugarCache")) {
-        $allFiles[] = findAllFiles("$from_dir/include/SugarCache", $allFiles);
+        $allFiles[] = findAllFiles("$from_dir/include/SugarCache", []);
     }
     if (file_exists("$from_dir/include/utils/external_cache.php")) {
         $allFiles[] = "$from_dir/include/utils/external_cache.php";
-    }
-    if (file_exists("$from_dir/include/upload_file.php")) {
-        $allFiles[] = "$from_dir/include/upload_file.php";
     }
     if (file_exists("$from_dir/include/file_utils.php")) {
         $allFiles[] = "$from_dir/include/file_utils.php";
@@ -2233,6 +2243,59 @@ if (!function_exists('validate_manifest')) {
     }
 }
 
+/**
+ * upgradeSugarCache
+ * @deprecated This function is unused and will be removed in a future release.
+ * change from using the older SugarCache in 6.1 and below to the new one in 6.2
+ */
+function upgradeSugarCache($file)
+{
+    global $sugar_config;
+    $cacheUploadUpgradesTemp = mk_temp_dir(sugar_cached('upgrades/temp'));
+    unzip($file, $cacheUploadUpgradesTemp);
+    if (!file_exists(clean_path("{$cacheUploadUpgradesTemp}/manifest.php"))) {
+        logThis("*** ERROR: no manifest file detected while bootstraping upgrade wizard files!");
+        return;
+    }
+    include(clean_path("{$cacheUploadUpgradesTemp}/manifest.php"));
+    $from_dir = "{$cacheUploadUpgradesTemp}/{$manifest['copy_files']['from_dir']}";
+    $allFiles = array();
+    if (file_exists("$from_dir/include/SugarCache")) {
+        $allFiles = findAllFiles("$from_dir/include/SugarCache", $allFiles);
+    }
+    if (file_exists("$from_dir/include/database")) {
+        $allFiles = findAllFiles("$from_dir/include/database", $allFiles);
+    }
+    if (file_exists("$from_dir/include/utils/external_cache.php")) {
+        $allFiles[] = "$from_dir/include/utils/external_cache.php";
+    }
+    if (file_exists("$from_dir/include/utils/sugar_file_utils.php")) {
+        $allFiles[] = "$from_dir/include/utils/sugar_file_utils.php";
+    }
+    if (file_exists("$from_dir/include/utils/sugar_file_utils.php")) {
+        $allFiles[] = "$from_dir/include/utils/sugar_file_utils.php";
+    }
+    if (file_exists("$from_dir/include/utils/autoloader.php")) {
+        $allFiles[] = "$from_dir/include/utils/autoloader.php";
+    }
+    foreach ($allFiles as $k => $file) {
+        $destFile = str_replace($from_dir . "/", "", $file);
+        if (!is_dir(dirname($destFile))) {
+            mkdir_recursive(dirname($destFile)); // make sure the directory exists
+        }
+        if (stristr($file, 'uw_main.tpl')) {
+            logThis('Skipping "' . $file . '" - file copy will during commit step.');
+        } else {
+            logThis('updating UpgradeWizard code: ' . $destFile);
+            copy_recursive($file, $destFile);
+        }
+    }
+}
+
+/**
+ * unlinkUploadFiles
+ * @deprecated This function is unused and will be removed in a future release.
+ */
 function unlinkUploadFiles()
 {
     return;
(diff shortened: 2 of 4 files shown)


pgorod:
I marked this as work-in-progress only because I would like to make further improvements to the upgrade code, especially to add more significant logging.

But what I've done so far should work, so if anyone wants to grab it and solve their upgrade problems, please go ahead and do it, and then come back here and tell us how it went. Thanks.

Mac-Rae:
@pgorod Will give it a test today at some point :)

Mac-Rae:
This works really good!!

So much quicker :+1:

Mac-Rae:
Other than the random `<?php` floating on line 2251 that needed to be removed before use

pgorod:
@Mac-Rae I added a couple of commits. You should see a 10x increase in speed in the "checking permissions" part. 🚀 

Hopefully this is going to make the upgrade tests so much faster that it will be measurable in the GDP of Scotland next year.

I'd like to ask you to please confirm if this works well on 

- older PHP's (5.6)
- Windows!

pgorod:
@Mac-Rae @samus-aran @Dillon-Brown I made significant changes, this needs to be retested.

The multiple copies of unzipped package in `cache/upgrades/temp` should be gone. Extra speed gains here, around 85% of the delays in an upgrade were.... bugs or bad code.

Let me just underline that I haven't tested this in Windows, I'm relying on you guys for this - and it's messing with paths so we should really make sure it works. 

I also made another dubious decision - completely dropping the `upgradeSugarCache` function, which was responsible for one the extra copies, just to grab a dozen files from the zip. I think these are not necessary to be copied beforehand, it looks like a hack specific to an upgrade from SugarCRM 6.1 to 6.2 (the comment says so). 

Please take a look and tell me what you think. If you think some (or all) of those files need to be pre-copied, we should just add them to the other list in `upgradeUWFiles`. 

I don't plan to do any more changes, unless some problem comes up. Thanks.

pgorod:
If I may suggest a more thorough test for these changes: instead of just checking whether the upgrade completes, if possible

1. Test the normal upgrade, before any changes. (BTW if possible, record all time spent waiting for processes to complete, excluding user time)

2. Grab the entire resulting `SuiteCRM` directory and store a copy of it somewhere.

3. Apply the changes in this PR into a new package and install it, also recording time.

4. Finish by making a `diff` between the results of one install and the results of another.

5. Repeat on Windows OS

6. Repeat for PHP 5.6 at least on Linux

Thanks.

Mac-Rae:
@pgorod You have been busy!! :1st_place_medal: 

Will take a look at this today :)

Mac-Rae:
@pgorod,
Looks good to me :) Assessed :+1:

modules/UpgradeWizard/systemCheck.php
```diff
@@ -58,7 +58,12 @@
     '.svn',
     '.git',
 );
-$files = uwFindAllFiles(getcwd(), array(), true, $skipDirs);
+
+$files = new RecursiveIteratorIterator(
```
Dillon-Brown: PSR-2:

```php
$files = new RecursiveIteratorIterator(
    new RecursiveDirectoryIterator(
        getcwd(),
        RecursiveDirectoryIterator::SKIP_DOTS | RecursiveIteratorIterator::SELF_FIRST)
);
```

modules/UpgradeWizard/systemCheck.php
```diff
@@ -74,6 +79,10 @@
 
 $isWindows = is_windows();
 foreach ($files as $file) {
+    if (dirInArray($file, $skipDirs)) {
+       continue;
```
Dillon-Brown: PSR-2
```suggestion
        continue;
```

modules/UpgradeWizard/uw_utils.php
```diff
@@ -118,8 +118,12 @@ function commitMakeBackupFiles($rest_dir, $install_file, $unzip_dir, $zip_from_d
 
     if (file_exists($rest_dir) && is_dir($rest_dir)) {
         logThis('backing up files to be overwritten...', $path);
-        $newFiles = findAllFiles(clean_path($unzip_dir . '/' . $zip_from_dir), array());
-
+        $newFiles = new RecursiveIteratorIterator(
```
Dillon-Brown: PSR-2:

```php
        $newFiles = new RecursiveIteratorIterator(
            new RecursiveDirectoryIterator(
                clean_path($unzip_dir . '/' . $zip_from_dir),
                RecursiveDirectoryIterator::SKIP_DOTS | RecursiveIteratorIterator::SELF_FIRST)
        );
```

modules/UpgradeWizard/uw_utils.php
```diff
@@ -187,7 +191,11 @@ function commitCopyNewFiles($unzip_dir, $zip_from_dir, $path='')
         }
     }
 
-    $newFiles = findAllFiles(clean_path($unzip_dir . '/' . $zip_from_dir), array());
+    $newFiles = new RecursiveIteratorIterator(
```
Dillon-Brown: PSR-2:
```php
    $newFiles = new RecursiveIteratorIterator(
        new RecursiveDirectoryIterator(
            clean_path($unzip_dir . '/' . $zip_from_dir),
            RecursiveDirectoryIterator::SKIP_DOTS | RecursiveIteratorIterator::SELF_FIRST)
    );
```

modules/UpgradeWizard/uw_utils.php
```diff
@@ -733,30 +744,27 @@ function upgradeUWFiles($file)
     }
     // upgradeWizard
     if (file_exists("$from_dir/modules/UpgradeWizard")) {
-        $allFiles[] = findAllFiles("$from_dir/modules/UpgradeWizard", $allFiles);
+        $allFiles[] = findAllFiles("$from_dir/modules/UpgradeWizard", array());
```
Dillon-Brown: Short array syntax is preferred:
```suggestion
        $allFiles[] = findAllFiles("$from_dir/modules/UpgradeWizard", []);
```

modules/UpgradeWizard/uw_utils.php
```diff
@@ -733,30 +744,27 @@ function upgradeUWFiles($file)
...
     }
     // moduleInstaller
     if (file_exists("$from_dir/ModuleInstall")) {
-        $allFiles[] = findAllFiles("$from_dir/ModuleInstall", $allFiles);
+        $allFiles[] = findAllFiles("$from_dir/ModuleInstall", array());
```
Dillon-Brown:
```suggestion
        $allFiles[] = findAllFiles("$from_dir/ModuleInstall", []);
```

modules/UpgradeWizard/uw_utils.php
```diff
@@ -733,30 +744,27 @@ function upgradeUWFiles($file)
...
     }
     if (file_exists("$from_dir/include/javascript/yui")) {
-        $allFiles[] = findAllFiles("$from_dir/include/javascript/yui", $allFiles);
+        $allFiles[] = findAllFiles("$from_dir/include/javascript/yui", array());
```
Dillon-Brown:
```suggestion
        $allFiles[] = findAllFiles("$from_dir/include/javascript/yui", []);
```

modules/UpgradeWizard/uw_utils.php
```diff
@@ -733,30 +744,27 @@ function upgradeUWFiles($file)
...
     }
     if (file_exists("$from_dir/HandleAjaxCall.php")) {
         $allFiles[] = "$from_dir/HandleAjaxCall.php";
     }
     if (file_exists("$from_dir/include/SugarTheme")) {
-        $allFiles[] = findAllFiles("$from_dir/include/SugarTheme", $allFiles);
+        $allFiles[] = findAllFiles("$from_dir/include/SugarTheme", array());
```
Dillon-Brown:
```suggestion
        $allFiles[] = findAllFiles("$from_dir/include/SugarTheme", []);
```

modules/UpgradeWizard/uw_utils.php
```diff
@@ -733,30 +744,27 @@ function upgradeUWFiles($file)
...
     }
     if (file_exists("$from_dir/include/SugarCache")) {
-        $allFiles[] = findAllFiles("$from_dir/include/SugarCache", $allFiles);
+        $allFiles[] = findAllFiles("$from_dir/include/SugarCache", array());
```
Dillon-Brown:
```suggestion
        $allFiles[] = findAllFiles("$from_dir/include/SugarCache", []);
```

modules/UpgradeWizard/uw_utils.php
```diff
@@ -2248,42 +2256,54 @@ function unlinkUploadFiles()
...
+ * @param string $folder         The directory path.
+ * @param bool   $keepRootFolder Whether to keep the top-level folder.
+ *
+ * @return bool TRUE on success, otherwise FALSE.
+ */
+function deleteTree($folder, $keepRootFolder = false) {
+	// Handle bad arguments.
+	if (empty($folder) || !file_exists($folder)) {
+		return true; // No such file/folder exists.
+	} elseif (is_file($folder) || is_link($folder)) {
```
Dillon-Brown: This can be converted into an if branch for lower complexity:

```php
    // Handle bad arguments.
    if (empty($folder) || !file_exists($folder)) {
        // No such file/folder exists.
        return true;
    }

    if (is_file($folder) || is_link($folder)) {
        // Delete file/link.
        return @unlink($folder);
    }
```

modules/UpgradeWizard/uw_utils.php
```diff
@@ -2248,42 +2256,54 @@ function unlinkUploadFiles()
...
 }
 
+/**
+ * Recursively deletes a directory tree.
+ *
+ * @param string $folder         The directory path.
+ * @param bool   $keepRootFolder Whether to keep the top-level folder.
+ *
+ * @return bool TRUE on success, otherwise FALSE.
+ */
```
Dillon-Brown: Spacing:

```
/**
 * Recursively deletes a directory tree.
 *
 * @param string $folder The directory path.
 * @param bool $keepRootFolder Whether to keep the top-level folder.
 *
 * @return bool TRUE on success, otherwise FALSE.
 */
```

modules/UpgradeWizard/uw_utils.php
```diff
@@ -2248,42 +2256,54 @@ function unlinkUploadFiles()
...
+function deleteTree($folder, $keepRootFolder = false) {
+	// Handle bad arguments.
+	if (empty($folder) || !file_exists($folder)) {
+		return true; // No such file/folder exists.
+	} elseif (is_file($folder) || is_link($folder)) {
+		return @unlink($folder); // Delete file/link.
+	}
+
+	// Delete all children.
+	$files = new \RecursiveIteratorIterator(
```
Dillon-Brown: PSR-2:

```
    $files = new RecursiveIteratorIterator(
        new RecursiveDirectoryIterator($folder, RecursiveDirectoryIterator::SKIP_DOTS),
        RecursiveIteratorIterator::CHILD_FIRST
    );
```

modules/UpgradeWizard/uw_utils.php
```diff
@@ -2248,42 +2256,54 @@ function unlinkUploadFiles()
...
+		new \RecursiveDirectoryIterator($folder, \RecursiveDirectoryIterator::SKIP_DOTS),
+		\RecursiveIteratorIterator::CHILD_FIRST
+	);
+
+	foreach ($files as $fileinfo) {
```
Dillon-Brown: Variables should be camelcase:
```php
    foreach ($files as $fileInfo) {
        $action = ($fileInfo->isDir() ? 'rmdir' : 'unlink');
        if (!@$action($fileInfo->getRealPath())) {
            // Abort due to the failure.
            return false;
        }
    }
```

modules/UpgradeWizard/uw_utils.php
```diff
@@ -2248,42 +2256,54 @@ function unlinkUploadFiles()
...
-                //logThis('removing dir ['.$dir.']', $path);
-                @rmdir($dir);
-            }
-        }
-        $cacheFile = sugar_cached("modules/UpgradeWizard/_persistence.php");
-        if (is_file($cacheFile)) {
-            logThis("Unlinking Upgrade cache file: '_persistence.php'", $path);
-            @unlink($cacheFile);
-        }
+    $cacheFile = sugar_cached("modules/UpgradeWizard/_persistence.php");
```
Dillon-Brown: Single quotes where possible:
```suggestion
    $cacheFile = sugar_cached('modules/UpgradeWizard/_persistence.php');
```

modules/UpgradeWizard/uw_utils.php
```diff
@@ -4357,58 +4377,6 @@ function add_unified_search_to_custom_modules_vardefs()
     }
 }
 
-/**
- * change from using the older SugarCache in 6.1 and below to the new one in 6.2
- */
-function upgradeSugarCache($file)
```
Dillon-Brown: We can deprecate this but we wouldn't be able to completely remove it in a minor release.

modules/UpgradeWizard/uw_utils.php
```diff
@@ -4547,6 +4515,23 @@ function whetherNeedToSkipDir($dir, $skipDirs)
     return false;
 }
 
+/**
+ * Whether directory exists within list of directories, matching only beginning of path
+ * @param string $dir dir to be checked
+ * @param array $dirsArray the list of directories to check 
+ * @return boolean
+ */
+function dirInArray($dir, $dirsArray)
```
Dillon-Brown: PSR-2:

```php
/**
 * Whether directory exists within list of directories, matching only beginning of path
 * @param string $dir dir to be checked
 * @param array $dirsArray the list of directories to check
 * @return boolean
 */
function dirInArray($dir, $dirsArray)
{
    $dir = clean_path($dir);
    foreach ($dirsArray as $aDir) {
        // Match the substring only at the beginning of the path:
        if (strpos($dir, getcwd() . '/' . $aDir) === 0) {
            return true;
        }
    }

    return false;
}
```

pgorod:
@Dillon-Brown finished making the requested changes. Thanks for the detailed code review.

<academic discussion>

Are you sure about removing the `\` referring the Iterator classes to their version from the global namespace? I put that without really minding it, just copying code snippets and checking they worked. 

I don't think this change will make any difference, since we're not redefining these classes anyway, I just wonder why they were there in the first place, maybe there are some scenarios with redefinitions where that might take effect...

</academic discussion>

Dillon-Brown:
@pgorod For the removal of the ```\```, I agree the change won't really make any difference (Except to remove the little yellow lightbulb in my IDE :D). Iirc at the opcode level the only difference is that calling functions in a namespace results in an additional opcode in order to check if the called function is present in the namespace and run it, or revert to the global namespace if it doesn't exist.

Technically you can gain an extremely small performance boost this way: https://github.com/nilportugues/php-backslasher but most major frameworks decided against implementing it as a standard.

pgorod:
Dillon, while testing, did you get a chance to record the time this was taking, versus what it is taking now? If we had that we might want to mention it on the posts describing "what's new" in the next release.

Dillon-Brown:
> Dillon, while testing, did you get a chance to record the time this was taking, versus what it is taking now? If we had that we might want to mention it on the posts describing "what's new" in the next release.

Not a bad idea, I believe @Mac-Rae will be testing this and some other upgrade fixes this week so I'll see if I can get him to record some times that we can then do a comparison with.

PR #8258, opened 2019-11-14
Fix "!" in pQuery and add tests

### LINKED ISSUE #7720: The pQuery method handles ! incorrectly for MySQL
#### Issue
I've been using the `pQuery` method on DBManager recently and I noticed that occasionally it seemed like it was breaking, and I couldn't figure out a pattern.

https://github.com/salesagility/SuiteCRM/blob/23782c12d8cfc927925a4582d067bed62921e115/include/database/DBManager.php#L1925

Today, I think I figured out the source of the issue. `!` is treated as a special character, and so the tokenizer treats it as such. Unfortunately, the tokenizer considers it a special character even when the `!` is part of a `!=` comparison operator.

#### Expected Behavior
```php
$db = DBManagerFactory::getInstance();
$db->pQuery("SELECT * FROM users WHERE id != '1'); #=> valid mysqli_object with data on all the other user records
$db->pQuery("SELECT * FROM users WHERE id <> '1'); #=> valid mysqli_object with data on all the other user records
```

#### Actual Behavior

```php
$db = DBManagerFactory::getInstance();
$db->pQuery("SELECT * FROM users WHERE id != '1'); #=> empty mysqli_object instance
$db->pQuery("SELECT * FROM users WHERE id <> '1'); #=> valid mysqli_object with data on all the other user records
```

#### Possible Fix
Improve the regular expressions/switch-case statements used when preparing the query so they don't count `!=` as a token.

#### Steps to Reproduce
1. Create code like that in the expected/actual behavior section.
2. Run the code and see what it outputs.
3. See that it's not valid.
4. See that, if you instead use `$db->pQuery("SELECT * FROM users WHERE id <> '1');`, it works fine. (`!=` and `<>` are aliases in MySQL)

#### Context
This pretty much breaks any ability to use the `pQuery` method since it's such a big landmine.

#### Your Environment
* SuiteCRM Version used: SuiteCRM 7.10.19
* Operating System and version (e.g Ubuntu 16.04): macOS Mojave 10.14.6

connorshea:
## Description

Fixes #7720 and adds tests for some basic pQuery functionality, to prevent it from regressing.

## Motivation and Context
A query like `SELECT foo FROM bar WHERE baz != '?';` would have failed when passed to pQuery because the `!` isn't handled in any special way when it precedes an `=`.

`!` can be used as a token, but it shouldn't conflict with valid SQL queries.

It changes the way pQuery works, by renaming `executePreparedQuery` to `createPreparedQuery` and instead handle the execution of the queries in pQuery itself. `executePreparedQuery` was only used in the `pQuery` function. The point of this change is to make pQuery easier to test. I wanted to make sure the queries were returned by a function as strings, since the results of the queries aren't relevant to me and would just add complexity to the tests.

## How To Test This
Make sure the automated tests pass.

## Types of changes

### Final checklist

### DIFF
--- a/include/database/DBManager.php
+++ b/include/database/DBManager.php
@@ -1851,6 +1851,12 @@ abstract class DBManager
                 case '?':
                 case '!':
                 case '&':
+                    // If the next character after the `!` is an `=`, don't
+                    // try to replace it. `!=` is valid SQL and shouldn't be
+                    // replaced like it's a token in a prepared statement.
+                    if ($val === '!' && strpos($tokens[$key + 1], '=') === 0) {
+                        break;
+                    }
                     $count++;
                     $sqlStr .= '?';
                     break;
@@ -1876,7 +1882,7 @@ abstract class DBManager
      * @param  array $data The array of data to replace the tokens with.
      * @return resource result set or false on error
      */
-    public function executePreparedQuery($stmt, $data = array())
+    public function createPreparedQuery($stmt, $data = array())
     {
         if (!empty($this->preparedTokens[$stmt])) {
             if (!is_array($data)) {
@@ -1895,7 +1901,7 @@ abstract class DBManager
             $query = '';
             $dataIndex = 0;
             $tokens = $pTokens['tokens'];
-            foreach ($tokens as $val) {
+            foreach ($tokens as $key => $val) {
                 switch ($val) {
                     case '?':
                         $query .= $this->quote($data[$dataIndex++]);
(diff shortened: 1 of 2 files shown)


include/database/DBManager.php
```diff
@@ -1928,7 +1931,9 @@ public function pQuery($sql, $data = array())
     {
         $stmt = $this->prepareQuery($sql);
 
-        return $this->executePreparedQuery($stmt, $data);
+        $query = $this->createPreparedQuery($stmt, $data);
+
+        return $this->query($query);
```
connorshea: What does `$this->query()` do if it's given `false`. 🤔  Should maybe switch this up so there's a check for `$query` being false.

Mac-Rae:
Assessed :+1:

include/database/DBManager.php
```diff
@@ -1876,7 +1879,7 @@ public function prepareQuery($sql)
      * @param  array $data The array of data to replace the tokens with.
      * @return resource result set or false on error
      */
-    public function executePreparedQuery($stmt, $data = array())
+    public function createPreparedQuery($stmt, $data = array())
```
Dillon-Brown: I don't think we'd be able to rename for backwards-compatibility reasons since this is a public function. We'd need to either keep it as ```executePreparedQuery``` or make a new ```createPreparedQuery``` function and deprecate this one.

connorshea:
@Dillon-Brown Rebased the branch on top of hotfix-7.10.x and restored the `executePreparedQuery` function. I've marked it as deprecated.

Mac-Rae:
Assessed :+1:

PR #8326, opened 2019-11-28
Fix #2464 - logo upload

### LINKED ISSUE #2464: Logo upload function is not working

pgorod:
## Description
Please see Issue #2464

Another user recently confirmed in the forums that this works
https://suitecrm.com/suitecrm/forum/suite-themes/27994-unable-to-replace-logo-with-png-extension#92257

### DIFF
--- a/include/utils.php
+++ b/include/utils.php
@@ -5201,7 +5201,7 @@ function verify_image_file($path, $jpeg = false)
  */
 function verify_uploaded_image($path, $jpeg_only = false)
 {
-    $supportedExtensions = array('jpg' => 'image/jpeg', 'jpeg' => 'image/jpeg');
+    $supportedExtensions = array('jpg' => 'image/jpeg', 'jpeg' => 'image/jpeg', 'tmp' => 'tmp');
     if (!$jpeg_only) {
         $supportedExtensions['png'] = 'image/png';
     }


Mac-Rae:
Assessed :+1:

include/utils.php
```diff
@@ -5201,7 +5201,7 @@ function verify_image_file($path, $jpeg = false)
  */
 function verify_uploaded_image($path, $jpeg_only = false)
 {
-    $supportedExtensions = array('jpg' => 'image/jpeg', 'jpeg' => 'image/jpeg');
+    $supportedExtensions = array('jpg' => 'image/jpeg', 'jpeg' => 'image/jpeg', 'tmp' => 'tmp');
```
Dillon-Brown: Seems fine but I wonder if there are security implications we'd need to consider here if we start allowing ```tmp``` files to be uploaded.

pgorod: I've wondered about that but I get the feeling it's not going to be a problem. 
This extension is only `tmp` because we name the logo file like that (we could change _that_ of course).
About people uploading _other_ files with `tmp` extension, I don't think that makes much of a difference anyway. `tmp` files are _never_ executable, I believe.

But we really should work for a solution here, it's ridiculous how difficult it's been to upload a Logo into SuiteCRM, for so long. It's the first thing most people do after installing...

Dillon-Brown: Agreed, although looking at how we verify uploaded images I think we should probably improve it, doesn't seem like it actually validates much.... This should be fine for a quick fix to get logo uploads working though.

PR #8331, opened 2019-11-29
Fixed #8281 - Quotes in list value not saved correctly

### LINKED ISSUE #8281: Quotes in list value not saved correctly.
Just found a bug in list value, in the studio list builder for dropdown.

Story :  I have a list with all french department (http://www.cartesfrance.fr/carte-france-departement/carte-france-departements.html)

If I save for exemple this one :
CotesdArmor [22 - Côtes d'Armor]

It will save correctly the value " 22 - Côtes d'Armor " **on the first time only**.

If I re save a sec time the list it became : " 22 - Côtes d "

So the quotes make the value truncate and delete everything after the quote.

Last version of Chrome
SuiteCRM : 7.11.10

604media:
Fixed #8281 - Quotes in list value not saved correctly


mattlorimer:
This has been manually Merged, thanks for the fix

PR #8346, opened 2019-12-03
Update links

pgorod:
### The next SuiteCRM version should contain updated Forums links 

... since it should be coming out after the migration is done.

(I suggest merging this with a squash since we don't need all these commits, it's just that it was easier for me to edit the files directly form Github and it uses a separate commit every time we save a file.)


Mac-Rae:
Assessed :+1:

PR #8349, opened 2019-12-03
FIX ControllerFactory crashes when the custom controller class name is wrong

strangedev:
## Description
ControllerFactory crashes when the custom controller class name is incorrect. In other cases, we already default to the SugarController. This is now the default for all cases where a controller could not be loaded for any reason. Whenever a custom controller file is found, but the controller can't be instantiated, appropriate log messages are emitted.

## Motivation and Context
I'm currently developing custom modules and was irritated by the behavior, as well as the absence of log statements, since the logic to check these error cases is already in place.
Also, there was some inconsistency in the logic of the if/else branches, which I wanted to clean up.

## How To Test This
Start a new instance of SuiteCRM, set log level to debug, load a custom controller with incorrect class name and observe the logs as well as the application not crashing anymore when trying to access the controller.

## Types of changes

Not sure if this could be considered breaking, since the default behavior in the case a controller couldn't be loaded was changed.

### Final checklist

At least I think so, but please point out any mistakes and I will fix them.
Also, I'm unsure about commit 93470b7, there seems to be a problem with `.gitattributes` or the affected files we're reverted to a version with CRLF line breaks upstream and were automatically converted by my git client on clone. Whichever way, this probably needs some investigation. 

Thank you for your time and happy coding.


pgorod:
Are you working from Windows?

Maybe this will help with the CRLF issues

https://help.github.com/en/github/using-git/configuring-git-to-handle-line-endings

Normally this is fixed with just a `git config core.autocrlf ...` command, but have a look at that link if you need additional steps. And thanks for the contribution!

strangedev:
Hi @pgorod ,
nice seeing you here, your gists have helped my tremendously in my customization efforts!
I'm working on Linux, so my OS should comply with the settings in `.gitattributes` (which specifies `LF` for `.js` files).

pgorod:
@Dillon-Brown do you have any idea what could be the problem?

strangedev:
Turns out it was my shuffling around of remotes which confused git.
Closing this in favor of https://github.com/salesagility/SuiteCRM/pull/8367

PR #8389, opened 2019-12-17
Fix #8388 - Add userID as optional argument in SugarLogger log method

### LINKED ISSUE #8388: Add optional argument $userID to log method in SugarLogger
#### Suggestion
In order to extend `SugarLogger` class and override `log` method it would be useful to have the `$userID` value passed as an optional argument to this method.
 
#### Issue
`SugarLogger` `log` method uses UUID as fixed, for user identification (`$userID`).
 
#### Expected Behavior
Extensibility of SugarLogger `log` method so `$userID` can use other values aside from the UUID. 

#### Actual Behavior
In the need of using any other `$userID` value (for example `user_name`) when logging, the only way to accomplish that is to overwrite the `log` method completely.

#### Possible Fix
Add an optional argument $userID to `log` method, with default value `null` or empty string. When this value is not empty, use it. Otherwise, use UUID (current behaviour).

marin-h:
Add an optional argument $userID to log method, with default value null.
When this value is not empty, use it. Otherwise, use UUID (current behaviour).

## Description
Fix #8388 

## Motivation and Context
See #8388 

## How To Test This
1. Extend SugarLogger class.
2. Override log method.
3. Provide any string, for example the current user's `user_name` as the userID parameter in log method.
4. Ensure this class writes the log.
5. Check log file to see that the user `UUID` is now replaced by `user_name` in each log line. 
6. When using SugarLogger directly nothing should change.

It can be tested this using the following example method in MyLogger class (extend it from SugarLogger) in `custom/include/SugarLogger/MyLogger.php`:

```php
    public function log($level, $message, $userName = null)
    {
        $userName = !empty($GLOBALS['current_user']->user_name) ? $GLOBALS['current_user']->user_name : null;
        parent::log($level, $message, $userName);
    }
```
And don't forget to turn off SugarLogger with log mapping level or configure it to be able to recognize the logs written by each class (SugarLogger and MyLogger). 

## Types of changes

### Final checklist

### DIFF
--- a/include/SugarLogger/SugarLogger.php
+++ b/include/SugarLogger/SugarLogger.php
@@ -196,15 +196,18 @@ class SugarLogger implements LoggerTemplate
      */
     public function log(
         $level,
-        $message
-        ) {
+        $message,
+        $userID = null
+    ) {
         global $sugar_config;
 
         if (!$this->initialized) {
             return;
         }
         //lets get the current user id or default to -none- if it is not set yet
-        $userID = (!empty($GLOBALS['current_user']->id))?$GLOBALS['current_user']->id:'-none-';
+        if (empty($userID)) {
+            $userID = (!empty($GLOBALS['current_user']->id))?$GLOBALS['current_user']->id:'-none-';
+        }
 
         //if we haven't opened a file pointer yet let's do that
         if (! $this->fp) {


Dillon-Brown (approved):

Dillon-Brown:
:+1: Nice work!

We should probably take a proper look at the ```SugarLogger``` stuff at some point, there's a lot that can improved (I believe we actually have another logger ```SuiteLogger``` that seems be have been added and then never used....).

marin-h:
@Dillon-Brown yes, I couldn't find any docs mentioning this `SuiteLogger`. Actually I found It by chance because I used that same name for the class I created to extend SugarLogger :nerd_face:   
If there is an intention to move forward and replace SugarLogger with SuiteLogger, it would be nice to have that information in https://docs.suitecrm.com/developer/logging, what do you think?

Dillon-Brown:
Yeah ideally I would want to deprecate one of them, replace all the current logger calls and then update the developer docs.

I'm curious as to why these ```Utility``` classes were added and then never used though, perhaps the implementation isn't up to standards and we'd need to replace them with something better.

marin-h:
> Yeah ideally I would want to deprecate one of them, replace all the current logger calls and then update the developer docs.
> 
I agree, 

> I'm curious as to why these `Utility` classes were added and then never used though, perhaps the implementation isn't up to standards and we'd need to replace them with something better first.

Me too... when this gets cleared up I'll be glad to collaborate.

PR #8399, opened 2019-12-19
Fixed #8393 - Remove "Maximum upload size" from System Settings

### LINKED ISSUE #8393: Remove "Maximum upload size" from System Settings
## Issue 

In `Admin` / `System Settings` we have a setting for `Maximum upload size`. 

This setting apparently only affects the way SuiteCRM tests file sizes before attempting upload; but we don't change the PHP setting from this value, `upload_max_filesize`, because that's not possible (`ini_set` doesn't work for this, [by design](https://www.php.net/manual/en/ini.list.php), since PHP 5.3).

Then there are the settings in php.ini which are the ones that actually take effect and cause PHP to FATAL out when trying a larger upload.

So this setting is ineffective and misleading. For example: our default is `30000000`. If the user has a larger value effective in `php.ini`, something quite common nowadays, some upload code paths (but not _all_ uploads in SuiteCRM) will fail with an error message saying that the upload limit is the problem, but it won't be obvious that it's the limit in `Admin` / `System Settings`, people will just think it's the `php.ini` value. But changing _that_ won't solve the error message.

On the other hand, if people have a higher value in `php.ini` than in System Settings, our error messages won't kick in, and they will get an uglier FATAL instead.

We also have a bunch of places in the code where we mess with this value, give it 8MB, 60MB, not really changing anything except the error threshold.

## Proposed solution

If the option isn't changing anything for real, it shouldn't exist.

- remove field from `Admin` / `System Settings`
- set `$sugar_config['upload_maxsize']` to whatever `ini_get('upload_max_filesize')` returns
- remove any instances of changing `$sugar_config['upload_maxsize']` in our code

... unless I'm missing something. I did review the code for everything I could find related to these values, but it is possible something escaped me.

Dillon-Brown:
## Description

Removed the various places ```$sugar_config['upload_maxsize']``` is set in core and replaced it with util functions to properly calculate max file size from the ```php.ini``` settings. 

## Motivation and Context

Issue reference: #8393 

## How To Test This

Regression test file uploads with various ```upload_max_filesize``` values set in ```php.ini```.

## Types of changes

### Final checklist


samus-aran:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
+ Solved 4
- Added 2
           

Complexity decreasing per file
==============================
+ include/UploadFile.php  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/samus-aran/SuiteCRM/pullRequest?prid=4708702&bid=15728410)

PR #8414, opened 2019-12-31
Handle aliases more appropriately

Jason-Dang:
## Motivation and Context

Sub Panel aliases experience some weird behaviour due to bean and list view query caching.
The first value gets processed correctly, but is then cached in the BeanFactory which ultimately causes the alias to be cached by the list view query. Ultimately this leads to a check on the cache always skipping any processing for subsequent record values, resulting in the values being set straight from the DB etc. Additionally whenever the logic is fired it coverts to user date time format regardless of field type meaning aliases can only really be used on date time columns. 

### Changes 
- Bypass cache check so aliased values always set.
- Check dbType before casting values to datetime.

## Types of changes

### Final checklist

### DIFF
--- a/include/ListView/ListViewSubPanel.php
+++ b/include/ListView/ListViewSubPanel.php
@@ -417,9 +417,16 @@ if (!defined('sugarEntry') || !sugarEntry) {
                                 $alias_field_def['name'] = $list_field['alias'];
                                 // Add alias field def into bean to can render field in subpanel
                                 $aItem->field_defs[$list_field['alias']] = $alias_field_def;
-                                if (!isset($fields[strtoupper($list_field['alias'])]) || empty($fields[strtoupper($list_field['alias'])])) {
+
+                                $fields[strtoupper($list_field['alias'])] = !empty($aItem->{$list_field['alias']})
+                                    ? $aItem->{$list_field['alias']}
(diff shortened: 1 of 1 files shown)


include/ListView/ListViewSubPanel.php
```diff
@@ -417,9 +417,16 @@ public function process_dynamic_listview_rows($data, $parent_data, $smartyTempla
                                 $alias_field_def['name'] = $list_field['alias'];
                                 // Add alias field def into bean to can render field in subpanel
                                 $aItem->field_defs[$list_field['alias']] = $alias_field_def;
-                                if (!isset($fields[strtoupper($list_field['alias'])]) || empty($fields[strtoupper($list_field['alias'])])) {
+
+                                $fields[strtoupper($list_field['alias'])] = !empty($aItem->{$list_field['alias']})
+                                    ? $aItem->{$list_field['alias']}
+                                    : $aItem->$field_name;
+
+                                if (($alias_field_def['dbType'] ?? '') === 'datetime') {
```
Dillon-Brown: No Null coalescing operators until we're on PHP 7 :(

Jason-Dang: True dat, fixed.

PR #8434, opened 2020-01-09
Allow abort of mark_deleted

ildarius:
Allow aborting of the delete operation via abort_mark_deleted flag that can be set inside before_delete logic hook method. Quick explanation https://youtu.be/p9xeikuZs3E

By setting $bean->abort_mark_deleted = 1; inside the logic hook method we prevent the deletion and destruction of relationships performed by mark_delete method. 

Currently there's no other way to stop the delete procedure inside the before_delete hook tried the alternative of using the $bean->mark_undeleted($id), but unfortunately it does not restore previously created relationships and even if it did it could cause the firing of additional logics hooks which may result in unexpected behaviour.

## Types of changes

### Final checklist

### DIFF
--- a/data/SugarBean.php
+++ b/data/SugarBean.php
@@ -5216,6 +5216,13 @@ class SugarBean
             // call the custom business logic
             $custom_logic_arguments['id'] = $id;
             $this->call_custom_logic("before_delete", $custom_logic_arguments);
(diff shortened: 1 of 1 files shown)


ildarius:
Wrong branch

PR #8539, opened 2020-02-12
Fix #7651: [WARN] CSS File Dawn/yui.css not found

### LINKED ISSUE #7651: [WARN] CSS File Dawn/yui.css not found
In current version (7.11.6), we are getting the following errors in out suitecrm.log:
[WARN] CSS File Dawn/yui.css not found

This appears to be related to the following line of code:
https://github.com/salesagility/SuiteCRM/blob/ef79908dad38b294d2f62924f91aeaa805d8f7a3/include/SugarTheme/SugarTheme.php#L626

There is, as the error log suggests, no yui.css file in that location.

YUI is, however, getting loaded independently, so it appears that this reference is redundant?

Looking at git blame, this line was re-injected in order to provide backwards compatibility for SuiteR/Suite7 themes, however as these are now removed, it appears that this attempt to include yui again (despite the file not existing) could be removed.

pgorod:
According to Dillon's comment here https://github.com/salesagility/SuiteCRM/issues/7651#issuecomment-534480674

... this can be removed (again) to prevent the log message, now that SuiteR and Suite7 are no longer part of SuiteCRM.

### DIFF
--- a/include/SugarTheme/SugarTheme.php
+++ b/include/SugarTheme/SugarTheme.php
@@ -623,7 +623,6 @@ class SugarTheme
         $html = '
             <!-- qtip & suggestion box -->
             <link rel="stylesheet" type="text/css" href="include/javascript/qtip/jquery.qtip.min.css" />';
-        $html .= '<link rel="stylesheet" type="text/css" href="'.$this->getCSSURL('yui.css').'" />';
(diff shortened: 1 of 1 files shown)


pjdm:
This previously fixed issue has re-appeared in V8.  This is from v8.4.0-beta in the suitecrm.log.
`Mon Aug  7 16:41:59 2023 [99483][1][WARN] CSS File Dawn/yui.css not found`

PR #8548, opened 2020-02-14
7.11.12 Release

mattlorimer:
7.11.12 Release

## Types of changes

### Final checklist


modules/Currencies/Currency.php
```diff
@@ -498,8 +498,6 @@ function format_money($amount, $for_display = true)
  */
 function get_number_seperators($reset_sep = false)
 {
-    LoggerManager::getLogger()->deprecated('get_number_seperators will be removed in a future release, please
-    update your code to use get_number_separators');
```
Dillon-Brown: Agreed that this would only really be adding more log spam since it's unlikely we'll actually remove it. Although I think it might be worth putting some kind of notice after the PHPDoc ```@deprecated``` tag saying to keep it for BC (I could see someone easily removing it later on without realising).

Issue #743, opened 2015-12-02
Accept status is not showing on the meetings subpanels. Version 7.4.3


recci:
Meetings subpanels for Contacts, Users, Leads does not show the accept status. Status is updated in the database just does not display.  I have first noticed this in 7.4.3 but most likely affects earlier version also.

ghermans:
+1

recci:
Anybody have any idea when this last worked?

recci:
Was working in 7.1.2 but not 7.2.2

PR #8551, opened 2020-02-15
fixes #1241 - Reverts commit b3246d3 / adds condition for lookup rela…

### LINKED ISSUE #1241: Regression: new SugarBean.php in 7.5.2 breaks relationship fields
I had working customization using a new custom field in a contact-to-contact relationship. The field is a multienum. 

Notice that the field is in the relationship, not in any of the contacts. It was done following these instructions: http://sugarmods.co.uk/how-to-add-custom-fields-to-a-relationship-table-and-display-them-in-a-subpanel-sugarcrm-suitecrm

The field was showing correctly in a subpanel. After the upgrade from 7.5.1 to 7.5.2 it stops showing. No error, just a blank space instead of showing the value. 

I think none of the information from the relationship is being loaded into the bean, but I am not sure, it might be just this field.

After 5 hours debugging, and checking each file, I found that if I copy back data/SugarBean.php from 7.5.1 the issue is fixed...

JanSiero:
This commit follows up on PR #61. It concerns relationship fields. These fields require fields of type "relationship_info" to obtain information on how to query them. 

Due to a bug in the display of the accept status in subpanels (bug #743) bugfix b3246d3 was committed, which effectively defeats its purpose.

This pull requests reverts this commit and adds an additional check to remediate #743

## Description

Relationship fields, such as the "role" field between Opportunities and Contacts require a relationship_info field that holds the information to which link a relationship field belongs. This field is therefore defined in the vardefs as well.

To be able to query this field, normally the relationship_info field is added as a "query_only" field so that the relationship field can effectively be queried.

## Motivation and Context

As the vardefs well-define how each field should be queried, with PR #61 the ability was added to the SugarBean class to query relationship fields, without the need to supply their belonging relationship_info field. Instead, when querying a relationship field, the SugarBean class looks up the required relationship info field by itself.

This would not only remove the requirement to add the relationship_info fields as query_only fields for subpanels, but would also enable to use studio to add existing relationship fields in studio to the Listview and Filters without the need to add any relationship_info field.

This change however broke the display of the accept status (issue #743) of meetings, calls and events in subpanels. It is caused by the fact that accept status have multiple relationship_info fields, each for Meetings, Calls and Events, and the lookup did not discern these.

With commit b3246d3 this was fixed, however this broke the ability of SugarBean to query relationship_info fields by itself (issue #1241), effectively undoing PR #61.

This PR reverts b3246d3 and adds a check to make sure only the relationship_info field matching the parent module is used. Effectively it replaces the condition

`(!isset($value['link_type']) || $value['link_type'] != 'relationship_info')`

with

`$this->load_relationship($value['link']) && $this->{$value['link']}->getRelatedModuleName() == $parentModule`

It thereby caters all scenarios:
 - relationship fields having only one relationship_info field
 - relationship fields having multiple relationship_info fields
 - with and without its relationship_info supplied as "query_only" field

## How To Test This
See #1241

## Types of changes

### Final checklist

### DIFF
--- a/data/SugarBean.php
+++ b/data/SugarBean.php
@@ -3629,22 +3629,24 @@ class SugarBean
         //walk through the fields and for every relationship field add their relationship_info field
         //relationshipfield-aliases are resolved in SugarBean::create_new_list_query
         // through their relationship_info field
-        $addrelate = array();
-        foreach ($fields as $field => $value) {
-            if (isset($this->field_defs[$field]) && isset($this->field_defs[$field]['source']) &&
-                $this->field_defs[$field]['source'] == 'non-db'
-            ) {
-                $addrelatefield = $this->get_relationship_field($field);
-                if ($addrelatefield) {
-                    $addrelate[$addrelatefield] = true;
+        if (isset($parentbean)) {
+            $addrelate = array();
+            foreach ($fields as $field => $value) {
+                if (isset($this->field_defs[$field]) && isset($this->field_defs[$field]['source']) &&
+                    $this->field_defs[$field]['source'] == 'non-db'
+                ) {
+                    $addrelatefield = $this->get_relationship_field($field, $parentbean->module_name);
+                    if ($addrelatefield) {
+                        $addrelate[$addrelatefield] = true;
+                    }
+                }
+                if (!empty($this->field_defs[$field]['id_name'])) {
+                    $addrelate[$this->field_defs[$field]['id_name']] = true;
                 }
             }
-            if (!empty($this->field_defs[$field]['id_name'])) {
-                $addrelate[$this->field_defs[$field]['id_name']] = true;
-            }
-        }
 
-        $fields = array_merge($addrelate, $fields);
+            $fields = array_merge($addrelate, $fields);
+        }
 
         foreach ($fields as $field => $value) {
             //alias is used to alias field names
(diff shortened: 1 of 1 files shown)


data/SugarBean.php
```diff
@@ -3629,22 +3629,24 @@ public function create_new_list_query(
...
                 }
             }
-            if (!empty($this->field_defs[$field]['id_name'])) {
-                $addrelate[$this->field_defs[$field]['id_name']] = true;
-            }
-        }
 
-        $fields = array_merge($addrelate, $fields);
+            $fields = array_merge($addrelate, $fields);
+        }
```
JanSiero: Check if $parentbean is set, which is a requirement to lookup relationship_info fields, and pass $parentbean->module_name on to the method get_relationship_field.

JanSiero:
Withdrawing this pull request because it doesn't resolve lookups for relationship fields other than in Subpanels. 

Due to the ambiguity of which relationship info field should be queried, this should not be resolved by SugarBean::create_new_list_query(), but should be defined beforehand by supplying the required field, as described in this comment: https://github.com/salesagility/SuiteCRM/issues/1241#issuecomment-210798022

I suggest to fix this by having Studio adding the required relationship_info field when a relationship field is inserted into a listview / subpanel, for which I will create a PR at a later moment.

PR #8559, opened 2020-02-24
Fix user action menu support link

Dillon-Brown:
## Description

Fix for global support forum link.

## How To Test This

1. Check that the link still takes you to community forums.

## Types of changes

### Final checklist

### DIFF
--- a/include/MVC/View/SugarView.php
+++ b/include/MVC/View/SugarView.php
@@ -504,7 +504,7 @@ class SugarView
                 $ss->assign("LOGOUT_LABEL", key($value['linkinfo']));//key value for first element.
                 continue;
             }
-
+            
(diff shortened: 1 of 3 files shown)


Mac-Rae:
Assessed :+1: 

I will note that the link redirects on the same page where previously I believe it opens a new tab

pgorod:
A new tab would definitely be better for this, imho

Dillon-Brown:
> A new tab would definitely be better for this, imho

Agreed, the problem here is that these are loaded in dynamically but I'll see if I can add in a blank ref specifically for this link so that it opens in a new tab.

I'm generally opposed to opening another website in a new tab for security reasons but in this case, since it will only ever contain a link to our forums I guess it should be fine.

Mac-Rae:
Assessed :+1:

PR #8560, opened 2020-02-24
redefine static function calls

tsmgeek:
Redefine static calls to methods within same class with `self`.
Define functions that are called statically with `static` keyword to prevent E_DEPRECATED warnings.
https://www.php.net/manual/en/language.oop5.static.php

Within `parse_template` there is a check if method is within instantiated class or statically called. This does not seem to be needed as all child method calls do not do any calls to $this.

### DIFF
--- a/modules/EmailTemplates/EmailTemplate.php
+++ b/modules/EmailTemplates/EmailTemplate.php
@@ -501,7 +501,7 @@ class EmailTemplate extends SugarBean
      * @param string $value
      * @return string
      */
-    public function _convertToType($type, $value)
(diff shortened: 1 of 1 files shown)


Mac-Rae:
Hi @tsmgeek,

Please rebase this pull request onto hotfix-7.10.x as pull requests directly to master are no longer accepted.
For more information please take a look at our [contribution guidelines](https://docs.suitecrm.com/community/contributing-code/).

If you would like assistance please ensure you select "Allow edits from maintainers" checkbox and leave us a message @tagging me! 

Thanks :+1:

PR #8586, opened 2020-03-03
Use smarty foreach key instead of item key

tsmgeek:
Use foreach key instead of item key else its indexed from 1 instead of 0.
https://www.smarty.net/docsv2/en/language.function.foreach.tpl

As parameters are indexed from 0 in AOR_Reports but in dashlet it starts at 1 so params do not work.

This is a bug in 7.8.x but code has not changed since so expect its the same.

### DIFF
--- a/modules/AOR_Reports/Dashlets/AORReportsDashlet/dashletConfigure.tpl
+++ b/modules/AOR_Reports/Dashlets/AORReportsDashlet/dashletConfigure.tpl
@@ -150,7 +150,7 @@
                     </script>
                 </td>
             </tr>
-            {foreach from=$parameters item=condition}
+            {foreach from=$parameters item=condition key=key}
             <tr>
                 <td scope='row'>
                     {$MOD.LBL_PARAMETERS}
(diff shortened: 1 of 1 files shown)


Dillon-Brown (approved):

jack7anderson7:
Hi tsmgeek,

When trying to test the issue behind your fix, I'm afraid I am unable to replicate the issue.

I started off by making a report with the following parameters:
Screenshot from 2021-07-30 14-23-27

After this I then added a Report Dashlet to the Homepage and the report loaded as expected:
Screenshot from 2021-07-30 14-24-13

I then, changed the parameters within the dashlet as shown:
Screenshot from 2021-07-30 14-24-32

And after saving the changes seem to have taken effect as expected:
Screenshot from 2021-07-30 14-24-43

I was wondering if you could shed some light on anything I may be doing wrong when trying to identify the issue to help with the assessing of this PR.

Thanks, Jack

tsmgeek:
Move your parameters around then check the dashlet.

jack7anderson7:
Hi tsmgeek,

Could you please provide us with a step-by-step as to how you have tested this as we are unable to replicate. Following your response, we have removed the parameters, changed the order to them and have still not had any luck in replicating an issue. If you could provide more information that would be helpful.

Thanks, Jack

PR #8587, opened 2020-03-03
Add cases to email object_arr

tsmgeek:
Cases parent_type to be pulled for Email Template


tsmgeek:
Actually thinking about it why do we even check against of allowed types, why not just try to load the bean if parent/id are supplied and its in list of beans?

tsmgeek:
Can we merge this in, so simple but would help make things easier in our pipeline.

Dillon-Brown (approved):

Dillon-Brown (approved):
Removed the merge commit, aside from that LGTM.

jack7anderson7 (approved):

samus-aran (approved):

PR #8615, opened 2020-03-14
Display checkboxes for boolean fields in PDF templates

QuickCRM:
## Description

Replace true/false hardcoded strings by checkbox images.

true/false are hardcoded.
1- Create a checkbox field in Quotes or Invoices (for example)
2- Add that checkbox field to a PDF template
3- Use any SuiteCRM translation
4- The field will be displayed as true or false whatever the language

## Motivation and Context

- fix an issue with hardcoded strings
- more generic than just fixing this language issue (sometimes you would want to display True/False, sometimes you would want to display Yes/No)

## How To Test This

1- Create a checkbox field in Quotes or Invoices (for example)
2- Add that checkbox field to a PDF template
3- The field is displayed as checked or unchecked images

## Types of changes

### Final checklist


Mac-Rae:
Assessed :+1:

pgorod:
Wouldn't it be better to use Font Awesome icons instead of adding two new files?

samus-aran:
> Wouldn't it be better to use Font Awesome icons instead of adding two new files?

Seems alot of effort for two icons.
https://stackoverflow.com/questions/30453168/fontawesome-icons-wont-show-on-generated-pdf-using-mpdf

Requires a number of steps to include fontawesome and embed it into mPDF. It would be nice yes, but I think in this case it would be overkill. Perhaps for a larger enhancement but this PR I feel resolves a bug. :)

QuickCRM:
I agree. And using image files allow easy customizations as we used get_custom_file_if_exists if users do not want an orange checked icon.

--
Benoit Luquet
NS-Team
Tel. +33 (0)5 61 73 64 14 - +33 (0)6 64 50 59 08

jack7anderson7:
Hi @QuickCRM, 

Could you please change your indentation from tabs to spaces following PSR-2/PSR-12: https://www.php-fig.org/psr/psr-12/#23-lines

Thanks

Issue #6553, opened 2018-11-30
Active Directory Groups LDAP


AOSGit:
#### Issue
Code used to define User membership to a group incorrectly uses bind_dn and doesn't permit users to be in a sub-OU of the bind_dn, yet authentication works find.

#### Expected Behavior
Group search should use the Users DN (minus the initial container) to use in its group comparison rather than bind_dn

#### Actual Behavior
Group searching uses Base_DN as defined in the tool

#### Possible Fix
https://suitecrm.com/suitecrm/forum/suitecrm-7-0-discussion/21150-active-directory-group-ldap

#### Steps to Reproduce

1.  Create User in OU=Level2, OU=Level1,DC=Joe,DC=Bloggs
2. Set Base_DN to use OU=Level1,DC=Joe,DC=Bloggs
3. Test that authentication works (it does)
4. Add user to a group, say, CN=SugarCRM,OU=Groups,DC=Joe,DC=Bloggs
5. Activate the Group functionality and test.
6) Adding debug log will show you that the code checks the group CD=SugarCRM using CN=Level1,DC=joe,DC=Bloggs which is based on the Base_DN, not the actual user.

The same is true for the User information pull.

#### Context
Correct Authorization was not correct and the layout of our AD could not be changed to reflect SuiteCRM.  
For Business, medium.  Low for us as I found a workaround (see forum link above).  Probably not the best code, but it works.

#### Your Environment

Version 7.10.10
Sugar Version 6.5.25 (Build 344)

* Internet Explorer/Firefox 64-bit (latest)
* MySQL, PHP 7.0
* CentOS 7

Issue #6299, opened 2018-09-11
Unable to provide client certificate for ldap authentication


mphilipps:
hi
#### Issue
Since the entirety of SuiteCRM's documentation on LDAP Authentication consists of:

> If LDAP Authentication is enabled, none of the SuiteCRM Password Management settings will apply. Passwords will be managed by LDAP settings.

I have to rely on this. How do I provide the necessary certificates to SuiteCRM for the LDAP authentication?
While there are omnious fields like "Encryption Key" with no explanation on how that key is used, it certainly lacking in the expected config options.

#### Expected Behavior
I would expect a way to specify the client certificate, the key to use with it and CA to validate against.

#### Actual Behavior
There is no documentation on how to configure the LDAP connect authentication.

#### Possible Fix
Instead of implementing it's own LDAP authentication SuiteCRM could use the external authentication provided by the webserver to login users. Apache is able to performer a proper LDAP authentication and SuiteCRM could just auto login users based on $_SERVER['REMOTE_USER']. 

#### Context
Our LDAP requires that TLS is done with a client certificate. This is used to prevent unauthorized users from connecting and directly binds authorized ones.

#### Your Environment
* SuiteCRM Version used: 7.8.20
* Environment name and version (e.g. MySQL, PHP 7): MySQL, PHP 7
* Operating System and version (e.g Ubuntu 16.04): Debian Stretch
* LDAP Server: OpenLDAP 2.4.44 with TLSVerifyClient demand

RossRogers:
As a fellow user, I ran across this ticket as I was trying to solve my problem.  What ended up working for me was a 16 character string with hexadecimaal letters as the "Password Encryption Key"  on both the Suite CRM server and the outlook plugin like 'deadbeefdeadbeef' -- same key on server and outlook plugin .  I don't know if you can use more characters or less, but I don't really care as my backend is on https.  Furthermore, I believe md5 is being used to "secure" the password, so you really need to have your backend on SSL anyways since md5 collisions are relatively easy to find.

Anyways, that wasn't the end of it for me, since we also seemed to be hitting a bug or maybe a difference between PHP 5.6 ( which we're on) and php 7:
https://store.suitecrm.com/support/SuiteCRM-official-outlook-plugin/1802

mphilipps:
[...]

Issue #6070, opened 2018-06-22
Feature: Update User infos from LDAP


gunnicom:
I would like to update the infos for users if they change in the LDAP directory (like phone number, email address).
I can do this either by cron job or on each login. Which one qould be preferred?
Is there interest in such a feature? If yes i would start to implement this and do a pull request.

pgorod:
> I can do this either by cron job or on each login. Which one would be preferred?

I would say Scheduled job is better, because it gives admins a straight-forward way to schedule it with the desired frequency, or even to disable it, easily. It also gives the feature visibility. 

But I don't know about the usefulness of the feature, I don't use LDAP, I leave that to others, to comment here if they are interested in it.

Thanks for offering to do this.

Issue #5107, opened 2018-02-06
More than a one ldap server


topesantos:
#### Issue
I have more than a ldap server and I don't know if all are uptime

#### Expected Behavior
Allow SuiteCRM to handle more than one ldap server

#### Possible Fix
               // this servers are in $GLOBALS['ldap_config']->settings['ldap_hostname'];
               // and can pass to an array
		$adServers=array("server1","server2","server3"); 

		$i=0;
		do {
			$ldap = ldap_connect($adServers[$i]);
			$i++;
		}while(!$ldap);

pgorod:
I agree it's useful, LDAP shouldn't be "single point of failure" that leaves the entire company unable to log in to other systems.

PR #8642, opened 2020-04-04
Fix deprecated ldap_connect calls, enable ldap(s):// style LDAP URIs

RustyDust:
## Description

Problems:
- The call syntax for `ldap_connect` has changed from `ldap_connect($host, $port)` to `ldap_connect($ldap_uri)` between PHP5 and PHP7
- Although the backwards compatible format is still supported it has the problem of not supporting secure connections easily, forcing users to use cleartext connections to their LDAP servers
- The old syntax doesn't support ldap(s):// style URIs
- Unfortunately the new syntax doesn't support the simple `$host:$port` syntax but **requires** the protocol identifier to be present.

Solution:
- Test whether '://' is present in the `$host` variable. If it is, use the new call syntax by concatenating the `$host` and `$port` variables into one string and use the new call syntax
- If there's no protocol identifier given in the `$host` var keep using the old call syntax, thus stay backwards compatible to existing and already running setups

## Motivation and Context
This fix allows users to configure LDAP URIs the way they should be configured, using the proper protocol identifier for cleartext and secure connections to their LDAP servers.

## How To Test This
This is a drop-in replacement and should work in existing setups without any further changes.

## Types of changes

### Final checklist


pgorod:
@RustyDust do you think this PR might be related to any of the [currently open LDAP issues](https://github.com/salesagility/SuiteCRM/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+ldap+)?

RustyDust:
@pgorod Short answer: No, except for #5107. 

Very long answer for #5107:
The new call syntax for `ldap_connect` would allow for what is requested in #5107. However, this poses another problem: The input fields in the admin interface only offer one `port` entry. So there's multiple pitfalls here:

1. User enters "ldap1.example.com ldap2.example.com" in the `host` field. This would result in the old call syntax getting used with the specified port for all servers. This would fail because the old call style doesn't support multiple hosts.

2. User enters "ldap://ldap1.example.com ldap://ldap2.example.com" for the hosts to be used. Now What should he put in the `port` field? As long as all the LDAP servers are listening on the same port that's easy because he'd just enter the port they all are using. The new call syntax would be used and the correct port added to the last host only right now. This would be easy to fix so that the port gets added to all hosts. But what if all hosts use different ports?

3) User enters "ldap://ldap1.example.com:7389 ldap://ldap2.example.com:389". This will fail because as it currently is the code would append either the port given by the user or ":389" because of the setting of the `DEFAULT_PORT` constant in `LDAPAuthenticateUser.php` => result would be an invalid LDAP URI.

What I certainly could do is expand the tests on the `hosts` variable. In that case I'd suggest to handle it like this: If the "ldap(s)" protocol identifier is present, ignore the `$port` variable and just use the `$host` variable. If the protocol identifier is **not** present and multiple hosts are given, emit a log message and proceed with the first host. This is potentially dangerous since it gives the user the false sense of fail-over security when in reality there isn't

In the long run, I think the better solution would be to get rid of the `$port` entry and just use the `$host`variable. Of course this poses the issue of handling existing setups but this should be easy because as far as I can tell right now there should only be plaintext connections via LDAP for the current code simply won't work with secure setups. So the migration path would be to read the config, prepend the existing `$host` variable with `ldap://` and either add the port given by the user or the default of `:389`.

As for the other LDAP related issues:
#3911 looks solved to me but I didn't test it - not related to or affected by my patch
#6070 is unrelated
I think #6299 is based on a misunderstanding and could be solved by adding the certificate of the LDAP server (it's CA to be precise) to the trusted certificates of the client machine.
#6553 isn't related but the request itself is (somewhat) valid although I think it isn't as easy as outlined there.

chris001:
@RustyDust  
Why close this?  Seems to be well along the way to solving the issue.  Only need to complete the two caveats by verifying that the `hostnames` resolve to `IP` addresses and the port numbers are valid.

RustyDust:
I wasn't aware of the two caveats you're talking about, can you give a bit more detail about what you mean exactly? As far as I can tell the original code didn't do those checks either and I'm not sure whether they're actually necessary.

As for closing the PR: I was scratching my own itch, solved the issue I was having and also solved an issue I wasn't having with the 2nd commit. Then the PR sat idle for over four months with no feedback and "1 more approval needed for merge". Also, the company I work for has moved on to a different solution in the meantime. Since I don't need the patch anymore and since there didn't seem to be any interest in merging it, why keep it?

chris001:
> I wasn't aware of the two caveats you're talking about, can you give a bit more detail about what you mean exactly? As far as I can tell the original code didn't do those checks either and I'm not sure whether they're actually necessary.

The Caveats are from this pull request, it's [the message on one the commits you wanted to merge](https://github.com/salesagility/SuiteCRM/pull/8642/commits/9a6a668fa7d6ef1a1a006f6a98e1522d591cb428):
```
 Add support for #5107 (multiple LDAP servers)

Changes to the original patch:
- If ldap(s):// is present in host var ignore `$port` variable and use
  new call syntax with the host(s) given
- If no protocol identifier is given still use the old syntax and
  stay compatible to existing setups
- Modify the help strings displayed in Admin / Configuration to
  make it clearer that the protocol identifier is now mandatory
- Add ":port" example to the help text
- Remove the port field from Admin page and enlarge the host field
  so that multiple LDAP servers can be added/reviewed more easily

Caveats:
- If the first host definition is valid but the following servers
  aren't specified in a valid format the connect will fail. But
  instead of pointing to the real source of the problem the error
  message will be something along the lines of "You need to enter
  valid user credentials".
- Likewise, if all servers are specified in a valid format but
  only the first is really reachable no error message will be
  shown and logins will work as long as the first server specified
  is up and running.
```
 
> As for closing the PR: I was scratching my own itch, solved the issue I was having and also solved an issue I wasn't having with the 2nd commit. Then the PR sat idle for over four months with no feedback and "1 more approval needed for merge". Also, the company I work for has moved on to a different solution in the meantime. Since I don't need the patch anymore and since there didn't seem to be any interest in merging it, why keep it?

I think they're distracted fixing other areas of the CRM, but this PR is still valid because it fixes bugs and updates the CRM to be compatible with today's much more commonly used secure `LDAPS` directory services, ie Active Directory, etc.  So it should probably be re-opened so that they'll  see it, test it, approve it, and merge it in for the next release.

RustyDust:
I see. See, in the meantime I even forgot about the caveats I wrote myself ;)
Well, for the first one, there's no easy way to check/fix this. The "check" in this scenario would actually be the test of the connection for I don't see any way to test right after input. That's why I documented the behavior in this case. The same is true for the second one. But if you have a good suggestion for an easy solution, I'm all ears.

And since there's at least some interest, I'll also reopen the PR.

chris001:
> Well, for the first one, there's no easy way to check/fix this. The "check" in this scenario would actually be the test of the connection for I don't see any way to test right after input. That's why I documented the behavior in this case. The same is true for the second one. But if you have a good suggestion for an easy solution, I'm all ears.

First, you split the string of possibly many URLs into individual string URLs, then for each URL string, you call :
```
$result = filter_var($testURL, FILTER_VALIDATE_URL));
if (!$result) {
   throw new Exception("error : bad format LDAP URL $testURL");
}
```

https://www.php.net/manual/en/filter.filters.validate.php
```
<?php
// Validate a domain name
var_dump(filter_var('mandrill._domainkey.mailchimp.com', FILTER_VALIDATE_DOMAIN));
// string(33) "mandrill._domainkey.mailchimp.com"

// Validate an hostname (here, the underscore is invalid)
var_dump(filter_var('mandrill._domainkey.mailchimp.com', FILTER_VALIDATE_DOMAIN, FILTER_FLAG_HOSTNAME));
// bool(false)

// Label ends with a hyphen
var_dump(filter_var('http://a-.bc.com', FILTER_VALIDATE_URL));
// bool(false)

// Label is more than 63 octets
var_dump(filter_var('http://toolongtoolongtoolongtoolongtoolongtoolongtoolongtoolongtoolongtoolong.com', FILTER_VALIDATE_URL));
// bool(false)
?>
```

RustyDust:
@chris001 
Correct me if I'm wrong but that sort of check would only test whether the inputs are (somewhat) syntactically correct. While I agree this is better than not checking anything, it doesn't solve the main problem: to my knowledge SuiteCRM doesn't provide any easily usable means to either inform the user of faulty inputs or to prevent saving of the settings if a faulty input is detected. 

The only way I can see is to extend the `addcheck` JavaScript function on the input page. But even that won't help much since the only checks that can be made there are again syntax checks. Also, the tests required for checking a string containing multiple entries aren't yet included in `jssource/src_files/include/javascript/sugar_3.js` which is what gets called by the onsubmit `check_form` call. So one would have to write those functions too.

So no matter how extensive your testing is, there's no easy way of giving feedback or detailed error messages to the user let alone do the real testing whether syntactically correct settings actually work. The only way to do that would be to implement a "Test my settings" button which with all the work attached to that endeavor would blow the patch way out of proportion.

chris001:
You're right, there's two levels of checking / verification:
1. Are the string of multiple LDAP URLs containing valid URLs, in other words, are they legit, no syntax errors, not too long, no invalid characters, in the right format: scheme colon slash slash hostname colon port ?
2. is the service, at each of these URLs, a valid LDAP service?  Are these services up?  Answering connections?  And, optionally, providing data back when queried?

The code I provided above is for level 1 validation.  

Level 2 validation would be nice, but it should be caught at runtime, when the CRM app tries to connect to the directory and fails, it must throw a detailed exception message to the user/admin, saying "Hey user, Hey admin, I had a serious problem connecting to the following LDAP directory URL!"

RustyDust:
I have no idea what is stuck here. I signed the CLA and it shows as signed when I click the "Details" link.

mattlorimer:
It only showing this as we had to transfer over to a new CLA system, In essence you can sign the agreement again on the new CLA or as you ave already sign the agreement previously we will be retro fixing this eventually and it will be transferred  and/or prior to review/merging this CR we will manually update it if required

RustyDust:
When I try to sign the CLA again it shows as already signed which is why I was wondering what's stuck there. That's the CL version from 2020-08-27, if that helps.

mattlorimer:
in that case its likely that there is an email address mismatch from the CLA to your github account, or the email address associated with the commits, I couldn't say for sure what matters the most at the moment.

Dillon-Brown (approved):

PR #8644, opened 2020-04-06
Fix #8643 - Reports: Display or filter custom fields in related modules

### LINKED ISSUE #8643: Reports do not work with custom fields in Users module
Display or filter a custom field in Users generates a wrong query

```
LEFT JOIN ``leads:assigned_user_link`_cstm` `leads:assigned_user_link_cstm` ON `leads:assigned_user_link`.id = `leads:assigned_user_link_cstm`.id_c 
```

Generated:
```
LEFT JOIN ``leads:assigned_user_link`_cstm` 
```
Should be:
```
LEFT JOIN `leads:assigned_user_link_cstm` 
```

#### Issue
Reports using custom fields in Users do not work.

#### Actual Behavior

An empty page is displayed

#### Steps to Reproduce

1. Create a custom field in Users
2. Create a report for Leads
3. Select this custom Field in Leads / Assigned User (display or condition)
4. Save this report
5. Check suitecrm log

#### Your Environment

* SuiteCRM Version used: SuiteCRM 7.10, SuiteCRM 7.11

QuickCRM:
Fixed Join query to allow display of custom fields in related module

## Description

#8643 

## Motivation and Context

Custom fields in related modules generate a wrong query and the report is blank.
## How To Test This

1. Create a report on Opportunities
2. Add Account / Geocode status or any custom field of Accounts
3. The report is displayed

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOR_Reports/AOR_Report.php
+++ b/modules/AOR_Reports/AOR_Report.php
@@ -1414,7 +1414,8 @@ class AOR_Report extends Basic
         if (!isset($query['join'][$alias])) {
             switch ($type) {
                 case 'custom':
-                    $customTable = $module->get_custom_table_name();
+                    $tmpModule = BeanFactory::getBean($module->module_dir);
+                    $customTable = $tmpModule->get_custom_table_name();
                     $query['join'][$alias] =
                         'LEFT JOIN ' .
                         $db->quoteIdentifier($customTable) .' '. $db->quoteIdentifier($alias) .


modules/AOR_Reports/AOR_Report.php
```diff
@@ -1414,7 +1414,8 @@ public function build_report_query_join(
         if (!isset($query['join'][$alias])) {
             switch ($type) {
                 case 'custom':
-                    $customTable = $module->get_custom_table_name();
+                    $tmpModule = BeanFactory::getBean($module->module_dir);
```
mattlorimer, line 1417: this make no difference, it's calling getBean in a bean that has already been passed in, so won't change the result from the get_custom_table_name and from my tests doesn't fix the issue

QuickCRM: @mattlorimer 
Sorry but it makes change.
In that case, $module->table_name has already been changed when the code reach that step.
add a log->fatal with $module->get_custom_table_name() and $tmpModule->get_custom_table_name() and you will see it.

QuickCRM:
Sorry Matt,

I don’t know what I did wrong in Github but your comment and my answer have disappeared.

This make a lot of change:
$module->table_name has already been changed to an alias when the code reaches that step.
So we need to get the real table_name and that’s what our change does.

We have tested that and already using that (of course not upgrade-safe).

Benoit

--
Benoit Luquet
NS-Team
Tel. +33 (0)5 61 73 64 14 - +33 (0)6 64 50 59 08

QuickCRM:
Again, it does not make a difference if the custom field is for the report module.
But it makes a difference if the custom field is on a related record (like custom field of Account for an Opportunity report).

--
Benoit Luquet
NS-Team
Tel. +33 (0)5 61 73 64 14 - +33 (0)6 64 50 59 08

modules/AOR_Reports/AOR_Report.php
mattlorimer: it may work in some scenarios, but it will create an intermittent issue where sometimes it will work and sometime it won't, as you have identified this requires a fix further up the chain.

QuickCRM:
Our customer have made a lot of reports using that change and we didn’t find any error so far.
Any idea of the cases where it wouldn’t work?

But anyway, the current code generates an SQL error, so even if not perfect, it would be better than the current situation that does not generate anything.

--
Benoit Luquet
NS-Team
Tel. +33 (0)5 61 73 64 14 - +33 (0)6 64 50 59 08

mattlorimer:
In case where the bean factory has already cached the bean, which I can see a few simple reports it has done, so the fix fails is those cases, but I can this issue is mainly to do with build_report_access_query, I will post an alternative fix and close this fix as I don't think it this correct approach.

mattlorimer:
closed in favour of https://github.com/salesagility/SuiteCRM/pull/8652 but thanks for you contribution

PR #7612, opened 2019-07-22
Fixed #7608 - A non-numeric value encountered at ListViewSubPanel.php

Dillon-Brown:
## Description

Fixes PHP notice ```A non-numeric value encountered at ListViewSubPanel.php, line 323```.

## Motivation and Context

Issue reference: #7608 

## How To Test This

1. Reports -> Schedule Report subpanel.
2. Check for PHP errors.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [X] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change to the documentation.
- [X] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

Issue #7608, opened 2019-07-19
A non-numeric value encountered at ListViewSubPanel.php, line 323


cherrador:
This has been already addressed on closed issue #3576 but it still causing problems on some situations.

#### Issue
While running AOR_Reports and trying to build the Scheduled Reports subpanel was not been properly defined. The function `$this->getOffset` returns an empty string because this calls down the function `$this->getSessionVariable` which if the session variable cannot be found, it returns back an empty string. 

#### Expected Behavior
No warning exceptions.

#### Actual Behavior
PHP logs reports: 

A non-numeric value encountered at ListViewSubPanel.php, line 323 

Because the $offset value is equal to "".

#### Possible Fix
Coercing the value to be an integer:
This could be fixed at superficial level at /include/ListView/ListViewSubPanel.php, line 264 replacing from:
```php
$offset = ($this->getOffset($html_varName)) === false ? 0 : $this->getOffset($html_varName);
```
to:
```php
$offset = ($this->getOffset($html_varName)) === false ? 0 : (int)$this->getOffset($html_varName);
```
Or from the root at /include/ListView/ListView.php at line 875 replacing from:
```php
return $offset;
```
to:
```php
return (int)$offset;
```

I will recommend the second solution as this could also fix problems waiting to happens on other areas where the result of `$this->getOffset` is expected to be a number.

#### Steps to Reproduce
1. Other bugs on AOR_Report were producing this error. The 'Schedule Report' subpanel was not properly created. After fixing those bug, now it is fine, but the it could come back on other areas.

#### Context
Although, in my case it is not failing any more after fixing other bugs, the logic still flawed and it could cause problems on the future. 

#### Your Environment
* SuiteCRM Version 7.8.30
* Chrome Version 75.0.3770.100 (Official Build) (64-bit)
* PHP 7.1.30
* Windows 2016 Server.

Issue #3576, opened 2017-05-21
A non-numeric value encountered in SUITECRM/include/ListView/ListViewSubPanel.php on line [264|546]


jobst:
A non-numeric value encountered in SUITECRM/include/ListView/ListViewSubPanel.php on line 546.

#### Issue

When checking the value of "$current_offset" when the error (subject line) appears its empty/null.
You cannot add 1 to a null/empty value.

#### Expected Behavior

Its probably an oversight of the person writing that code not expecting the person using the function NOT checking whether the passed value is actually an integer value.

#### Actual Behavior

error and stack trace shows up in the log.

#### Possible Fix

Replace this line (264)

           $offset = $this->getOffset($html_varName);

with this line

           $offset = empty($this->getOffset($html_varName)) ? 0:$this->getOffset($html_varName);

Replace this line (546)

            $start_record = $current_offset + 1;

with these three lines

            if( !isset($current_offset) || empty($current_offset) )                                                                    
              $current_offset=0;
            $start_record = $current_offset + 1;

#### Steps to Reproduce

1. open a ANY list of any subpanel
1a. in my case a user that had no role and security management defined
2. if the list that was opened has no member the error arises

#### Context

Its ugly code and throws exceptions everywhere a member list of any subpanel is zero.
It will also slow down the processing as the exceptions needs to be dealt with.

#### Your Environment

* SuiteCRM Version used: 7.8.2
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): N/A
* Environment name and version (e.g. MySQL, PHP 7): N/A
* Operating System and version (e.g Ubuntu 16.04): N/A

chris001:
Good find!  Make a PR.

daniel-samson:
See https://suitecrm.com/wiki/index.php/Contributing_to_SuiteCRM for more information about contributing to SuiteCRM.

jobst:
@chris001 I have a problem submitting this fix - when creating a PR for this it seems wanting to submit my previous PR's as well. 
How can I stop this from happening?

chris001:
[...]

PR #8677, opened 2020-04-24
Fix getOffset recession from #7608, subpanel "end" navigation

pgorod:
## Description

This started with #3576, continued with #7608, leading to a fix #7612.

But the ghastly `getOffset` function handles all kinds of values and types, like 0, 10, '', or 'end', and the fix doesn't take that into account.

https://community.suitecrm.com/t/campaign-module-sub-panel-navigation-is-not-working/71790

So it breaks the "end" navigation in _all_ subpanels. So I am fixing _that_ by moving back to a code that is closer to the original, and fixes the first bug in a less impacting manner... so that I can hope I am not breaking something else. 🤷‍♂️ 

## How To Test This
- to facilitate testing, set a low subpanel limit by adding something like `$sugar_config['list_max_entries_per_subpanel'] = '3';` to `config_override.php`
- get any subpanel to show excessive items, so that a second page is needed.
- click the "end" navigation button on the subpanel:
image

- test also the previous issues that I link above, to check for regressions to the regression 

## Types of changes

### Final checklist

### DIFF
--- a/include/ListView/ListView.php
+++ b/include/ListView/ListView.php
@@ -8,7 +8,7 @@ if (!defined('sugarEntry') || !sugarEntry) {
  * SugarCRM, Inc. Copyright (C) 2004-2013 SugarCRM Inc.
  *
  * SuiteCRM is an extension to SugarCRM Community Edition developed by SalesAgility Ltd.
- * Copyright (C) 2011 - 2018 SalesAgility Ltd.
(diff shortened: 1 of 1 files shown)


pgorod:
User in Forums confirms it works: https://community.suitecrm.com/t/campaign-module-sub-panel-navigation-is-not-working/71790/20?u=pgr

mattlorimer (approved):

PR #8679, opened 2020-04-27
Support regex email subject case ID parsing

tsmgeek:
Allow using regex to parse subject to extract CaseID.
We have seen many cases where clients seem to slightly change the subject to the point it does not parse correctly, or they have translated `CASE` into another language, removed the semicolon or square brackets etc.
This adds ability to use regex string instead of plain string.
Partial backwards compatibility has been added to check for `/` at the start and end of the macro string to detect if its regex, of-course if someone has used these as their ident then may fail.

### DIFF
--- a/modules/InboundEmail/InboundEmail.php
+++ b/modules/InboundEmail/InboundEmail.php
@@ -6013,29 +6013,44 @@ class InboundEmail extends SugarBean
     public function getCaseIdFromCaseNumber($emailName, $aCase)
     {
         //$emailSubjectMacro
-        $exMacro = explode('%1', $aCase->getEmailSubjectMacro());
-        $open = $exMacro[0];
-        $close = $exMacro[1];
-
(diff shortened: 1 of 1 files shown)


Dillon-Brown (approved):
Squashed commits and changed:

```php
if (substr($macro, 0, 1) == '/' && substr($macro, -1) == '/'){
```

To

```php
if ($macro[0] === '/' && substr($macro, -1) === '/') {
```

Rest of the changes LGTM!

Dillon-Brown (approved):

jack7anderson7:
Hi tsmgeek,

I was wondering if you had an example of regex we could use to test this PR correctly.

Thanks, Jack

tsmgeek:
@jack7anderson7 standard preg named groups either it tries to find 'id' or uses the first match. Currently it's case sensitive. 

'/[C(\d+)]/'
'/[C(?<id>\d+)]/'

Issue #8680, opened 2020-04-28
Scheduled Report overlap issue


JackVBuchanan:
#### Issue
When the cron is run on every minute it schedules a job for the scheduled reports
When the report is generated for a long time the scheduler continues to schedule the report on next minute and a while you have multiple cronjobs running and every of them creating the same report

#### Expected Behavior
Overlap should be detected before new job starts

#### Actual Behavior
Multiple jobs create the same report in an overlap

#### Steps to Reproduce

1. Create large reports
2. Set up cron to run every minute
3. See scheduled reports start to overlap due to large report

#### Context
Multiple large report running at the same time can cause high useage/overload

#### Your Environment

* SuiteCRM Version used: 7.10.22
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Chrome latest
* Environment name and version (e.g. MySQL, PHP 7):MySQL, PHP 7.2
* Operating System and version (e.g Ubuntu 16.04): Ubuntu 18.04

PR #8681, opened 2020-04-28
Fix issue 8680 where scheduled reports overlap in cron

JackBuchanan:
Add database check for jobs that are running before starting new job

## Description
#8680 
When the cron is run on every minute it schedules a job for the scheduled reports
When the report is generated for a long time the scheduler continues to schedule the report on next minute and a while you have multiple cronjobs running and every of them creating the same report

## Motivation and Context
Multiple large report running at the same time can cause high useage/overload

## How To Test This

Create large reports
Set up cron to run every minute
See scheduled reports start to overlap due to large report

## Types of changes

### Final checklist

### DIFF
--- a/modules/Schedulers/_AddJobsHere.php
+++ b/modules/Schedulers/_AddJobsHere.php
@@ -762,6 +762,7 @@ function performLuceneIndexing()
 function aorRunScheduledReports()
 {
     require_once 'include/SugarQueue/SugarJobQueue.php';
+    $db = DBManagerFactory::getInstance();
     $date = new DateTime();//Ensure we check all schedules at the same instant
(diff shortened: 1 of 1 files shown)


modules/Schedulers/_AddJobsHere.php
```diff
@@ -777,6 +778,11 @@ function aorRunScheduledReports()
             if (empty($scheduledReport->aor_report_id)) {
                 continue;
             }
+            $queued = $db->fetchOne("SELECT count(*) cnt FROM job_queue WHERE data=".$db->quoted($scheduledReport->id)." and deleted=0 and status = 'running' and execute_time >= DATE_ADD(now(), INTERVAL -6 hour)");
```
mattlorimer, line 781: This won't work in SQL server, you would need to use the conversions at least with DATE_ADD here
plus 6 hours is kind of random

PR #8685, opened 2020-04-30
Only init Currency once when saving 

tsmgeek:
Patch only init Currency class once when saving products as its attached to focus object which does not change per loop.

### DIFF
--- a/modules/AOS_Products_Quotes/AOS_Utils.php
+++ b/modules/AOS_Products_Quotes/AOS_Utils.php
@@ -29,6 +29,8 @@ if (!defined('sugarEntry') || !sugarEntry) {
 
 function perform_aos_save($focus)
 {
+    $currency = fetch_aos_currency($focus);
+
     foreach ($focus->field_defs as $field) {
         $fieldName = $field['name'];
         $fieldNameDollar = $field['name'].'_usdollar';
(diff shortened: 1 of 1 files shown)


tsmgeek:
Can we apply this performance patch?

Dillon-Brown:
Hi @tsmgeek, I've pushed a small change to this just to add a PHPDoc to the new method and change the direct call to:

```php
$currency = new Currency();
``` 

to 

```php
$currency = BeanFactory::newBean('Currencies');
```

I've also squashed the commits down to one as we are looking to use an "Atomic" commit strategy meaning in general most PRs should have a single commit that looks something like:

```
Fix #9999 - Bug in reports
```

Just something to keep in mind for future PRs, don't worry too much about all the ones you've already made, it's something we can easily fix when it comes time to merge (I believe we still need to document this in our how to contribute too).

> Can we apply this performance patch?

Thanks for the reminder, I've seen that you've made some really nice fixes recently (Especially to emails), that I think are worth bringing in asap to our minor releases if we can, I'll highlight this internally and see if we can get some of them across.

Dillon-Brown (approved):

jack7anderson7 (approved):

samus-aran (approved):

PR #8686, opened 2020-04-30
Improved diagnostics

pgorod:
## Description
* improves the output of **Admin / Diagnostics** screen to actually provide meaningful information that is regularly needed in support and troubleshooting
* adds a quick view of the full `phpinfo` directly from that page, without requiring the user to download a zip
* adds a `diagnostics:quick-report` Robo task to produce the same report from the command-line, useful for servers where you can't even get to the **Admin / Diagnostics** page (uninstalled, or broken installations).
* adds a generic mechanism to query the web server side of PHP from the CLI side. This is the first step to having (in the future) a more integrated management of settings, permissions management, etc., across the entire app.

## Motivation and Context
So many posts in the Forums are just to get users to provide basic information... this way it's all there, together in one place.

## How To Test This
1. Start with a QR&R to get the language LBL added.
1. Visit the Diagnostics screen.
1. In that screen, press the new "View phpinfo" **button**.
1. From the command-line: run the Robo `diagnostics:quick-report`

Interesting variations for the tests:
1. Windows / Linux
1. In `php.ini`, add `exec` to `disabled_functions`, common in many shared hostings
1. Using `port` number in `site_url` in `config.php`

## Suggested Future developments
If somebody wants to pick up on this work later, I suggest:
* check permissions and actual directory writeability, and report on that. Make it obvious to users when their installation is still not operational. I don't suggest a useless full check, like the one we have in the Upgrade Wizard, which is dumb (complains in many set ups where access _is_ working). I mean a practical check in a few crucial directories (write file, read it, delete it).
* Add also the reverse mechanism allowing to run CLI jobs from the web server side by persisting something that a custom scheduler will then pick up and execute. This is not that hard to achieve, and will allow us to bridge the two PHP worlds effectively.
* Make better use of the `setPhpIniSettings` function, [called](https://github.com/salesagility/SuiteCRM/blob/master/include/entryPoint.php#L103) in every entry-point,
so that users can select basic PHP settings from within our UI, and have them applied automatically by SuiteCRM, both in Web server and CLI. No more `php.ini` editing!

## Types of changes

### Final checklist


pgorod:
@samus-aran @clemente-raposo 

This is the PR that saves time for people helping out in the forums with installation problems. I know that, because I did more of that work than anyone else.

The SuiteCRM 8 release thankfully caused a flood of such issues, and we have out lead developer answering all that. Maybe we want to save his time for better things?

For example, how many posts do you think this thread will take, for a totally insignificant, well-known issue, before it gets resolved? https://github.com/salesagility/SuiteCRM-Core/issues/29

"Which php.ini shoud I change, which are the effective values, which is my web server user, how can I get the web app settings from the cli, or the cli settings from the web app?" All that is already solved in this PR, and I didn't even start with the permissions issues, which can also be much improved...

include/MVC/Controller/entry_point_registry.php
```diff
@@ -98,4 +98,5 @@
     'sendConfirmOptInEmail' => array('file' => 'include/entryPointConfirmOptInConnector.php', 'auth' => true),
     'saveGoogleApiKey' => array('file' => 'modules/Users/entryPointSaveGoogleApiKey.php', 'auth' => true),
     'setImapTestSettings' => ['file' => 'include/Imap/ImapTestSettingsEntry.php', 'auth' => true],
+    'queryWebServerFromCLI' => ['file' => 'include/entryPointQueryWebServerFromCLI.php', 'auth' => false],
```
clemente-raposo, line 101: Hi @pgorod, what is the need to have this a non-authenticated.
I would prefer to not have it. The less information we expose the better.

clemente-raposo: Hi @pgorod, I think I understand it better now. The idea is to get the info from the webserver, which may be different from the cli one.

lib/Utility/Diagnostics.php
```diff
@@ -0,0 +1,223 @@
...
+{
+    public function checkToken() {
+
+        global $sugar_config;
+
+        if (!isset($_REQUEST['callerToken']) || !isset($sugar_config['cache_dir'])) {
+            return false;
+        }
+
+        $tokenFilename = $sugar_config['cache_dir'] . $_REQUEST['callerToken'] . '.tmp';
```
clemente-raposo, line 55: Hi @pgorod, we should validate `$_REQUEST['callerToken']` and check if it has the format we expect

lib/Utility/Diagnostics.php
```diff
@@ -0,0 +1,223 @@
...
+            fclose($handle);
+
+            $post_data = [
+                'callerToken' => $token,
+                'module' => 'Administration',
+                'action' => 'DiagnosticQuickReport',
+            ];
+            // always make the call as localhost, but get any port number from the site_url, if present
+            $url = preg_replace('/(.*):\/\/(.*?)((?::[0-9]+)|\/)(.*)/', '$1://localhost$3$4', $sugar_config['site_url']);
+            $url .= '/index.php?entryPoint=queryWebServerFromCLI';
```
clemente-raposo, line 90: Hi @pgorod, calling an entryPoint together with an action is no longer allowed

modules/Administration/controller.php
```diff
@@ -166,4 +166,9 @@ public function action_callRebuildSprites()
             $GLOBALS['log']->error($mod_strings['LBL_SPRITES_NOT_SUPPORTED']);
         }
     }
+
+    public function action_diagnosticquickviewphpinfo() {
+        phpinfo();
```
clemente-raposo, line 171: Hi @pgorod, could we add a check here to make sure the user is an admin user, please?

lib/Utility/Diagnostics.php
pgorod: Ok, I will add check for length (since it is constant, always 64 chars) and remove all non-alpha-numeric chars, to avoid any path tricks.

lib/Utility/Diagnostics.php
pgorod: I actually don't need the action there, I will remove it.

modules/Administration/controller.php
pgorod: Ok, will do. I will also add it to the other actions in this controller that are missing it.

pgorod:
I had to do this from Github, not from PhpStorm, so maybe someone can squash the commits for me? Thanks

clemente-raposo (dismissed):
LGTM

modules/Administration/Diagnostic.tpl
```diff
@@ -37,93 +37,128 @@
...
 <form name="Diagnostic" method="POST" action="index.php">
-<input type="hidden" name="module" value="Administration">
-<input type="hidden" name="action" value="DiagnosticRun">
- 
-<table cellpadding="0" cellspacing="0" border="0" width="100%" class="actionsContainer">
-	<tr>
-	<td>
-	<input title="{$MOD.LBL_DIAG_EXECUTE_BUTTON}" class="button" onclick="this.form.action.value='DiagnosticRun';" type="submit" name="button" value="  {$MOD.LBL_DIAG_EXECUTE_BUTTON}  " >
-	<input title="{$MOD.LBL_DIAG_CANCEL_BUTTON}" class="button" onclick="this.form.action.value='index'; this.form.module.value='Administration'; " type="submit" name="button" value="  {$MOD.LBL_DIAG_CANCEL_BUTTON}  "></td>
+    <h4>Download Diagnostic Files</h4><br/>
```
likhobory, line 67: @pgorod 
Could you pls add corresponding label for 'Download Diagnostic Files' ?

modules/Administration/Diagnostic.tpl
```diff
@@ -37,93 +37,128 @@
  * reasonably feasible for technical reasons, the Appropriate Legal Notices must
  * display the words "Powered by SugarCRM" and "Supercharged by SuiteCRM".
  */
+*}
 
-/**
+<form name="Diagnostic" method="POST" action="index.php">
+    <h4>Quick Report</h4><br/>
```
likhobory, line 43: @pgorod
Could you pls add corresponding label for 'Quick Report' ?

pgorod: Done, thanks

modules/Administration/Diagnostic.tpl
pgorod: Done, thanks. I re-used a previously existing label.

pgorod:
@jack7anderson7 please don't forget this one!

dwaltsch:
@clemente-raposo is there any reason there has been no progress lately especially having "adds a quick view of the full phpinfo directly from that page, without requiring the user to download a zip" in suitecrm would be very useful

shubham-pawar:
@pgorod  thank you for the contribution!

Is this already merged into the codebase?

1667647592:
退订

pgorod:
@shubham-pawar no, sadly the PR was never merged. 

Same for #10342

shubham-pawar:
@johnM2401 Could we please review this PR when you have time?

PR #8696, opened 2020-05-01
Proof of Concept: Run PHP-CS-Fixer on the entire codebase

connorshea:
## Description

A draft to show what we could do with PHPCsFixer, see the first commits if you want to just see the `.php_cs.dist` changes.

## Motivation and Context

It cleans up a ton of inconsistencies and problems throughout the codebase, e.g. bad indentation, unnecessary blank lines, removing unnecessary else blocks, etc.

It's not perfect, but I'd say it's almost definitely better than before.

If we changed all the necessary files so the `if (!defined('sugarEntry') || !sugarEntry)` stuff came after the header comment, we could also use the linter to add the header comment automatically.

You can see all the configurable rules in the php-cs-fixer repo: https://github.com/FriendsOfPHP/PHP-CS-Fixer

## How To Test This
Just review the changes.

## Types of changes
Linting/code formatting.

### Final checklist


pgorod:
There are a few serious downsides to running big automatic operations on the whole codebase... 

1. Sometimes things get changed that you are not really happy with, but which you only realize some time later; 

1. Every file's history is polluted with a new commit, and when you're tracking "blame" views and looking for the correct commit that broke something, that can be a big nuisance. 

1. Our upgrade packages would become huge, because the diffs would include basically every file.

connorshea:
@pgorod I think those drawbacks are all worth it. If we resist this then it'll realistically never improve.

pgorod:
I would agree with you for the more serious programming best-practices, code-quality issues, but not for the formatting corrections like adding spaces and moving braces around. When I scanned this PR (only the first dozens of files), that's mostly what I saw.

I'm all for advancing decisively in the direction of code quality, but don't underestimate the _caveats_ I listed above. We get a lot of regressions, and typically these sweeping PR's _do_ bring that sort of regression. Then looking for those regressions in commit histories _is_ a serious challenge, and one that directly affects code quality. Finally, the upgrade packages growing by 20MB or so? We're talking about literally hundreds of broken installations and upgrades. These would be easy to work around with minimal admin skills, but the fact of the matter is that tons of people out there are installing, upgrading and administering SuiteCRM _without_ minimal admin skills...

So this kind of thing has us running around in circles (bug, fix, regression, revert, bug, fix, ...), sometimes for years. I would favor human based, focused approaches. For example: how long would take for a developer to read the code of a single module, and do some basic refactoring? We could use tools like Code-sniffer as an aid, of course. I think you could get very decent improvements to a module in a single sprint. Eventually we would get around to another module, then another... there are many of us.

connorshea:
There are 3600 files changed here, and changes take days or weeks to get merged into the repo. The idea that we'd be able to do this in pieces over time is entirely unrealistic, to be completely honest.

I could definitely see an argument for splitting this up into commits of 5 files at a time so you could bisect it when trying to fix regressions, but I think that's what we should do (still in one PR) rather than doing it in very small chunks over time.

For the packages being too big, is the size limited by PHP's INI or by the SuiteCRM config? If it's the latter, we should fix that. If it's the former, we should probably provide warnings for users right now in current releases about increasing the upload size allowed by PHP. I don't think limiting ourselves like this is doing anything good for the product overall.

pgorod:
Your idea of breaking it into bisectable commits sounds positive.

The upgrade packages problem is with the upload max size directive, as you say, but unfortunately that is not the only problem.

The most critical problem is the number of files (not the file size), as explained in this fix: #8257

But unfortunately all that is still flaky due to #8664 and #8665... plus the fact that some of the previous bugs still have their consequences out there. It sometimes takes two upgrade-cycles to make the upgrader fixes effective. Some people have cluttered `cache/upgrades/temp` directories that they are not aware of, and won't be able to start the upgrades that would fix their upgrader without a visit to the forums and some troubleshooting...

Plus, the installer still has some usages of the old `findAllfiles` code that is a time-bomb, with the current file count in our packages. The fixes I made for the upgrader would need to be made for the installer too.

I think unfortunately we still need a few cycles to develop more "community immunity" to the upgrade package problems, before we can risk a challenging upgrade like this...

> There are 3600 files changed here, and changes take days or weeks to get merged into the repo. The idea that we'd be able to do this in pieces over time is entirely unrealistic, to be completely honest.

I don't thin we need all of this. Have you tried a different scenario where the minor cosmetic changes are not made? Things like adding or removing spaces are not a good reason to incur these penalties I mentioned. Other things might be more important and really help with code-quality.

Maybe we should take this discussion to a list of exactly which Code sniffer rules we should be applying, trying to minimize touched files, and maximize code benefits. What do you think?

connorshea:
In that case, I guess unfortunately those problems can't be worked around as easily as I had hoped :/ 

I think the changes to the spacing and such are still valid in that they make the code easier to read but I guess I won't die on that hill for now. The most important changes are to the vardef/viewdef files. e.g. [this one](https://github.com/salesagility/SuiteCRM/blob/4ee24c78ebfa6d2c4ad6ec6e9bc8b9e590f7858d/modules/vCals/vardefs.php) is a mess that makes a lot more sense when reformatted.

chris001:
It'd be very good to activate it on Travis, to fix bad coding style on all PRs:
https://github.com/FriendsOfPHP/PHP-CS-Fixer#using-php-cs-fixer-on-ci

tsmgeek:
I would limit what sort of things is checks. But would need to be in a warning mode only for quite a while.

- array short vs long
- class methods missing scope
- method args type hits
- no tabs, 4 space indents
- all functions to have a return statement even if its `return $this` in classes
- return type declarations, very much depends on minimum targeted versions (https://mlocati.github.io/articles/php-type-hinting.html) ps I would favour removing support for older PHP versions at least min 7.2 now as its the only supported version
- check for deprecated function calls for the minimum php version allowed

Loads of things but as already said it will throw up loads of issues for some time before it should be allowed to block merges.

chris001:
When you install `php-cs-fixer` on Travis, it'll only auto fix the bad coding style on the modified files of the PR.

horus68:
I'm ok that you start with a clean up of the language files!

gunnicom:
If you do automatic fixing on every PR, then it would be more difficult to find the real changes of that PR, because many lines will be marked as changed although only visual fixes.
Therefore i would prefer one big cleanup.
For the Patchpacket upload size maybe this big cleanup could just be done before a big version change.
So only one big patch packet like 7.11.x -> 7.12.0 and after that the 7.12.x -> 7.12.y would be smaller again.

Issue #3922, opened 2017-07-19
Does anyone know how to get the group emails actually working? Are they broken?


jbeima:
I have done a fresh install of 7.9.2. Then I read that the patch to 7.9.3 carried a bug fix for group emails so I confirmed the challenge still exists in 7.9.3.

Here is what happens and what I have done:

Do a clean install.

Then create 3 admin users. The only data I entered was their first and last names and their positions with the company.

Then I configured the Outgoing email account and made sure it was working. All good.

Then I logged in as each of the 3 admin users one by one and setup their individual private email accounts. All worked perfectly.

Then I logged back in as my initial admin user, went into the Admin - Inbound Email.

Under this menu I created a group email account and entered all the settings for the company "info" email account. All staff member read and respond to the "info" email account.

I then logged in as one of the Admin Users and went into their email. Via All - Email

At the top I clicked on the "gear" to edit the email settings. On the user page I went down and clicked on the "Settings" button in the Email Settings section.

On the Mail Accounts Tab the group info account appeared, and I checked the box.

The on the General Tab under Folder Management the "INBOX" for the Info account showed up.

I went back to "All - Email" and by clicking on the folder at the top this user can now switch between both INBOX and email accounts perfectly. Also the user can select who they want to compose the email from just fine.

NOW here is where the challenge happens.

When I logout and log in as the second user.

I go into All - Email... Click The Gear... Click the Settings Button...

On the Mail Accounts Tab the Info account shows up, so I can click it.

HOWEVER back on the General Tab the INBOX for the General Account does NOT show up.

I have confirmed this is the same for any other user on the system.

The INBOX for the Group Email account only shows up for One User.

How do we get the INBOX to he selected by all the users which have the Group account activated?

Is this a bug? Am I doing something wrong?

John Beima
CDCI Development Inc.
jbeima@cdcidevelopment.com

jbeima:
I should have added.

The site is being run on a CentOS V:7.X web server with PHP V:7.1.6.

I am not sure if this is a PHP challenge, a CSS challenge, a JavaScript Challenge.

I am also using Google Chrome to access the site. Since I know THAT could be a factor as well.

sergio91pt:
@jbeima What bug fix are you talking about?

You created a group email account, but the account is [not associated to a valid (group) user](https://github.com/salesagility/SuiteCRM/pull/3812#discussion_r126318550), so you probably can only open it with the account that created it due to #3719.

[...]

Issue #3719, opened 2017-06-15
Every user can select all privat email folders


wundersause:
Whats happend? 

If an user has setup an imap-account in the user settings, everybody can assign to all existing private mailfolders. / Hotfix suiteCRM 7.9.x

More detailed:
Go to  Profile/Email Settings -> settings. It appears a popup with preferences, visual setting and folder management. 

The available User Folders are assignable for everybody, independent from groups, roles or admin users.

Expected Behavior:
Private Mailfolders should only assignable by the logged user, nobody else. This should be the difference to the groupmail folders. 

Actual Behavior:
Everybody can assign to everybodies mailbox folders. 

Possible Fix:
Cant find any hint.

Steps to Reproduce:

1. Go to Profile/Email Settings -> Settings
2. Choose a folder from the folder management in the popup overlay
3. Its possible to take every private mailfolder from every users account.

For security issues and for a professional use of the Email modul it should be important to fit up the access rights possibilty from the Mailfolder Management

Environment

- CentOs 6.8 x86_64
- MySQL  5.6.35
- suiteCRM 7.9.0 Hotfix

PR #8702, opened 2020-05-06
Incorrect check for group folder

tsmgeek:
If user is not admin and not the user that created group folder then check is `$item->isgroup` instead of `$item['is_group']`.

This leads to no group account being visible in folders list for non admin/created_by users.

SuiteCRM 7.10.25

### DIFF
--- a/include/SugarFolders/SugarFolders.php
+++ b/include/SugarFolders/SugarFolders.php
@@ -741,7 +741,7 @@ class SugarFolder
         $secureReturn = [];
 
         foreach ($return as $item) {
-            if ($item->isgroup === 1 || $item['created_by'] === $user->id || is_admin($user)) {
+            if ($item['is_group'] == 1 || $item['created_by'] === $user->id || is_admin($user)) {
                 $secureReturn[] = $item;
             }
         }


pgorod:
I just ran into this bug with `isgroup` vs `is_group` 🤦 

This is a very flaky area of the code (affects our infamous "From" dropdown in Compose Email window), it would be quite essential to get fix in ASAP (and the other email fixes that @tsmgeek provided 👍 . Thanks everyone.

tsmgeek:
Yup, we have found loads more issues with searching/filtering between IMAP/saved when emails are not imported. We have some fixes but there is an inherent problem with the code imo. 
The old way was in a way better and may be better to go back.

pgorod:
What do you mean by "the old way"? Are you talking about the Email module up to version 7.9, before it was rewritten?

tsmgeek:
> What do you mean by "the old way"? Are you talking about the Email module up to version 7.9, before it was rewritten?

In part, ie where it imports everything. IMAP can be slow. 
Might be an idea to look at how email clients pull in headers etc and x bytes of message etc.
Moving to the Horde IMAP client would also be better as it has support for special flags to help with fast sync etc. 

Another reason it's not so good is that I wanted to write my own email parser to link certian emails to different models and also use regex to parse header (already PR for that) etc but as emails are not imported I can't do this without overriding the main import script. Needs to be more hooks.

greenlanegreb:
Just tried this code (having backed up the old files) will it cause any harm to leave this code in situ in the longer term?

I think the inability to be able to make group mailboxes available to regular users is actually CSS related as it displays differently within Chrome, Firefox and Opera. Could it be that an apply button is actually being hidden by incorrect CSS too? Naturally any access rule changes wouldn't save in this scenario. "None" can be selected as well as all of the other inboxes, singular mailboxes cannot be selected and if one is clicked on, all highlight again. 

Code above has not fixed this. 

Thanks and Kind Regards, 

Beanie.

greenlanegreb:
Any further exploration of this issue please or temporary fixes? Thanks!

greenlanegreb:
This code stops mail from being pulled through to inboxes at least when you change between mailboxes. I will do some further testing including changing the files back from their backups to confirm. It would appear that this patch causes further breakage currently. Many thanks.

pstevens71:
on this change: https://github.com/salesagility/SuiteCRM/pull/8702/commits/5132dc54f7199f8f60ee510183a2f8320325eb8b

That won't quite work, I tried.  the === 1 needs to be in quotes like this:  

`if ($item['is_group'] === '1' || $item['created_by'] === $user->id || is_admin($user)) {`

chris001:
@pstevens71 So does it fix the issue with the `=== '1'` ?

chris001:
That's an unexpected crash, because setting `$this->new_with_id = false;` is [supposed to have the bean **`UPDATE`** the data row in the database table, not **`INSERT`** it](https://github.com/salesagility/SuiteCRM/blob/d0f078b7b75179bcad1cf5f8e7c2d9f5c52df2c0/data/SugarBean.php#L87) as if it were a new bean/row.

pstevens71:
thanks @chris001 I am making some progress on the issue of the group_id not getting populated.  I found exactly where the folders table row for the folder is created and/or updated.  I'm struggling a bit to get the value of is_personal from the inbound email address because the function works in the scope of the folder.  $this->....  references the folder and not the inbound email account which has me struggling.  Once I find how to do that, it's just a matter of populating "is_group" from the row creation and adding an update for "is_group" in the case of an update.  Kind of a novice so I'm struggling a bit to get the value I need.  The function is:  public function save($addSubscriptions = true)  in SugarFolders.php.

I made a video for you to demo the issue with this pull request:  https://github.com/salesagility/SuiteCRM/commit/c9821ba5c2dee77325b7e4d347c3981736c33223

https://vimeo.com/826210605/cb29f87204?share=copy

chris001:
It'd be great to get some input about this bug from...
1. @tsmgeek author of this PR, 
2. @mattlorimer last to commit [SugarFolders.php](https://github.com/salesagility/SuiteCRM/blob/hotfix/include/SugarFolders/SugarFolders.php) about [2.5 years ago](https://github.com/salesagility/SuiteCRM/commits?author=mattlorimer), and 
3. @kgaillot last to commit [_AddJobsHere.php](https://github.com/salesagility/SuiteCRM/blob/hotfix/modules/Schedulers/_AddJobsHere.php) about  [7 months ago](https://github.com/salesagility/SuiteCRM/commits?author=kgaillot).

pstevens71:
I've been doing all kinds of testing/investigation in group emails this week.  There are all kinds of issue regarding the folders, how they get created, updated ,etc.

Here's where I'm at so far:

1) Group emails folders cannot be seen by users because of: https://github.com/salesagility/SuiteCRM/pull/8702/commits/5132dc54f7199f8f60ee510183a2f8320325eb8b  I've thoroughly tested this and it works.  (wish I found it sooner took me a few days to get to this point).

2) Group folders do not get marked as "is_group" on creation or edit.  This is not a bug as far as I can see.  The functionality just doesn't exist.  I have modified SugarFolders.php to get the group folder status of the email account and add it to "is_group" in the folders table on creation or update of the inbound account.  I've tested my modifications pretty well and plan to do a PR to fix this.

So.... 1 and 2 above will at least get users seeing group folders.  ALL Group folders!

However, there are more problems I haven't got to yet.  The more I look the more I find.

3) There seems to be functionality in SugarFolders.php to remove folders when an inbound email account gets deleted.  I've tested and this is not firing at all, so it' s not being called.  I'm trying to figure out the best place to call from.  Currently, deleted folders are still in the database and visible to the user.

4) There's a whole other can of worms I haven't even looked at  yet around the user preferences interface and selecting which group folders you want the user to see as admin and/or the user selects themselves to see.   This whole functionality seems to be broken.  When you (as admin) select  folders for a user it creates subscriptions for ALL of the group folders.  It doesn't seem to "save" the choices.  As a user, same thing, they don't even see the choices in their preferences and nothing gets saved or updated even though the popup says so.

So far, that's what I've found.  I think 1,2 are solved, and 3 I think I can solve.  Haven't got to 4 yet, that seems more complex and involves the JS popup, etc.  but I'll do my best.

chris001:
Found these!
https://github.com/salesagility/SuiteCRM/issues/3719

https://github.com/salesagility/SuiteCRM/issues/3922
https://github.com/salesagility/SuiteCRM/issues/3922#issuecomment-371135598
> This issue should now be resolved. Please re-open if this is still an issue you can replicate and provide clear steps for. Thanks."
@Dillon-Brown

https://github.com/salesagility/SuiteCRM/pull/3722

https://github.com/salesagility/SuiteCRM/blob/d0f078b7b75179bcad1cf5f8e7c2d9f5c52df2c0/modules/InboundEmail/SaveGroupFolder.php#L58

https://github.com/salesagility/SuiteCRM/blob/d0f078b7b75179bcad1cf5f8e7c2d9f5c52df2c0/include/language/en_us.lang.php#LL1174C4-L1174C4

https://github.com/salesagility/SuiteCRM/blob/d0f078b7b75179bcad1cf5f8e7c2d9f5c52df2c0/modules/InboundEmail/EditGroupFolder.php#L119

https://github.com/salesagility/SuiteCRM/blob/d0f078b7b75179bcad1cf5f8e7c2d9f5c52df2c0/modules/Emails/EmailUIAjax.php#L1049

https://github.com/salesagility/SuiteCRM/blob/d0f078b7b75179bcad1cf5f8e7c2d9f5c52df2c0/modules/Emails/include/ListView/ListViewDataEmails.php#L167

https://github.com/salesagility/SuiteCRM/blob/d0f078b7b75179bcad1cf5f8e7c2d9f5c52df2c0/modules/Emails/Email.php#L1900

https://github.com/salesagility/SuiteCRM/blob/d0f078b7b75179bcad1cf5f8e7c2d9f5c52df2c0/modules/InboundEmail/EditView.html#L63

https://github.com/salesagility/SuiteCRM/blob/d0f078b7b75179bcad1cf5f8e7c2d9f5c52df2c0/modules/Emails/templates/editAccountDialogue.tpl#L46

https://github.com/salesagility/SuiteCRM/blob/d0f078b7b75179bcad1cf5f8e7c2d9f5c52df2c0/modules/InboundEmail/language/en_us.lang.php#L112

https://github.com/salesagility/SuiteCRM/blob/d0f078b7b75179bcad1cf5f8e7c2d9f5c52df2c0/modules/Emails/EmailUI.php#L1258

https://github.com/salesagility/SuiteCRM/blob/d0f078b7b75179bcad1cf5f8e7c2d9f5c52df2c0/modules/InboundEmail/DetailView.php#L186


https://github.com/salesagility/SuiteCRM/blob/d0f078b7b75179bcad1cf5f8e7c2d9f5c52df2c0/modules/Emails/language/en_us.lang.php#L300
**Group</b>: Email account accessible by members of specified teams. Team members can manage and import emails from this account.**

https://github.com/salesagility/SuiteCRM/blob/d0f078b7b75179bcad1cf5f8e7c2d9f5c52df2c0/modules/InboundEmail/Save.php#L405

https://github.com/salesagility/SuiteCRM/blob/d0f078b7b75179bcad1cf5f8e7c2d9f5c52df2c0/modules/UpgradeWizard/uw_utils.php#L3575

https://github.com/salesagility/SuiteCRM/blob/d0f078b7b75179bcad1cf5f8e7c2d9f5c52df2c0/modules/InboundEmail/InboundEmail.php#L7778

https://github.com/salesagility/SuiteCRM/blob/d0f078b7b75179bcad1cf5f8e7c2d9f5c52df2c0/modules/TableDictionary.php#L60

https://github.com/salesagility/SuiteCRM/blob/d0f078b7b75179bcad1cf5f8e7c2d9f5c52df2c0/tests/unit/phpunit/modules/InboundEmail/InboundEmailTest.php#L2512

gunnicom:
related? https://github.com/salesagility/SuiteCRM/pull/9816

chris001:
> related? #9816

Yes however the difference is the operators `$item['is_group'] ==` (equality of value)  vs `$item['is_group'] ===` (equality and same data type).

pstevens71:
Wow thanks @chris001 for putting that compendium together.  I had read most of them but it's super convenient to have everything in one place.  Hopefully  we can fix all these group emails issues in one push.

chris001:
@pstevens71 
From reading the comments and documentation site, it looks like, the way it says it's supposed to work, is the group email accounts are for your entire organization, and to refine down who can access these group accounts, you use `ACLgroups` or `SecurityGroups` to restrict each group email account to only users belonging to a certain "team" e.g. the "Sales" team (or ACLgroup) gets to access the "`sales@______.com`" group email account, the "Support" team gets to access the "`support@______.com`" group email account, etc.  

So, you shouldn't need to select who's allowed to see a group inbox on the group inbox page.  Instead, you select users who can access it, from the "ACLgroups" or "Security Groups" page.  

Does this make sense?

pstevens71:
@chris001 that would be wonderful!  And so much easier!  However, it doesn't work that way.  There is no relationship between Inbound Emails and Security Groups.... and no way to create one through the user interface.  


I've traced all this code over the past week.   What actually happens is when you create a group account, nothing really happens in terms of who sees it.  The admin can always see it (the filter in is_admin returns true for the folder in listview).  When an admin goes into a user account, the admin can see ALL mail boxes and thus can subscribe the individual user to the folder.  It then creates an entry in folders_subscriptions table and relates the user to the folder.  When the user clicks on the INBOX button in inbound email listview, they can see a list of folders they are "subscribed" to governed by 2 conditions : 1) they created the folder (this is a whole other problem because if admin sets up their email account  they will never be able to see it).  2) There is a group folder (is_group =1) AND they  have a record in folder_subscription table in the DB in which case they are subscribed.

So, what is broken in this scenario is a few things:

1) The is_group condition in listview never returned true because of the incorrect syntax  (fixed byhttps://github.com/salesagility/SuiteCRM/commit/5132dc54f7199f8f60ee510183a2f8320325eb8b) and users could never see group folders, only admins.

2) Record in folders table in the DB never gets marked as "is_group" on creation or update of an inbound_email record, the functionality is missing to do this.  I have written it an will submit a PR this week.  Without the folder getting marked as "is_group" in the folders table it will never appear in the users listview for inbound emails.

3) When an inbound email account is deleted it doesn't remove them from the folders table or the folders_subscription table and the users can see deleted inbound accounts in their listview  (assuming 1 and 2 are fixed above) .  There is a function to do this in SugarFolders.php but is never gets called.  I've tested this already.  Just need to find where the appropriate place to call it is.

4) When admin selects folders in the user account to subscribe to, it doesn't save correctly.  However, it does create subscriptions in the folder_subscriptions table for ALL folders, so the user can see all folders based on the condition (is_group =1 assuming 1 and 2 are fixed above).  When the admin tries to limit the folders, it still creates subscriptions for all folders regardless.  Haven't troubleshooted this one yet, other than knowing what's not working.   What needs to happen is records in the folders_subscription table are only created  for those which the admin has selected in the user account and records removed for those which are not.   This would limit the the folders the user has access to (or at least ones that are visible, I don't think any of the above actually limits access to the folders other than there is no way for the user to navigate to them).  This code exists, I just haven't got to it yet to see why it's not working properly.

pstevens71:
Ok working on problem When an inbound email account is deleted it doesn't remove them from the folders table or the folders_subscription table and the users can see deleted inbound accounts in their listview (assuming 1 and 2 are fixed above) . There is a function to do this in SugarFolders.php but is never gets called. I've tested this already. Just need to find where the appropriate place to call it is.

I found that when an inbound email address get's deleted, it runs \suitecrm\modules\InboundEmail/Delete.php

This is great!  Now I know where to call the function delete() from SugarFolders.php.

HOWEVER, there is no relationship between the inbound email account and the Folders!!!!

As an example Inbound email account ID = 9fd6f2b6-d27e-3a30-3860-646916d7b037
It has:  Group Folder ID = 995b9d97-1d42-ef53-d1a6-64691630c162   and Group ID = 9959fa3c-0e7f-103d-5416-6469165df7eb  (in inbound_email table)

Neither of which correspond to anything else in the database!

The ID of the parent folder (in folders table) that should be related to this is: a6ae099b-15d2-f2bc-9a67-64691642f9bf

There's no way to ID the folders to be deleted from the inbound email account.  I guess I could look for the email address, but that's not exact.  Am I missing something does anyone know how I'd relate the Inbound email record in the DB to the folder record?  (you'd think it would be by group folder ID but it's not!).

chris001:
> ...does anyone know how I'd relate the Inbound email record in the DB to the folder record? (you'd think it would be by group folder ID but it's not!).

Hope these two could help us here:

1. @daniel-samson who if I remember right, wrote much of this new Email implementation, and
3. @mattlorimer , last to commit [SugarFolders.php](https://github.com/salesagility/SuiteCRM/blob/hotfix/include/SugarFolders/SugarFolders.php) about [2.5 years ago](https://github.com/salesagility/SuiteCRM/commits?author=mattlorimer)

pstevens71:
Thanks I think I got it.  There is what I thought was a perent folder which had no relation to the account.  However, the inbox folder has the same ID as the email account.  All children have a parent ID of the inbox.  That works.  I'm able to get the parent folder ID in delete.php I'm just stuck trying to call delete() function from sugarfolders.php but I'm close I think.

chris001:
@pstevens71 Which IDE are you using, if any?

pstevens71:
Considering I had to Google what an IDE was I think the answer is none.  I'm not a dev,  just kind of took this on as a personal challenge to fix this issue.  I learned to code 40 years ago on a PET computer in BASIC so learning PHP and taking on this is a bit of a stretch for me.

PR #8705, opened 2020-05-07
Fixed #6046 - SQL 'rank' is a reserved word, can't be used as a field…

### LINKED ISSUE #6046: SQL "rank" is a reserved word, can't be used as a field name in MySQL 8.0.2 and later

hannenule:
… name in MySQL 8.0.2 and later

Function oneColumnSQLRep has been modified to add quotation marks to column names for SQL queries.

## Description

Quotation marks added to column names to avoid issues with reserved words in MySql 8 or higher.
Fix Issue #6046 

## Motivation and Context
This change allows us to use the most recent versions of MySQL.

## How To Test This
The most direct way to test is to perform a new installation using MySQL 8.0.19 and see that it installs successfully.

## Types of changes

### Final checklist

### DIFF
--- a/include/database/MysqlManager.php
+++ b/include/database/MysqlManager.php
@@ -828,6 +828,9 @@ class MysqlManager extends DBManager
             $ref['default'] = '';
         }
 
+        // Quote the name column incase it has been reserved by dbms
+        $ref['name'] = $this->quoteIdentifier($ref['name']);
+
         if ($return_as_array) {
             return $ref;
         } else {


murasame78:
Hi.

I used your fix and managed to get passed the pre-check but then I get this message :

Sun Jun 21 16:03:21 2020 [80161][-none-][FATAL] SugarBean::populateDefaultValues $field_defs should be an array
Sun Jun 21 16:03:21 2020 [80161][-none-][FATAL] "name" field does not exists in field definition.
Sun Jun 21 16:03:21 2020 [80161][-none-][FATAL] Mysqli_query failed.
Sun Jun 21 16:03:21 2020 [80161][-none-][FATAL] Error creating table: : Query Failed: CREATE TABLE  (``   NULL  ) CHARACTER SET utf8 COLLATE utf8_general_ci: MySQL error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '(``   NULL  ) CHARACTER SET utf8 COLLATE utf8_general_ci' at line 1

Ubuntu 20.04 Fresh install. 
SuiteCRM 7.11.15
MySQL 8.0.20
PHP 7.4

hannenule:
Hello @murasame78

Check if you did the steps correctly. The error indicates is that a array is expected and is returning something else.
I also wanted to comment that the 'rank' problem occurs during the installation not at the  @pre-check.

Have a great day.

murasame78:
Hello @hannenule,

I can't see how I could make it in a different order. What are you suggesting?

As for my latest tries, I was unable to pass pre-check until I applied your fix.

I have installed several suitecrms on different platform and I just seem to be unlucky with my latest config. (everything freshly installed)

I'd happily follow your instructions / ideas / comments and give you feedback.

hannenule:
Good morning @murasame78, the order is important because the function that is modified returns two different types. but in this case it is indifferent.

The error that is coming out, is because the system are trying to create a table without name and without columns. I don't know what the reason for this is.

I tried to install SuiteCRM 7.11.15 and 7.11.14 (no fix added) and I got the same error.

What I can recommend is to install version 7.11.13 and then upgrade to version 7.11.15.
I did that test and I was able to perform the installation correctly.

Remember to apply my fix (rank error) to version 7.11.13 before starting the installation.

I hope this is helpful to you. Have a nice day.

murasame78:
Good evening @hannenule 

Ok. By order you meant : 

1- Install 7.11.13 (with your fix applied)
2- Upgrade to 7.11.15

I can confirm **it works** in that order. 

Why 7.11.15 won't install directly is out of my understanding... but at least, there is a way for now to run SuiteCRM in a PHP 7.4 and MySQL 8.02 environment.

Thanks for your follow up. Let me now if you need me for further testing. On my side, I'll report any issues.

Have a nice evening.

blackhawkworks (approved):

mattlorimer (approved):

PR #8708, opened 2020-05-07
Fix issues with IMAP InboundEmail

tsmgeek:
Office365 does not support charsets other than US-ASCII, PHP does not throw exception for BADCHARSET which means you will get no search results back at all but instead a literal FALSE.

- This patch will in that event check the last error message and switch to US-ASCII for the rest of the session.


tsmgeek:
Added fixes to imap_sort charset as well.

Removed changes to IncomingEmail and split to another PR

mattlorimer (approved):

PR #8715, opened 2020-05-11
Fix issues with saving drafts and new messages

tsmgeek:
SuiteCRM 7.10.25

When replying to a received message, if you press `Save Draft` or `Discard Draft` it affects the received message instead of creating a new one and making changes to that.

Also when sending new messages the `date_sent_received ` is incorrect based on original message date.

Trying to discard a draft before it has even been saved removes the original message instead of throwing error.

### DIFF
--- a/modules/Emails/Email.php
+++ b/modules/Emails/Email.php
@@ -4677,6 +4677,12 @@ eoq;
                 $bean->new_with_id = true;
                 $bean->type = 'out';
                 $bean->status = 'draft';
+		$bean->date_entered = null;
+		$bean->date_modified = null;
+		$bean->uid = null;
+		$bean->date_start = null;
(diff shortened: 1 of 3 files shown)


tsmgeek:
What im not clear about is the difference between `date_sent` and `date_sent_received`, after migration my old messages have `date_sent` but my new messages do not and vica versa?

Is this related to legacy code or did the migration script fail to move the data across?

Dillon-Brown (dismissed):
Squashed and fixed some minor PSR-2 spacing issues + code cleanup. 

Rest of the changes here LGTM.

Dillon-Brown (approved):

samus-aran:
@jack7anderson7 to assess :+1:

Issue #8645, opened 2020-04-06
Export report fails with a One Of condition set as parameter


QuickCRM:
#### Issue

Export generates an empty CSV when the condition of a dropdown field is set as parameter

#### Steps to Reproduce

0. Go to SuiteCRM Demo
1. Create a report based on Opportunities
2. Add condition on Sales Stage one of Prospecting, Qualification
3. Check that condition as parameter
4. Save
5. Export the report (don't need to change the one of selection)
Empty CSV

6. Uncheck parameter in the condition
7. Save
8. Export the report
The CSV is correctly exported

#### Your Environment

* SuiteCRM Version used: 7.10, 7.11
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)):
* Environment name and version (e.g. MySQL, PHP 7):
* Operating System and version (e.g Ubuntu 16.04):

Mac-Rae:
Hi @QuickCRM,

This issue has been confirmed to exist and is replicable in the latest 7.10.x version.

Thanks :+1:

tsmgeek:
Ive created fix #8717 for this but there is also #8245

Issue #8228, opened 2019-11-11
Report parameters using multi-select don't work


JackBuchanan:
#### Issue
Using a multi-select field works when report runs until user presses update on the parameter on detail view

#### Expected Behavior
Report should filter results based on param conditions

#### Actual Behavior
Report uses incorrect value when making query

#### Possible Fix
WHERE used in sql query is generated as 'Value' when it should be '^Value^'

#### Steps to Reproduce

1. Make multi select field 
2. Set records to use value from multi-select
3. Run report with multi-select as param 
4. Press update and see no records are returned

#### Your Environment

* SuiteCRM Version used: 7.11.9
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Chrome latest 
* Environment name and version (e.g. MySQL, PHP 7): MySQL, PHP 7.1
* Operating System and version (e.g Ubuntu 16.04): Ubuntu 18.04

J-Light:
I would like to add that our ability to use reports have been impacted by this bug and #8645. We are using 7.11.10

pgorod:
@J-Light can you confirm if the proposed fix works for you?

tsmgeek:
Ive created fix #8717 for this but just realised there is also #8245

PR #8717, opened 2020-05-12
AOR_Report - Fix Multi condition not working

tsmgeek:
SuiteCRM 7.10.25

Multi `One of` condition not working on report due to incorrect checking of elements.

https://github.com/salesagility/SuiteCRM/issues/8645


tsmgeek:
Sorry noticed that there was another PR #8228 after creating this but I think mine seems less code changes.

PR #8720, opened 2020-05-12
SugarFolder doing Insert instead of Update

tsmgeek:
SuiteCRM 7.10.25

OLD: https://github.com/salesagility/SuiteCRM/blob/v7.8.31/include/SugarFolders/SugarFolders.php#L891

Seems there was a regression from old 7.8.x on how SugarFolders handles updates/inserts of folders.

Ive reverted it with some tweeks to prevent running empty update statements when there is no `parent_folder`.


Dillon-Brown (approved):
Squashed, rebased, and did a minor code cleanup. 

LGTM!

jack7anderson7 (approved):

PR #8721, opened 2020-05-13
Fix User Incoming Email folder subscriptions

tsmgeek:
SuiteCRM 7.10.25

If you unsubscribe to all IncomingEmail accounts the system will give you all accounts due to use of  `LIKE ''` statement when finding folders to subscribe to.
This prevents that by..
- changing `LIKE` to an equal comparison
- doing array_unqiue+array_filter to prevent duplication


Dillon-Brown (approved):
Squashed and rebased, changes LGTM!

PR #8727, opened 2020-05-13
Replace Crypt_Blowfish with phpseclib

tsmgeek:
SuiteCRM 7.10.25

Replace PEAR Crypt Blowfish with phpseclib in composer.

Added internal caching of Blowfish so sequential calls will not need to reload library each time


tsmgeek:
Can we get this replaced to ease 7.4

tsmgeek:
Reopening to trigger travis-ci

Dillon-Brown (approved):
Squashed, rebased, and added an extra commit that re-adds the ```Blowfish.php``` + ```DefaultKey.php``` files. 

Aim here being to ensure that custom code using: 
```php
require_once('include/Pear/Crypt_Blowfish/Blowfish.php');
```

still works along with aliasing the new classes to the old so that code so that:

```php
function blowfishEncode($key, $data)
{
    $bf = new Crypt_Blowfish($key);
    $encrypted = $bf->encrypt($data);
    return base64_encode($encrypted);
}
```

Also works but gives a deprecation warning.

PR #8728, opened 2020-05-14
Remove all deprecated functions from utils.php

tsmgeek:
Remove all functions marked as @deprecated.
Check all other modules/includes for needed changes.

## Motivation and Context
Clean up dead code


tsmgeek:
Codacy is complaining about things ive not changed

pgorod:
I think Codacy re-examines the _entire_ files that you touch.

About the Deprecateds - I don't think you can just remove them thinking they're not used in the core code. They should stay there (for a while) because they might be used in custom code and add-ons. You'd have to examine how long they have been deprecated. 

I also think `for each` is more performant than `array_map`, and these functions are used everywhere, in every request. I am not sure I would change them just for a slight syntactic convenience (for each is syntactically ok too).

This is just a personal opinion, and I think it's nice that you brought up this discussion.

tsmgeek:
The flag "@deprecated" does say it can be removed in future version but does not stipulate when, maybe is should that have tag on it saying "not before version 7.X.X" then people know when looking at it.

From what ive read array_map and foreach are about the same performance now since PHP 7 (oh that debate again), but the amount of times it called in these methods is low, IMO cleaner code.

chris001:
It's best stick with "Semantic Versioning" aka "SemVer": https://semver.org/ where the version numbers mean something important about changes to the software:

"Given a version number MAJOR.MINOR.PATCH, increment the:

    MAJOR version when you make incompatible API changes,
    MINOR version when you add functionality in a backwards compatible manner, and
    PATCH version when you make backwards compatible bug fixes.

Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format."

So when you **remove a deprecated public API function** or change it in a breaking-backward-compatible way, wait to remove/breaking-change that until you've bumped the MAJOR version, e.g. to 8.0.  So that you don't break third party module code that currently depends on that public API function.

tsmgeek:
I agree with semver guide, but unless someone goes through each and every function removed and what version release etc it will be a faf. Moving forwards indicating which version to remove it in/from will help make task easier. But note I also found deprecated functions which have not applied across the existing code base. As soon as deprecation is added to function then I think the search/replaced on existing code should be done. That's just my op.

chris001:
**Deprecated private** util/helper functions and class/API methods, can be removed at any time because by definition private are for internal use only.

**Deprecated public and protected** util/helper functions and API/class methods should only be removed (or changed in a breaking backwards compatibility way), at major version numbers, e.g. 8.0, in agreement with SemVer, so that derived classes depending on those protected functions, or third part code depending on the public API/helper functions, can upgrade at the major version number.

You can update/change a public and protected function/API/class method, without breaking backwards compatibility, by adding parameters that declare a default value in the function signature, and then just call it with values that override those default values, to access your updated/changed functionality.

tsmgeek:
@chris001 that does not work with utils.php as its not a class file there is lots like this atm

chris001:
@tsmgeek  If third party module publishers are calling functions in `utils.php`, and those functions are removed or changed to break BC, then the module code will crash before next major version 8.0 which would break SemVer.

connorshea:
SuiteCRM doesn't follow traditional semver AFAIK, it deprecates functions in "minor" versions since those are essentially major versions. Moving the goalposts would slow the project down even more than it already is.

Dillon-Brown (changes requested):
Squashed and rebased, @tsmgeek If you're still interested in getting this one across I would recommend we do the following here:

1. Take the replace deprecated method with new method changes such as ```get_select_options``` -> ```get_select_options_with_id``` along with the ```create_guid```, ```is_guid```, ```clean_incoming_data``` and ```array_map``` changes into a new PR to hotfix-7.10.x.
2. Take the removal of the deprecated methods changes into a new PR to develop.

We wouldn't be able to remove these deprecated methods in a minor release as they are public and likely used by custom code (Admittedly we don't always follow semVer perfectly but removing all of these at once would be a huge change). Even to develop we will need to consider the impact this will have for developers that upgrade to 7.12.x and eventually 8.x.

PR #8736, opened 2020-05-21
Update authentication module to utilise get_custom_file_if_exists

markvp:
## Description
The authentication module has many local file includes which makes it very challenging to customise any of the functionality in the /custom folder.
This PR replaces all local includes/requires to use the get_custom_file_if_exists util function allowing any individual file to be customised.
There is also a lot of repetition of folder names that I have extracted into constants to avoid magic-strings.

## Motivation and Context
It is really frustrating to customise any files in this module without having to replicate the entire module into the custom folder.
For me this was related to implementing this: https://github.com/salesagility/SuiteCRM/pull/6850#issuecomment-555936024

## How To Test This
Standard login / logout through each of the authentication modules.

## Types of changes

### Final checklist

### DIFF
--- a/include/entryPoint.php
+++ b/include/entryPoint.php
@@ -130,7 +130,7 @@ require_once 'modules/ACL/ACLController.php';
 require_once 'modules/Administration/Administration.php';
 require_once 'modules/Administration/updater_utils.php';
 require_once 'modules/Users/User.php';
-require_once 'modules/Users/authentication/AuthenticationController.php';
+require_once get_custom_file_if_exists('modules/Users/authentication/AuthenticationController.php');
 require_once 'include/utils/LogicHook.php';
 require_once 'include/SugarTheme/SugarTheme.php';
 require_once 'include/MVC/SugarModule.php';
--- a/modules/Users/authentication/AuthenticationController.php
+++ b/modules/Users/authentication/AuthenticationController.php
@@ -82,17 +82,7 @@ class AuthenticationController
         if ($type == 'SugarAuthenticate' && !empty($GLOBALS['system_config']->settings['system_ldap_enabled']) && empty($_SESSION['sugar_user'])) {
             $type = 'LDAPAuthenticate';
         }
-
-        // check in custom dir first, in case someone want's to override an auth controller
-        if (file_exists('custom/modules/Users/authentication/'.$type.'/' . $type . '.php')) {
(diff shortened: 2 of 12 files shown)


pgorod:
We shouldn't add `get_custom_file_if_exists` in places where customization by overriding classes is possible. We shouldn't encourage and inferior type of customization (replacing the entire file) in place of a superior one (inheritance, partial redefinition, ability to call `parent`, ability to consolidate changes from several sources into a single coherent whole).

Did you try the class override? what problems did you run into?

chris001:
The core sugar specs require that when a user installs custom code into the `custom` directory, that the core detect this and automatically load the custom class instead of the core class file.  This should be consistently working everywhere.  PSR-0 and PSR-4 should also be working everywhere.  One should take precedence over the other.  
A good reference to pattern after by default would be the Sugar 9.0 developer guide where it explains how Sugar 9 uses a custom handler with composer for auto loading code etc:
https://support.sugarcrm.com/Documentation/Sugar_Developer/Sugar_Developer_Guide_9.0/Introduction/Composer/

markvp:
@pgorod can you please be specific where you think that applies here?
I need to customise the onelogin/settings.php file but because of the inconsistent retrieval of custom files, I have to customise SAML2Authenticate.php to load that custom file, but then because it uses relative directory references, I have to copy SugarAuthenticate.php to the custom folder as well. I don't see how overriding a class is the simplest or best solution here? All I want to do is customise a settings file...
With this code change, I am ensuring that any file can be placed in the custom folder and it will be retrieved correctly.

@chris001 I agree, any custom code should be loaded. That's what I'm trying to implement here. Can you give me an example of the pattern you're referring to? I can't see a pattern in your reference.

pgorod:
@markvp my comment was preliminary and was more of a question than a statement. If you (or we) conclude that this is the right way to do it, great. I don't see any problem in using a `get_custom_file_if_exists` for a settings file.

I am not sufficiently familiar with the architecture at this point in the app to have a final opinion on this... 

My only objection would be to replacing a superior customization scheme with an inferior one, if that was the case. For example, customizing single methods in controllers via the Extension Framework is easy, and shouldn't be replaced with a `get_custom_file_if_exists` of the entire controller file. As long as you check that none of your changes are of this kind, I'm fine with them :+1:

PR #3740, opened 2017-06-18
Fixed #3594 - AOR Report- Datetime is not converted to UTC timezone thus providing inaccurate report

ChangezKhan:
## Description

When you create a report with date parameter of type 'Period' with any value, then the resultant report is not accurate.

If you log the query, which gets generated for this report, then you can see that the datetime doesn't get converted to UTC timezone and because of this, inaccurate reports gets generated.

So, to fix this issue, changed timezone of date parameter to UTC/GMT.

Steps to reproduce:

1. Create a new **Report** with any datetime parameter as a condition and type as a _Period_ and value as a _Last Quarter_
2. Log the query and check the datetime parameter should be converted to UTC/GMT timezone.

Issue no. #3594 

## Motivation and Context

It converts datetime to UTC/GMT timezone.

## How To Test This

Steps to reproduce:

1. Create a new **Report** with any datetime parameter as a condition and type as a _Period_ and value as a _Last Quarter_
2. Log the query and check the datetime parameter should be converted to UTC/GMT timezone.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [x] My code follows the code style of this project found [here](https://suitecrm.com/wiki/index.php/Coding_Standards).
- [ ] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://suitecrm.com/wiki/index.php/Contributing_to_SuiteCRM) guidelines.

Issue #3594, opened 2017-05-25
AOR Report- Datetime is not converted to UTC timezone, thus providing inaccurate report


ChangezKhan:
#### Issue

When you create a report with date parameter of type 'Period' with any value, then the resultant report is not accurate.

If you log the query, which gets generated for this report, then you can see that the datetime doesn't get converted to UTC timezone and because of this, inaccurate reports gets generated.

I have created one report with date_entered as parameter of 'Period' type with value of 'Last Quarter'. And this is the query got generated.

    SELECT [leads].do_not_call AS 'Do_Not_Call0', [leads].first_name AS 'First_Name1', 
    [leads].last_name AS 'Last_Name2', [leads].date_entered AS 'Date_Created3', 
    [leads].id AS 'leads_id' FROM [leads]  WHERE (  
    [leads].date_entered BETWEEN '2017-01-01 00:00:00' AND '2017-04-01 00:00:00' ) 
    AND  leads.deleted = 0 

Here in above query, you can see that data is extracting of 1st quarter, but it should check the date_entered between `2016-12-31 18:30:00` and '2017-03-31 18:30:00'. 

#### Expected Behavior

Datetime should get converted to UTC timezone and return the accurate report.

#### Actual Behavior

Datetime doesn't get converted to UTC timezone and return the wrong report.

#### Possible Fix

Convert datetime to UTC timezone before returning the value to query.

#### Steps to Reproduce

1. Create a report with a date parameter condition
2. Select Period as type and choose any value
3. Inaccurate report gets generated.

#### Your Environment

* SuiteCRM Version used: 7.8.2
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): NA
* Environment name and version (e.g. MySQL, PHP 7): PHP 5.3, MSSQL Server 2008
* Operating System and version (e.g Ubuntu 16.04): Windows 7

Mausino:
@Dillon-Brown this is big issue.. because if paralyses all reports on period base. :( and solution we have here more as 1 year. It should be in next milestone 7.11.3 👍

tsmgeek:
Can we get this merged in as reports are problematic

johnM2401:
Hey!
Thank you for raising this!

This appears to have since been resolved.
Notably with the PRs: https://github.com/SuiteCRM/SuiteCRM/pull/9552 and https://github.com/SuiteCRM/SuiteCRM/pull/9806

I'm going to close this off for now.

[...]

PR #8764, opened 2020-06-10
Fix date calculation for quarters/periods

tsmgeek:
When you create a report with date parameter of type 'Period' with any value, then the resultant report is not accurate due to period end dates being the 1st instead of last Day of month.

#3594
OLD PR #3740

### DIFF
--- a/tests/unit/phpunit/modules/AOR_Reports/aor_utilsTest.php
+++ b/tests/unit/phpunit/modules/AOR_Reports/aor_utilsTest.php
@@ -1,0 +1,37 @@
+use SuiteCRM\Test\SuitePHPUnitFrameworkTestCase;
+
+require_once __DIR__ . '/../../../../../modules/AOR_Reports/aor_utils.php';
+
+/**
+ * Class aor_utilsTest
+ */
+class aor_utilsTest extends SuitePHPUnitFrameworkTestCase
+{
+    public function testsetMaxTime()
+    {
+        $dateTime = DateTime::createFromFormat(
+            'Y-m-d H:i:s',
+            '1970-01-01 12:00:00'
+        );
+        setMaxTime($dateTime);
+
+        self::assertEquals(86399, $dateTime->getTimestamp());
+
+        if (PHP_VERSION_ID >= 70100)
+            self::assertEquals(999999, intval($dateTime->format('u')));
+    }
+
+    public function testsetMinTime()
+    {
+        $dateTime = DateTime::createFromFormat(
+            'Y-m-d H:i:s',
+            '1970-01-01 12:00:00'
+        );
+        setMinTime($dateTime);
+
+        self::assertEquals(0, $dateTime->getTimestamp());
+
+        if (PHP_VERSION_ID >= 70100)
+            self::assertEquals(0, intval($dateTime->format('u')));
+    }
+}


tsmgeek:
@mattlorimer  ive updated this as also noticed issues with filtering using periods, ie `Last Month` would filter date as ...

`date BETWEEN 2020-01-01 00:00:00 AND 2020-02-01 00:00:00`

This can be problematic when filtering on Date only fields. 
Ive added in helper functions to set DateTime with microsecond granularity if PHP >=7.1 to 23:59:59.

modules/AOR_Reports/aor_utils.php
```diff
@@ -334,81 +336,85 @@ function getPeriodDate($date_time_period_list_selected)
...
-            $datetimePeriod = new DateTime('next year first day of january');
-            $datetimePeriod->setTime(0, 0, 0);
+            $datetimePeriod = new DateTime('this year last day of december');
+            setMaxTime($datetimePeriod);
             break;
         case 'last_year':
-            $datetimePeriod = new DateTime("this year first day of january");
-            $datetimePeriod->setTime(0, 0, 0);
+            $datetimePeriod = new DateTime("last year last day of december");
+            setMaxTime($datetimePeriod);
```
Dillon-Brown, line 413: I would suggest adding a default case here, maybe something like:

```php
        default:
            throw new InvalidArgumentException(
                $dateTimePeriodListSelected . ' is not a valid date time period'
            );
```

modules/AOR_Reports/aor_utils.php
```diff
@@ -334,81 +336,85 @@ function getPeriodDate($date_time_period_list_selected)
...
             break;
     }
+    
+    $datetimePeriod = $timedate->tzGMT($datetimePeriod);
 
     return $datetimePeriod;
```
Dillon-Brown, line 419: This can be returned directly:

```suggestion
   return $timedate->tzGMT($datetimePeriod);
```

modules/AOR_Reports/aor_utils.php
```diff
@@ -570,3 +585,19 @@ function convertToDateTime($value)
...
     return $dateTime;
 }
+
+function setMinTime(DateTime $dt){
+    if (version_compare(PHP_VERSION, '7.1.0') >= 0) {
+        $dt->setTime(0, 0, 0, 0);
+    } else {
+        $dt->setTime(0, 0, 0);
+    }
+}
```
Dillon-Brown, line 595: Can this be dropped? I don't see it being used anywhere.

modules/AOR_Reports/aor_utils.php
```diff
@@ -570,3 +585,19 @@ function convertToDateTime($value)
...
+
+function setMaxTime(DateTime $dt){
+    if (version_compare(PHP_VERSION, '7.1.0') >= 0) {
+        $dt->setTime(23, 59, 59, 999999);
+    } else {
+        $dt->setTime(23, 59, 59);
+    }
+}
```
Dillon-Brown, line 603: Seeing as this is used by just reports, do we really need to add the overhead of running version compares to include micro-seconds?

```suggestion
/**
 * @param DateTime $dt
 */
function setMaxTime(DateTime $dt)
{
    $dt->setTime(23, 59, 59);
}
```

Dillon-Brown (commented):
Left a few comments.

modules/AOR_Reports/aor_utils.php
tsmgeek: My thought is from the DB perspective, as 59 vs 59.999999 are different.

I modeled it similar to LocalDateTime which has min/max constants but in PHP microseconds have only been supported after 7.1. If minimum ver was set then it could be simplified.

modules/AOR_Reports/aor_utils.php
tsmgeek: I have a tenancy to like having the inverse method to setMaxTime.
Setting microtime is just to make sure we are always at a known microtime of 0, as say now() may return microseconds above 0 and you will have DB issues when doing >=0 vs >=0.124

modules/AOR_Reports/aor_utils.php
tsmgeek: Agreed makes sense.

modules/AOR_Reports/aor_utils.php
Dillon-Brown: True, however I believe the only places we call ```getPeriodEndDate``` and ```getPeriodDate``` are here: 

https://github.com/salesagility/SuiteCRM/blob/hotfix-7.10.x/modules/AOR_Reports/AOR_Report.php#L1896
https://github.com/salesagility/SuiteCRM/blob/hotfix-7.10.x/modules/AOR_Reports/AOR_Report.php#L1740
https://github.com/salesagility/SuiteCRM/blob/hotfix-7.10.x/modules/AOR_Reports/AOR_Report.php#L1897

Which all format down to the second. Also it appears as though the only method to call ```calculateQuarters()```

https://github.com/salesagility/SuiteCRM/blob/hotfix-7.10.x/modules/AOR_Reports/aor_utils.php#L238

Is returning with:

```php
    $datetime_period = $datetime_period->setTime(0, 0, 0);

    return $timedate->tzGMT($datetime_period);
```

So unless I'm mistaken there isn't actually anywhere that will run against the DB using ```999999``` even if it's present.

tsmgeek: We can not for sure know what data users are writing directly to the DB or the column format they are using, this is just a belts and braces approach to make sure that queries against the DB are inclusive of all rows even if people are using internal methods.

Sub-second accuracy especially when you want to make sure audit logs are in order is important, this just covers the bases.

Ive got a custom module that is externally populated and tracks external events, to maintain order im using subsecond timestamps, but then we report on it so need to make sure that its correct.

Of course this would not be a problem if the SQL query was "date >= XXXX and < XXXX (rounded up to next second)` but when its using `<=` or `BETWEEN` the problem can become apparent.

Not that we use PG for SuiteCRM (though a few have asked for that compatibility to be added in the past), sub second accuracy is fairly common place with datetime field.

To that end its futureproofing IMO.

tsmgeek: This post pointed to the prior issue of pushing the second forward due to use of `BETWEEN` so I corrected it with microsecond precision. We were having reports including things that were exactly on `00:00:00.000000`.

https://github.com/salesagility/SuiteCRM/pull/8764#issuecomment-818815096

Dillon-Brown: Yeah I see your point, in that case I would suggest the following:

```php
/**
 * @param DateTime $dt
 * @noinspection DateTimeSetTimeUsageInspection
 */
function setMaxTime(DateTime $dt)
{
    if (PHP_VERSION_ID >= 70100) {
        $dt->setTime(23, 59, 59, 999999);
    } else {
        $dt->setTime(23, 59, 59);
    }
}
```

Just to remove the function call and set a noinspection. It may also be worth adding even a basic test for this so we can check nothing crazy happens across PHP versions, maybe something like:

```php
use SuiteCRM\Test\SuitePHPUnitFrameworkTestCase;

require_once __DIR__ . '/../../../../../modules/AOR_Reports/aor_utils.php';

/**
 * Class aor_utilsTest
 */
class aor_utilsTest extends SuitePHPUnitFrameworkTestCase
{
    public function testsetMaxTime()
    {
        $dateTime = DateTime::createFromFormat(
            'Y-m-d H:i:s',
            '1970-01-01 00:00:00'
        );
        setMaxTime($dateTime);
        $timeStamp = $dateTime->getTimestamp();

        self::assertEquals(86399, $timeStamp);
    }
}
```

Inside ```tests/unit/phpunit/modules/Aor_reports/aor_utilsTest.php```

tsmgeek: Ive modified and added UT

PR #8770, opened 2020-06-12
IncomingEmail PHP errors

tsmgeek:
Fixes to prevent some log error messages when...
- if false is returned then `count()` will fail so there is a minor check if is_array.
- check if we have a Note to save as attachment else skip to prevent errors

### DIFF
--- a/include/database/MysqlManager.php
+++ b/include/database/MysqlManager.php
@@ -828,6 +828,9 @@ class MysqlManager extends DBManager
             $ref['default'] = '';
         }
 
+        // Quote the name column incase it has been reserved by dbms
+        $ref['name'] = $this->quoteIdentifier($ref['name']);
+
         if ($return_as_array) {
             return $ref;
         } else {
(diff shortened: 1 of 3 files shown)


tsmgeek:
`Warning [2]: Declaration of AOPInboundEmail::handleCreateCase(Email $email, $userId) should be compatible with InboundEmail::handleCreateCase($email, $userId) in /var/www/html/modules/InboundEmail/AOPInboundEmail.php on line 27`

In which method should we fix this error. Personally I would update InboundEmail to match as I can not see why it should be a problem anyway.

tsmgeek:
Reopening to trigger travis-ci

tsmgeek:
@mattlorimer Was this merged in then lost?

Dillon-Brown:
> @mattlorimer Was this merged in then lost?

I don't think this has been merged. Most likely the references seen here on GitHub about force pushing hotfix-7.10.x is because there are other commits in the PR like https://github.com/salesagility/SuiteCRM/pull/8770/commits/9dc49c7a8544c6ed560af4f72048b846284052a0 that have been merged. Going to squash and rebase this.

Dillon-Brown (approved):
LGTM

PR #8774, opened 2020-06-16
Fix #8771 - Silent failure when no PHP-json module installed

### LINKED ISSUE #8771: Silent failure when no PHP-json module installed
Silent failure when no PHP-json module installed

#### Issue
When installing Suite CRM, a number of modules are checked for on the first page, the ability to call the json_encode function is not one of these, if this module is not installed then there is a failure later in the process whereby a blank 500 page is returned. It took some time to figure out why this was the case (I spent a fair bit of time assuming it was a permissions issue in writing the configuration files) and it would seem that the problem wouldn't have occurred if the dependency was checked in the same way, as, say, xml_parser_create is scanned for.

#### Expected Behavior
I'd expect the page that checks for dependencies to check that the relevant JSON functions are present and tell me if they aren't.

#### Actual Behavior
No error is given on-screen about missing JSON dependencies.
A 500 error is returned with little context to the user after putting in config details.

The following appears in a log:

    PHP Fatal error:  Uncaught Error: Call to undefined function json_encode() in /home/crm/www/install/performSetup.php:53 Stack trace:
    #0 /home/crm/www/install/performSetup.php(58): installStatus('', NULL, true, '')
    #1 /home/crm/www/install.php(793): require('/home/crm/www/i...')
    #2 {main}

#### Possible Fix
Add JSON as a dependency for the dependency screen the user hits.

I think some extra code in installSystemCheck.php near where the block that starts:

    if (!function_exists('xml_parser_create')) {

with something that starts:

    if (!function_exists('json_encode')) {

would be appropriate.

#### Steps to Reproduce
1. Install CentOS 8.1
2. Install PHP etc. (but not the php-json RPM)
3. Observe that the dependencies screen shows no issues
4. Put in valid configuration details
5. Press next, a 500 error is returned to the browser.

#### Context
I was trying to install SuiteCRM for the first time. I'd suspect that many operating systems have JSON encoding functions as the base install of PHP, Centos 8.1 does not do this.

#### Your Environment
* SuiteCRM Version used: SuiteCRM-7.10.27
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Chrome on Linux
* Environment name and version (e.g. MySQL, PHP 7): php-7.2.11 , mariadb-10.3.17
* Operating System and version (e.g Ubuntu 16.04): CentOS Linux release 8.1.1911

Note that to reproduce the bug, it's critical not to have php-json installed.

maaarghk:
## Description
PHP can be compiled without JSON support and this will cause
SuiteCRM to throw 500 errors. Add a check to the installer to
advise the user of the missing dependency.

## Motivation and Context
Bugfix #8771 

## How To Test This
Before PR:
1. Install CentOS 8.1
1. Install PHP etc. (but not the php-json RPM)
1. Observe that the dependencies screen shows no issues
1. Put in valid configuration details
1. Press next, a 500 error is returned to the browser.

After PR:
1. Install CentOS 8.1
1. Install PHP etc. (but not the php-json RPM)
1. The dependencies screen will show a json issue.

## Types of changes

### Final checklist


maaarghk:
cc'ing @mattlorimer as you seem to be the main maintainer based on a 15 second scan of the last month of commits. FYI The tool behind the above link does not work (disabled select box labelled "version" only). Please consider this comment as my acceptance of the terms of the agreement.

samus-aran:
Hi @maaarghk. Thanks for your comment and we will add you to list to import into the CLA signature list.

However if you had a chance before then to try it out again as there was a API rate limit hit (due to rechecking the PRs) that would be much appreciated. Thanks for your cooperation.

mattlorimer (approved):

PR #8830, opened 2020-07-27
Fix handling of file names with underscores in download.php

pgorod:
## Description
From Forums topic: photo files with extra underscores weren't downloading correctly.
https://community.suitecrm.com/t/broken-photo-for-id-like-zcrm-12345/74087/21?u=pgr

## Motivation and Context
The code in `download.php` is using a simplistic method to extract the ending in `_photo` or `photo_c` used by image fields. It assumes the filename has no additional underscores, which is not always the case.

## How To Test This
- Studio: Add Photo field to detail views in Contacts module (which has an existing `photo` field by default, it just needs adding to view)
- Studio: Add Photo field Products module (which does not have an existing `photo` field by default) and add it to Detail view
- QR & R
- add a couple of photos in records from both modules
- check that the detail view displays the uploaded photos correctly in both modules
- try the same with Bean ids that are not UUIDs, but are set from the database with values including underscores (this is typical of import situations and integration solutions).

(User in Forums thread has confirmed the fix works well)

## Types of changes

### Final checklist

### DIFF
--- a/download.php
+++ b/download.php
@@ -112,10 +112,13 @@ if ((!isset($_REQUEST['isProfile']) && empty($_REQUEST['id'])) || empty($_REQUES
             sugar_die("Remote file detected, location header sent.");
         }
     } // if
-    $temp = explode("_", $_REQUEST['id'], 2);
+
+    // split the id by the last underscore, isolating names like 'photo' or 'photo_c' for handling below:
+    $temp = explode('_', $_REQUEST['id']);
     if (is_array($temp)) {
-        $image_field = isset($temp[1]) ? $temp[1] : null;
-        $image_id = $temp[0];
+        $image_field = (count($temp) > 1) ? array_pop($temp) : null;
+        $image_field = ($image_field === 'c') ? array_pop($temp) . '_c' : $image_field;
+        $image_id = implode('_', $temp);
     }
     if (isset($_REQUEST['ieId']) && isset($_REQUEST['isTempFile'])) {
         $local_location = sugar_cached("modules/Emails/{$_REQUEST['ieId']}/attachments/{$_REQUEST['id']}");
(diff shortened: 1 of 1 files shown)


mattlorimer (approved):

cjmny:
It looks like this merge broke display of images in the 7.10.28 update.  For example, I have a field "storefront_photo_c" in a custom module.  When you try and view the record, this error is generated in the log and the image is broken.

Query Failed: SELECT photo_c FROM sr_shopreports LEFT JOIN sr_shopreports_cstm cstm ON cstm.id_c = sr_shopreports.id WHERE sr_shopreports.id= '5ec8d813-a148-47de-c523-5f96fd14c88f_storefront': MySQL error 1054: Unknown column 'photo_c' in 'field list'

pgorod:
Sorry about the inconvenience. 

I think you might be hitting a quirk of the naming system. What this code is doing (not a good idea, but it's probably been like this since the beginning), is inferring the field name from the file name. 
So a file name of `12345_photo_c` indicates it's from a field called `photo_c`. 

SuiteCRM appends those terminations, `photo` and `photo_c`, by itself.

Now, my guess is that you named your field "Storefront Photo" and Studio turned that into `storefront_photo_c`. Note that now the `photo` part is given by you, it's not an appendix added to the file names. And this is confusing the code in an unexpected way. 

I'd like to check if my preliminary theory above is correct. 

1. Do you have files in `upload` folder called something like `c6d669b4-d53a-d38c-f8d1-5e6fa5de5918_storefront_photo_c` or are they called `c6d669b4-d53a-d38c-f8d1-5e6fa5de5918_storefront_photo_photo_c`?

2. Exactly which photo fields do you have in database table `sr_shopreports`?

3. Exactly which photo fields do you have in database table `sr_shopreports_cstm`?

Thanks

cjmny:
No inconvenience at all.  I'm up and running with this merge reversed.

I think your theory is correct.  

1) In my upload folder, I have files named like this: 710caa35-228e-529f-8b87-5f9c3b5be204_storefront_photo_c

2) sr_shopreports has no photo (image) fields.  

3) sr_shopreports_cstm has 4 (and they all had the same display issue): lobby_photo_c, storefront_photo_c, work_area_photo_c, womm_photo_c

Let me know if I can provide anything else to help.  Thanks!

pgorod:
Can you please test the fix to the fix, my new version is here:

https://github.com/pgorod/SuiteCRM/blob/patch-9/download.php

cjmny:
Just uploaded the new download.php and it seems to be working!  Thank you!  If I notice any quirks I'll let you know.

PR #8864, opened 2020-09-11
Fixed #8852 - Bounce Handling account is not correctly set in email campaigns

### LINKED ISSUE #8852: Bounce Handling account is not correctly set in email campaigns
**Issue Description**
When defining an email campaign Suite CRM allows to define the following 3 different accounts to be used:
- Bounce Handling Account, which is supposed to be used as "return path" to catch the bounces;
- From Address, which is supposed to be used as sender;
- Reply-to Address, which is supposed to be used as default address for potential answers.
The problem is that the "Bounce Handling Account" is not used correctly and, in its place, the "From Address" is always set as "return path".

**Expected Behavior**
Given the campaign definition as follows:
- Bounce Handling Account = Address 1
- From Address = Address 2
- Reply-to Address = Address 3

The email source code should contain the following information:
- Return path = Address 1
- From = Address 2
- Reply-to = Address 3

**Current Behavior**
Given the campaign definition as follows:
- Bounce Handling Account = Address 1
- From Address = Address 2
- Reply-to Address = Address 3

The email source code contains the following information:
- Return path = Address 2
- From = Address 2
- Reply-to = Address 3

**Steps to Reproduce**
You can easily reproduce the issue by creating a test campaign with 3 different addresses set as "bounce", "from" and "reply-to" and verify the wrong assignments by looking at the source code of the email that the system sends.

**My Environment**
* SuiteCRM Version used: Version 7.11.15 / Sugar Version 6.5.25 (Build 344)
* Browser name and version: Mozilla Firefox 79.0 (64 bit)
* Environment name and version (e.g. MySQL, PHP 7): PHP Version 7.3.19-1~deb10u1
* Operating System and version: Windows 10'

stenfalp:
Corrected bounce address logic in email campaigns.

## Description

Switched the if condition to what seems more rational.
[See issue 8852](https://github.com/salesagility/SuiteCRM/issues/8852)

## Motivation and Context

Email bounces go to the email address specified in the Return-Path header field. The current code gets the address from the From field which can and should be different from the Return-Path field.  Furthermore the current code has a meaningless if condition because the ``` $this->current_emailmarketing->from_addr```  is mandatory.

## How To Test This

Create a bounce handling account.
Create an email campain with a couple of not existant email address.
See if the bounce count match. (I've tested it on my system obviously)

## Types of changes

### Final checklist

### DIFF
--- a/modules/EmailMan/EmailMan.php
+++ b/modules/EmailMan/EmailMan.php
@@ -1037,7 +1037,7 @@ class EmailMan extends SugarBean
 
             $mail->ClearAllRecipients();
             $mail->ClearReplyTos();
-            $mail->Sender = $this->current_emailmarketing->from_addr ? $this->current_emailmarketing->from_addr : $this->mailbox_from_addr;
+            $mail->Sender = $this->mailbox_from_addr ? $this->mailbox_from_addr : $this->current_emailmarketing->from_addr;
             isValidEmailAddress($mail->Sender);
             $mail->From = $this->current_emailmarketing->from_addr ? $this->current_emailmarketing->from_addr : $this->mailbox_from_addr;
             isValidEmailAddress($mail->From);


pjyoti562 (approved):
It works so well. Thanks a lot.

mattlorimer (approved):
LGTM 👍🏻

johnM2401 (changes requested):
Hey!
Thank you for your PR!

I've given this a test on the most recent release, 7.14.6, and unfortunately this PR seems to no longer be applicable.
(Unless I've perhaps misunderstood the issue/fix?)

<br>

I believe this PR was submitted around the 7.11.x release window.

However, there have been significant changes to the Email behaviour since then, (particularly in 7.13.0), that I think make it so this fix does not work on the most recent releases.

See gif:

bounceMail

After applying the fix, I send a Campaign.
My System Email is "test@test.com", I set my from to "zzzzzz@zzz.com"
However, after sending, both the "return-path" and "from" match, where they should not.

Thank you again for your contribution!

PR #8875, opened 2020-09-24
Fix the module display permission of the home dashboard

296721605:
The permission control method of the home dashboard: function checkDashletDisplay(...), when the permission is judged, it will always be true, I think this is wrong

## Motivation and Context
I need to show and hide some modules in the home dashboard based on permissions。
image
This code will cause the permission control to always be true, in this way, the permission control is invalid

## How To Test This

Install the permission plug-in, and then install another plug-in. At this time, SecurityGroups will be placed on the home dashboard, even if the permissions that cannot be viewed are configured, it will be invalid
## Types of changes

### Final checklist

### DIFF
--- a/include/MySugar/MySugar.php
+++ b/include/MySugar/MySugar.php
@@ -72,22 +72,18 @@ class MySugar
 
     public function checkDashletDisplay()
     {
-        if ((!in_array($this->type, $GLOBALS['moduleList'])
-                && !in_array($this->type, $GLOBALS['modInvisList']))
-                && (!in_array('Activities', $GLOBALS['moduleList']))) {
-            $displayDashlet = false;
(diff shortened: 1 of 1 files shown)


include/MySugar/MySugar.php
```diff
@@ -80,8 +80,9 @@ public function checkDashletDisplay()
             $bean = SugarModule::get($this->type)->loadBean();
             if (!ACLController::checkAccess($this->type, 'list', true, $bean->acltype)) {
                 $displayDashlet = false;
+            }else{
+                $displayDashlet = true;
```
pgorod, line 84: This is just a suggestion, but I would simplify the function by moving the `$displayDashlet = true;` to the top of the function, at the every beginning, and then remove the two elses.

pgorod (dismissed):

PR #8958, opened 2020-11-13
Use HTML entity for ellipsis instead of regular triple dots

BKPepe:
## Description
In language file, there is used HTML entity for non-breaking space (`&nbsp;`),
so I replaced triple dots with HTML entity (`&hellip;`) for it to be semantically
correct.

There is a difference between using ... (triple dots) or rather ellipsis char …

## Motivation and Context
Fix broken HTML

## Types of changes

### Final checklist


pgorod:
With all the over-zealous security code we have all over the place, I'd be careful testing this PR to make sure nothing gets lost along the way...

horus68:
What is the point to have html entities in language files? It should be plain text.
Besides, this change will cause strings involved to be translated again on Crowdin.
3pontos

Issue #7255, opened 2019-05-05
Suggestion: change SuiteP font from Lato to Open Sans


okdotnz:
#### Issue

When switching the interface language to romanian, characters like ă, î, ș are rendered using Arial.
#### Expected Behavior

Latin extended characters to be rendered using the main font.
#### Actual Behavior

Latin extended characters are rendered using the (fallback) Arial font.
#### Possible Fix

Changing the font from Lato to Open Sans (or other font with latin extended support)
#### Steps to Reproduce

1.  Set the interface language to romanian after installing the language pack.

#### Your Environment

* SuiteCRM Version used: 7.11.4 
* Browser name and version: Firefox 60.6.1 ESR
* Environment name and version:  MySQL 5.7, PHP 7.2
* Operating System and version: Ubuntu 18.04

pgorod:
It seems there's a font called `LatoLatin` which has the missing characters.

BKPepe:
I have same the same issue in the Czech language, we have diacriticts. This is a little bit problematic, when some company has in their name "Česká republika" (translated to Czech Republic). 
![obrazek](https://user-images.githubusercontent.com/4096468/99146570-b258e280-2679-11eb-85d1-957384cc540f.png)

Here are steps to reproduce:
I used as name "+ěščřžýáíé=" while filling form, it was OK.
![obrazek](https://user-images.githubusercontent.com/4096468/99146536-53936900-2679-11eb-8870-70b880a1f386.png)
But I saved it and this is how it looks:

![obrazek](https://user-images.githubusercontent.com/4096468/99146544-660da280-2679-11eb-9a93-612e5263cd4e.png)

I'm using SuiteCRM version Version 7.11.18; Sugar Version 6.5.25 (Build 344)

As OP correctly pointed out, this is caused by using Lato, which does not support these characters.
This can be checked by using [Google Fonts](https://fonts.google.com/specimen/Lato?query=Lato).

BKPepe:
The issue still persists. Why this was marked as completed? Is there any reference that it fixes this PR?

pgorod:
@BKPepe I don't know the answer, this thing with the SuiteBot closing issues is new, but from what I've seen from other issues it has closed, it seems to be targeting issues labeled as "suggestion" that have been inactive for some time.

[...]

PR #8962, opened 2020-11-14
theme SuiteP: Use LatoLatin instead of Lato

BKPepe:
## Description
Some characters used in Czech like ěčř or Romanian language are not present in
Lato font and they are rendered in Arial, which does not look good.

There exists LatoLatin [1] from the same author, which includes glyphs
for European Latin-based languages.

[1] https://www.latofonts.com/lato-free-fonts/

## Motivation and Context

Before:
obrazek

After:
obrazek

obrazek

Fixes:  #7255 

## How To Test This
If you are using CloudFlare then you need to purge cache and if it does not work then try to disable DNS proxy.
But that's not all. You need to enable Developer mode in SuiteCRM, which worked in the end for me according to SuiteCRM's Discourse forum: https://community.suitecrm.com/t/editing-css-how-to-refresh-cache/33675/

## Types of changes

### Final checklist


BKPepe:
It would worth a shot to try the latest version of [nomanize.css](http://github.com/necolas/normalize.css), which is using correct indentation and it could bring more improvements, but this is not part of this pull request and someone could take a look at it later.

pgorod:
Initially I thought that LatoLatin was a larger font file, with _more_ characters in it, but from what I just read online, it's the opposite, it's supposed to be a subset of Lato. I find this confusing, if it is a subset, then why does it bring in characters that Lato failed to display?

I wonder if moving to LatoLatin creates other kinds of problems, namely with Cyrillic (@likhobory can you please check?)

likhobory:
It looks good.

изображение

BKPepe:
> I find this confusing, if it is a subset, then why does it bring in characters that Lato failed to display?

This could be caused that Google Fonts offers v1, which didn't contain Latin chars, but v2 contains them, but while looking at the size of each font it is incredibly large (almost two times bigger for each font) and someone (I do not) could be using SuiteCRM on phone and while using broadband connection every byte counts. That's why I jumped on LatoLatin, which was almost equal to font size, which is in the repository.

This is preview of Latin fonts, which I downloaded for website from https://www.latofonts.com/lato-free-fonts/, so the fonts are there.
obrazek

> Another thing that might make sense is to list both: 'Lato', Lato, LatoLatin, ...

As said above, potentially we can do it, but you don't want to ship two versions of fonts and each variant of it (bold, italic, and such). Still, I wonder, why we can not use some font, which is already presented in almost any system.

chris001:
Is `master` the right branch?  Or should it be to `hotfix`?

BKPepe:
You tell me, since there is a conflict, I need to solve the merge conflict
and rebase it.

Dne po 19. 2. 2024 18:14 uživatel Chris Coleman ***@***.***>
napsal:

chris001:
@BKPepe  The default branch for pull requests is `hotfix`... if you could rebase to `hotfix`, they should accept this easier.

PR #8965, opened 2020-11-16
Show popup help field in EditView.

peterkappelt:
When clicking on the "help popup" button in the EditView, the help text shall be shown. There is no reason for setting it's size to -1

## Description

When clicking on the small "i" button next to a field in the EditView, it's help description shall be shown in a popup window.

Also, consistent use of Smarty brackets is recommended.

## Motivation and Context

We want to show additional field explanations and guidelines to our users.

## How To Test This

Set the "popupHelp" property for a field of your choice. When opening the edit view for the entity which the field belongs to, there shall be a "i" button next to the field. When clicking on this i, a popup with the set description shall open.

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

### Final checklist

### DIFF
--- a/themes/SuiteP/include/EditView/tab_panel_content.tpl
+++ b/themes/SuiteP/include/EditView/tab_panel_content.tpl
@@ -50,15 +50,15 @@
                                 {{/if}}
                                 {{if isset($colData.field.popupHelp) || isset($fields[$colData.field.name]) && isset($fields[$colData.field.name].popupHelp) }}
                                     {{if isset($colData.field.popupHelp)}}
-                                        {{capture name="popupText" assign="popupText"}}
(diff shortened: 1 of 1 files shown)


pgorod:
Watch out for the double brackets

https://stackoverflow.com/questions/7455123/is-there-a-difference-between-using-a-single-vs-double-brace-delimiters-in-su

I didn't test, but does it work well with the single brackets around **capture**?

peterkappelt:
Yes, it is working. `capture` shall not be cached, thus the single brackets. 

Also, in the elseif-block, capture was used with single brackets in the upstream branch for at least 2 years. So doesn't make sense to not use it in the upper if-block, does it?

The caching that's active with double brackets even prevents the capture method to work properly.

PR #8996, opened 2020-12-15
Allow Onelogin override from sugar_config

tsmgeek:
- Add ability to define IDP entityId in config.
- Allow full override of onelogin settings from config for specific idp setup

This will make it easier for to implement more secure SAML setups without having to change default files which may be overwritten with updates.
ie. Not all admins wants to use the spBase as the entityId or they want to used signed saml etc.


tsmgeek:
This will now show a message in Administration that there is custom onelogin config if its defined in config.php and hide the manual input fields.

Dillon-Brown (approved):
Squashed, rebased, and did a minor code reformat. 

Rest of the code LGTM!

Issue #4435, opened 2017-10-21
<pagebreak /> disapear after saving PDF template


hubyhuby:
#### Issue
`< pagebreak />`  is a function of the PDF template and it doesn t work

#### Expected Behavior
`< pagebreak />` should do a page break

#### Actual Behavior
When I use `< pagebreak />`  in a PDF template document
When I edit it again , it has not been recorded.
Everything else seams to be recorded.

#### Possible Fix
Reverte to 7.8.7 where it was working ?

#### Steps to Reproduce
1)Edit a template
2)use `< pagebreak />`  and save (under google chrome OR firefox )
3)save
4)edit the template : page brake is not in the code of the page 
5) "print quote" you get a PDF without page breaks

#### Context
a page break

#### Your Environment
Ubuntu 16.04
apache 2.4
firefox and chrome
suitecrm 7.8.8

PS: sorry for the previous confusing bug the `< pagebreak />` are interpreted and deleted by github ! MAybe the same bug :)

pgorod:
There's a work around in the [forum thread](https://suitecrm.com/forum/installation-upgrade-help/16286-pdf-template-pagebreak-disapearing-after-upgrade-to-7-8-8#55085).

I don't know if it's suitable to be used as a definitive fix...

Jason-Dang:
Cannot reproduce in version 7.8.x, inserting page breaks in a PDF Template appears to work fine in both Chrome and Firefox.

hubyhuby:
Tested on 7.8.18; it now works perfectely

rgarcia-tcnx:
Tested on 7.10.14: <pagebreak /> in PDF Templates is not working. When you insert a pagebreak in a template, after do the save action, the <pagebreak /> label is removed.

More info: https://suitecrm.com/suitecrm/forum/installation-upgrade-help/16286-pdf-template-pagebreak-disapearing-after-upgrade-to-7-8-8

jack7anderson7:
Tested on 7.11.8 : When saved the Pagebreak then disappears after re-editing the record and doesn't take effect.

CrazyGoose:
Please fix this issue.

bunglex:
i consider this a high priority,  without pagebreaks ALL of our PDF's are poorly formatted, which means we cannot output from suitecrm directly to customers.

pgorod:
[...]

PR #9046, opened 2021-02-27
Maintain pagebreak tag

ildarius:
If tag is 'pagebreak' then do not run SugarCleaner on it

## Description
Solution for fixing this bug. The pagebreak tag is still unfortunately getting removed when the PDF template is generated.
More details: https://github.com/salesagility/SuiteCRM/issues/4435

## Motivation and Context
The pagebreak gets removed when a template get saved in PDF Templates module

## How To Test This
PDF Template > Create > Create a template and insert the pagebreak tag 

tag

## Types of changes

- [ X] Bug fix (non-breaking change which fixes an issue)

### Final checklist

### DIFF
--- a/data/SugarBean.php
+++ b/data/SugarBean.php
@@ -2452,8 +2452,10 @@ class SugarBean
                     $type .= $def['dbType'];
                 }
 
-                if (isset($def['type']) && ($def['type'] == 'html' || $def['type'] == 'longhtml')) {
-                    $this->$key = htmlentities(SugarCleaner::cleanHtml($this->$key, true));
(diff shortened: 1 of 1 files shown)


pgorod:
I agree with using this fix to resolve the problem, but I have to say this entire HTML cleaning is wrong, it shouldn't be done like this, and makes PDF Templates (and many other features of SuiteCRM) break.

HTML should be cleaned before output as HTML, not when going into the database.

ildarius:
closing as per above comments

pgorod:
Maybe I didn't explain well enough, but I wasn't suggesting this should be closed. 

I was just pointing out a larger, underlying issue. But I'm afraid that larger issue might not be addressed any time soon, so these smaller, specific fixes might still be useful. They solve user's problems.

PR #9050, opened 2021-03-01
Fixed #9049 - Remove the Prospsects and ProspectLists from the modInv…

### LINKED ISSUE #9049: Prospects and ProspectLists not available through the V8 modules API
#### Issue
I need to query a target (Prospect) by email as a response to a campaign, however, it seems there is no access to Prospects or ProspectList API, when I issue a {suite.url}/V8/meta/modules request neither Prospects (Targets) nor ProspectLists (Target lists) are returned in the available modules.

#### Expected Behavior
I should be able to interact with Prospects in the same way I can with Leads or Contacts.

#### Actual Behavior
Nothing as the modules are not available

#### Possible Fix
My first assumption was that the this was an ACL issue as the module appeared in the $moduleList, however, after step by step debugging through the whole process I found that that actual issue was the $modInvisList (also in globals) contained both Prospects and ProspectLists (and a number of other modules), which implies at some point it was decided not to include these, by removing both of these from the $modInvisList they became available and during my testing (mainly create, search and get) they seem to function exactly as Leads and Contacts.

I am raising this bug report so that I can provide a hot-fix for review but also (as I am very new to this product) to ensure that I have not missed something fundamental.

#### Steps to Reproduce
Any call to {suite.url}/V8/meta/modules.

[Please see this link for the full commentary on the issue](https://community.suitecrm.com/t/prospects-targets-not-available-when-performing-a-meta-module-list/78524)

#### Context
I am using my workaround until this is updated as I have an extremely complex target management process that runs across a number of different systems, so for me personally this was a V. HIGH requirement, but for the general use I guess this would be MEDIUM

#### Your Environment
* SuiteCRM Version used: 7.11.18 - Sugar Version 6.5.25 (Build 344)

shaung999:
…isList - make them visible - will result in them being available for a meta/modules call and subsequent module calls

## Description
Removed Prospects and ProspectLists from the module invisible list (making them visible) fixing bug 9049

## Motivation and Context
Ability to manage targets just like Leads and Contacts through the API

## How To Test This
After this change the Prospects and ProspectLists appear in the meta modules list and can be managed using the same module methods as used for Leads and Contacts

## Types of changes

### Final checklist

### DIFF
--- a/include/modules.php
+++ b/include/modules.php
@@ -266,8 +266,6 @@ $modInvisList = [
     'CampaignTrackers',
     'CampaignLog',
     'EmailMan',
-    'Prospects',
-    'ProspectLists',
     'Groups',
     'InboundEmail',
     'ACLActions',


jack7anderson7 (approved):

clemente-raposo (approved):
code looks ok

PR #9054, opened 2021-03-05
Add missing scheduler label for trimSugarFeeds

tsmgeek:
Scheduler label missing.


pgorod:
This is related to **Group F** of this larger issue: #4291

But since the entire group seems to be already there, the simple thing to do is exactly what this PR does, add the single label that is missing.

jack7anderson7:
Hi tsmgeek,

The label you are trying to add already exists within the file you're trying to add it to just under the name 'LBL_OOTB_SUITEFEEDS'.

tsmgeek:
@jack7anderson7  OOTB labels are only used when the schedulers are created by default to set the `Name` field, ie on a new install.
Im talking about when you go to create a scheduler and click the list of functions, in there it will use the other label.

https://github.com/salesagility/SuiteCRM/blob/e523df75ffaeaa593a1eebb4f7306b9e9c81759c/modules/Schedulers/EditView.php#L156

jack7anderson7 (approved):

mattlorimer (approved):

PR #9057, opened 2021-03-11
Composer PHP Version Constraint Changes

simonr44:
## Description
1) Remove the artificial php version constraint from composer.json
```
"platform": {
      "php": "5.5.9"
    },
```
2) Define the maximum php version supported.

## Motivation and Context
The platform php key seems to be used to define the minimum supported php version. This value artificially locks the php version in composer and will only allow packages to be installed which support this version. 

Locking the php version to 5.5.9 prevents package installs and updates that are version tested against the installed php version, limiting potential bug fixes and new features.

For example if SuiteCRM is installed on a host with PHP v7.3.15, it is not possible to install a package which requires >=5.6.0 even though the actual installed php version supports this.

The work around for this is to bypass the platform requirement with the ``` --ignore-platform-reqs ``` flag but doing so may have  unintended consequences.

I don't believe this was the intended outcome and propose we remove this and utilise the required key to set the minimum and maximum supported php versions. Doing so will allow composer to do its job and manage versions based on the actual installed software and their version constraints as defined in composer.json

Here is the composer documentation for reference:
https://getcomposer.org/doc/06-config.md#platform

**Please note**
Keen to hear other opinions, before we consider merging this. It has been tested on several new and existing installations with no sign of problems but I would appreciate the communities assistance in confirming this.

## How To Test This
Run the composer commands below on a new installation
```
composer update --lock
composer install
```

## Types of changes

### Final checklist
- [] My change requires a change to the documentation.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     "sort-packages": true
   },
   "require": {
-    "php": ">=5.5.9",
+    "php": "5.5.9 - 7.3",
     "ext-curl": "*",
     "ext-gd": "*",
     "ext-imap": "*",
--- a/composer.json
+++ b/composer.json
@@ -15,9 +15,6 @@
   },
   "config": {
     "vendor-dir": "vendor",
-    "platform": {
-      "php": "5.5.9"
-    },
     "optimize-autoloader": true,
     "sort-packages": true
   },
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "f03ead2eccf4a5773120321c0fb719d2",
+    "content-hash": "0fc509cf9f52903f92fdbf1b8dc37d49",
     "packages": [
         {
             "name": "consolidation/annotated-command",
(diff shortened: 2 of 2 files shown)


composer.json
```diff
@@ -15,14 +15,11 @@
...
     "vendor-dir": "vendor",
-    "platform": {
-      "php": "5.5.9"
-    },
     "optimize-autoloader": true,
     "sort-packages": true
   },
   "require": {
-    "php": ">=5.5.9",
+    "php": "5.5.9 - 7.3",
```
Dillon-Brown, line 22: I think this would have some BC concerns. It's possible that some people may be running PHP 7.4, 8, etc, even though we don't officially support it yet. Aside from that, removing the platform reqs looks like a good idea.

simonr44: I had the same concerns. The platform constraint was the main point of this MR and i'm more than happy to remove this part to get it over the line.

I considered it worthwhile to define the upper limit as an aid to new installs.
If people are running >=7.4 then they hopefully have the knowledge to edit composer manually to overcome this. If they don't, then directing to 7.3 may be the best thing for everyone.

pgorod: I guess that the upper limit might be a good thing. If people are using 7.4, they have a real problem, many things will break, not just this. And I guess that if they're messing with **composer** they can tweak that value if they really need to. I'm not sure about this, "I'm just sayin..."

PR #9084, opened 2021-04-07
Set default perms on new log file

tsmgeek:
Depending on the umask default for your system new files can be created with permissions that do not allow log files to be written to if they were created from a cron vs UI action.

ie. if cron runs and creates a log file I get `0644 / root / www-data` as my crons are running within a custom docker container which is root, but the UI will create them as `0644 / www-data / www-data`. The issue comes if the UI tries to write to log file originally created by the cron due to `0644` perms it will not write it but with `0664` it will.

Just set `0664` at the point the log file is created.

Also there is no need to do `file_exists` and `is_writable` when you are checking if the log file exists, a total of three checks are done and now ive reduced it to one, if its not writable it then checks if it exists and creates a new file, tests again its writable (no reason it should not be).

### DIFF
--- a/include/SugarLogger/SugarLogger.php
+++ b/include/SugarLogger/SugarLogger.php
@@ -68,6 +68,7 @@ class SugarLogger implements LoggerTemplate
     protected $filesuffix = "";
     protected $date_suffix = "";
     protected $log_dir = '.';
+    protected $defaultPerms = 0664;
 
 
     /**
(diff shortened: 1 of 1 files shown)


Dillon-Brown (dismissed):
Rebased and did a small code cleanup.

LGTM!

Dillon-Brown (approved):

jack7anderson7 (approved):

PR #9101, opened 2021-04-20
Fix LangText exception breaking ElasticSearch

Dillon-Brown:
## Description

- Fixed an issue where LangText was throwing an exception if there were any missing keys. As ElasticSearch is indexing potentially every module, there could be many missing keys that we need to log but not die on so the user can continue to see ElasticSearch results.

## How To Test This

1. Index ElasticSearch

## Types of changes

### Final checklist

### DIFF
--- a/lib/Search/UI/SearchResultsController.php
+++ b/lib/Search/UI/SearchResultsController.php
@@ -88,7 +88,7 @@ class SearchResultsController extends Controller
     public function display()
     {
         $headers = $this->getListViewHeaders();
-        
+
         $total = $this->results->getTotal();
         if ($total > 1) {
             $size = $this->query->getSize();
@@ -116,7 +116,7 @@ class SearchResultsController extends Controller
             }
         }
         $this->view->getTemplate()->assign('total', $total);
-        
+
         $this->view->getTemplate()->assign('headers', $headers);
         $this->view->getTemplate()->assign('results', $this->results);
         $this->view->getTemplate()->assign('resultsAsBean', $this->results->getHitsAsBeans());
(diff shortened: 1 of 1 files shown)


lib/Search/UI/SearchResultsController.php
```diff
@@ -202,7 +202,7 @@ protected function getListViewHeaderLabel(SugarBean $bean, $fieldValue, $fieldDe
                 $fieldValue['label'],
                 null,
                 LangText::USING_ALL_STRINGS,
-                true,
+                false,
                 true,
```
clemente-raposo, line 206: @Dillon-Brown was looking into the LangText code. if we set log as `false`, throw will never happen. That said maybe best to also set `throw` as false here, to avoid having people confused when looking at this code in the future.

Dillon-Brown: Ah, nice spot. I actually intended to set throw as false and keep logging. Pushed.

clemente-raposo: yup, I though of that also, but if we set the log to true, the text will be null (or similar). However, when you set the log to false, it sets the text to be the same as the key.

Should it be empty or have the key?

```php
        if (!$text) {
            if ($this->log) {
                ErrorMessage::handler('A language key does not found: [' . $this->key . ']', self::LOG_LEVEL, $this->throw);
            } else {
                $text = $this->key;
            }
        }
```

Dillon-Brown: Should be empty, wouldn't want to display invalid keys as text, better to just log it and move on imo.

clemente-raposo: agreed!

clemente-raposo (approved):
code looks ok

jack7anderson7 (approved):

Issue #6932, opened 2019-02-12
7.11.1: Newer version of PHPMailer is not compatible with Email:email2Send method


dawid-zaroda:
SuiteCRM 7.11 comes with newer version of PHPMailer. Method **addAttachment** in PHPMailer object got new condition: **isPermittedPath** which blocks all pathes with ://

Email:emal2Send for $request['documents'] and $request['templateAttachments'] uses pathes:
`$fileLocation = "upload://{$GUID}";`

which will be blocked by isPermittedPath.

`SugarPHPMailer encountered an error: Could not access file: upload://c2a5ebd1-5a6d-0f2f-eafe-5c6286a20e74`
#### Your Environment
* SuiteCRM Version used: 7.11.1
* Environment name and version (e.g. MySQL, PHP 7): PHP 7.1
* Operating System and version (e.g Ubuntu 16.04): Ubuntu 16.04.4 LTS

Mausino:
Is your issue related to #6423 or duplicate of it ?

dawid-zaroda:
I think no, because there: https://github.com/salesagility/SuiteCRM/blob/v7.8.22/include/phpmailer/class.phpmailer.php
is still older version of PHPMailer, which works fine with issue reported by me. This file is not present in 7.11.1

Dillon-Brown:
Thought I had fixed this with #6791 but I might have missed a few places.

Mausino:
I did those changes #7154, did R&R in admin and still if the email template has attachment and is used in email... the attachment isn't added to sent email. Also in compose view isn't visible that attachment was added from Email Template

PR #6938, opened 2019-02-13
Fixed #6932 - 7.11.1: Newer version of PHPMailer is not compatible with Email:email2Send method

Dillon-Brown:
## Description

```
`SugarPHPMailer encountered an error: Could not access file: upload://c2a5ebd1-5a6d-0f2f-eafe-5c6286a20e74`
```

## Motivation and Context

Issue reference: #6932

## How To Test This

1. Send an email with an attachment.
2. Send an email with a document attachment.
3. Sent an email with template attachments.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [X] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change to the documentation.
- [X] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

PR #9123, opened 2021-05-11
Upgrade PHPMailer to 6.4.x

Dillon-Brown:
## Description

- Moved PHPMailer to composer.
- Upgrade PHPMailer to 6.4.x.
- Removed old phpmailer references and kept classes for BC.

## How To Test This

1. Regression test emails.

## Types of changes

### Final checklist


modules/Emails/Email.php
```diff
@@ -1278,7 +1278,7 @@ public function email2Send($request)
                     if (!empty($note->id)) {
                         $filename = $note->filename;
                         $noteGUID = preg_replace('/[^a-z0-9\-]/', "", $note->id);
-                        $fileLocation = "upload://{$noteGUID}";
+                        $fileLocation = "upload/{$noteGUID}";
```
clemente-raposo, line 1281: Hi @Dillon-Brown,

Why do we need this change?

Dillon-Brown: It's for compatibility with PHPMailer 6.4.x. I've pulled in the change from: https://github.com/salesagility/SuiteCRM/pull/6938. Which was to fix https://github.com/salesagility/SuiteCRM/issues/6932 when we upgraded to 6.4.x in 7.11.x.

clemente-raposo: ok, thanks Dillon

clemente-raposo (approved):
Changes look ok.

Mac-Rae (approved):

Mac-Rae:
Assessed :+1:

pgorod:
@Dillon-Brown @clemente-raposo I think we had issues a similar change before, though I don't fully remember the details. There is a configuration variable called [upload_dir](https://github.com/salesagility/SuiteCRM/search?q=upload_dir&type=) which lets people change the directory for uploads. 

Internally, I think that is implemented with something called "streams". Changing these **`upload://`** paths to just **`upload/`** is basically hard-coding the path again. In some places, not all, which can get a bit messy...

Dillon-Brown:
> There is a configuration variable called upload_dir which lets people change the directory for uploads.

That's actually a really good point, I don't see how this would work if you had that configured unless it just doesn't apply to these files and they always go into a hardcoded paths. It's definitely something we'll need to double-check but as these changes have been in 7.11.x for quite a while I don't think it would be too high of a priority.

pgorod:
I confess I have some trouble getting my head around these PHP streams concept, in the way that it is used for the uploads. But apparently it is registered with [stream_wrapper_register](https://www.php.net/manual/en/function.stream-wrapper-register.php) from here:

https://github.com/salesagility/SuiteCRM/blob/master/include/entryPoint.php#L141-L142
which calls 
https://github.com/salesagility/SuiteCRM/blob/master/include/UploadStream.php#L148

From that point on, I believe the expression **`upload://`** will refer our own "protocol" and invoke our [UploadStream class](https://github.com/salesagility/SuiteCRM/blob/master/include/UploadStream.php), which I suppose will use whatever directory is in `$GLOBALS['sugar_config']['upload_dir']`.

So the sneaky part is that when we're calling regular PHP file functions, that `UploadStream` class is actually intervening, if the filename happens to use the `upload::/` protocol.

Honestly, I think we should just leave this as it was. Exactly which compatibility problem were you getting with PHP Mailer?

Another way to remove the protocol but keep the configurability is to use some construct like this:
`$location = $GLOBALS['sugar_config']['upload_dir']."/" . $filename;`

We actually find constructs like this all over the code, which happens whenever the developer wasn't understanding how to actually use the Streams... I've done this mistake myself in my code...

Dillon-Brown:
> Honestly, I think we should just leave this as it was. Exactly which compatibility problem were you getting with PHP Mailer?

@pgorod So really all I was doing here was backporting the PHPMailer 6.4.x upgrade from 7.11.x to 7.10.x. The extra commit which changes the ```upload://``` I pulled in from here: https://github.com/salesagility/SuiteCRM/pull/6938. Which was a fix intended to fix an issue raised when we did the 6.4.x upgrade on 7.11.x here: https://github.com/salesagility/SuiteCRM/issues/6932.

pgorod:
Ok
so I guess the correct fix for this would be to resolve to a real path before sending it into the PHPMailer package.

There's a method in **UploadStream.php** which appears to do just this:

```php
    /**
     * Get real FS path of the upload stream file
     * @param string $path Upload stream path (with upload://)
     * @return string FS path
     */
    public static function path($path)
    {
        $path = substr($path, strlen(self::STREAM_NAME) + 3); // cut off upload://
        $path = str_replace("\\", "/", $path); // canonicalize path
        if ($path == ".." || substr($path, 0, 3) == "../" || substr($path, -3, 3) == "/.." || strstr($path, "/../")) {
            return null;
        }

        return self::getDir() . "/" . $path;
    }

```

This would leave the configurable `upload_dir` system intact.

PR #9140, opened 2021-05-21
Sales stage tracking logic hook

### LINKED ISSUE #9139: Provide Skeleton Logic Hook: Skeleton template to add sales pipeline date and duration tracking template
There are a number of uses for tracking the amount of time an opportunity spends in each sales stage. This information can be used in reports, or to determine sales cycle times and also useful conversion information. This seems to be a feature request that comes up a couple of times.

This pull request provides a (commented out) logic hook that can be easily modified to support this kind of reporting. It also provides a starting point for implementing custom logic.

https://community.suitecrm.com/t/sales-workflow/37301

#### Issue

There are two issues here; there isn't any skeleton or example code for how to implement a custom logic hook. And, there isn't an easy way to keep track of how the amount of time an opportunity spends in each sales stage.

This issue resolves both, by providing a commented example of a logic hook, which users can subsequently modify to meet their own needs.

#### Expected Behavior

Ideally, there should exist a sample template or dummy code that people can use when first learning about logic hooks. This would provide a better entry and allow people to test and develop their own logic hook implementations.

#### Actual Behavior

There aren't really any examples of functioning logic hooks.

#### Possible Fix

Submitting a PR that provides this example code, along with reasonably inline comments.

#### Context

We spent around 3 weeks trying to implement this feature for our own uses. During this time, we found that this was a reasonably common use case on the forums, but that concrete and detailed instructions for exactly how to implement a logic hook weren't readily available.

Instead, we had to piece it together from a number of different sources.

This commit, provides an example of a (commented out) functioning logic hook, albeit one that would need to be modified to support the default sales stages. It needs to be modified to support the default sales stages, and also requires some fields to be added, but otherwise works.

vmw:
This PR closes #9139.

It provides a template logic hook, with extensive comments, which allows new users to have an example which they can use to help build their own logic hooks.

While it's not particularly extensive, it does go a long way to providing a concrete example of a functioning logic hook which can later be extended or modified as required. We tried to use the default sales stages, so that everyone would have a common starting point, but there are a couple of fields that need to be added.

### DIFF
--- a/custom/modules/SetOppDates.php
+++ b/custom/modules/SetOppDates.php
@@ -1,0 +1,109 @@
+<?php
+
+/**
+* This update is used to track the amount of time an opportunity spends in each sales stage. 
+* This information can be used in reports, or to determine sales cycle times and also useful conversion information. 
+* The file is called with a before_save module hook from the logic hooks file in the Opportunities module folder.
+* 
+* Prerequisites:
+* - create an integer field for each sales stage to keep track of the amount of time in that stage. (ie. time_in_qualification_c)
+* - create a DateTime field called sales_stage_date_set_c to keeep track of the last time the sales stage time informatio  was set.
+* - create a DateTime field for each sales stage to keep track of the last time the sales stage was set to that particular stage. (datetime_qualification_set_c) 
+* - create a DropDown field called last_sales_stage_c to keep track of the last sales stage the opp was in. 
+
+* Note that time in each stage is stored in minutes. Feel free to change for your purposes.
+*/
+                class Calc {
+                function setTime($bean, $event, $arguments) {
+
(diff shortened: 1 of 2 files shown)


mattlorimer:
Hi @vmw @gwendo12 

Thanks for your PR, It not the normal kind we receive and not something we would usually include in the core Codebase. It would be good to understand where you see it going?

From our perspective it may be better to include in the documentation site https://docs.suitecrm.com/ https://github.com/salesagility/SuiteDocs, maybe with a [technical blog](https://docs.suitecrm.com/blog/) post of what you have in the issue. We think this would be really useful addition to the developer guide, an ideal way for you to get credit for your work and to share it with the wide community.

Let us know your thoughts

Regards
Matt

PR #9159, opened 2021-05-31
Implement standard SearchEngines

Dillon-Brown:
## Description

- Implemented BasicSearchEngine
- Implemented LuceneSearchEngine
- Standardised SearchEngine output
- Many miscellaneous bug fixes and improvements.

## Motivation and Context

- All SearchEngines will now output the same results and come from a common core.
- I've left most of the original code for the Basic and AOD searching in ```modules/home```, there are perhaps 4-5 classes/files that can be totally removed from here but I've left them in for some degree of BC.
- In the interest of clean design I've separated ```UnifiedSearchAdvanced``` into two separate engines. Users will still be able to select the usual ```BasicAndAodEngine``` options and all current settings are still respected. How this works for a user using ```BasicAndAodEngine``` is that the SearchQuery will determine a likely default based on if the user has ```AOD``` enabled along with looking at the request data for older sugar searches. The front-end will then also display the old functionality of being able to switch between ```AOD``` and ```Basic``` based on what the users default has been set to.

All search pages (Elastic/AOD/Basic) now look like the following:

SuiteSearch

With multiple engines available, you will now be able to select directly on the results page:

resultsEngine

 
## How To Test This

1. Test ElasticSearch
2. Test BasicSearch
3. Test LuceneSearch

## Types of changes

### Final checklist


lib/Search/AOD/LuceneSearchEngine.php
```diff
@@ -0,0 +1,178 @@
...
+     *
+     * @param string $queryString
+     * @return array
+     */
+    private function runLucene(string $queryString): array
+    {
+        $cachePath = 'cache/modules/AOD_Index/QueryCache/' . md5($queryString);
+        if (is_file($cachePath)) {
+            $mTime = filemtime($cachePath);
+            if ($mTime > (time() - 5 * 60)) {
```
clemente-raposo, line 104: Hi @Dillon-Brown, does this mean we are caching query results for a given amount of time?

lib/Search/AOD/LuceneSearchEngine.php
```diff
@@ -0,0 +1,178 @@
...
+
+    /**
+     * @param mixed $hits
+     * @return array
+     */
+    private function parseHits(array $hits): array
+    {
+        $searchResults = [];
+
+        foreach ($hits as $i => $array) {
```
clemente-raposo, line 157: Hi @Dillon-Brown, minor suggestion could we rename `$array` to `$hit`, makes it easier to understand, and with `$array` one may get confused with the reserved word. Anyway, just a sugestion

lib/Search/AOD/LuceneSearchEngine.php
```diff
@@ -0,0 +1,178 @@
...
+            $recordModule = $array->record_module;
+            $searchResults[$recordModule][$i] = $array->record_id;
```
clemente-raposo, line 159: Hi @Dillon-Brown, does the index in `$searchResults[$recordModule][$i]` matter? because using `$i` for the index may lead to gaps/jumps in the indexes. If we want incremental indexes without jumps we could just do`$searchResults[$recordModule][]  = $array->record_id;`

lib/Search/BasicSearch/BasicSearchEngine.php
```diff
@@ -0,0 +1,526 @@
...
+     */
+    protected function getSearchModules(): array
+    {
+        $unifiedSearchModuleDisplay = $this->getUnifiedSearchModulesDisplay();
+
+        require_once 'include/ListView/ListViewSmarty.php';
+
+        global $beanList, $current_user;
+
+        if (!empty($_REQUEST['advanced']) && $_REQUEST['advanced'] !== 'false') {
```
clemente-raposo, line 112: Hi @Dillon-Brown 

I imagine this was here before, though some questions:

- Is the code expecting to receive things from `$_REQUEST`?
- The engine is one of the possible implementations right? it can be this one or lucene or other?
- If we have a generic engine that can have several implementation, we probably should receive everything to the parameters. So that all engines receive it.

what do you think?

lib/Search/BasicSearch/BasicSearchEngine.php
```diff
@@ -0,0 +1,526 @@
...
+        if (!empty($searchQuery)) {
+            foreach ($modulesToSearch as $moduleName => $beanName) {
+                require_once $beanFiles[$beanName];
+                $seed = new $beanName();
+
+                $listViewData = new ListViewData();
+
+                // Retrieve the original list view defs and store for processing in case of custom layout changes
+                require(__DIR__ . '/../../../modules/' . $seed->module_dir . '/metadata/listviewdefs.php');
+                $orig_listViewDefs = $listViewDefs;
```
clemente-raposo, line 175: Hi @Dillon-Brown probably this was here before, but we could we rename `$orig_listViewDefs` to keep with the standard, something like: `$originalListViewDefs` or similar?

lib/Search/BasicSearch/BasicSearchEngine.php
```diff
@@ -0,0 +1,526 @@
...
+                    if (isset($def['db_field'])) {
+                        foreach ($def['db_field'] as $dbfield) {
+                            $where_clauses[] = $dbfield . " LIKE '" . DBManagerFactory::getInstance()->quote($searchQuery) . "%'";
+                        }
+                        $params['custom_select'] .= ", $dbfield";
+                        $params['distinct'] = true;
+                    }
+                }
+
+                if (!empty($where_clauses)) {
```
clemente-raposo, line 245: Hi @Dillon-Brown could we also rename `$where_clauses` to `$whereClauses` or similar please?

lib/Search/BasicSearch/BasicSearchEngine.php
```diff
@@ -0,0 +1,526 @@
...
+                    $where = '((' . implode(' ) OR ( ', $where_clauses) . '))';
+                } else {
+                    $where = '';
+                }
+
+                $listData = $listViewData->getListViewData($seed, $where, 0, -1, [], $params);
+
+                $moduleCounts[$moduleName] = $listData['pageData']['offsets']['total'];
+
+                foreach ($listData['data'] as $i => $array) {
```
clemente-raposo, line 255: Hi @Dillon-Brown if could also rename `$array` to `$hit` or `$entry` or `$result`, please?

lib/Search/BasicSearch/BasicSearchEngine.php
```diff
@@ -0,0 +1,526 @@
...
+            'modules' => $moduleResults
+        ];
+    }
+
+    /** @noinspection PhpIncludeInspection */
+    public function buildCache(): void
+    {
+        global $beanList, $beanFiles, $dictionary;
+
+        $supported_modules = [];
```
clemente-raposo, line 272: Hi @Dillon-Brown could we rename `$supported_modules` to `$supportedModules` to keep with standards please?

lib/Search/BasicSearch/BasicSearchEngine.php
```diff
@@ -0,0 +1,526 @@
...
+                }
+
+                foreach ($searchFields[$moduleName] as $field => $def) {
+                    if (!empty($def['force_unifiedsearch'])) {
+                        $fields[$field] = $def;
+                    }
+                }
+
+                if (!empty($fields)) {
+                    $supported_modules [$moduleName] ['fields'] = $fields;
```
clemente-raposo, line 339: Hi @Dillon-Brown minor formatting change needed on `$supported_modules [$moduleName] ['fields'] `.

lib/Search/AOD/LuceneSearchEngine.php
Dillon-Brown: Yes, I've kept this functionality the same as before, only real change here is that I'm calling ```filemtime``` directly instead of using a method to calculate DST manually (Due to a PHP bug a decade ago).

lib/Search/AOD/LuceneSearchEngine.php
Dillon-Brown: Agreed, pushed.

lib/Search/AOD/LuceneSearchEngine.php
Dillon-Brown: Nope it doesn't matter, good point. Pushed.

Dillon-Brown: Nope it doesn't matter, good point. Pushed.

lib/Search/BasicSearch/BasicSearchEngine.php
Dillon-Brown: Removed, we won't be using request data here, this engine will always "BasicSearch" so it doesn't need to check for advanced.

lib/Search/BasicSearch/BasicSearchEngine.php
Dillon-Brown: Pushed

lib/Search/BasicSearch/BasicSearchEngine.php
Dillon-Brown: Pushed

lib/Search/BasicSearch/BasicSearchEngine.php
Dillon-Brown: Pushed

lib/Search/BasicSearch/BasicSearchEngine.php
Dillon-Brown: Pushed, strange my IDE doesn't want to fix these though 🤔

lib/Search/AOD/LuceneSearchEngine.php
clemente-raposo: Hi @Dillon-Brown, thank you. I maybe wrong, though Ideally we should cache the query results? we may get results that are out of date right?

lib/Search/AOD/LuceneSearchEngine.php
clemente-raposo: thank you

lib/Search/AOD/LuceneSearchEngine.php
clemente-raposo: thank you

lib/Search/BasicSearch/BasicSearchEngine.php
clemente-raposo: thank you

lib/Search/BasicSearch/BasicSearchEngine.php
clemente-raposo: thank you

lib/Search/UI/templates/search.form.tpl
```diff
@@ -44,54 +50,46 @@
...
     <script>
         {literal}
-            var searchForm = {
-                onSubmitClick: function(e) {
-                    // jump to the first page on new results list
-                    $('input[name="search-query-from"]').val(0);
-                    return true;
-                }
-            };
+        let searchForm = {
```
clemente-raposo, line 86: Hi @Dillon-Brown I'm not sure if its safe already to use `let` without a transpiler. Though it seems to be supported by most browsers in the latest versions:
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let

lib/Search/SqlSearch/SimpleSqlSearchEngine.php
```diff
@@ -83,16 +89,16 @@ public function search(SearchQuery $query)
...
         $filteredFields = [];
 
         foreach ($fields as $name => $type) {
-            if (strpos($type, 'varchar') == 0) {
+            if (str_starts_with($type, 'varchar')) {
                 $filteredFields[$name] = 'varchar';
             }
 
-            if (strpos($type, 'text') == 0) {
+            if (str_starts_with($type, 'text')) {
```
clemente-raposo, line 101: Hi @Dillon-Brown I think `str_starts_with` its only supported in php8

lib/Search/SearchQuery.php
```diff
@@ -154,39 +199,67 @@ private static function filterArray(array $array, $key, $default, $filter = null
...
+    /**
+     * Get the default engine by checking the config
+     *
+     * @return null|string
+     */
+    public function getDefaultEngine(): ?string
+    {
+        global $sugar_config;
+
+        if (!empty($sugar_config['search']['defaultEngine'])) {
```
clemente-raposo, line 230: Hi @Dillon-Brown, what should this method return if `$sugar_config['search']['defaultEngine']` is empty

lib/Search/AOD/LuceneSearchEngine.php
clemente-raposo: @Dillon-Brown as you mentioned this is keeping existing behavior. This is something we can go over on a separate conversation

lib/Search/UI/templates/search.form.tpl
Dillon-Brown: Changed back to var, will need to look into this further at some point as my IDE wants to auto-replaces these to either ```let``` or ```const```.

lib/Search/SqlSearch/SimpleSqlSearchEngine.php
Dillon-Brown: We have polyfills for these, I'm going to revert anyway though as I believe the last Suite-7 release we had some issues with the polyfills when it came to running the upgrade wizard. Nice spot.

lib/Search/SearchQuery.php
Dillon-Brown: Changed to return ```BasicSearchEngine``` when the config is not set. I think that would make the most sense since it's the only engine that doesn't require any setup and works OOTB.

Dillon-Brown:
@clemente-raposo Pushed two additional changes.

1. Deprecated ```UnifiedSearchAdvanced``` as none of this code is now used (Kept for BC but would prefer to remove completely, thoughts?)
2. Removed PHP 7.2 from Travis config and replaced it with PHP 7.3. Also pointed the codecoverage to 7.3 for performance reasons.

Dillon-Brown:
@clemente-raposo Latest changes include a new way to select engines from the results page. This also works for any custom engines the user may have added:

extendedEngines

clemente-raposo (dismissed):
code looks ok

Dillon-Brown:
Latest changes are just rebasing onto latest develop, all conflicts due to https://github.com/salesagility/SuiteCRM/pull/9170 which is just PHPUnit changes.

Mac-Rae (approved):

clemente-raposo (approved):
latest changes look ok

PR #9095, opened 2021-04-15
Consolidate global search settings (AOD, Basic)

Dillon-Brown:
## Description

- Removed a duplicate "Global Search" link in the admin panel.
- Moved AOD to "Search Settings" page.
- Moved SearchModules to "Search Settings" page. 
- Removed AOD enable option, this now comes from the search engine drop-down.
- Minor coding standard fixes, code simplification, cleanup, etc.
- Minor styling fixes, each of the search settings should be using standard panels:
- Cleaned up flaky ElasticSearch tests (These should be improved further / overhauled if we upgrade ElasticSearch versions).

searchsettings2

## Motivation and Context

- This is to consolidate and simplify the search options within the CRM. All settings should work exactly the same as before just now in one location.

## How To Test This

1. Check AOD works as expected.
3. Check basic search works as expected.
4. Check module selection works as expected.
5. Check that all settings visible on the page work and that nothing is missing.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [X] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [X] My change requires a change to the documentation.
- [X] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

PR #9171, opened 2021-06-07
Upgrade ElasticSearch to 7.x

Dillon-Brown:
## Description

- Follow on from https://github.com/salesagility/SuiteCRM/pull/9159 -> https://github.com/salesagility/SuiteCRM/pull/9095 ->https://github.com/salesagility/SuiteCRM/pull/9171.
- Upgrades ElasticSearch to 7.x.

## Mapping types
As mapping types have now been removed from ElasticSearch, I have instead implemented an Index per document type design. Previously we were using a single index ```private $index = 'main';``` along with an accompanying ```defaultParams.yml``` file which specified the mapping.

E.g. Instead of storing accounts and contacts in a single ```main``` index, we now store accounts in the accounts index and contacts in the contacts index. Indices are completely independent of each other and so there will be no conflict of field types between indices.

A few benefits of this design:

- Data is more likely to be dense and so benefit from compression techniques used in Lucene.
- The term statistics used for scoring in full text search are more likely to be accurate because all documents in the same index represent a single entity.
- Each index can be sized appropriately for the number of documents it will contain: you can use a smaller number of primary shards for accounts and a larger number of primary shards for contacts.

## Upgrading from ElasticSearch 5.x.
- As both mapping types and the index names have changed, users will be required to run a full re-index before being able to use the new 7.x Engine.
- We will also need to either document for users or handle on upgrade the ```cache/ElasticSearchIndexer.lock``` file. If attempting to run a full re-index from a freshly upgraded instance that still contains the old cache, the system will instead launch a differential index (By design), normally this would be fine but as the user won't have any of the new indexes, elasticsearch will fail.

## How To Test This

1. Test ElasticSearch functionality using a 7.x ElasticSearch server.

## Types of changes

### Final checklist


lib/Search/ElasticSearch/ElasticSearchEngine.php
```diff
@@ -88,7 +82,7 @@ public function search(SearchQuery $query): SearchResults
         $end = microtime(true);
         $searchTime = ($end - $start);
 
-        return new SearchResults($results, true, $searchTime, $hits['hits']['total']);
+        return new SearchResults($results, true, $searchTime, $hits['hits']['total']['value']);
```
clemente-raposo, line 85: Hi @Dillon-Brown is this related with the index change? does seem so? Has the structure changed on elastic search?

lib/Search/ElasticSearch/ElasticSearchIndexer.php
```diff
@@ -454,28 +464,11 @@ private function indexBatch($module, array $beans)
...
     {
         return $this->differentialIndexing && $this->lastRunTimestamp !== false;
     }
 
-    /**
-     * Retrieves the default params to set up an optimised default index for Elasticsearch.
-     *
-     * @return array
-     */
-    private function getDefaultMapParams()
```
clemente-raposo, line 467: Hi @Dillon-Brown, the reason for no longer needing the map params is related with elastic search 7?

clemente-raposo (approved):
Code looks ok. 

Added some comments to have a better understanding of the changes

lib/Search/ElasticSearch/ElasticSearchEngine.php
Dillon-Brown: Yes structure has changed on ElasticSearch, I've also changed what comes back from ```parseHits``` as it now needs to include ```module_source``` in it's array so that we can determine the index.

lib/Search/ElasticSearch/ElasticSearchIndexer.php
Dillon-Brown: Yup exactly, ElasticSearch will now reject anything that comes with mappings like this.

Mac-Rae (dismissed):

Dillon-Brown:
Latest push is rebasing onto develop and a minor change to the ```composer.json```:

- Removed ```"elasticsearch/elasticsearch": "dist"``` and ```"rm -Rf vendor/elasticsearch/elasticsearch/tests/Elasticsearch/Tests"``` as the tests directory is no longer included.

clemente-raposo (approved):
latest changes look ok

mattlorimer:
Build passing --> https://travis-ci.com/github/mattlorimer/SuiteCRM/builds/232272625

clemente-raposo (approved):
@Dillon-Brown latest changes look ok

Mac-Rae:
Assessed :+1:

Mac-Rae (approved):

PR #9172, opened 2021-06-07
Malicious File Scanning

simonr44:
Adds optional malicious file scanning capabilities to file uploads.

## Description
This feature gives the option to configure and enable malicious file scanning using third party software. So far we have support for ClamAV over TCP and Sophos locally.

I picked up from [this](https://github.com/salesagility/SuiteCRM/pull/5381) PR. I built upon that concept to support a multi-server or container based environment with Clam AV TCP Scanning. Much of the credit for the original design should go there.

As per the original design, developers can create providers which implement the AntiMalwareFileScanner interface and throw a MalwareFound exception. This will make it really easy to support many command line based antivirus programs.

## Motivation and Context
There was no existing malicious file scanning integrated with the application.

## How To Test This
Sophos
This can be installed locally on your server and configured in your config.php or config_override.php. The default values for path and options should be compatible with most linux systems?

ClamAV
To enable ClamAV you must install and configure clamd daemon to listen on a tcp socket. Please see the [ClamAV documentation](https://www.clamav.net/documents/scanning) for details on how to complete this step. This can be locally on the same machine or remotely on any server or container, provided a connection can be established with the ClamAV Daemon on the configured tcp port (default is 3310). The path value should be null but the

Regardless of the scanner configured, this has been written to fail safe, so if a scanner is enabled in the config and a good scan cannot be confirmed, the file upload will fail.

Once a scanner is configured and enabled, its simply a matter of uploading a test file which will trigger the scan to fail. Commonly the [EICAR test file](https://en.wikipedia.org/wiki/EICAR_test_file) is used for this purpose.

## Types of changes

### Final checklist
- [] My change requires a change to the documentation.

### DIFF
--- a/include/utils.php
+++ b/include/utils.php
@@ -108,6 +108,26 @@ function make_sugar_config(&$sugar_config)
 
     $sugar_config = array(
         'admin_export_only' => empty($admin_export_only) ? false : $admin_export_only,
+        'anti_malware_scanners' => array(
+            'SuiteCRM\Utility\AntiMalware\Providers\ClamTCP' => array(
+                'name' => 'ClamAV TCP',
+                'support_page' => 'https://www.clamav.net/',
+                'enabled' => false,
+                'path' => null,
+                'options' => array(
+                    'ip' => '127.0.0.1',
+                    'port' => 3310,
+                    'type' => 'local'
+                ),
+            ),
+            'SuiteCRM\Utility\AntiMalware\Providers\Sophos' => array(
+                'name' => 'Sophos Anti Virus (Linux)',
+                'support_page' => 'https://www.sophos.com/en-us/products/free-tools/sophos-antivirus-for-linux.aspx',
+                'enabled' => false,
+                'path' => '/opt/sophos-av/bin/savscan',
+                'options' => '-ss'
+            )
+        ),
         'export_delimiter' => empty($export_delimiter) ? ',' : $export_delimiter,
         'cache_dir' => empty($cache_dir) ? 'cache/' : $cache_dir,
         'calculate_response_time' => empty($calculate_response_time) ? true : $calculate_response_time,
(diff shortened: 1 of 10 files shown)


include/utils.php
```diff
@@ -109,6 +109,26 @@ function make_sugar_config(&$sugar_config)
...
+                ),
+            ),
+            'SuiteCRM\Utility\AntiMalware\Providers\Sophos' => array(
+                'name' => 'Sophos Anti Virus (Linux)',
+                'support_page' => 'https://www.sophos.com/en-us/products/free-tools/sophos-antivirus-for-linux.aspx',
+                'enabled' => false,
+                'path' => '/opt/sophos-av/bin/savscan',
+                'options' => '-ss'
+            )
+        ),
```
Dillon-Brown, line 131: I believe this can be removed. From what I can see this ```$sugar_config``` array is intended to populate values during setup, as we aren't actually populating anything here we only need to use the ```$sugar_config_defaults``` below.

include/utils.php
```diff
@@ -279,6 +299,26 @@ function get_sugar_config_defaults(): array
 
     $sugar_config_defaults = [
         'admin_export_only' => false,
+        'anti_malware_scanners' => array(
```
Dillon-Brown, line 302: Short array syntax is preferred :).

```suggestion
        'anti_malware_scanners' => [
```

index.php
```diff
@@ -38,6 +38,8 @@
  * display the words "Powered by SugarCRM" and "Supercharged by SuiteCRM".
  */
 
+
+
```
Dillon-Brown, line 42: Can this index.php change be reverted? @jack7anderson7

tests/_data/AntiMalwareTestFile.txt
```diff
@@ -0,0 +1 @@
+X5O!P%@AP[4\PZX54(P^)7CC)7}$EICAR-STANDARD-ANTIVIRUS-TEST-FILE!$H+H*
```
Dillon-Brown, line 1: @jack7anderson7 Thinking about this further, we should probably remove this. We don't publish tests with release packages but this is still going to impact anyone pulling down from the repo directly (Setting off AV, etc).

If we need to test this we'll likely need to look at mocking instead.

Dillon-Brown (changes requested):

Dillon-Brown:
@jack7anderson7 Also this should be rebased down to one atomic commit if possible.

Dillon-Brown (approved):

jack7anderson7 (dismissed):

lib/Utility/AntiMalware/Providers/Sophos.php
```diff
@@ -0,0 +1,213 @@
...
+     * @param string $path - location in the file system
+     * @throws InvalidArgumentException
+     * @throws MalwareFoundException|Exception
+     */
+    public function scanFilePath(string $path): void
+    {
+        $this->validatePath($path);
+
+        $command = $this->getExecutablePath() . ' ' . $this->getOptions() . ' ' . $path;
+        $lineOutput = shell_exec($command);
```
Dillon-Brown, line 119: Almost missed this one 😅, If I'm understanding how the command is built here then I think we could run into security issues here.

E.g. What if I named my file something like ```&& rm -rf .*```.

Dillon-Brown (dismissed):
LGTM, might hit some issues with windows and the shell_exec but seeing as Sophos is Linux and the path + options are fully configurable, it shouldn't cause any real issues.

jack7anderson7 (dismissed):

mattlorimer:
Build passing --> https://travis-ci.com/github/mattlorimer/SuiteCRM/builds/232273613

jack7anderson7 (dismissed):

Dillon-Brown (approved):

Issue #9192, opened 2021-06-15
Email - folders_rel fills with duplicate entries


tsmgeek:
https://community.suitecrm.com/t/database-table-folders-rel-runs-full-for-each-email-inbound-check/73490/13

Ive just come across the same issue as indicated in forum post above.
My folders_rel table currently has 99million rows+ and filling each and every run checking the inbound folder.

As always this comes back to a terribly broken email module.

SuiteCRM: 7.10.29

tsmgeek:
https://github.com/salesagility/SuiteCRM/blob/d57e91389d97791fe621d811f03fe05f8f5a7f78/include/SugarFolders/SugarFolders.php#L656-L678

In this function `addBean` there is no duplicate detection but is allowed to create the same entry multiple times which normally would be ok if further up the code it would not fire this off multiple times.

https://github.com/salesagility/SuiteCRM/blob/7.10.x/modules/Schedulers/_AddJobsHere.php#L626

- This line is looped for each and every message, if email is imported via `returnImportedEmail`
- `returnImportedEmail` always returns message ID even if email was already imported which is a change to prior `returnOneEmail`, now there is no possibility for calling function to see if the email is a duplicate or not
https://github.com/salesagility/SuiteCRM/blob/9cb957e4f41562eb44f6ce8c982e2a3c169fc951/modules/InboundEmail/InboundEmail.php#L5487
- Check will always pass on previously imported emails causing `addBean` to be run all the time
https://github.com/salesagility/SuiteCRM/blob/7.10.x/modules/Schedulers/_AddJobsHere.php#L623

kylix34:
Would the fix also work for SuiteCRM v7.11.18?

Mac-Rae:
Closing this issue as it has been resolved in the 7.10.34 release :+1:

SuiteBot:
This issue has been mentioned on **SuiteCRM**. There might be relevant details there:

https://community.suitecrm.com/t/suitecrm-stopped-working-folders-rel-folder-filled-up-the-entire-disk-space/85264/8

PR #9194, opened 2021-06-16
#9192 Fix InboundEmail.php returning ID for already imported emails

tsmgeek:
Do not return EmailID if email was not imported.


Dillon-Brown (approved):

tsmgeek:
@Dillon-Brown 
!! DO NOT IMPORT THIS !!

Ive got to do an additional fix as while it fixes one area when importing via CLI, it now does not work when doing manual imports from the UI.
Ile get it fixed.

tsmgeek:
@Dillon-Brown Not sure exactly how to fix this without breaking existing method compatibility.
On the one hand during scheduler import we need to know if an email was imported, not if a duplicate.
On the other hand  during manual UI import (which is more an Assign To than Import), it needs to get the ID back from the `returnImportedEmail` method so it can then apply the assignment with the `setAfterImport` method which is passed the emailID.

I can fix it but I was thinking ...
- to create a 2nd function `importEmail` that just imports and returns ID or returns false.
- to modify `returnImportedEmail` to return an array `['new'=>true,'id'=>'xxxx']` then update all the code
- rename `returnImportedEmail` to say `getImportedEmail`, create new `returnImportedEmail` as helper function, create new `importEmail` as a helper function that just returns ID/false.

The last way I think will maintain compatibility at the sake of messy code. Reality is how many of these `public` calls are really public or because its so broken can we just repair it.

Whichever way without a full rewrite this Email stuff is a total mess.

tsmgeek:
@Dillon-Brown which way do you think would be best path to fix this issue?

Dillon-Brown:
> @Dillon-Brown which way do you think would be best path to fix this issue?

Not entirely sure, think I'd need to spend some time looking into this to say for certain. However in regards to your questions around making BC changes, I'd recommend the following:

1. Leave any public methods you want to rename / refactor like ```returnImportedEmail ``` as is, just mark them as deprecated with a PHPDoc annotation and log a warning with ```LoggerManager::getLogger()->deprecated($deprecationMessage)```.
2. Create new methods and remove all the calls to the old ones.

Agreed that emails are a huge mess and need a full rewrite 😅

tsmgeek:
Ive redone the function to be more flexible and now this is running in our production system.

PR #9188, opened 2021-06-14
Fix PDF_Lib constructors

Dillon-Brown:
## Description

- Fixed duplicate class properties and constructor syntax issues throughout PDF_Lib that were introduced by an old-style PHP constructor fix PR. I don't expect any of this PDF_Lib code to be in 7.12.x so this should be fairly low risk, we only really need these classes as a reference for building an interface.

## How To Test This

1. Check that MPDF works as expected.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [X] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change to the documentation.
- [X] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

PR #9208, opened 2021-06-21
Implement standard PDFEngines

### LINKED ISSUE #7111: Update mPDF and move it to Composer

Dillon-Brown:
## Description

- Branched from #9188.
- Implemented base PDF engine ```lib/PDF```.
- Refactored old mPDF calls to use new ```PDFWrapper```.
- Removed the old method of setting the ```error_reporting``` to 0 to avoid having PDFs show PHP errors in favor of using ```@``` when the mPDF lib is called directly, this should also fix the issues mentioned in #7111.
- Added autogenerated font files to ```.gitignore```.

## Motivation and Context

- This is intended to be the minimum base required to complete further PDFEngine work (Adding new engines, engine selection, etc).

## How To Test This

1. Fully regression test all MPDF functionality (Reports, bulk action pdf, quotes, etc).

## Types of changes

### Final checklist

### DIFF
--- a/lib/PDF/PDFWrapper.php
+++ b/lib/PDF/PDFWrapper.php
@@ -1,0 +1,175 @@
+<?php
+/**
+ * SugarCRM Community Edition is a customer relationship management program developed by
+ * SugarCRM, Inc. Copyright (C) 2004-2013 SugarCRM Inc.
+ *
+ * SuiteCRM is an extension to SugarCRM Community Edition developed by SalesAgility Ltd.
+ * Copyright (C) 2011 - 2021 SalesAgility Ltd.
+ *
+ * This program is free software; you can redistribute it and/or modify it under
+ * the terms of the GNU Affero General Public License version 3 as published by the
+ * Free Software Foundation with the addition of the following permission added
+ * to Section 15 as permitted in Section 7(a): FOR ANY PART OF THE COVERED WORK
+ * IN WHICH THE COPYRIGHT IS OWNED BY SUGARCRM, SUGARCRM DISCLAIMS THE WARRANTY
+ * OF NON INFRINGEMENT OF THIRD PARTY RIGHTS.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU Affero General Public License along with
+ * this program; if not, see http://www.gnu.org/licenses or write to the Free
+ * Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301 USA.
+ *
+ * You can contact SugarCRM, Inc. headquarters at 10050 North Wolfe Road,
+ * SW2-130, Cupertino, CA 95014, USA. or at email address contact@sugarcrm.com.
+ *
+ * The interactive user interfaces in modified source and object code versions
+ * of this program must display Appropriate Legal Notices, as required under
+ * Section 5 of the GNU Affero General Public License version 3.
+ *
+ * In accordance with Section 7(b) of the GNU Affero General Public License version 3,
+ * these Appropriate Legal Notices must retain the display of the "Powered by
+ * SugarCRM" logo and "Supercharged by SuiteCRM" logo. If the display of the logos is not
+ * reasonably feasible for technical reasons, the Appropriate Legal Notices must
+ * display the words "Powered by SugarCRM" and "Supercharged by SuiteCRM".
+ */
+
+namespace SuiteCRM\PDF;
+
+use SuiteCRM\PDF\Exceptions\PDFEngineNotFoundException;
+use SuiteCRM\PDF\MPDF\MPDFEngine;
+
+if (!defined('sugarEntry') || !sugarEntry) {
+    die('Not A Valid Entry Point');
+}
+
+/**
+ * Class PDFWrapper
+ * @package SuiteCRM\PDF
+ */
+class PDFWrapper
+{
+    /**
+     * @var array stores an associative array matching the PDF engine class name with the file it is stored in.
+     */
+    private static $engines = [
+        'MPDFEngine' => [
+            'name' => 'MPDFEngine',
+            'FQN' => MPDFEngine::class,
+            'filepath' => 'lib/PDF/MPDF/MPDFEngine.php'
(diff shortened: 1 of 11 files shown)


modules/AOS_PDF_Templates/generatePdf.php
```diff
@@ -139,28 +138,40 @@ function ($matches) {
...
     try {
-        $orientation = ($template->orientation == "Landscape") ? "-L" : "";
-        $pdf = new mPDF('en', $template->page_size . $orientation, '', 'DejaVuSansCondensed', $template->margin_left, $template->margin_right, $template->margin_top, $template->margin_bottom, $template->margin_header, $template->margin_footer);
-        $pdf->SetAutoFont();
-        $pdf->SetHTMLHeader($header);
-        $pdf->SetHTMLFooter($footer);
-        $pdf->WriteHTML($printable);
-        if ($task == 'pdf') {
-            $pdf->Output($file_name, "D");
+        $orientation = ($template->orientation === "Landscape") ? "-L" : "";
```
clemente-raposo, line 145: Hi @Dillon-Brown, was wondering if it would be best to pass a standard `orientation`, like `landscape` or `portrait` instead of `-L` joined on the `format`. Then each pdf engine implementation would now know to set it, as it may differ from one to another.

lib/PDF/PDFWrapper.php
```diff
@@ -0,0 +1,175 @@
...
+    private static $engines = [
+        'MPDFEngine' => [
+            'name' => 'MPDFEngine',
+            'FQN' => MPDFEngine::class,
+            'filepath' => 'lib/PDF/MPDF/MPDFEngine.php'
+        ],
+    ];
+
+    /** @var string Path to the folder where to load custom engines from */
+    private static $customEnginePath = __DIR__ . '/../../custom/Extension/PDFEngines/';
```
clemente-raposo, line 67: Hi @Dillon-Brown does this work similar to the regular extension framework? what configuration is needed to add a new engine?

Dillon-Brown:
@clemente-raposo Just pushed a single line change to the tests (PHP 7.4 was failing due to a deprecation notice inside the mpdf lib).

lib/PDF/PDFWrapper.php
Dillon-Brown: So at the moment, everything in the PDFWrapper is hardcoded to use MPDFEngine. The intention here being that later on we will be able to specify an engine in the ```config.php``` that users can set through the front-end.

For the extension framework, you can just drop a new engine file inside ```custom/Extension/PDFEngines``` and the wrapper will pick it up as an option. I've kept this pretty much exactly the same as the GlobalSearch ```SearchWrapper``` (So much so that I'm thinking of extracting this out further into a common ```WrapperInterface``` for everything inside of ```lib/```).

clemente-raposo: ok thank you

modules/AOS_PDF_Templates/generatePdf.php
Dillon-Brown: Agreed, looking at mPDF it looks like it accepts orientation in ```Landscape``` which is what we generate with the template vars anyway.

Pushed along with a small PHP notice fix in the AOR_Reports controller when trying to count null:

```php
        if (is_countable($graphs)) {
            $countOfCharts = count($graphs);
        }
```

Dillon-Brown:
@clemente-raposo Travis build here: https://travis-ci.com/github/Dillon-Brown/SuiteCRM/builds/230164475

modules/AOS_PDF_Templates/generatePdf.php
clemente-raposo: thank you

clemente-raposo (dismissed):
code looks ok

tsmgeek:
I would include ability to pass in some custom array of params from config to set things like ttf font paths, default pdf params, just merge it in with the configurator.

We support watermarking for PDF with mPDF.
```php
              if(array_key_exists('watermark', $params) && !empty($params['watermark'])){
                  $pdf->SetWatermarkText($params['watermark']);
                  $pdf->showWatermarkText = true;
              }
```

I like this idea as we played with `wkhtmltopdf` and we had to bypass all the existing stuff, below is a base code we tried when testing this out.

```php
            $pdf = new mikehaertl\wkhtmlto\Pdf(array(
                'binary' => '/usr/bin/wkhtmltopdf'                    
            ));
            $pdf->setOptions(array(
                'page-size' => $template->page_size,
                'orientation' => ($template->orientation === 'Landscape' ? 'Landscape' : 'Portrait'),
                'dpi' => 300,
                'image-quality' => 75,
                'no-outline', // Make Chrome not complain
                'margin-top' => $template->margin_top,
                'margin-right' => $template->margin_right,
                'margin-bottom' => $template->margin_bottom,
                'margin-left' => $template->margin_left,
                'header-spacing' => '0',
                'enable-external-links',
                'disable-smart-shrinking'
            ));
    
            if(array_key_exists('watermark', $params) && is_string($params['watermark'] !== FALSE)){
                $GLOBALS["log"]->warn("watermark not supported by wkhtmltopdf");
            }
            
            $pdf->addPage($header . $printable . $footer);
            
            try{
                $ret = $pdf->saveAs($tmpfilepdf);
                if($ret===TRUE){
                    return $tmpfilepdf;
                }else{
                    echo $pdf->getError();
                }
                return false;
            }catch(Exception $e){
                return false;
            }
```

Dillon-Brown:
> I would include ability to pass in some custom array of params from config to set things like ttf font paths, default pdf params, just merge it in with the configurator.

Agreed, very good idea. At the moment my ```configurePDF``` method is more just a hacky way for me to get the default mPDF constructor params we need to support than it is a real configurator.

Will be adding multiple engines and a dedicated PDFConfigurator class as a follow-up to this :).

tsmgeek:
@Dillon-Brown  rather than extending an abstract class `PDFEngine` why not just make it an implementation and then test as `implements PDFEngineInterface`?

Dillon-Brown:
> @Dillon-Brown rather than extending an abstract class `PDFEngine` why not just make it an implementation and then test as `implements PDFEngineInterface`?

My thinking at the time was that I'm not sure if there is going to be common functionality between the PDFEngines that I would want to implement as part of ```SearchEngine```. If it turns out that there's not and ```SearchEngine``` remains solely comprised of abstract methods then I'll definitely be converting it to a pure interface.

clemente-raposo (approved):
latest changes look ok

Mac-Rae (approved):

PR #9227, opened 2021-07-11
CSS fixes: campaign wizard icons + alerts popup

YonatanRosemarin:
Fix CSS issues of 2 cases:
* Campaign wizard first step
* Alerts notifications pop up

## Description

Campaign - Before:
image

Campaign - After:
image

Alerts - Before:
image

Alerts - After:
image

## Motivation and Context

To fix display issue of the 2 cases described above

## How To Test This

**Need to first build CSS files from SCSS**
1. For alerts - create new alert
2. For Campaigns - go to Campaigns and click on "Create Campaign"

## Types of changes

### Final checklist

### DIFF
--- a/themes/SuiteP/css/suitep-base/navbar.scss
+++ b/themes/SuiteP/css/suitep-base/navbar.scss
@@ -1064,7 +1064,7 @@
 }
 
 .desktop_notifications > .dropdown-menu {
-  left: auto;
+  left: -82%;
 }
 
 .desktop_notifications a {
--- a/themes/SuiteP/css/suitep-base/suitepicon-glyphs.scss
+++ b/themes/SuiteP/css/suitep-base/suitepicon-glyphs.scss
@@ -85,6 +85,10 @@
 .suitepicon-action-megaphone:before {
     content: "\f11d";
 }
+.suitepicon.suitepicon-action-megaphone:after {
+    content: "\a";
+    white-space: pre;
+}  
 .suitepicon-action-minus:before {
     content: "\f11e";
 }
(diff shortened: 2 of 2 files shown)


themes/SuiteP/css/suitep-base/navbar.scss
```diff
@@ -1064,7 +1064,7 @@
 }
 
 .desktop_notifications > .dropdown-menu {
-  left: auto;
+  left: -82%;
```
samus-aran, line 1067: This also works for mobile? I dislike having these percentages as large as that, it's almost masking the real problem I feel. If I enlarge my text on my browers (cause I often take my glasses off) will it still be in a reasonable position?

samus-aran (commented):
The campaign wizard fixes look good and fairly straight forward to me :+1: 

Though I'm abit on the fence about the alerts. Should we instead look at where the anchor is of those desktop notifications rather than having to have a percentage (a very custom looking one) to position.

align_1
Is this the proposed anchor and we doing a minus from there.... Would this cause issues with mobile view? or text larger browser settings?
My thinking would it be better to move the anchor so that we are not so dependant on the right hand of the screen being in perfect alignment like so:
align_2

Hope I make sense :sweat_smile:

themes/SuiteP/css/suitep-base/navbar.scss
YonatanRosemarin: @samus-aran - you are correct, we shouldn't use it like that.
Do you have a suggestion on how to fix it correctly? Do you prefer I will remove it from the PR for now (I'm not sure when I will have the time to find the best solution)?

serhiisamko091184:
Hello @YonatanRosemarin ,
thanks for your contribution, 
it's recreated due to the last comment (fix focused on the icons only): 
https://github.com/salesagility/SuiteCRM/pull/10203

Thanks! 

Regards,
Serhii

Issue #9232, opened 2021-07-15
Content of custom fields of type URL (Generate link) is not displayed in listviews


gody01:
If one create custom field in studio of type URL and assign it in listview, no content is showed in the corresponding column in listview.

#### Issue
In studio I created custom field of type URL in Accounts module.
Then I assigned it to DetailView and ListView.
When I go to DetailView I see content of new field.
When I go to ListView there is no content in detail view.

#### Expected Behavior
If field of Type URL is assigned to ListView, conten should be displayed

#### Actual Behavior
There is no content in column on the ListView

#### Possible Fix
It is acctualy problem, that in include/ListView/ListViewData.php main_query returns empty string, since URL field is generated according to vardefs.

So one solution/workaround is to create logic_hook to populate correct data (it turn's out it is very simple):

```
<?php

     class AlterCustomURL_fieldname
     {
         function AlterURL($bean, $event, $arguments)
         {
             $bean->fieldname_c = $bean->field_name_map['fieldname_c']['default'] ;
         }
     }

?>
```

Template than already replaces what should be replaced.

#### Steps to Reproduce
1. Create field in studio of type URL
2. Add Field to ListView (DetailView, ...)
3. Populate relevant data (if it is generated URL from existing {id} for example nothing needs to be done.
4. Go to ListView

#### Context
I tried to make customization and relied to documentation, that I can put cutom fields to ListView 

Apart from tme spent searching and debuging for cause, workaroujnd turns out to be prety simple. 
So perhaps some documentation/example would be enough.

But it would be better if procesing of vardefs definiton and above code would be introduced somewhere in include/ListView/ListViewData.php so that this would work without need for logic_hook for each new custom field of type URL.

#### Your Environment

* SuiteCRM Version used: 7.11.20
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Independent of browser
* Environment name and version (e.g. MySQL, PHP 7): Independent of Mysql/PHP
* Operating System and version (e.g Ubuntu 16.04): tested on OpenSuSE Leap & Tumbletweed and CentOS 7 & 8.

johnM2401:
Hey there, 

I'm not able to replicate this currently, i'm afraid.

It appears to render as expected on Listview:
![image](https://user-images.githubusercontent.com/13675281/126644639-50ce66a5-5704-4ec7-9d19-58ea6c8dac6b.png)


Here are the steps I'd taken, please let me know if anything was missed:

[...]

PR #9234, opened 2021-07-15
9232 listview display url content type

gody01:
Changes to display content of generated field of type 'url' in ListView.

## Description

Fix for issue #9232 - Content of custom fields of type URL (Generate link) is not displayed in listviews

1. In include/ListView/ListViewDisplay.php 
 In function setupFilterFields 
  - added check if 'default' value for 'url' type fields contains dependend fields
  - add dependend fields to $filter_fields, so they are retrieved from DB

2. In data/SugarBean.php
 In function get_list_view_array
 - added check if  value of 'default' for 'url' type fields contains dependend fields (that means, that 'url' field is generated)
 - return value of 'default' as field value

## Motivation and Context

To display 'url' generated fields in listviews.

## How To Test This

- define URL type field with checked generated box
- select filed to generate
- define field to be displayed in ListView
- Check correct or any URL is displayed in ListView

## Types of changes

### Final checklist

- [X ] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ X] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

### DIFF
--- a/data/SugarBean.php
+++ b/data/SugarBean.php
@@ -5723,6 +5723,8 @@ class SugarBean
                     } else {
                         $return_array[$cache[$field]] = $this->$field;
                     }
+                } elseif ( (!empty($value['type']) && ($value['type'] == 'url' ) ) && preg_match ('|{(.*)}|U' , $value['default'])) {
+                    $return_array[$cache[$field]] = $value['default'];
                 } else {
                     $return_array[$cache[$field]] = $this->$field;
                 }
--- a/include/ListView/ListViewDisplay.php
+++ b/include/ListView/ListViewDisplay.php
@@ -223,6 +223,13 @@ class ListViewDisplay
                         }
                     }
                 }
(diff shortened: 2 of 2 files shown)


include/SugarFields/SugarFieldHandler.php
```diff
@@ -80,9 +80,11 @@ public static function fixupFieldType($field)
                case 'date':
                     $field = 'datetime';
                     break;
+/* #9232 user's selected column not displayed properly with link template
```
samus-aran, line 83: Hey @gody01 Just asking why is this commented out? Is there still more work to be done to be done?

samus-aran (commented):
I'm abit weary renaming tpl files in case others use it directly. Is it necessary? I know it would be inline with other fields name formats but is the effort doing it worth a potential breakage just for a lettercase. :)

PR #9245, opened 2021-08-03
add ssl support to mysqli connection

oc666:
MySQLi connection is not support ssl encryption.

## Description

This PR add support to MySQLi/MariaDB encrypted connection.
## Motivation and Context

Add security for database connection.
## How To Test This

Use MySQL or MariaDB that required SSL encryption.

Simple configuration to use in _config_override.php_:
```
$sugar_config['dbconfig']['db_ssl'] = true;
$sugar_config['dbconfig']['db_ssl_key'] = '<PATH>/client-key.pem';
$sugar_config['dbconfig']['db_ssl_certificate'] = '<PATH>/client-cert.pem';
$sugar_config['dbconfig']['db_ssl_ca_certificate'] = '<PATH>/server-ca.pem';
$sugar_config['dbconfig']['db_ssl_ca_path'] = null;
$sugar_config['dbconfig']['db_ssl_cipher_algos'] = null;
$sugar_config['dbconfig']['db_ssl_connect_options'][] = MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT;
```

## Types of changes

### Final checklist

### DIFF
--- a/include/database/MysqliManager.php
+++ b/include/database/MysqliManager.php
@@ -313,13 +313,48 @@ class MysqliManager extends MysqlManager
                 $dbport = substr($configOptions['db_host_name'], $pos + 1);
             }
 
-            $this->database = @mysqli_connect(
-                $dbhost,
-                $configOptions['db_user_name'],
-                $configOptions['db_password'],
-                isset($configOptions['db_name']) ? $configOptions['db_name'] : '',
-                $dbport
-            );
+			
+            if (!empty($configOptions['db_ssl'])) {
+                try {
+                    $this->database = @mysqli_init();
+                    mysqli_ssl_set(
+                        $this->database,
+                        isset($configOptions['db_ssl_key']) ? $configOptions['db_ssl_key'] : '',
+                        isset($configOptions['db_ssl_certificate']) ? $configOptions['db_ssl_certificate'] : '',
+                        isset($configOptions['db_ssl_ca_certificate']) ? $configOptions['db_ssl_ca_certificate'] : '',
+                        isset($configOptions['db_ssl_ca_path']) ? $configOptions['db_ssl_ca_path'] : null,
+                        isset($configOptions['db_ssl_cipher_algos']) ? $configOptions['db_ssl_cipher_algos'] : null
+                    );
+                } catch (Exception $ex) {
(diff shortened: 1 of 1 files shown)


tsmgeek:
These are my points on code.

-  `mysqli_connect` is deprecated in favour of `mysqli_real_connect`
- Just have one connect method and pass a generic `flags` to it, when ssl is enabled the required flags are added.
- seems that if there is an issue doing an ssl connection it falls back to non-ssl, this may be a security risk, either have this as a option `db_ssl_required` default `true` or always force ssl if config is defined, if not required and fails then log, if failed and is required SugarException and stop
- I would probably just nest the sugar config for ssl, ie `db_ssl=>['required'=>true]` etc.
- if you are doing `$db = mysql_init()` then you can do `$db->real_connect()` instead, same with `$db->ssl_set()`, using oop flow.

oc666:
> mysqli_connect is deprecated in favour of mysqli_real_connect
Just have one connect method and pass a generic flags to it, when ssl is enabled the required flags are added.

We try to reduce backward compatability changes as much as we can.

> seems that if there is an issue doing an ssl connection it falls back to non-ssl, this may be a security risk, either have this as a option db_ssl_required default true or always force ssl if config is defined, if not required and fails then log, if failed and is required SugarException and stop

No. This is wrong assumption as the old connect (without ssl) is under else. This means that if there is no change in the configuration and ssl setting was not added, the system will use the same code. If ssl settings is on it will try to use only the ssl option without drop back to non-ssl.

> if you are doing $db = mysql_init() then you can do $db->real_connect() instead, same with $db->ssl_set(), using oop flow.

You cannot put in real_connect method the ssl certificates, but only flags. This is the reason we used ssl_set method before.

tsmgeek:
I forgot that the link returned from Procedural commands can not be used with oop.

You have your mysqli_ssl_set in a try/catch, if that fails then it will fall though to connect without it being set, ie would it give exception if cert file was missing the connect without ssl.

chris001:
@oc666 Please change target branch from [master](https://github.com/salesagility/SuiteCRM/tree/master) to [hotfix](https://github.com/salesagility/SuiteCRM/tree/hotfix).

PR #9294, opened 2021-09-21
Fix/noon styling issues

balumahendranms:
## Description
Noon Styling Issue fixes

## Motivation and Context
CSS variable is not assigned to the color. So the change is required

## How To Test This
In SuiteCRM pages

## Types of changes

- The close button of some models have the same foreground & background colour
- On the dashboard, the page numbers on dashlets are not centred
- When using the mass update menu it has the same foreground & background colour
- Some buttons within models have the same foreground & background colour
- Fields are not rounded in relate popups

### Final checklist


themes/SuiteP/css/suitep-base/listview.scss
```diff
@@ -941,7 +941,7 @@ a.listViewTdToolsS1 {
 
 #selectLinkTop > .sugar_action_button > .subnav a {
   background-color: $list-view-action-menu-bg;
-  color: $default-btn-color;
+  color: $white;
```
Dillon-Brown, line 944: Are these meant to be white on all the other themes like Dawn/Day/Dusk? since this is changing the base it will impact every theme.

Dillon-Brown:
@balumahendranms It looks like the style.css is unminified., you should be able to minify it by using ```compressed``` in the compile command.

themes/SuiteP/css/suitep-base/listview.scss
balumahendranms: Hi dillon i have cross verified the all the themes with built command seperatly. Its working fine without impacting the color changes.

balumahendranms:
> @balumahendranms It looks like the style.css is unminified., you should be able to minify it by using `compressed` in the compile command.



> @balumahendranms It looks like the style.css is unminified., you should be able to minify it by using `compressed` in the compile command.

I have minified

themes/SuiteP/css/suitep-base/listview.scss
Dillon-Brown: So looking at Day we have:

```
$default-btn-color: $color-51;
$color-51: #F5F5F5;
```

And then in Dusk we have:

```
$default-btn-color: $color-3;
$color-9: #D3D3D3;
```

These are close enough to white that it's pretty unnoticeable, however I think we lose some ability to customise by changing the variables here to ```$white```. Would it not make sense to keep them as ```$default-btn-color``` and then just change the variable assignment specifically in the Noon theme? or would that have a larger impact?

themes/SuiteP/css/Dawn/variables.scss
```diff
@@ -939,6 +939,8 @@ $default-btn-bg-hover: $color-28;
 $text-hover-listview: $color-80;
 $subpanel-text-hover: $color-90;
 
+$text-color-white: $color-90;
```
Dillon-Brown, line 942: Could we change the name of this variable to remove the explicit color? it's possible that ```$color-90``` might be changed to be something other than white.

It would also be good to squash the 4 commits here down into one if we can.

Dillon-Brown (approved):

Mac-Rae:
Assessed :+1:

Mac-Rae (approved):

Dillon-Brown (approved):

Mac-Rae (approved):

PR #9361, opened 2021-11-12
Fix Issue #9192: Fix duplication of folders_rel table entries.

MikeyJC:
## Description

Fixes a bug when importing emails and using the Sugar Folders module where it would add an identical relationship between Sugar Folders and Inbound Email accounts.

## Motivation and Context

Stops duplicating data in the database for each email imported.

## How To Test This

If the folders_rel table is empty, when importing emails, the Inbound Email account in use should create a new relationship between itself and a Sugar Folder.
Once that relationship exists, it should be the only one of it's kind in the folders_rel table, even after importing more emails with the same Inbound Email account.

## Types of changes

### Final checklist

### DIFF
--- a/include/SugarFolders/SugarFolders.php
+++ b/include/SugarFolders/SugarFolders.php
@@ -695,6 +695,29 @@ class SugarFolder
             return false;
         }
 
+        /*
+        Fix issue #9192 - Duplicating rows for folders_rel
+        First check if a row with the same data already exists
+        If so, return false
+        */
+
+        $q = "SELECT id FROM folders_rel WHERE".
+            " folder_id = ".$this->db->quoted($this->id).
+            " AND polymorphic_module = ".$this->db->quoted($bean->module_dir).
+            " AND polymorphic_id = ".$this->db->quoted($bean->id).
+            " AND deleted = 0";
+
+        $result = $this->db->fetchByAssoc($this->db->query($q));
+        if ($result === false){
+            $GLOBALS['log']->debug("Error in query to check for existing email folders");
+            return false;
+        }
+
+        if($result) {
+            $GLOBALS['log']->debug("*** FOLDERS: addBean() is trying to create an already existing relationship");
+            return false;
+        }
+        
         $guid = create_guid();
 
         $query = "INSERT INTO folders_rel " .


jack7anderson7 (dismissed):

pgorod:
Since this will happen for every email added to the database, and the amounts can be significant, wouldn't it be better to try to get this into a single query? The SELECT + INSERT could become an INSERT with a WHERE clause.

Better yet, get the restrictions to be handled by the database, so that the inserts would simply be rejected if they violated uniqueness constraints. I say this would be better, but I didn't study it thoroughly, this could get confusing if the different databases (MYSQL, MSSQL, etc) don't have similar mechanisms for this.

MikeyJC:
While the way it's currently built likely wouldn't cause a major impact on performance, that may be a good idea to try.

A bit of clarification regarding what would happen during email imports. The addBean() functionality was called to create a relationship between the Inbound Email Account and a Sugar Folder for every email, which was the cause of the duplication bug. With this change, if the relationship already exists, it will only ever run the SELECT query, and it would only run through both SELECT + INSERT queries once for each Inbound Email Account.

Regardless, it may be a good idea to try moving them into one query anyway, as it's called during moveEmails() as well.

include/SugarFolders/SugarFolders.php
```diff
@@ -695,6 +695,23 @@ public function addBean(SugarBean $bean)
...
+        If so, return false
+        */
+
+        $q = "SELECT id FROM folders_rel WHERE".
+            " folder_id = ".$this->db->quoted($this->id).
+            " AND polymorphic_module = ".$this->db->quoted($bean->module_dir).
+            " AND polymorphic_id = ".$this->db->quoted($bean->id).
+            " AND deleted = 0";
+
+        if($this->db->fetchByAssoc($this->db->query($q))) {
```
clemente-raposo, line 710: Hi @MikeyJC ,

Thank you for the pull request.

I was thinking that probably we should also handle errors when running the query. If the `fetchByAssoc` returns `false` that means there was an error running the query. Thus, in that case we should probably also log and return.

What do you think?

jack7anderson7 (dismissed):

jack7anderson7 (approved):

clemente-raposo (approved):
changes look ok

PR #9388, opened 2021-12-03
Optimise Export Memory Usage

simonr44:
## Description
This PR passes records to the export function in export_utils in small batches rather than all at once to minimise the memory consumption of the csv export process.

## Motivation and Context
When exporting lots of records at one time, the php memory consumption can quickly exceed the default memory_limit
This change will process the records in small batches, consuming less memory.

Large exports of +/- 100k records can easily exceed 750MB peak memory consumption. With this change consumption remains below the php default value of 128MB ( Approx 74MB in testing, but will vary based on record contents )

## How To Test This
Select all records on the list view and perform a bulk action export
Best performed on a test DB with many thousands of records.

## Types of changes

### Final checklist

### DIFF
--- a/export.php
+++ b/export.php
@@ -49,6 +49,7 @@ require_once('include/export_utils.php');
 global $sugar_config;
 global $current_user;
 global $app_list_strings;
+global $db;
 
 $the_module = clean_string($_REQUEST['module']);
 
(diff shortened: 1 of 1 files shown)


jack7anderson7 (dismissed):

clemente-raposo (dismissed):
LGTM

chris001:
Would be great to have this on import also, if not already there, so that users wouldn't have to continue to import large number of records by the command line.

johnM2401:
Hey @simonr44 !
After some testing, this does indeed seem to cut resource costs quite notably.
On a dataset of about 100k records locally, it seems to cut ~400mb of memory usage.

However, I've noticed one possible issue.

It seems as though it adds the Header row for each batch of records, back into the CSV.


I've grabbed a CSV from pre-fix and post-fix. (left and right respectively), using the same dataset.
However, the post-fix CSV appears to have 110 more rows:
image

<br>

After sorting A->Z on the "Name" header row, I noticed that there were now 110 extra rows that contained the header row items
See screenshot:
image
image


I imagine the CSV should only have the initial header row?
This might cause data issues if the CSV is ever re-imported to a CRM.
Could you have a look when you get a chance?

Thank you!

Issue #5197, opened 2018-02-14
Importing Contacts saves every contacts email as the first email in the csv.


rokacom:
There was the same issue in a previous version, but users reported it to be fixed.  I confirmed it was not in version 7.9.9, then I upgraded to 7.9.12 and the issue appeared.

#### Issue

When importing contacts via the "Import Contacts" method each contact is saved with the first line's email address. The same email address is saved for more than the correct contact.

#### Expected Behavior

The correct email should be saved in each contact.

#### Actual Behavior
When importing a csv with the following contents.

Email Address, First Name, Last Name
test1@domain.com, First1, Last1
test2@domain.com, First2, Last2
test3@domain.com, First3, Last3

When imported It adds the records as follows
First Name       Last Name    Email
First1                 Last1            test1@domain.com
First2                Last2            test1@domain.com
First3                Last3            test1@domain.com

#### Possible Fix
Unknown

#### Steps to Reproduce

1. Go to Contacts List
2. Click on Import Contacts
3. Import a file with the following information

Email Address, First Name, Last Name
test1@domain.com, First1, Last1
test2@domain.com, First2, Last2
test3@domain.com, First3, Last3

4. Look at the contacts. Each contact will have the same email address.

#### Context

We cannot import our contacts into a new installation.

#### Your Environment

* SuiteCRM Version used: 7.9.12. (It was working in 7.9.9 before I upgraded)
* Browser name and version Firefox 58.0.2 (64-bit) On Mac
* Environment name and version  PHP 7 
* Operating System and version: Ubuntu 16.04

sasha2002:
Same here with Leads

horus68:
Related: https://github.com/salesagility/SuiteCRM/issues/5200

PR #5203, opened 2018-02-14
Fixed #5197 - Importing Contacts saves every contacts email as the first email in the csv.

Dillon-Brown:
## Description

When importing contacts via the "Import Contacts" method each contact is saved with the first line's email address. The same email address is saved for more than the correct contact.

## Motivation and Context

Issue reference: #5197 

## How To Test This

1. Fresh install and then check that the accounts list-view has been correctly populated.
2. Enter contacts module and click "import"
3. Import a csv with the following data:
Email Address, First Name, Last Name
test1@domain.com, First1, Last1
test2@domain.com, First2, Last2
test3@domain.com, First3, Last3

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [X] My code follows the code style of this project found [here](https://suitecrm.com/wiki/index.php/Coding_Standards).
- [ ] My change requires a change to the documentation.
- [X] I have read the [**How to Contribute**](https://suitecrm.com/wiki/index.php/Contributing_to_SuiteCRM) guidelines.

PR #9391, opened 2021-12-07
Fixed #7842 Do not reset email addresses list upon saving #9390

### LINKED ISSUE #7842: Email Addresses Automatically Deleted By Workflows Relating to Contacts
If a workflow runs and makes changes to a contact, if the contact has more than 2 emails, all but 2 emails are deleted

There are some issues (specifically issue #6983, although I don't know if the cause is related) that are similar, but nothing that specifically relates to workflow. This has been tested on demo.

#### Issue
When a workflow is created, and modifies a contact record, if the contact has more than 2 emails, all but 2 emails are deleted. This is not related to only one type of workflow and is reproducible on save, on scheduler, on all, and doesn't matter which module the workflow is related to, as long as the workflow ends up modifying some field on contact. 

I have pictures detailing the contact before and after.

Screen Shot 09-10-19 at 08 46 AM

Screen Shot 09-10-19 at 08 48 AM

#### Expected Behavior
Workflow doesn't delete email addresses by itself
#### Actual Behavior
What is occurring is the SugarEmailAddress.php file is setting one out of  x > 2 addresses to primary, one as a reply to (?) and then deleting any others left in the $check_list array.  You can see this all in the logs when debug mode is turned on.

#### Possible Fix
I don't have an exact fix but this is as close as I got:
It would seem that all addresses are placed in the $check_list array in SugarEmailAddress.php, then at line 698 they enter an if statement where a primary address is selected, and also a reply to address (This part I'm not sure about, but I'm guessing that's what happens at 701). These addresses are unset at line 704, leaving all the rest of the addresses in the $current_links array.  When the array reaches 730, the array is not a conversion, and is not empty, so it passes the conditionals and those links are then deleted. Issue #6983 seemed to believe the if statement at 730 shouldn't always pass, but I don't know enough to about the intended functionality to determine whether this is the issue, or if the issue is that these addresses were supposed to be removed from $current_links before it reached that point.

https://github.com/salesagility/SuiteCRM/blob/12a5b11ed870d61307c8e67c48629a77c146fe3c/include/SugarEmailAddress/SugarEmailAddress.php#L697-L738

#### Steps to Reproduce
1. Create a contact with more than 2 emails (or use an existing contact with more than 2 emails)
2. Create a workflow relating to anything besides contacts, and modify any field in contacts as one of the actions (in my case I used Tasks and updated the relatives Contact's Primary Street Address, although any combination should work)
3. Create a Task that meets the conditions of your new workflow
4. Check the associated contact after workflow runs.

#### Context
This bug has ended up deleting about a hundred contact emails for my company.  We did not realize the leak, as most contacts don't have 3 or more emails.  I personally from that standpoint would say that this is high priority due to the potential for data loss.  In the meantime, we will have to strip most workflow functionality. 

#### Your Environment

* SuiteCRM Version used: Admin Demo, 7.10.18
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Chrome newest
* Environment name and version (e.g. MySQL, PHP 7): MariaDB, PHP 7.3
* Operating System and version (e.g Ubuntu 16.04): Ubuntu 18.04 LTS

clemente-raposo:
## Description

- Fix: https://github.com/salesagility/SuiteCRM/issues/7842
- Fixes issue in workflows, where email addresses where being removed.
-- this happened if record had more than 2 email addresses
- - Revert 41ee520

## Motivation and Context

- Fixes issue https://github.com/salesagility/SuiteCRM/issues/7842

## How To Test This

- See `Steps to reproduce` in https://github.com/salesagility/SuiteCRM/issues/7842
-  Also re-test https://github.com/salesagility/SuiteCRM/issues/5197  

## Types of changes

### Final checklist

### DIFF
--- a/include/SugarEmailAddress/SugarEmailAddress.php
+++ b/include/SugarEmailAddress/SugarEmailAddress.php
@@ -239,7 +239,7 @@ class SugarEmailAddress extends SugarBean
             || !isset($_REQUEST[$bean->module_dir . '_email_widget_id'])
             || !isset($_REQUEST['massupdate'])
         ) {
-            if (empty($this->addresses) || (!empty($bean->email1))) {
+            if (empty($this->addresses)) {
                 $this->addresses = array();
                 $optOut = (isset($bean->email_opt_out) && $bean->email_opt_out == '1');
                 $invalid = (isset($bean->invalid_email) && $bean->invalid_email == '1');


clemente-raposo:
Caused by fixes in following MR:

- https://github.com/salesagility/SuiteCRM/pull/5203#discussion_r168165185
    -  the coded highlighted in the above comment is causing data loss.

Reverted the code mentioned on the comment aboce. Re-testied the problem in https://github.com/salesagility/SuiteCRM/issues/5197 and wasn't able to replicate.

mattlorimer (dismissed):

Mac-Rae (approved):

mattlorimer (approved):

jack7anderson7 (approved):

PR #9436, opened 2022-01-19
Fix #9435 - Dropdown doesn't return empty selected value

### LINKED ISSUE #9435: Workflow: Dropdown always returns default value when empty item is selected
The editview of a workflow is built using a Smarty template. When editing an existing workflow with an action that sets to blank an enum field which has a not blank default value, the template does not show the blank value (as expected) but the default value. 

See image bellow:

WF_defaultValueDropdown

#### Issue
After opening a saved workflow that contains a condition using a dropdown with the empty item selected, the dropdown field always displays the default value if the default value isn't empty.

#### Expected Behavior
The empty ítem should be displayed when the saved workflow record is loaded

#### Actual Behavior
The workflow actually runs well, using the condition "empty" for the dropdown, but the displaying in the EditView is wrong.

#### Possible Fix
Forcing the value " ' ' " into the vardef definition. Some changes in modules/AOW_WorkFlow/aow_utils.php

#### Steps to Reproduce
1. Set any item as a default value for the dropdown Category in Documents
2. Create a Workflow based in Documents
3. Add condition of the field Category selecting the empty value
4. Save and open the workflow, check that the value displayed in the condition is the default and not the empty one.

#### Context
This can be very confusing for users

#### Your Environment

SinergiaCRM:
Closes #9435 

## Description
As described in the issue, the editview of a workflow is built using a Smarty template. When editing an existing workflow with an action that sets to blank an enum field which has a not blank default value, the template does not show the blank value (as expected) but the default value. 

In order to properly show the expected blank value, we hack the vardef definition overriding default value with blank value when the stored workflow value (now in $value) is blank.

## Motivation and Context
This can be confusing for the users and it needs to be fixed.

## How To Test This
1. Set any item as a default value for the dropdown Category in Documents
2. Create a Workflow based in Documents
3. Add condition of the field Category selecting the empty value
4. Save and open the workflow, check that the value displayed in the condition is the empty one.

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOW_WorkFlow/aow_utils.php
+++ b/modules/AOW_WorkFlow/aow_utils.php
@@ -533,6 +533,11 @@ function getModuleField(
         if (isset($fieldlist[$name]['options']) && is_array($fieldlist[$name]['options']) && !isset($fieldlist[$name]['options'][''])) {
             $fieldlist[$name]['options'][''] = '';
         }
+	// Fix #9435 - In order to properly show the expected blank value, we hack the vardef definition overriding default value 
+        // with blank value when the stored workflow value (now in $value) is blank.
+        if ($fieldlist[$name]['type'] == 'enum' || $fieldlist[$name]['type'] == 'multienum' || $fieldlist[$name]['type'] == 'dynamicenum') {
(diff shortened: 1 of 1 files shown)


SinergiaCRM:
Rebased branch to hotfix to #9522

PR #9441, opened 2022-01-20
Fix #9440 - Forcing default null value for numeric core fields.

### LINKED ISSUE #9440: Saving "0" value in numeric fields when it's empty
In a new module (from the ModuleBuilder) that contains a non-required non-custom/core numeric field (integer or float). If the record is saved without filling that field, the field will display and contain (in DB) a "0" value.

After, if you edit the record, remove the 0 and save it again, it saves fine.

#### Issue
The numeric fields with default empty value are wrongly filled with a "0" after the saving the record. This happens with non-required non-custom/core numeric field (integer or float) from newly created modules from Module builder

#### Expected Behavior
It should be saved "null" if no value is set.

#### Actual Behavior
A "0" value appears on that field.

#### Possible Fix
Change behaviour of the DBManager for those fields.

#### Steps to Reproduce
1. Create a new module in Module Builder
2. Add an Integer or Decimal field
3. Add the field to Edit and Detail View
4. Deploy module
5. Create a new record to the module, filling only the required name
6. Check that the numeric field has a "0" value

#### Context
This is very confusing. It isn't the same a value 0 than an empty one. This should be fixed.

#### Your Environment

SinergiaCRM:
Closes #9440 

## Description
As described in the issue, non-custom/core non-required numeric fields aren't saving properly a default empty value in newly created modules.

## Motivation and Context
(int) and (float) summoning always return 0 if the value is empty. Therefore we add a condition to return "null" if it's empty.

## How To Test This
1. Create a new module in Module Builder
2. Add an Integer or Decimal field
3. Add the field to Edit and Detail View
4. Deploy module
5. Create a new record to the module, filling only the required name
6. Check that the numeric field has an empty value

## Types of changes

### Final checklist


SinergiaCRM:
Rebased branch to hotfix to #9524

PR #9490, opened 2022-02-03
Fixed #9489 - Remove cookies from $_REQUEST super global variable

### LINKED ISSUE #9489: Export functionality breaks because of cookie with JSON value
#### Issue
We had an advance search on Contacts and then we wanted to export the filtered elements however all elements were in the exported file. After investigation it turned out that the server has `request_order = "GPC"` configuration in php.ini and we had another cookie on the same domain that has json as value.
```
Cookie name: cck1
Cookie value: {"cm":true,"all1st":true,"closed":false}
```

image

#### Expected Behavior
To only export what was selected by the filter.

#### Actual Behavior
The exported file contains all the contacts because the `where` clause generation fails in `include/export_utils.php`

#### Possible Fix
It seems when `$_REQUEST` global variable includes cookies as well then this ends up in the `current_query_by_page` too, however I think it's not needed because cookies are stored and sent anyway and they do not need to be included into `current_query_by_page`.
Maybe they are there because of the multi-tab massupdate?

#### Steps to Reproduce
1. Create a filter on Contacts
2. Select all the filtered elements and export them -> you should get the proper export
3. Introduce the above mentioned cookie to the domain via the browser for example
4. Select all the filtered elements again and export them -> you should get the full list in the export instead of the filtered elements

#### Context
This line:
`$searchForm->populateFromArray(json_decode(html_entity_decode($query), true));`
from `include/export_utils.php` fails because the `$query` variable also contains the json text from the cookie because of `current_query_by_page`

#### Your Environment
- SuiteCRM Version used: 7.11.13
- Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Chrome
- Environment name and version (e.g. MySQL, PHP 7): PHP 7.3.20
- Operating System and version (e.g Ubuntu 16.04): Ubuntu 20.04.3 LTS

gabor-harsanyi:
## Description
If request_order contains cookies as well, like for example "GPC" then the cookie values end up in $current_query_by_page and hence allowing other cookies from the same domain to break SuiteCRM with their json content during json decoding.

see: https://github.com/salesagility/SuiteCRM/issues/9489

## How To Test This

1. Create a filter on Contacts
2. Select all the filtered elements and export them -> you should get the proper export
3. Introduce a cookie to the domain via the browser for example with the value: {"cm":true,"all1st":true,"closed":false}
4. Select all the filtered elements again and export them -> you should get the full list in the export instead of the filtered elements

## Types of changes

### Final checklist

### DIFF
--- a/include/ListView/ListViewDisplay.php
+++ b/include/ListView/ListViewDisplay.php
@@ -589,7 +589,7 @@ class ListViewDisplay
         global $app_strings;
         unset($_REQUEST[session_name()]);
         unset($_REQUEST['PHPSESSID']);
-        $current_query_by_page = htmlentities(json_encode($_REQUEST));
+        $current_query_by_page = htmlentities(json_encode(array_diff_key($_REQUEST, $_COOKIE)));
 
         $js = <<<EOF
             if(sugarListView.get_checks_count() < 1) {
--- a/include/MassUpdate.php
+++ b/include/MassUpdate.php
@@ -111,7 +111,7 @@ class MassUpdate
         unset($_REQUEST['current_query_by_page']);
         unset($_REQUEST[session_name()]);
         unset($_REQUEST['PHPSESSID']);
-        $query = json_encode($_REQUEST);
+        $query = json_encode(array_diff_key($_REQUEST, $_COOKIE));
 
         if (!isset($_REQUEST['module'])) {
             LoggerManager::getLogger()->warn('Undefined index: module');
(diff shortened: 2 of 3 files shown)


serhiisamko091184:
Hello @gabor-harsanyi 
thanks for your contribution, 
would you be so kind to update me is the issue still present in the latest versions of SuiteCRM (the latest is 7.14.0 at the moment)? 

If yes, could you, please, be a bit more speific, how to replicate the issue, as we didn't manage to reproduce it. 
https://github.com/salesagility/SuiteCRM/issues/9489

Thanks in advance. 

Regards,
Serhii.

serhiisamko091184:
Hello, @gabor-harsanyi 

The issue has been marked as stale because there has been no recent activity. It will be closed if no further activity occurs.
Thanks for your contributions.

Regards,
Serhii

serhiisamko091184:
Hey @gabor-harsanyi 

Thank you.

Regards,
Serhii

PR #9498, opened 2022-02-07
Fix #10629 - Duplicating Surveys

### LINKED ISSUE #8947: Duplicate Surveys is Broken
#### Issue

There are actually two issues here although I suspect they are related.
1) Duplicating a survey with entries added, fails with an error message about editing surveys with entries
2) Duplicating an empty survey copies the survey but MOVES the questions (ie the questions appear okay on the new survey but are deleted from the old one)

#### Expected Behavior

At least in 2 above the both the original survey and the new one should have the questions attached.

#### Actual Behavior

The questions in the original survey are missing after the duplication. I suspect that the attempt to move rather than duplciate is why the error in 1) above happens.

#### Steps to Reproduce

1. Create a Survey with questions
2.Duplicate the survey
3.Check the original survey and see the questions missing
4.

#### Your Environment

* SuiteCRM Version used:7.11

### LINKED ISSUE #10629: When duplicating a survey with a workflow, the questions in the duplicated survey are duplicated
### Issue

It is checked if we have a workflow and duplicate a survey, the duplicate survey will also have duplicate questions

Image

### Possible Fix

_No response_

```bash
1. WorkFlow Module: Surveys
2. Run: Only On Save
3. Run On: New Records
4. Action:
  4.1. Modify the survey itself, for example, change from draft to published status.
  4.2. Calculate fields, for example, calculate the name.
5. When duplicating a survey, the duplicate survey will also have duplicate questions.
```

### Context

_No response_

### Version

7.14.6

Chrome

### Browser Version

Chrome

### Environment Information

PHP 7.4

Ubuntu 22.04

SinergiaCRM:
Closes https://github.com/salesagility/SuiteCRM/issues/8947

As the Issue mentions, there is a problem when duplicating Surveys. Basically, what goes wrong when duplicating:
- If the base survey has questions and no answers, the duplicated survey will take all the questions, leaving the base survey without questions
- If the base survey has questions and answers, the duplicated survey won't take/keep the questions, neither the answers. Although the Editview that appears after duplicating will contain the questions and won't allow it to modify them:
Selection_430

Closes https://github.com/salesagility/SuiteCRM/issues/10629

Having a WorkFlow and duplicating a survey or create also duplicated the questions, as the code was executed without any workflow verification. This was because:
- When duplicating survey X to Y or creating survey X, having WorkFlow active, the code saved the questions normally.
- But because there was no control over the workflow, when the workflow was executed:
   - The code re-executed the foreach loop that saves the questions.
   - As a result, the questions were saved twice: once during the normal save and once during the workflow execution.

## Description
This PR fixes the Duplicate functionality for the Surveys module, as we understand it should work:
- If the base survey has questions and no answers, the duplicated survey will keep the questions, duplicating them.
- If the base survey has questions and answers, the duplicated survey ~won't~ will take/keep/duplicate the questions ~or~ but not the answers. And other new Questions can be added during the first edition after duplication.
- Fixed the workflow problem by adding a control variable `already_saved` which is used to verify that the method is not re-executed. If `already_saved` is set, it means that the method has already been executed and reprocessing of the questions is avoided.

## Motivation and Context
With this change, duplicating Surveys will be consistent.

## How To Test This
1- Create a Survey
2- Add questions
3- Duplicate it.
4- Repeat adding answers to the survey.

- Workflow Issue:
1. WorkFlow Module: Surveys
2. Run: Only On Save
3. Run On: New Records
4. Action:
  4.1. Modify the survey itself, for example, change from draft to published status.
  4.2. Calculate fields, for example, calculate the name.
5. Duplicate the survey from step 1, checking the number of questions it has.
6. Check that the survey has been duplicated without duplicating the questions and that it has the same questions as the survey from step 1.
7. Create a survey with several questions and check that the surveys are not duplicated.

## Types of changes

### Final checklist


jack7anderson7:
Hi @SinergiaCRM,

The fix looks good, there's just one change that would be ideal. When duplicating surveys it is preferred if the questions are always carried across even when they have answers, just the answers wouldn't.

Thanks,
Jack

SinergiaCRM:
> Hi @SinergiaCRM,
> 
> The fix looks good, there's just one change that would be ideal. When duplicating surveys it is preferred if the questions are always carried across even when they have answers, just the answers wouldn't.
> 
> Thanks, Jack

Hi @jack7anderson7,

Yes, this was actually what we wanted to implement. Updated the main PR message and the code. :+1:

jack7anderson7 (approved):

SinergiaCRM:
Updated the code, there was a syntax error

phil8900 (approved):
Fix is working and should be merged I think. @jack7anderson7 could you take a look at this? :)

serhiisamko091184:
Hello @SinergiaCRM ,
thanks a lot for your contribution, 

would you be so kind to combine commits into one? 

Thank you in advance! 

Regards, 
Serhii

SinergiaCRM:
> Hello @SinergiaCRM , thanks a lot for your contribution,
> 
> would you be so kind to combine commits into one?
> 
> Thank you in advance!
> 
> Regards, Serhii

Done!

PR #9524, opened 2022-02-25
Fix #9440 - Forcing default null value for numeric core fields.

### LINKED ISSUE #9440: Saving "0" value in numeric fields when it's empty

SinergiaCRM:
Rebased branch to hotfix from #9441 

Closes #9440 

## Description
As described in the issue, non-custom/core non-required numeric fields aren't saving properly a default empty value in newly created modules.

## Motivation and Context
(int) and (float) summoning always return 0 if the value is empty. Therefore we add a condition to return "null" if it's empty.

## How To Test This
1. Create a new module in Module Builder
2. Add an Integer or Decimal field
3. Add the field to Edit and Detail View
4. Deploy module
5. Create a new record to the module, filling only the required name
6. Check that the numeric field has an empty value

## Types of changes

### Final checklist

### DIFF
--- a/include/database/DBManager.php
+++ b/include/database/DBManager.php
@@ -2209,7 +2209,7 @@ abstract class DBManager
                         return 0;
                     }
 
-                    return (int)$val;
+                    return $val === '' ? 'NULL' : (int)$val; // Fix #9440 - Forcing default null value for numeric fields.
                 case 'bigint':
                     $val = (float)$val;
                     if (!empty($fieldDef['required']) && $val == false) {
@@ -2230,7 +2230,7 @@ abstract class DBManager
                         return 0;
                     }
 
-                    return (float)$val;
+                    return $val === '' ? 'NULL' : (float)$val; // Fix #9440 - Forcing default null value for numeric fields.
                 case 'time':
                 case 'date':
                     // empty date can't be '', so convert it to either NULL or empty date value


clemente-raposo (approved):
LGTM

jack7anderson7 (approved):

johnM2401 (approved):
Each of the four core number-type fields are rending as "null" when saved as empty now.

I have noticed that this same initial issue happens for other field-types too (ie, saving as " ", rather than NULL):
image

However, I assume this is fine as-is. (As the difference between 0 and NULL for numbers is more distinct)

Assuming this is okay, LGTM!

gunnicom:
I think I fix this like this since years (I tried to integrate that in the original Sugar years ago)
Does take the "" => NULL into account, too.
```
diff --git a/include/database/DBManager.php b/include/database/DBManager.php
index 5d172ba93..0fe7644c4 100644
--- a/include/database/DBManager.php
+++ b/include/database/DBManager.php
@@ -543,7 +543,7 @@ abstract class DBManager
                     continue;
                 }

-                if (isset($data[$field])) {
+                if(isset($data[$field]) && (strlen($data[$field])>0 || $data[$field]===false)) { 
                     // clean the incoming value..
                     $val = from_html($data[$field]);
                 } else {
```

PR #9552, opened 2022-03-28
Fix #9551 - Update date period to include the users TZ.

### LINKED ISSUE #9551: Timezone ignored in report period
#### Issue

The users timezone appears to be ignored when using the period conditions (spotted on the "Last Month" condition). I'm unsure if this is intended or not.

#### Expected Behavior

A user viewing a report with the condition "Last Month" should only see records for the last month.

#### Actual Behavior

Records that fall within the timezone offset are shown. This means the user may see entries in the report for the last day of the month two months ago,

#### Possible Fix

`getPeriodDate` in `aor_utils` could subtract the users timezone offset before returning the date. This would mean a user will always see the correct list of results but also means that users in different timezones will see different report results.

#### Steps to Reproduce

1. Create a new report on calls with date created = Last Month and add date created to the columns
2. Switch TZ to Arizona
3. Modifiy a call so it's date created is 23:00 on the last day of two months ago (i.e when running the report in March created date should be 31st Jan 23:00
4. Check the report - the call that is outside of the period will appear in the results and the date created will show as outside of the conditions.

#### Context

Trying to create a report that is viewed by users in a different timezone.

#### Your Environment

* SuiteCRM Version used:
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Firefox
* Environment name and version (e.g. MySQL, PHP 7): MySQL, PHP7
* Operating System and version (e.g Ubuntu 16.04): Ubuntu 20.04

JimMackin:
## Description

This updates the AOR Utils Report Periods to include the users Timezone

## Motivation and Context

This fixes issue #9551. Note that, although this does fix #9551 it does slightly change behaviour - different users will may see different results from a report depending on their Timezone.

## How To Test This

- Create a report as described in #9551 - there should no longer be any records that fall outside the date period.

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOR_Reports/aor_utils.php
+++ b/modules/AOR_Reports/aor_utils.php
@@ -324,7 +324,7 @@ function getPeriodDate($date_time_period_list_selected)
     // set time to 00:00:00
     $datetime_period = $datetime_period->setTime(0, 0, 0);
 
-    $datetime_period->add(DateInterval::createFromDateString($timedate->getUserUTCOffset().' minutes'));
+    $datetime_period->sub(DateInterval::createFromDateString($timedate->getUserUTCOffset().' minutes'));
 
     return $datetime_period;
 }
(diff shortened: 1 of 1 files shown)


jack7anderson7 (approved):

clemente-raposo (approved):
LGTM

PR #9581, opened 2022-05-11
Fix #3157 - Add default option to enable session_gc

### LINKED ISSUE #3157: SuiteCRM fails to adequately housekeep session files potentially resulting in server failure

clemente-raposo:
## Description

- Enable php session_gc by default
- Add config to not enable it
- Add config to set values for gc_probability and gc_divisor

## Motivation and Context

- When using custom `session_dir`, the session files aren't cleaned up.
- see #3157 

## How To Test This

- on config.php, change `session_dir` to something different from the default
- Login in the app
- Check that session files were added to the directory defined in `session_dir`
- logout from the session
- check if some of the session files were removed. You may need to refresh the page some times. (the session garbage collector is done randomly with a probability of 1/100)
- In config.php, change the value of `enable_session_gc` to false.
- Check that after login in and logging out the session files aren't removed 

- see #3157 

## Types of changes

### Final checklist

### DIFF
--- a/include/entryPoint.php
+++ b/include/entryPoint.php
@@ -161,6 +161,15 @@ if (empty($GLOBALS['installing'])) {
         }
     }
 
+    $sessionGCConfig = $sugar_config['session_gc'] ?? [];
+    if (!isset($sessionGCConfig['enable']) || isTrue($sessionGCConfig['enable'])) {
+        $gcProbability = $sessionGCConfig['gc_probability'] ?? 1;
+        $gcDivisor = $sessionGCConfig['gc_divisor'] ?? 100;
+
+        ini_set('session.gc_probability', $gcProbability);
+        ini_set('session.gc_divisor', $gcDivisor);
+    }
+
     if (!empty($sugar_config['session_dir'])) {
         session_save_path($sugar_config['session_dir']);
     }
--- a/include/utils.php
+++ b/include/utils.php
@@ -547,6 +547,11 @@ function get_sugar_config_defaults(): array
             'min_cron_interval' => 30, // minimal interval between cron jobs
         ],
         'strict_id_validation' => false,
+        'session_gc' => [
+            'enable' => true,
+            'gc_probability' => 1,
+            'gc_divisor' => 100,
+        ]
     ];
 
     if (!is_object($locale)) {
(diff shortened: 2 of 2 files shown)


johnM2401 (approved):
Looks good to me

Appears to work according to the Divisor & Probability calculated chance, as well as the max life limit set in php.ini

PR #9582, opened 2022-05-11
Fix #9437 default cookie path

### LINKED ISSUE #9437: Generating unnecessary cookies due to wrong default cookie path
For a few months we were experiencing forced logouts from our SuiteCRM instances, we explain this case here: https://community.suitecrm.com/t/cookies-amount-for-each-suitecrm-installation/81535

Due to a bug in the code, SuiteCRM is generating unnecessary cookies, each SuiteCRM installation can generate up to 20-30 or 40 cookies. If you work with just one installation in each subdomain, that isn't a problem. But if you have more than 5 instances inside the same domain you are closely hitting the "Max Cookie Limit per Domain" that the browsers implements (Chrome 180 cookies http://browsercookielimits.iain.guru/). When this limit is reached, the browser will do some "cleaning" of cookies of that domain, and the user will be most likely kicked out from the SuiteCRM session.

This is happening since the merge of the PR https://github.com/salesagility/SuiteCRM/pull/7523 . Although not sure that this might be the best solution.

We reverted changes in our systems since August 2021, and it works flawlesly. Here a screenshot of the cookies comparing with SuiteCRM demo and SuiteCRM 8 demo:

Selection_402
Selection_403
Selection_404

As you can see, demo.suiteondemand.com is generating several cookies with same name, but different paths. demo.sinergiacrm.org is generating one cooking for each, pointing always to "/" path:
Selection_405

#### Issue
SuiteCRM generates unnecessary cookies. This might potencially cause reaching the Max cookies per Domain limit of the browsers.

#### Expected Behavior
Just one cookie per type should be created, all within same path "/"

#### Actual Behavior
For each path, it creates a different cookies, but using the same name. This is inconsistent.

#### Possible Fix
If the "path" parameter of PHP function setcookie is set to "null", each of the cookies will be generated into different browsers cookie paths. We can fix this forcing the "/" path in all cases. 

We implemented it this way:

https://github.com/salesagility/SuiteCRM/blob/0201c36b1468a16eb89218e7c798cf0ce2adac5c/include/MVC/SugarApplication.php#L429
Replace: 
```setcookie('sugar_user_theme', $theme, time() + 31536000, "/", null, isSSL(), true); // expires in a year ```
Adding "/" parameter, we force using the root of the browser cookie path for each of the cookies generated by SuiteCRM

https://github.com/salesagility/SuiteCRM/blob/0201c36b1468a16eb89218e7c798cf0ce2adac5c/include/MVC/SugarApplication.php#L787
Replace:
```$path = "/",```
In this way, each setCookie() function call that doesn't specify the path, will be always using the root path.

#### Steps to Reproduce
1. Login into any SuiteCRM instance
2. Browse within the CRM
3. Check the cookies for that domain

#### Context
This was a very frustrating situation, when the session ends unexpectedly. We have implemented this solution in August 2021, and so far no troubles.

#### Your Environment

clemente-raposo:
## Description

- Change cookie_path handling
    - in setCookie if no cookie_path value provided
        - use session.cookie_path if defined
        - Otherwise fallback to `/`
- Use application setCookie across the app
    - Replace direct php setcookie calls with application setCookie
 - see #9437 

## Motivation and Context

- Cookie path set was not set to `/`
- see #9437 

## How To Test This

- Go to app
- Clear all cookies
- Refresh. Check cookies and path
- Login. Check cookies and path
- Navigate through the app. Check cookies and path

## Types of changes

### Final checklist

### DIFF
--- a/include/MVC/SugarApplication.php
+++ b/include/MVC/SugarApplication.php
@@ -426,7 +426,7 @@ class SugarApplication
         }
 
         if (!is_null($theme) && !headers_sent()) {
-            setcookie('sugar_user_theme', $theme, time() + 31536000, null, null, isSSL(), true); // expires in a year
+            self::setCookie('sugar_user_theme', $theme, time() + 31536000, null, null, isSSL(), true); // expires in a year
         }
 
         SugarThemeRegistry::set($theme);
--- a/include/MVC/SugarApplication.php
+++ b/include/MVC/SugarApplication.php
@@ -800,6 +800,15 @@ class SugarApplication
             }
         }
 
+        $defaultCookiePath = ini_get('session.cookie_path');
+        if ($path === null) {
+            if(empty($defaultCookiePath)) {
+                $path = '/';
+            } else {
+                $path = $defaultCookiePath;
+            }
+        }
+
         if (!headers_sent()) {
             setcookie($name, $value, $expire, $path, $domain, $secure, $httponly);
         }
--- a/include/SubPanel/SubPanelTiles.php
+++ b/include/SubPanel/SubPanelTiles.php
@@ -195,7 +195,7 @@ class SubPanelTiles
         $default_div_display = 'inline';
         if (!empty($sugar_config['hide_subpanels_on_login'])) {
             if (!isset($_SESSION['visited_details'][$this->focus->module_dir])) {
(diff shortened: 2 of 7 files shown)


mattlorimer (approved):
Lgtm

PR #9610, opened 2022-06-12
AOW Conditions do not work on_save action

gunnicom:
When saving a record AOW does call check_valid_bean in the logic hook. So the bug does not show in the scheduler.
In line 828 following line does not compare to "false" and as of that if compare_condition returns "0" it interprets it as "false"
 if (!($this->compare_condition($field, $value, $condition->operator))) {

In compare_conditions we give back the result of strpos. This is "0" if the comparision is true, but for the first char.
Example:
Fieldvalue: Myspecialfieldname
Condition: Starts_With "M"
Should be true, Record should run the Workflow, but does not run because strpos gives back "0" interpreted as false.

On a side note: I do not get on first sight why Starts_with uses strrpos and not strpos.

### DIFF
--- a/modules/AOW_WorkFlow/AOW_WorkFlow.php
+++ b/modules/AOW_WorkFlow/AOW_WorkFlow.php
@@ -864,9 +864,9 @@ class AOW_WorkFlow extends Basic
             case "Less_Than":  return $var1 <  $var2;
             case "Greater_Than_or_Equal_To": return $var1 >= $var2;
             case "Less_Than_or_Equal_To": return $var1 <= $var2;
-            case "Contains": return strpos($var1, $var2);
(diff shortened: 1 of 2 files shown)


gunnicom:
Related 3 years old and still open: 
https://github.com/salesagility/SuiteCRM/issues/7030
https://github.com/salesagility/SuiteCRM/pull/7042

gunnicom:
Renamed my branch, Github does not like that, and closes it. Reopened.

jack7anderson7:
Hi @gunnicom,

Thanks you for the contribution.

However, closing this as the fix has been merged via another [PR 9614](https://github.com/salesagility/SuiteCRM/pull/9614)

PR #9710, opened 2022-08-01
Fix #8963 - Script tags evaluated twice when AJAX UI is enabled

### LINKED ISSUE #8963: Script tags evaluated twice when AJAX UI is enabled
#### Issue
We have a number of custom modules which add additional script tags into the page during load but have noticed that since upgrading from 7.11.15 these scripts (and many of the core ones) are being eval'd twice.

#### Expected Behavior
Scripts tags should be processed once.

#### Actual Behavior
Scripts tags are processed twice.

#### Possible Fix
The bug seems to have been introduced with the changes to 7.11.16 for the YUI removal.
In /jssource/src_files/include/javascript/sugar3.js, Sugar.util.evalScript() was switched from YUI.use() to $.ajax() but I think the duplicate run issue has been masked by the core files having additional code to prevent re-runs. 

#### Steps to Reproduce
This is the simplest explanation although it's a global issue and not related to only the Accounts module.
1. Insert a console.log() statement into "/modules/Accounts/Account.js"
2. Create/edit any an account through the AJAX UI
3. Notice log statement appears more than once

#### Context
High priority perhaps, we can manage our own stuff and implement a workaround but I'm sure there are plenty of other addins out there that are suddenly showing issues.

#### Your Environment
* SuiteCRM Version used: 7.11.18
* Browser name and version: Chrome 86.0.4240.193 (64-bit)

SinergiaCRM:
- Closes #8963

Credits to @FR-JS from the issue #8963

## Description

After removing and replacing YUI3 (https://github.com/salesagility/SuiteCRM/blob/3f332b87ce496b8887e359e205948b234f15911d/jssource/src_files/include/javascript/sugar_3.js), there was included an issue when loading JS files in views that use AJAX UI. This issue makes the custom JS files, loading twice. 

In this PR we remove the line that is causing the issue, as suggested in https://github.com/salesagility/SuiteCRM/issues/8963#issuecomment-775894649

## Motivation and Context

When loading JS code in a module, it can be problematic to load it twice in the same view.

## How To Test This

1- Add a console.log at the begining of the file modules/Accounts/Account.js (ex: console.log('test');
2- Open the browser console and navigate to an Edit or Detail view of an Account
3- Check in the browser console that the console.log is displayed twice.

## Types of changes

### Final checklist


cyb456:
thanks

PR #9769, opened 2022-09-14
Fix #9768 - Disable relative URLs in TinyMCE

### LINKED ISSUE #9768: Disable relative URLs in TinyMCE
#### Issue
When inserting, in TinyMCE or Mozaik, a link that contains the schema and domain of the site itself, both the schema and domain are deleted on save the link from the Create Link popup view.

Example: When creating the link https://midomain.org/index.php it will be transformed into index.php

DesactivateRelativeURLs

This causes an error when clicking the link in the received mail because the URL does not have the schmea and the domain

#### Expected Behavior
The link does not lose the schema and the domain

#### Actual Behavior
The link loses the schema and the domain

#### Possible Fix
Add the following code in the TinyMCE configuration 
```
relative_urls: false,
remove_script_host : false,
```

in both modules, Campaigns and Email:
- include/SuiteEditor/SuiteEditorConnector.php
- modules/Emails/include/ComposeView/EmailsComposeView.js

To reproduce the solution in Emails you also have to add the following configuration EmailsComposeView.js:
```
plugins: ["link"],
toolbar: ['fontselect | fontsizeselect | bold italic underline | forecolor backcolor | styleselect | outdent indent | link'],
```

This code resolves the issue in TinyMCE and Mozaik

#### Steps to Reproduce
1. Create a Campaign an go to Templates step
2. Open the popup view and insert a link with the URL of the site itself in the body of the email.
3. Open the popup view of the newly created link and check that the schema and domain have been lost

#### Context
Unable to link CRM content in emails to send

#### Your Environment
SuiteCRM Version used: Version 7.12.6
Browser name and version: Chrome Versión 100.0.4896.88 (Build oficial) (64 bits)
Environment name and version: MySQL, PHP 7.4
Operating System and version: Ubuntu 20.04

SinergiaCRM:
## Description
Solves #9768 

As described in the issue, the PR adds the necessary configuration in TinyMCE to disable relative URLs and not remove the scheme and domain from the URL when creating a link.

## Motivation and Context
Being able to link to CRM content in emails.

## How To Test This
1. Create a campaign and go to the Templates step.
2. Open the popup view and insert a link to the site's URL in the body of the email.
3. Open the popup view of the newly created link and verify that the scheme and domain have not been removed.

## Types of changes

### Final checklist

### DIFF
--- a/modules/Emails/include/ComposeView/EmailsComposeView.js
+++ b/modules/Emails/include/ComposeView/EmailsComposeView.js
@@ -1387,7 +1387,8 @@
   $.fn.EmailsComposeView.defaults = {
     "tinyMceOptions": {
       menubar: false,
-      toolbar: ['fontselect | fontsizeselect | bold italic underline | forecolor backcolor | styleselect | outdent indent'],
+      plugins: ['link'],
+      toolbar: ['fontselect | fontsizeselect | bold italic underline | forecolor backcolor | styleselect | outdent indent | link'],
       formats: {
         bold: {inline: 'b'},
         italic: {inline: 'i'},
@@ -1396,6 +1397,8 @@
       convert_urls: true,
       relative_urls: false,
       remove_script_host: false,
+      relative_urls: false,
+      remove_script_host : false,      
     }
   };
 }(jQuery));
--- a/include/SuiteEditor/SuiteEditorConnector.php
+++ b/include/SuiteEditor/SuiteEditorConnector.php
@@ -78,6 +78,8 @@ class SuiteEditorConnector
                 height : '480',
                 plugins: ['code', 'table', 'link', 'image'],
                 toolbar: ['fontselect | fontsizeselect | bold italic underline | forecolor backcolor | styleselect | outdent indent | link image'],
+                relative_urls: false,
+                remove_script_host : false,
             }"
         );
     }


modules/Emails/include/ComposeView/EmailsComposeView.js
```diff
@@ -1396,6 +1397,8 @@
       convert_urls: true,
       relative_urls: false,
       remove_script_host: false,
+      relative_urls: false,
```
jack7anderson7, line 1400: Hi @SinergiaCRM I believe you have some duplicate code here, the lines you have added are already there above the lines you added.

ManuSinergiaCRM: Hi @jack7anderson7, you are right. 

We have also seen that the proposed solution is not good since now the schema and the domain are added to any URL that is indicated, even in the tracking URLs.

We are going to close this PR for being wrong and open another one with the following solution and configuration:

`convert_urls=false`

Thanks!

SinergiaCRM: Hi @jack7anderson7, you are right.

We have also seen that the proposed solution is not good since now the schema and the domain are added to any URL that is indicated, even in the tracking URLs.

We are going to close this PR for being wrong and open another one with the following solution and configuration:

`convert_urls=false`

Thanks!

SinergiaCRM: https://github.com/salesagility/SuiteCRM/pull/9777

PR #9770, opened 2022-09-15
Fix #99568: Ignore int len when comparing vardefs in newer MySQL versions

### LINKED ISSUE #9568: MySQL8: Queries in QR&R never take effect (due to integer display lengths deprecated)
#### Issue
MySQL is deprecating using any sort of length for all kinds of integer fields (shortint, integer, etc). It just changes all such fields to `int` without any specified length.

This causes the QR&R to always detect an issue with lengths if the vardefs specifies them. Just an example of one field:
``` 
/* Table : acl_actions */
/*COLUMNS*/
/*MISMATCH WITH DATABASE - aclaccess -  ROW [name] => 'aclaccess'  [type] => 'int'  [len] => ''  */
/* VARDEF - aclaccess -  ROW[name] => 'aclaccess'  [vname] => 'LBL_ACCESS'  [type] => 'int'  [len] => '3'  [reportable] => '1'  [comment] => 'Number specifying access priority; highest access "wins"'  [dbType] => 'int'  [required] => ''  */
/* INDEXES */
ALTER TABLE acl_actions   modify COLUMN `aclaccess` int(3)  NULL ;
```

Running the query directly in phpMyAdmin causes this warning to be issued:

`Warning: #1681 Integer display width is deprecated and will be removed in a future release.` 

Even after the UPDATE query, the field definition will **not** be modified to `int(3)`, it will stay as just `int`.

#### Expected Behavior
QR&R "Execute query" should only be required once.

#### Actual Behavior
No matter how often the query is run, the database stays in a state that will provoke the same query to be proposed again and again.

#### Possible Fix
Somewhere in DBManager making the code ignore this case?

#### Steps to Reproduce
1. Upgrade a previously existing instance to MySQL v8
2. Go to QR & R
3. Execute Query to solve vardef mismatches - note the presence of many queries just fixing lengths of integers (all kinds of integers)
4. Go to the QR & R again and the query is re-proposed...

#### Your Environment
* SuiteCRM Version used: 7.12.5
* Environment name and version: PHP 7.3, MySQL 8.0
* Operating System and version: Ubuntu 20.04

JimMackin:
## Description

Add a check on the MYSQL version and ignore int lengths in versions >= 8.0.19
## Motivation and Context

This fixes #9568 - namely some DB changes always appearing in the QR&R log due to MySQL ignoring int lengths. https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-19.html
## How To Test This

- Run a QR&R on an instance using MySQL > 8.0.19. Plenty of changes should appear (add an int field with a length in studio if not
- Run the DB changes - note that this has no effect on the QR&R results
- Apply fix - run QR&R and the changes will be gone

## Types of changes

### Final checklist

### DIFF
--- a/include/database/MysqliManager.php
+++ b/include/database/MysqliManager.php
@@ -427,4 +427,21 @@ class MysqliManager extends MysqlManager
     {
         return function_exists("mysqli_connect") && empty($GLOBALS['sugar_config']['mysqli_disabled']);
     }
+
+    public function compareVarDefs($fielddef1, $fielddef2, $ignoreName = false)
+    {
+        /**
+         * Int lengths are ignored in MySQL versions >= 8.0.19 so we need to ignore when comparing vardefs.
+         */
+        if($fielddef1['type'] == 'int') {
+            $db_version = $this->version();
+            if (!empty($db_version)
+                && version_compare($db_version, '8.0.19') >= 0
+                && strpos($db_version, "MariaDB") === false
+            ) {
+                unset($fielddef2['len']);
+            }
+        }
+        return parent::compareVarDefs($fielddef1, $fielddef2, $ignoreName);
+    }
 }


sadnblueish:
This fixed the issue for me. Running SuiteCRM 7.12.7 on MySQL 8.

Issue #9680, opened 2022-07-21
The related record selected from subpanel edit view is not updated


SinergiaCRM:
#### Issue
Taking the 1-N relationship between the Contacts and Projects modules as an example, when a record (a project) is edited from the corresponding subpanel (projects subpanel in a contact) and the value of the related record is changed (the person changes) , after clicking on save the record:

- The change on the related field is not applied (the person has not changed).
- The rest of the fields of the record (project) have changed.

#### Expected Behavior
The related record should have changed and contain the record indicated in the edit form

#### Actual Behavior
The related record does not save the record indicated in the edit form

#### Possible Fix

In file _themes/SuiteP/include/EditView/header.tpl_, change line 84:

`<input type="hidden" name="relate_id" value="{$smarty.request.return_id}">`

to the following code:

`<input type="hidden" name="relate_id" value="{$fields.id.value}">`

#### Steps to Reproduce
1. Create two contacts.
2. Go to the detail view of one of them and create a new project from the subpanel or edit an existing one.
3. Change various fields, including the person.
4. Check that the related person is still the initial one and the rest of the project fields have been modified

#### Context
Failure to update related field causes data inconsistencies

#### Your Environment
SuiteCRM Version used: Version 7.12.6
Browser name and version: Versión 100.0.4896.88 (Build oficial) (64 bit
Environment name and version: MySQL, PHP 7
Operating System and version: Ubuntu 20.04

PR #9779, opened 2022-09-27
Fix salesagility#9680 - The related record selected from subpanel edit view is not updated

SinergiaCRM:
Solves #9680 

## Description
Assign the record indicated in the subpanel form instead of the main record (record from which the subpanel was opened) as they may be different.

We handle the two types of relationships that we have identified as failing so far:
- Relationships with ID in the _ida field of the relationship different from the parent record
- Special relationships like member_accounts where parent_id is not equal to parent record ID

We have created this code at the end of the function `set_relationship_info()`, although due to the fact that we are not experts in the core of SuiteCRM and the existing complexity in relationship management, we have the doubt if the proposed code can be more efficient in another part of the function or of SuiteCRM code

## Motivation and Context
Failure to update related field causes data inconsistencies

## How To Test This
1. Create two accounts.
2. Go to the detail view of one of them, create a new contact or a new opportunity from the subpanel and change the related contact. 
3. Check that the contact has been updated with the contact indicated in the edit form
4. Edit the newly created contact or opportunnity from the subpanel and change the related contact.
5. Check that the contact or opportunnity has been updated with the account indicated in the edit form.

## Types of changes

### Final checklist

### DIFF
--- a/data/SugarBean.php
+++ b/data/SugarBean.php
@@ -2841,6 +2841,18 @@ class SugarBean
             }
         }
 
+        // Assign the record indicated in the subpanel form instead of 
+        // the parent record (record from which the subpanel was opened) since they might be different
+        if (!empty($new_rel_id) 
+            && ( // Relationships with an ID in the _ida field of the relationship different from the parent record
+                 (!empty($this->{$new_rel_link}) && is_string($this->{$new_rel_link}) && $new_rel_id != $this->{$new_rel_link})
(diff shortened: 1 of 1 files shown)


SinergiaCRM:
Hi, this PR required to comply with the license/cla although we did not understand the reason why it will remain in that status... Once the PR is closed and reopened, that requirement is already fulfilled, so we understand that we can continue with the resolution of the issue.

PR #9806, opened 2022-10-26
Fix #9805 - Use timezone offset for datetime only

### LINKED ISSUE #9805: Report Period conditions use timezone offset for Date field types
#### Issue

When a User has set a different timezone and they attempt to report on Date fields, the CRM queries the database using a datetime value with an offset for the timezone. This provides incorrect results as the query will assume a date type has a midnight timestamp.
#### Expected Behavior
When the the user has set a timezone that is not the same as the database timezone.
The Date Period condition will return the correct results for "Today", "Yesterday" etc

#### Actual Behavior

When the the user has set a timezone that is not the same as the database timezone.
Date Period conditions of "Yesterday" will return dates from "Today" if the timezone is behind the db timezone.

#### Possible Fix

If the field type is "datetime" then provide an offset value, else use the midnight value for "date"

#### Steps to Reproduce

1. Set your timezone to America/Chicago in your profile.
2. Create a date field in Contacts
3. Create a Contacts record with the newly created date field set to today.
4. Create a Report on Contacts with a condition on the newly created date field equal to the period "Today"
5. The Contacts record will not be returned in the report

#### Your Environment

* SuiteCRM Version used: SuiteCRM 7.12.7
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Firefox 105
* Environment name and version (e.g. MySQL, PHP 7): PHP7.4
* Operating System and version (e.g Ubuntu 16.04): Ubuntu 20.0.5

ojs87:
## Description

See https://github.com/salesagility/SuiteCRM/issues/9805

When a User has set a different timezone and they attempt to report on Date fields, the CRM queries the database using a datetime value with an offset for the timezone. This provides incorrect results as the query will assume a date type has a midnight timestamp.

## Motivation and Context
Users with custom timezones may have incorrect report results.

## How To Test This
1. Set your timezone to America/Chicago in your profile.
2. Create a date field in Contacts
3. Create a Contacts record with the newly created date field set to today.
4. Create a Report on Contacts with a condition on the newly created date field equal to the period "Today"
5. The Contacts record will not be returned in the report

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOR_Reports/aor_utils.php
+++ b/modules/AOR_Reports/aor_utils.php
@@ -226,7 +226,7 @@ function getConditionsAsParameters($report, $override = array())
  * @param $date_time_period_list_selected
  * @return DateTime
  */
-function getPeriodDate($date_time_period_list_selected)
+function getPeriodDate($date_time_period_list_selected, $type = '')
 {
     global $sugar_config, $timedate;
     $datetime_period = new DateTime();
@@ -324,7 +324,9 @@ function getPeriodDate($date_time_period_list_selected)
     // set time to 00:00:00
     $datetime_period = $datetime_period->setTime(0, 0, 0);
 
-    $datetime_period->sub(DateInterval::createFromDateString($timedate->getUserUTCOffset().' minutes'));
+    if($type === 'datetime') {
+        $datetime_period->sub(DateInterval::createFromDateString($timedate->getUserUTCOffset() . ' minutes'));
+    }
 
     return $datetime_period;
 }
@@ -334,7 +336,7 @@ function getPeriodDate($date_time_period_list_selected)
  * @param $date_time_period_list_selected
  * @return DateTime
  */
-function getPeriodEndDate($dateTimePeriodListSelected)
+function getPeriodEndDate($dateTimePeriodListSelected, $type = '')
 {
     global $timedate;
     switch ($dateTimePeriodListSelected) {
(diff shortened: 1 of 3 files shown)


jack7anderson7 (approved):

modules/AOR_Reports/aor_utils.php
```diff
@@ -226,7 +226,7 @@ function getConditionsAsParameters($report, $override = array())
  * @param $date_time_period_list_selected
  * @return DateTime
  */
-function getPeriodDate($date_time_period_list_selected)
+function getPeriodDate($date_time_period_list_selected, $type)
```
clemente-raposo, line 229: Hi @ojs87, thank you for your PR. Do we want to make `$type` mandatory? should we define it like the following?

```
function getPeriodDate($date_time_period_list_selected, $type = '')
```

ojs87: Hi @clemente-raposo , thanks for your feedback. I've updated the variable to make it non-mandatory as suggested!

clemente-raposo (dismissed):
LGTM

johnM2401:
Hey @ojs87 !

I've had a look and it seems the issue is partly resolved for Today/Yesterday
However, the issue is still present for the other Period Parameters

See:

As an example using "This/Last Year"

This contact has a Date set on the first day of the year:
image


However, they show in both "This Year" and "Last Year" periods:
image

image


<br>
<br>


Additionally, the date value always seems to show the previous day.

<br>

For Example, using "Today", it is set as 2023 - 5th of June
image

But the report displays June 4th for the "Today" period:
image


<br>

Could you have another look, when you get a chance?
Thanks!

ojs87:
Hi John,

I have updated this PR with fixes for the issues you found.
Reports should now display the correct Date on the report results.
The record should now only appear in the correct Period for the date.

clemente-raposo (approved):

PR #9822, opened 2022-11-16
Module header in search results untranslated

gunnicom:
Module header in search results untranslated


cyb456:
Can you just make this properly so it can be accepted?

It's such a nobrainer obvious problem and fix I don't understand why this has not been done, even if not from this PR.

PR #9823, opened 2022-11-16
Fix #9823 - Wrong link to related objects in search result

gunnicom:
In search result some related objects get the same id in the link as the found object. As example you find object of type custom type 
mmm_mymodule id 12345678
related to mmm_realtedmodule 87654321

Search results gives two links if related module is in listview

index.php?module=mmm_mymodule &id=12345678
index.php?module=mmm_realtedmodule &id=12345678

So the second link will lead to an unexistent object. I do not know exactly why this happens only in some cases. Maybe the relation is defined in the mmm_relatedmodule. Maybe it is because it is custom module.

The search result page anyways is totally butchered by someone. Checkboxes in listview get displayed as "0" and "1" etc.

### DIFF
--- a/lib/Search/SearchResults.php
+++ b/lib/Search/SearchResults.php
@@ -209,7 +209,12 @@ class SearchResults
     {
         $relField = $idName;
         if (isset($obj->$link)) {
-            $relId = $obj->$link->getFocus()->$relField;
+            $linkedbeans=$obj->$link->getBeans();
+            if(count($linkedbeans)==1){
+                $relId = array_keys($linkedbeans)[0];
+            }else{
+                $relId = $obj->$link->getFocus()->$relField;
+            }
             if (is_object($relId)) {
                 if (method_exists($relId, "getFocus")) {
                     $relId = $relId->getFocus()->id;


jack7anderson7:
Hi @gunnicom, 

Thank you for your Pull Request and for the other PR's you have raised, which we very much appreciate.

However, we have noticed that when raising a PR, you are removing/ ignoring the required PR template, making it difficult for us to clearly understand the reason for the PR.  To clarify, the goal of the template is to show a formatted and in depth explanation of the Pull Request that all contributors are asked to complete.

We would therefore ask that any future PRs raised, are based on the template, where each section is completed as requested.    

Thanks, Jack

gunnicom:
@jack7anderson7 Yes, I do not use the template, because why should I waste my time, when my experience is, that PRs and issues get not recognised. Look at the open PRs and issues. And then I saw another day, that even a PR of me was merged but never got into the code because in the same merge the code was overwritten.
So if you want to make Suite better, take my work. If you don't, then just close my PRs. I will not put hours of work into Suite anymore if I do not see the PRs and issues raised get recognized.

mattlorimer:
Alternative pull request merged

Issue #9788, opened 2022-10-12
Database Failures in Subpanel (MySQL Error)


pstevens71:
<!--- Provide a general summary of the issue in the **Title** above -->
<!--- Before you open an issue, please check if a similar issue already exists or has been closed before. --->
<!--- If you have discovered a security risk please report it by emailing security@suitecrm.com. This will be delivered to the product team who handle security issues. Please don't disclose security bugs publicly until they have been handled by the security team. --->
<!--- Please be aware that as of the 31st January 2022 we no longer support 7.10.x.
New issues referring to 7.10.x will only be valid if applicable to 7.12.x and above.
If your issue is still applicable in 7.12.x, please create the issue following the template below -->
#### Issue
<!--- Provide a more detailed introduction to the issue itself, and why you consider it to be a bug -->
Database error in subpanels with the error:  ERROR 1250 (42000): Table ‘…’ from one of the SELECTs cannot be used in global ORDER clause.  The issue pertains to use the Table name in the sort clause.  This seemed to work fine until today (Oct. 10. 2022).  So must have been a recent update in mySQL that no longer permits this syntax.  I was able to get the subpanel to display contacts again by sorting on a different column.  Example below (screenshots) is in the Cases module and Contacts Subpanel.  However, this issue is throughout  SuiteCRM and not just this module, and not just contacts.
![sort problem](https://user-images.githubusercontent.com/29544935/195383235-d940e040-4b12-495b-8b7f-bb7f0afa4496.png)
When I sort on "Accounts" it works:
![working sort problem](https://user-images.githubusercontent.com/29544935/195383318-aff4c101-e63a-44e1-8fd7-2713a6eb00e2.png)

<!--- Ensure that all code ``` is surrounded ``` by triple back quotes. This can also be done over multiple lines -->

#### Expected Behavior
<!--- Tell us what should happen -->
The contact in the subpanel should display without error.

#### Actual Behavior
<!--- Tell us what happens instead -->
Contacts in subpanel display with database error.
[...]

PR #9843, opened 2022-12-06
Fix salesagility#9788 - Database Failures in Subpanels throughout SuiteCRM (MySQL Error)

hbartel:
Fix Issue https://github.com/salesagility/SuiteCRM/issues/9788

Description
Changes implemented in MySQL 8.0.31 cause SQL queries that pull data for subpannels to fail. If the outer ( ) is removed from the query, it will complete normally.

The code that adds the outer ( ) to the query is located on lines 876 and 930 of Suite CRM/data/SugarBean.php. By removing these lines the query if fixed.

876: $final_query = '(' . $tmp_final_query . ')';
930: $query = '(' . $query . ')'; '''

Motivation and Context
If the MySQL environment is updated to MySQL 8.0.31 the subpannels that depend on these SQL queries will fail.

How To Test
Update MySQL to 8.0.31 on host server
Create Account with linked subpanel data (contacts, opportunities, leads, ect)
Subpanels must have data for the error to occur.
Open Account detail page & suitecrm.log to see error.
Remove specified lines and test

## Types of changes

### Final checklist

### DIFF
--- a/data/SugarBean.php
+++ b/data/SugarBean.php
@@ -873,7 +873,7 @@ class SugarBean
                     $final_query .= ' UNION ALL ( ' . $tmp_final_query . ' )';
                 } else {
                     $final_query_rows = '(' . $parentbean->create_list_count_query($tmp_final_query, $parameters) . ')';
-                    $final_query = '(' . $tmp_final_query . ')';
+                    $final_query = $tmp_final_query;
                     $first = false;
                 }
             }
@@ -927,7 +927,6 @@ class SugarBean
                     $query = ' UNION ALL ( ' . $query . ' )';
                     $final_query_rows .= " UNION ALL ";
                 } else {
-                    $query = '(' . $query . ')';
                     $first = false;
                 }
                 $query_array = $subquery['query_array'];
@@ -3511,7 +3510,7 @@ class SugarBean
         if (isset($_SESSION['show_deleted'])) {
             $show_deleted = 1;
         }
-        
+
         $query = $this->create_new_list_query(
             $order_by,
             $where,
@@ -4472,7 +4471,7 @@ class SugarBean
         if (isset($_SESSION['show_deleted'])) {
             $show_deleted = 1;
         }
-        
+
         $query = $this->create_new_list_query($order_by, $where, array(), array(), $show_deleted, $offset);
 
         return $this->process_detail_query($query, $row_offset, $limit, $max, $where, $offset);
@@ -6159,10 +6158,10 @@ class SugarBean
 
         return $args->access
             && ACLController::checkAccess(
-                $this->module_dir, 
-                $args->view, 
-                $args->is_owner, 
-                $this->acltype, 
+                $this->module_dir,
+                $args->view,
+                $args->is_owner,
+                $this->acltype,
                 $args->in_group
             );
     }


jack7anderson7:
Hi @hbartel,

Thanks for your contribution!

Just on thing, could you please change your target branch from `master` to `hotfix-7.13.x`.

Thanks,
Jack

pgorod:
This REALLY needs to be merged ASAP.

But according to the discussion on the Issue, part of this fix si not correct yet.

Instead of commenting out this line:
```
//$final_query = '(' . $tmp_final_query . ')';
```
I think it should be left there, only without the brackets:
```
$final_query =  $tmp_final_query;
```

chris001:
Looks good to me...

pgorod:
@clemente-raposo, this PR should really not be delayed...

clemente-raposo (approved):
@pgorod @pstevens71 @hbartel @chris001   this is an issue in mysql, which has been fixed in version `8.0.33`.

The query that SuiteCRM is generating is valid.

We risk breaking other queries in the app because of a bug in a mysql version.

Feedback is welcome.


See: https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-33.html

> A table reference in an ORDER BY outside the parenthesized query block in which the table was used, and which query block had no LIMIT or ORDER BY of its own, raised an error. (Bug #34704011)
> References: This issue is a regression of: Bug #103954, Bug #32980249.



EDIT (reverting the following comment):


 ~Approving this PR. However, please have in mind:~
 
~It is very hard to predict the impact on other queries, this method is called from different places.~
 
~For the best of my knowledge the way the `final_query` is being used and then inserted in to other queries won't break. Tough its not a small risk.~ 
~@johnM2401 several different types of subpanels should be checked, more complex ones like activities and history and others with custom queries~

clemente-raposo:
@johnM2401 could we test this with mysql 8.0.33+ please?

clemente-raposo:
@pgorod @pstevens71 @hbartel @chris001

Some more data: Tested with 8.0.31, 8.0.34, 8.1.0, 8.2.0. Only 8.0.31 fails. Adding info about the tests below.

### Testing with 8.0.31

**docker setup**

image from mysql dockerhub
```
  mysql8031:
    image: mysql:8.0.31
    container_name: mysql8031
    restart: always
    ports:
      - "3310:3306"
    volumes:
      - ./mysql8031:/var/lib/mysql
    environment:
      - MYSQL_ROOT_PASSWORD=<pass>
      - MYSQL_DATABASE=<db>
      - MYSQL_USER=<user>
      - MYSQL_PASSWORD=<pass>
```

**adminer**

image

**result**

image


### Testing with 8.0.34

image from mysql dockerhub
```
  mysql80:
    image: mysql:8.0
    container_name: mysql80
    restart: always
    ports:
      - "3309:3306"
    volumes:
      - ./mysql80:/var/lib/mysql
    environment:
      - MYSQL_ROOT_PASSWORD=<pass>
      - MYSQL_DATABASE=<db>
      - MYSQL_USER=<user>
      - MYSQL_PASSWORD=<pass>
```

**adminer**

image

**result**
image


### Testing with 8.1.0

image from mysql dockerhub
```
  mysql8:
    image: mysql:8.1
    container_name: mysql8
    restart: always
    ports:
      - "3308:3306"
    volumes:
      - ./mysql8:/var/lib/mysql
    environment:
      - MYSQL_ROOT_PASSWORD=<pass>
      - MYSQL_DATABASE=<db>
      - MYSQL_USER=<user>
      - MYSQL_PASSWORD=<pass>
```

**adminer**
image


**result**
image

### Testing with 8.2.0

image from mysql dockerhub
```
  mysql82:
    image: mysql:8.2
    container_name: mysql82
    restart: always
    ports:
      - "3312:3306"
    volumes:
      - ./mysql82:/var/lib/mysql
    environment:
      - MYSQL_ROOT_PASSWORD=<pass>
      - MYSQL_DATABASE=<db>
      - MYSQL_USER=<user>
      - MYSQL_PASSWORD=<pass>
```

**adminer**

image

**result**
image

pstevens71:
I find too its also dependent on your sort.  Sometimes even with the fix above it fails.  (however the fix does solve alot of problems).

For example:
image

image

cripton:
We tested the mysql version 8.0.34 and it fails when a subpanel have a subpaneldefs setting to "sort_by" to a field that is not a column in the subpanel, so the query is formed with an ORDER BY to this field which is not in the fields generated in the Parenthesized Query, that were introduced in mysql 8.0.31: (https://dev.mysql.com/doc/refman/8.0/en/parenthesized-query-expressions.html).

Example:
(SELECT a, b, c FROM t) ORDER BY x;

The fix proposed by @hbartel probably fixes this particular problem.

osiris:
Hello everyone, on a _SuiteCRM_ `v8.5.1` instance we applied the
proposed patch by @hbartel and has been working without problems since
2024-02-16.

lukio:
Hello @mattlorimer @pgorod @chris001 @clemente-raposo @osiris

Why this PR was not yet been merged? The problem persist at SuiteCRM 7.15.x and of course on SuiteCRM-8.x 

Has been closed by @mattlorimer without any explanation.

If the problem was to change the branch to be merged, I believed that we can create a new PR and to be apply to hotfix. 

The matrix compatibility says that mysql-8.4 is supported, and until this change will not be merged, the matrix compatibility is not true.

lukio:
Hi @osiris @hbartel @chris001 @clemente-raposo @jack7anderson7 @mattlorimer @pgorod @cripton 

I open new PR with sames changes https://github.com/SuiteCRM/SuiteCRM/pull/10792. 

Please, comment inside this new PR.

Thanks!

PR #9862, opened 2022-12-21
 Fix #9861 - Error calculating the end date and time of a call

### LINKED ISSUE #9861: Error calculating the end date and time of a call
#### Issue
As we have seen, the condition that determines if the hour or the minutes of a call fails, so that if the value 0 or 00 is set in the hours or minutes, it is evaluated as empty and does not enter the if that it calculates the value of date_end field.
In this case, the end date of the call remains the same as its start date.

#### Expected Behavior
The call end date should calculate correctly regardless of the value entered for the call duration

#### Actual Behavior
The calculation is not produced correctly

#### Steps to Reproduce
1) In the calls module, add the end_date field to the edit view or detail view.
2) Create a record indicating in the duration field 0 in hours or minutes.
3) Check that the value of end_date does not add the duration to the start date and time of the call

#### Your Environment
* SuiteCRM Version used: 7.12.6

juanSTIC:
resolve #9861
## Description
When a call is saved, the value of the end_date field is not calculated correctly if any of the values of the duration field is 0 (hours or minutes), since in this case, the code flow does not enter the if that performs the calculation .
In this PR, the conditions are modified to evaluate whether or not to calculate end_date

## Motivation and Context

It is important to correctly calculate the end time of the call correctly

## How To Test This
In the calls module, add the end_date field to the edit view or detail view.
Create a record indicating in the duration field 0 in hours or minutes.
Check that the value of the end_date field has been correctly calculated
## Types of changes

### Final checklist

### DIFF
--- a/modules/Calls/Call.php
+++ b/modules/Calls/Call.php
@@ -174,7 +174,7 @@ class Call extends SugarBean
         global $timedate;
 
         if (!empty($this->date_start)) {
-            if (!empty($this->duration_hours) && !empty($this->duration_minutes)) {
+            if (!empty($this->duration_hours + $this->duration_minutes)) {
                 $td = $timedate->fromDb($this->date_start);
                 if ($td) {
                     $this->date_end = $td->modify(


SinergiaCRM:
Error. Invalid PR

PR #9866, opened 2022-12-21
Fix #9865 -  Tabs with validation error are not highlighted

### LINKED ISSUE #9865: Tab with validation error is not highlighted
#### Issue
In the edit view of any module, if there is a required field in a tab other than the first one and it is left blank, when trying to save the record, the tab is not highlighted and no information is displayed to the user.

This can cause confusion for the user who does not know where the error is coming from.

#### Expected Behavior

If the required field is not on the first tab and there is a validation error, the tab should be highlighted, indicating where the error is.

#### Actual Behavior
Tab is not highlighted.

#### Steps to Reproduce
1) Create a record in a module whose edit view contains more than one tab and which has some required field in a tab other than the first one.
2) Save the record. Check that the save is blocked without the user seeing the reason for the problem until they navigate to the affected tab.

#### Your Environment
* SuiteCRM Version used: 7.12.6

SinergiaCRM:
- resolve #9865
## Description
When trying to save a record from any module, which contains validations in tabs other than the first, the tabs are  highlighted so that the user can easily find the error.

To do this, the javascript code that provides this functionality and that previously depended on the YUI library, now obsolete, is modified.

## Motivation and Context
The objective is to facilitate the detection of errors and improve the user experience.

## How To Test
1) Create a record in a module whose edit view contains more than one tab and has some required field in a different tab than the first one.
2) Save the record. Check that the save is locked, but the affected tabs are highlighted

### Final checklist


jssource/src_files/include/javascript/sugar_3.js
```diff
@@ -799,22 +799,24 @@ function add_error_style(formname, input, txt, flash) {
...
-              if (inputsWithErrors.length == 1)
-                tabView.selectTab(i);
+      if (typeof window[formname + "_tabs"] != "undefined") {
+        if ($(".validation-message")) {
+          // Reset tab error styles
+          $("#EditView_tabs li a").removeAttr("style");
+          counter = 0;
+          $(".validation-message").each(function () {
+            // Get error validation tabs ids
+            CurrentValidationTabId = $(this).closest(".tab-pane-NOBOOTSTRAPTOGGLER").attr("id").split('-').slice(-1);
```
clemente-raposo, line 809: Hi @SinergiaCRM, 2 small thing:
- don't we need to use `var` here like `var currentValidationTabId`?
- Also the variable names should start with lowecase.

jssource/src_files/include/javascript/sugar_3.js
```diff
@@ -799,22 +799,24 @@ function add_error_style(formname, input, txt, flash) {
...
+            // Get tabs
+            $CurrentValidationTabA = $("#EditView_tabs a#tab" + CurrentValidationTabId);
```
clemente-raposo, line 811: Hi @SinergiaCRM, similar to the above. 

- don't we need to use `var` here like `var $currentValidationTabA`?
- Also the variable names should start with lowecase.

jssource/src_files/include/javascript/sugar_3.js
SinergiaCRM: Done

jssource/src_files/include/javascript/sugar_3.js
SinergiaCRM: Done

SinergiaCRM:
Hi @serhiisamko091184 ,

Can we finish validating this PR?

Issue #9818, opened 2022-11-15
Custom validation do not run on inline edit (7.12.7)


gunnicom:
#### Issue
If you define a custom validation in vardefs, like:
```
    'size' => '33',
    'validation' => 
    array (
      'type' => 'callback',
      'callback' => 'function(formname,nameIndex){
                           var fieldvalue=$("#" + formname + " input[name=" + nameIndex + "]").val();
                           var returncode=true;
                           if(fieldvalue.trim()!=fieldvalue){
                               add_error_style(formname,nameIndex,"Space at the start or end of value not allowed " + nameIndex); 
                               returncode= false;
                           }
                           return returncode;
			}',
    ),
```
It won't run on inline edt. It will though work on normal edit.

#### Expected Behavior
Validations should run also on inline edit.

#### Actual Behavior
Only "system" validations like validation if it is a number seems to run, but custom ones do not.

#### Possible Fix
x

#### Steps to Reproduce
1. Add custom validation code to vardefs.php of a field
2. Try to save in inline edit with invalid value
3. Value save although it should be invalid
4. Save the same value in normal edit view, validation works, won't save

#### Your Environment

* SuiteCRM Version used: 7.12.7
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Chrome 107
* Environment name and version (e.g. MySQL, PHP 7): MariaDB 10.2.x, PHP 7.4
* Operating System and version (e.g Ubuntu 16.04): Debian 11

gunnicom:
Easier to reproduce:
- Add an integer field via Studio
- Parameter the field with min value 0 and max value 10
- Deploy the field in the views
- Edit the object in edit mode: Validation works, you can not enter values greater than 10
- Edit the object via inline edit: Validation does not work. You can enter a value like 99

SuiteBot:
This issue has been mentioned on **SuiteCRM**. There might be relevant details there:

https://community.suitecrm.com/t/make-field-required-based-on-another-field/90250/7

PR #9868, opened 2022-12-22
#9818 disable inline edit for fields with validations to prevent faulty data

gunnicom:
## Description
Validations of fields do not work on inline edit so disable inline edit for them to prevent faulty data.

## Motivation and Context
Data quality

## How To Test This
See #9818 

## Types of changes

### Final checklist


pgorod:
Why don't validations work, isn't that the real bug we should be fixing? I see that inline edits call a function called getValidationRules, so I expect it was planned to work... although I can't confirm that I ever saw it working.

gunnicom:
@pgorod Of course, but I did not see why it does not work. I assume it was never tested when introducing the feature. I do not have the time and motivation now to look for the real issue so I think disabling is better than wrong data nobody is aware about.

PR #9881, opened 2023-01-02
Fix #9880 -  Error when importing currency fields with a decimal separator

### LINKED ISSUE #9880: Error when importing currency fields with a decimal separator set to a comma
#### Issue
If the decimal separator set in the user's settings is a comma “,” an error occurs when importing currency fields containing decimals. If this happens, during the import process the decimal separator is lost and the imported value is multiplied by 100.

#### Expected Behavior
The value of the field should be maintained regardless of the decimal separator used.

#### Actual Behavior
If a currency field with the value 1,12 is imported, the result will be 112.

#### Steps to Reproduce

1. Set the decimal separator to a comma “,” in the user's settings.
2. Import a record from the Opportunities module with the Opportunity Amount set to a decimal value, for example 1,12.
3. Verify that after the import, the Opportunity Amount field has been set to 112.

#### Your Environment
* SuiteCRM Version used: 7.12.8

SinergiaCRM:
resolve #9880
## Description
If the decimal separator is set to a comma "," in the user's settings, an error occurs when importing currency fields that include decimals because the decimal separator is removed and the result is the value of the field multiplied by 100.
The error was caused by the filter set for importing currency fields being incorrect, as it passes the same filter as float fields, when a currency field, after the currency symbol is removed, should follow the same treatment as a decimal field. This is what has been implemented in the solution.

## Motivation and Context
Prevent errors when importing currency fields with different types of decimal separators.

## How To Test
1. Set the decimal separator to a comma “,” in the user's settings.
2. Import a record from the Opportunities module with the Opportunity Amount set to a decimal value, for example 1,12.
3. Verify that after the import, the Opportunity Amount field has been set to 1,12.

## Types of changes

### Final checklist

### DIFF
--- a/include/SugarFields/Fields/Currency/SugarFieldCurrency.php
+++ b/include/SugarFields/Fields/Currency/SugarFieldCurrency.php
@@ -95,7 +95,14 @@ class SugarFieldCurrency extends SugarFieldFloat
         ) {
         $value = str_replace($settings->currency_symbol, "", $value);
         
-        return $settings->float($value, $vardef, $focus);
+        $value = str_replace($settings->num_grp_sep, "", $value);       
+
+        if (isset($vardef['len'])) {
+            // check for field length
+            $value = sugar_substr($value, $vardef['len']);
+        }
+
+        return $value;
     }
 
     /**


serhiisamko091184:
Hello, @SinergiaCRM ,
Your change appears to be defining `$value` twice, on line 96 and then on 98.

Thanks,
Serhii

juanSTIC:
> Hello, @SinergiaCRM , Your change appears to be defining `$value` twice, on line 96 and then on 98.
> 
> Thanks, Serhii

Yes, the assignment of `$value` on line 98 is modifying the value obtained on line 96, which in turn modifies the value of the argument received in the function, I think that should not be a problem.

include/SugarFields/Fields/Currency/SugarFieldCurrency.php
```diff
@@ -95,7 +95,14 @@ public function importSanitize(
...
         
-        return $settings->float($value, $vardef, $focus);
+        $value = str_replace($settings->num_grp_sep, "", $value);       
+
+        if (isset($vardef['len'])) {
+            // check for field length
+            $value = sugar_substr($value, $vardef['len']);
+        }
+
+        return $value;
```
clemente-raposo, line 105: Hi @SinergiaCRM,

Thank you for your PR.

Don't we also want to make sure the decimal separator is a "."? or is that what causes the problem on currencies at the moment?

clemente-raposo: Hi @SinergiaCRM, thank you. Makes sense.

clemente-raposo (approved):
LGTM

johnM2401:
Hey!

I've given this some testing, and it seems like the fix does indeed resolve issues when importing new records.

However, I seem to hit an error when trying to Import & Update records.

I've set my User's Thousands Operator to "."
And decimal operator to ","
image

<br>

Then, Importing to Update two Opportunities, I received the error:
"A non-numeric value encountered on line 3093 in file [crm]/include/database/DBManager.php"
"A non-numeric value encountered on line 3094 in file [crm]/include/database/DBManager.php"

image

<br>

This doesn't seem to happen when importing NEW records, only when updating existing records.

CSV Used attached, for reference:
[Opportunitiesz.csv](https://github.com/salesagility/SuiteCRM/files/11123510/Opportunitiesz.csv)

<br>

What are your thoughts?

SinergiaCRM:
> Hey!
> 
> I've given this some testing, and it seems like the fix does indeed resolve issues when importing new records.
> 
> However, I seem to hit an error when trying to Import & Update records.
> 
> I've set my User's Thousands Operator to "." And decimal operator to "," image
> 
> Then, Importing to Update two Opportunities, I received the error: "A non-numeric value encountered on line 3093 in file [crm]/include/database/DBManager.php" "A non-numeric value encountered on line 3094 in file [crm]/include/database/DBManager.php"
> 
> image
> 
> This doesn't seem to happen when importing NEW records, only when updating existing records.
> 
> CSV Used attached, for reference: [Opportunitiesz.csv](https://github.com/salesagility/SuiteCRM/files/11123510/Opportunitiesz.csv)
> 
> What are your thoughts?

Hi @johnM2401, 

What versión of PHP, SuiteCRM, MySQL/MariaDB are you using?

johnM2401 (approved):
Apologies for the delay, @SinergiaCRM 

I've retested this after merging into the most recent version locally, (7.14), and cannot seem to replicate the issue I've raised above.

After some testing with various combinations of special-character Decimal/Thousands separators, it appears to be working well.

LGTM!

PR #9884, opened 2023-01-03
Fix #9883 -  Security Groups do not work with modules whose name exceeds 36 characters.

### LINKED ISSUE #9883: Security Groups do not work with modules whose name exceeds 36 characters.
#### Issue
Security groups do not work properly with modules whose name **exceeds 36 characters**. In the detail view, the GS subpanel is correctly displayed and the assignment does not cause any errors in the log, but the selected GS is not displayed in the subpanel.

#### Expected Behavior
The operation should be correct regardless of the number of characters in the module.

#### Actual Behavior
If the name of a module exceeds 36 characters, security groups simply do not work for that module.

#### Steps to Reproduce

1. In the module builder, create a package with a module whose name (including the prefix) exceeds 36 characters and install it.
2. Create any security group.
3. Assign any security group to a record of the newly created module.
4. Check in the detail view of a record of the new module for the error (the GS will not be visible in the subpanel after assigning it).

#### Your Environment
* SuiteCRM Version used: 7.12.8

SinergiaCRM:
resolve #9883
## Description
It has been found that the problem is caused by the _module_ field in the _securitygroups_records_ table being defined as _char(36)_, so the name of the module will be truncated in the database when a record is created in this table, that is, when any Security group is assigned to a record of the same. In this PR, the type of the module field is changed to _varchar(100)_ in the _securitygroups_records_ table, so that longer module names are stored correctly. It has been considered that setting the limit of the field at 100 is sufficient for any reasonable module name.

## Motivation and Context
It is common to use module names that, along with the prefix, exceed 36 characters and this must be corrected.

## How To Test

1. Remove or rename the securitygroups_record table and quick repair. The table should be created with the module field set to varchar(100).
2. Verify that the assignment of security groups to records of a module whose name exceeds 36 characters works correctly, specifically that the name of the module is not truncated in the database.

## Types of changes

### Final checklist

### DIFF
--- a/metadata/securitygroups_recordsMetaData.php
+++ b/metadata/securitygroups_recordsMetaData.php
@@ -5,7 +5,7 @@ $dictionary['securitygroups_records'] = array(
         array('name' =>'id', 'type' =>'char', 'len'=>'36', 'required'=>true, 'default'=>'')
       , array('name' =>'securitygroup_id', 'type' =>'char', 'len'=>'36')
       , array('name' =>'record_id', 'type' =>'char', 'len'=>'36')
-      , array('name' =>'module', 'type' =>'char', 'len'=>'36')
+      , array('name' =>'module', 'type' =>'varchar', 'len'=>'100')
       , array('name' =>'date_modified','type' => 'datetime')
       , array('name' =>'modified_user_id', 'type' =>'char', 'len'=>'36')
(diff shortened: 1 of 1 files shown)


clemente-raposo (approved):
LGTM

johnM2401 (approved):
LGTM

PR #9887, opened 2023-01-04
There is no module "Quotes" (anymore)

gunnicom:
Obsolete realtion definition. There are quite some more of these. Will do more pull requests like this, after this one here is merged.


pgorod:
There is no module Quotes anymore? This surprises me. I just checked the 7.13 live demo and it's there...

gunnicom:
@pgorod You maybe mean the module "AOS Quotes" which is there with its database table "aos_quotes" and its module folder "modules/AOS_Quotes". But there is no more module "Quotes" with the table "quotes" and a module folder "modules/Quotes" as it is in the relation definition I refer here.

PR #9940, opened 2023-02-09
Fix #9939 - Consistent Null/Empty handling on Reports and Workflows

### LINKED ISSUE #9939: Null/Empty value handling inconsistent across modules
#### Issue
When comparing empty and null strings, the current handling of this is inconsistent between `Reports` and `AOW_Workflows`

Reports:
Checks if a value is empty or null during an "equal to" empty string value.
Checks if a value is empty during a "not equal to" empty string value.

Workflows:
Checks if a value is empty during an "equal to" empty string value.
Checks if a value is empty during a "not equal to" empty string value.

#### Expected Behavior
The handling of empty and null values should be consistent across these modules.
As a null value could easily be interpreted by a user as empty, null could be checked at the same time in both types and modules.

#### Actual Behavior
Null is considered empty in one of the four situations.

#### Possible Fix
Check for empty or null on all four situations. Merge request to follow.

#### Steps to Reproduce
1. Create a report which checks against a nullable field with empty values
2. Behaviour will be inconsistent between "equal to" and "not equal to" across the two modules discussed.

#### Context
Records returned between these two modules are different with the same conditions

#### Your Environment
* SuiteCRM Version used: 7.13.1
* Environment name and version (e.g. MySQL, PHP 7): PHP7.4, MariaDB
* Operating System and version (e.g Ubuntu 16.04): Debian 11

simonr44:
## Description
When comparing empty and null strings, the current handling of this is inconsistent between Reports and AOW_Workflows

Reports:

- Checks if a value is empty or null during an "equal to" empty string value.
- Checks if a value is empty during a "not equal to" empty string value.

Workflows:

- Checks if a value is empty during an "equal to" empty string value.
- Checks if a value is empty during a "not equal to" empty string value.

This change will:

- Check for null or empty strings on an "equal to"
- Check for not null or empty strings on an "not equal to"

As null will now be considered empty where it was not previously, this may lead to different (hopefully more accurate?) report / workflow results depending upon the data set. Still something to consider.

Null values can still be checked with the null, not null condition parameters and are unaffected by this change.

## Motivation and Context
Records returned between these two modules are different with the same conditions

## How To Test This

1. Create a report which checks against a null-able field with empty values
2. Behaviour will be consistent between "equal to" and "not equal to" across the two modules discussed.

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOR_Reports/AOR_Report.php
+++ b/modules/AOR_Reports/AOR_Report.php
@@ -1859,6 +1859,19 @@ class AOR_Report extends Basic
                         $field = "(" . $field;
                     }
 
+                    if ($condition->value_type == 'Value' && !$condition->value && $condition->operator == 'Not_Equal_To') {
+                        if (!isset($value)) {
+                            $GLOBALS['log']->warn(
+                                $condition->field
+                                . ' value is not set, assuming empty string value'
+                            );
+                            $value = '';
+                        }
+
+                        $value = "{$value} OR {$field} IS NOT NULL)";
+                        $field = "(" . $field;
+                    }
+
                     if (!$where_set) {
                         if ($condition->value_type == "Period") {
                             if (array_key_exists($condition->value, $app_list_strings['date_time_period_list'])) {
--- a/modules/AOW_WorkFlow/AOW_WorkFlow.php
+++ b/modules/AOW_WorkFlow/AOW_WorkFlow.php
@@ -591,6 +591,16 @@ class AOW_WorkFlow extends Basic
                     $where_set = true;
                     break;
                 case 'Value':
+                    if(empty($condition->value) && $condition->operator === 'Equal_To') {
+                        $value = "'' OR {$field} IS NULL)";
+                        $field = "(" . $field;
+                        break;
+                    }
+                    if(empty($condition->value) && $condition->operator === 'Not_Equal_To') {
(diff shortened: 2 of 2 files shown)


serhiisamko091184:
Hi @simonr44 ,
thanks for your contribution!

Regards,
Serhii

clemente-raposo (approved):

johnM2401:
Hey @simonr44 

Apologies if i've perhaps misunderstood functionality, but it seems like there may still be a scenario that has issue.

If you are using "Not Equal to" on an [Empty] value, it seems to incorrectly return the record in both Reports and Workflow modules.

See:

I have the following dataset:

Screenshot 2023-09-13 102924

<br>

Setting the Condition for:
```Fax - Equal to - Empty```

Returns the "BothEmpty" record, which is correct.
Screenshot 2023-09-13 102859

<br>

Setting the Condition for:
```Fax - Not Equal to - Empty```

Returns the "BothEmpty" record, despite this value being empty in the database.
Screenshot 2023-09-13 102910


Setting the Condition for both:
```Fax - Not Equal to - Empty```
```Office Phone - Not Equal to - Empty```

Returns the "BothEmpty" record, despite both values being empty in the database.
image


Could you have a look when you get a moment?

Thanks!

Issue #9516, opened 2022-02-22
Global search error after upgrade


dtosun61:
#### Issue

After upgrade from 7.10.15 to 7.12.4, global search doesn't work some modules like Contacts. There is an error like this;
```
Error
An unknown error has occurred while performing the search. Contact an administrator if the problem persists. More information available in the logs.
```
![image](https://user-images.githubusercontent.com/22373035/155119328-4fb7ac25-4e83-4163-bbf0-ca590e83f5ba.png)

![image](https://user-images.githubusercontent.com/22373035/155119404-25eae1c1-0f2e-4275-94e7-21e331b85aa0.png)

If i remove Contacts module from settings, it is working good. When i try to do for new fresh installation, it is working good. May be some files missing in upgrade package?

**This is log;**

`Tue Feb 22 14:03:42 2022 [2430483][134bd9c7-f8ed-de37-4ab0-4b9de86a2a13][FATAL] [ERROR] [SearchThrowableHandler] An unknown error has occurred while performing the search.
Tue Feb 22 14:03:42 2022 [2430483][134bd9c7-f8ed-de37-4ab0-4b9de86a2a13][FATAL] [ERROR] [SearchThrowableHandler] TypeError: Return value of SuiteCRM\Search\SearchResults::getRelatedId() must be of the type string, null returned in /var/www/crm/sractivities/lib/Search/SearchResults.php:227
Stack trace:
#0 /var/www/crm/sractivities/lib/Search/SearchResults.php(179): SuiteCRM\Search\SearchResults->getRelatedId()
#1 /var/www/crm/sractivities/lib/Search/SearchResults.php(162): SuiteCRM\Search\SearchResults->updateObjLinks()
#2 /var/www/crm/sractivities/lib/Search/SearchResults.php(145): SuiteCRM\Search\SearchResults->updateFieldDefLinks()
#3 /var/www/crm/sractivities/lib/Search/UI/SearchResultsController.php(132): SuiteCRM\Search\SearchResults->getHitsAsBeans()
#4 /var/www/crm/sractivities/lib/Search/SearchEngine.php(102): SuiteCRM\Search\UI\SearchResultsController->display()
#5 /var/www/crm/sractivities/lib/Search/SearchEngine.php(78): SuiteCRM\Search\SearchEngine->displayResults()
#6 /var/www/crm/sractivities/lib/Search/SearchWrapper.php(94): SuiteCRM\Search\SearchEngine->searchAndDisplay()
[...]

PR #9983, opened 2023-03-17
Fix issue #9516

QuickCRM:
## Description

getRelatedId is supposed to return a string but returns null in some conditions. Return an empty string instead

Fix issue #9516 

## Motivation and Context

Global Search can fail with an error message:
An unknown error has occurred while performing the search. Contact an administrator if the problem persists. More information available in the logs.
Return value of SuiteCRM\Search\SearchResults::getRelatedId() must be of the type string, null returned in /var/www/crm/sractivities/lib/Search/SearchResults.php:227
## Types of changes

### Final checklist

### DIFF
--- a/lib/Search/SearchResults.php
+++ b/lib/Search/SearchResults.php
@@ -224,6 +224,9 @@ class SearchResults
             LoggerManager::getLogger()->warn('Unresolved related ID for field: ' . $relField);
         }
 
+        if (!$relId) {
+            $relId = '';
+        }
         return $relId;
     }
 


QuickCRM:
Hi @pgorod 
To me, it's just a PHP issue.
Being defined as protected function getRelatedId(SugarBean $obj, string $idName, string $link): **string**
PHP does not allow returning a null value.
So, it already breaks it.

If you prefer, we could just remove :string in the function definition

pgorod:
You are right.

I checked all the other gerRelatedId calls I was seeing, and they are a different function, related to email, with 2 arguments, not 3.

This one should definitely return a string. 👍

PR #10032, opened 2023-04-24
Fix #10030 - Workflow - HTML decoding of parameters to be used in formulas

### LINKED ISSUE #10030: Workflow - HTML decoding of parameters to be used in formulas
#### Issue

When creating a workflow that performs an action on calculated fields with functions such as _substring_, _length_, _uppercase_ or _replace_, some special characters (such as á,à,ä,ñ) are miscoded. 

#### Expected Behavior
Correct encoding of special characters is expected.

#### Actual Behavior

For example, when passing `montaña` to the _uppercase_ function the result is` MONTA&NTILDE;A `(`&NTILDE;` is not an HTML entity, the correct one for Ñ is `&Ntilde;`). Using the _length_ function in this same word calculates `14` instead of the expected `7`. This does not happen with _lowercase_ because `MONTAÑA` becomes `MONTA&Ntilde;A` and then `monta&ntilde;a` which in this case can be reconverted to `montaña`.

#### Steps to Reproduce

1.Create different workflows that use the action of calculated fields with functions such as length, or uppercase.
2.Create a record on which the workflow acts that contains special characters such as á,à,ä,ñ.
3.Check the encoding problem.

#### Your Environment

* SuiteCRM Version used: 7.12.10
* Operating System and version: Ubuntu 20.04

SinergiaCRM:
- Closes #10030  

## Description

 The HTML values of the bean fields to be passed as parameters to the workflow formula are decoded to avoid mis-encoding of special characters.

## How To Test This

1.Create different workflows that use the action of calculated fields with functions such as length, or uppercase.
2.Create a record on which the workflow acts that contains special characters such as á,à,ä,ñ.
3.Check that the coding problem is fixed.

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOW_Actions/actions/actionComputeField.php
+++ b/modules/AOW_Actions/actions/actionComputeField.php
@@ -161,7 +161,23 @@ class actionComputeField extends actionBase
                     }
                 }
             } else {
-                $resolvedParameters[$i] = ($bean->{$parameters[$i]} == null) ? "" : $bean->{$parameters[$i]};
+                $type = '';
+                if (isset($bean->field_name_map[$parameters[$i]]['type'])) {
+                    $type = $bean->field_name_map[$parameters[$i]]['type'];
+                }
+                if (isset($bean->field_name_map[$parameters[$i]]['dbType'])) {
+                    $type .= $bean->field_name_map[$parameters[$i]]['dbType'];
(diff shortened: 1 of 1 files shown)


pgorod:
Why are these encoded in the first place?

PR #10036, opened 2023-04-25
Fix #10033 PHP Fatal error Uncaught TypeError PHP8

### LINKED ISSUE #10033: "Send Email" Workflow Action fails in PHP8.0+
#### Issue

When the user has a "Send Email" Workflow Action set.

If a record meets the set Conditions and the Workflow fires, (either by On-Save or Scheduler), it fails to send.

On-save records will show a white screen.

Cron runs show the following error:
image

#### Expected Behavior
Users should be able to send Emails via Workflow in PHP8

#### Actual Behavior
Emails cannot be sent via Workflow in PHP8, as an error occurs.

#### Steps to Reproduce

1. Ensure System Emails are set up via Admin->Email Settings
2. Create a Workflow on an Email-valid module, (ie: accounts)
3. Set an Action to "Send Email" to a valid address
4. Cause the workflow to fire, via On-save or Scheduled Job
5. See that the Mail is not sent, and the item gets stuck at "Running"
image

#### Your Environment

* SuiteCRM Version used: 7.13.3
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Firefox 112.0.1
* Environment name and version (e.g. MySQL, PHP 7): PHP8.0/Mariadb 10.4
* Operating System and version (e.g Ubuntu 16.04): Ubuntu 18.04

chris001:
The `$array` parameter passed to `implode($separator, $array)` was `null`. 
Previously in PHP 7 and less, this gave only a `Notice` or `Warning` error.
In PHP 8+ passing a null array parameter to implode now gives a `FATAL` error and PHP exits because the call matches a different overloaded function signature, expecting an `array` in first position, but it gets the `separator` which is a `string`.

## Description

Fixes bug #10033 when running on PHP 8+

## Motivation and Context

PHP 7 and below only emitted a `Notice` or `Warning` to the log when you call `implode` with a `NULL` for the `$array`.
PHP 8 and above now will throw an `Exception` and exit PHP because the call signature matches a different overload and the `$separator` is passed in where it expects the `$array`.

## How To Test This

See issue.
Basic code example proving this is `FATAL` on PHP 8+
```
<?php
$emailsArray = null;
$emailsToLine = implode (',' , $emailsArray);
?>
// When run on PHP 7, this will exit without `Exception`.
// On PHP 8, this will throw a PHP `Fatal uncaught TypeError` Exception
```

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOW_Actions/actions/actionSendEmail.php
+++ b/modules/AOW_Actions/actions/actionSendEmail.php
@@ -484,7 +484,7 @@ class actionSendEmail extends actionBase
         $mail->handleAttachments($attachments);
         $mail->prepForOutbound();
 
-        if (empty($emailTo)) {
+        if ((empty($emailTo)) || (!is_array($emailTo))) {
             return false;
         }
         foreach ($emailTo as $to) {
@@ -499,6 +499,13 @@ class actionSendEmail extends actionBase
             foreach ($emailBcc as $email) {
                 $mail->AddBCC($email);
             }
+        }
+        if (!is_array($emailCc)) {
+            $emailCc = [];
+        }
+
+        if (!is_array($emailBcc)) {
+            $emailBcc = [];
         }
 
         //now create email


serhiisamko091184:
Hello @chris001 ,
thanks a lot for the contribution. 

Regards,
Serhii

clemente-raposo (approved):
LGTM

PR #10059, opened 2023-05-15
Update Folders to Mark Group Folders =1 RE: #9028

pstevens71:
This is a fix so that when an Inbound email account is created or edited, the folder gets updated with is_group =1 if it is a group folder.  This is necessary because only group folders with is_group=1 show up for users to view (if they have access) in email ListView when switching between folders.

## Description

The changes check to see if the current folder being updated (after a change or Inbound Account creation) is a parent folder or not.  If a parent folder, it checks to via the ID of the folder to see if the associated Inbound account is a "group" account.  If so, it marks the parent folder with is_group =1.

If it is not a parent folder, then it uses the parent folder ID to lookup the associated Inbound Account and whether or not it is  a group account.  If it is a group account, it marks the folder with is_group=1.

This is to address https://github.com/salesagility/SuiteCRM/issues/9028

## Motivation and Context

Currently only admins can see group folders.  The filter that makes folders available for viewing on the front end selects only folders that is_group=1 or is created by the user.  (unless the user is admin and then it shows all).  This will allow folder to be marked accordingly so that they can be seen by users (and they are subscribed to the folder).

## How To Test This

This change is needed in conjunction with:  https://github.com/salesagility/SuiteCRM/pull/8702
And hotfix:  https://github.com/salesagility/SuiteCRM/pull/8702/commits/5132dc54f7199f8f60ee510183a2f8320325eb8b
pull #8702 is also a problem because it has a syntax error.  Once 8702 is applied along with this hotfix users will be able to see group folders in Inbound Email ListView when clicking on the folders button to switch folders.

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [x ] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change to the documentation.
- [ x] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

Issue #9028, opened 2021-02-08
Accessing Group Mailboxes Does Not Work


greenlanegreb:
There is no ability to assign access to Group Mailboxes to users via user admin or the group mailbox itself. 

If I am not mistaken (and one of the Devs John, Maintainer) hasn't had any luck either, then a CRM without an ability to communicate and interact is a very serious issue and, I would propose that whatever issues there are with the functionality needs to be resolved as a priority. I am running the latest version of Suite CRM on a Linux Server with MySQL (although not obviously relevant). 

Thanks and Happy New Year.

johnM2401:
Links to relevant threads:
https://community.suitecrm.com/t/new-cases-not-being-created-from-email-and-return-emails-not-being-sent/76254/17
https://community.suitecrm.com/t/how-to-assign-inbox-group-email-account-to-users/70220/

tsmgeek:
Might be fixed by #8702

greenlanegreb:
Please see my notes added to bug #8745 - I think CSS as a possible hidden assign button may be central to the issue. Cheers.

greenlanegreb:
> Might be fixed by #8702

@tsmgeek good day Ricardo, has this fix been rolled out in the latest update please? Thank you.

greenlanegreb:
@tsmgeek - i've updated #8702 as your code doesn't fix this but thanks so much for your hard work. Is your code OK to leave in situ on a long term basis? I have backed up the original files but I again shout CSS issue as explained just a moment ago on #8702 

Thanks and Kind Regards, 

Beanie.

greenlanegreb:
Any further exploration of this issue please or temporary fixes? Thanks.

tsmgeek:
I use all my patches in production where I work, problem is that the Email module IMO is deeply flawed after it was all reworked and even with my patches it still needs more work to fix it properly.

greenlanegreb:
Thanks tsmgeek - it's categorised as critical, it's been open 4 days and nobody is keeping us up to date with whether anything is going on. I'm probably going to leave SuiteCRM once and for all. It's been an absolute disaster in one way or another for so long now.

SuiteBot:
This issue has been mentioned on **SuiteCRM**. There might be relevant details there:

https://community.suitecrm.com/t/basic-setup-with-group-email/77776/3

pstevens71:
[...]

PR #10084, opened 2023-05-30
Update Delete.php to mark group folders as deleted RE:#9028 #10059

pstevens71:
When group email accounts are deleted the associated folders don't get marked as deleted=1 in the folders table.  This means they show up for users in folder selection because is shows all folder not equal to deleted=1 (among a few other conditions).  They should be marked as deleted when the inbound group email account is deleted.  There is in fact, functionality to do this in SugarFolders.php but is is not called when an inbound group email is deleted.  This additional code just calls the existing function delete() when an inbound email account is deleted.

## Description

This is an addtion to delete.php file simply to get the ID of the folder that is being deleted and then to call and pass it along to SugarFolders.php.  There are two functions that then get processed:  delete() which also then calls deleteChildrenCascade($id) to delete the child folders.

This is related to the general issue #9028 there are various problems related to group emails.  This is one.

#9028 #10059 

## Motivation and Context

This change is required so that the delete function gets called properly and the folders from the folders table are marked deleted as they should be.

## How To Test This

Create a group inbound email account. Ensure you have subscribed the user to folders so they are created.
Delete the inbound group email account, after application of this change, the INBOX (parent folder) and the children folders of the inbound email account should now be updated such that deleted="1" in the folders table.

## Types of changes

### Final checklist

- [x ] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [x ] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

### DIFF
--- a/modules/InboundEmail/Delete.php
+++ b/modules/InboundEmail/Delete.php
@@ -53,4 +53,21 @@ if (empty($_REQUEST['record'])) {
     $focus->retrieve($_REQUEST['record']);
     $focus->mark_deleted($_REQUEST['record']);
     header("Location: index.php?module=".$_REQUEST['return_module']."&action=".$_REQUEST['return_action']."&record=".$_REQUEST['return_id']);
+    //delete related folders by calling delete() in SugarFolders.php
+    // Retrieve the ID of the inbound email account being deleted
+    // Create an instance of the SugarFolders class to call the delete() function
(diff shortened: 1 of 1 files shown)


pstevens71:
Note:  this change is applicable to both 7.12 and 7.13+ branches.

Issue #10099, opened 2023-06-27
Workflow action create record on CRON set incorrect date field values from related entity


re8260:
#### Issue
Workflow action Create Record works different from on screen save and CRON.
On cron run it set incorrect date values based on CRON user date time format
When save from screen - all works correct
Example WF:
![image](https://github.com/salesagility/SuiteCRM/assets/18328936/5eebcfb8-13a7-4f6f-a8d9-6188b0fe2020)

#### Expected Behavior

All START and END date should be same when saving from screen or by cron and should not be based on CRON user datetime format and timezone.
#### Actual Behavior

![image](https://github.com/salesagility/SuiteCRM/assets/18328936/40a402c7-103c-4138-901b-d9a2e15f2ce7)
There is same workflow runs before and after fix with 3 different sets:
1. When workflow runs from trigger ON_SAVE system user date format not effect on the result
2. When workflow runs from trigger In_Scheduler with system user date format DD/MM/YYYY+3
3. When workflow runs from trigger In_Scheduler with system user date format YYYY-MM-DD+3

#### Possible Fix

https://github.com/salesagility/SuiteCRM/pull/10101
#### Steps to Reproduce

1. Create Workflow with action create record
2. set bean datetime field from workflow bean date field (example screenshot attached)
3. run cron

#### Your Environment

* SuiteCRM Version used:
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)):
* Environment name and version (e.g. MySQL, PHP 7):
* Operating System and version (e.g Ubuntu 16.04):

pgorod:
Cron doesn't execute for any specific user.

Suppose your company has users in multiple timezones, which one would you use in CRON?

re8260:
hi @pgorod I just cloned this repo from here, create Workflow, and run the cron, to show this issue and fix.

By "CRON USER" in my case I meant the current CRM user and not system user that running the cron function.
public/cron.php
![image](https://github.com/salesagility/SuiteCRM/assets/18328936/6322b95d-41c7-4f75-a4d8-2501ccd0ba75)
by default is the user ID:1, admin user
public/modules/Users/User.php
![image](https://github.com/salesagility/SuiteCRM/assets/18328936/7b7c5e6c-e2d2-413d-8279-78ca5f706e63)

When workflow runs with **same user** from trigger by screen or from cron running the result is different, the result must be the same.

pgorod:
[...]

PR #10101, opened 2023-06-28
#10099 Workflow action create record by Scheduler set incorrect date field values from related entity

re8260:
## Description

Workflow action Create Record works different from on screen save and CRON.
On cron run it set incorrect date values based on CRON user date time format
When save from screen - all works correct
https://github.com/salesagility/SuiteCRM/issues/10099
image

There is same workflow runs before and after fix with 3 different sets:
1. When workflow runs from trigger ON_SAVE system user date format not effect on the result 
2. When workflow runs from trigger In_Scheduler with system user date format DD/MM/YYYY+3
3. When workflow runs from trigger In_Scheduler with system user date format YYYY-MM-DD+3

## Motivation and Context

When action "Create record" runs on bean saved from screen, the date in bean are not converted to the user date format, all dates are always in DB GMT format
On cron this action receive bean with **already converted dates** to user format - the current user is cron user as usually ADMIN and the this date passed to PHP function `strtotime` that fail or works different with different date formats and time zone
## How To Test This

Steps in issue: https://github.com/salesagility/SuiteCRM/issues/10099
## Types of changes

### Final checklist

### DIFF
--- a/modules/AOW_Actions/actions/actionCreateRecord.php
+++ b/modules/AOW_Actions/actions/actionCreateRecord.php
@@ -285,7 +285,8 @@ class actionCreateRecord extends actionBase
                                     $date = $params['value'][$key][0];
                                 } else {
                                     $dateToUse = $params['value'][$key][0];
-                                    $date = $bean->$dateToUse;
+                                    $bean->retrieve($bean->id);
+                                    $date = $timedate->fromUser($bean->$dateToUse)->asDB();
                                 }
 
                                 if ($params['value'][$key][1] !== 'now') {


SinergiaCRM:
Hello @re8260,
we have also encountered this issue and had an alternative solution, which used the value stored in `$bean->fetched_row.` if is set.
Upon further analysis, we prefer your proposal.
Thank you.

pgorod:
I don't think this can be evaluated without precise information about your _effective_ **`date.timezone`** parameter for both

- the CRON **`php.ini`**
- the web server **`php.ini`**

When you call **`fromUser`** function from CRON, which user is that? Why is that user's timezone appropriate to use here?

SinergiaCRM:
Okay, true, in the case of different time formats, this solution may not necessarily work.

Our proposed solution captures the value obtained in `$bean->fetched_row`, which is empty if the action is executed upon saving the record. However, in that case, the date contains the correct value:
```php
// The following workaround uses the date of fetched_row if is set
// When a workflow is executed "after save", the date format is correct, and the fetched_row is empty
$date = (isset($bean->fetched_row) && isset($bean->fetched_row[$dateToUse])) ? $bean->fetched_row[$dateToUse] : $bean->$dateToUse;
```

re8260:
Cron always has logged-in user, as usually its user Admin ID:1

image

and user has time zone settings in his profile.

_Current user_ is required when running scheduler from cron to set "created by" \ "modified by" and the change log

serhiisamko091184:
Thanks for your PR, @re8260!

jack7anderson7 (approved):

PR #10114, opened 2023-07-14
Fix parameter userTime method in class TimeDate

gpibarra:
Another alternative would be to change this function
```
    /**
     * Format DateTime object as user time
     * Note: by default does not convert TZ!
     *
     * @param DateTime $date
     * @param boolean $tz Perform TZ conversion?
     * @param User $user
     * @return string
     */
    public function asUserTime(DateTime $date, $tz = false, User $user = null)
    {
        if ($tz) {
            $this->tzUser($date, $user);
        }

        return $date->format($this->get_time_format($user));
    }
``Σ

### DIFF
--- a/include/TimeDate.php
+++ b/include/TimeDate.php
@@ -617,7 +617,7 @@ class TimeDate
                 return $this->asUserDate($date, true, $user);
                 break;
             case 'time':
-                return $this->asUserTime($date, true, $user);
+                return $this->asUserTime($date, $user);
                 break;
             case 'datetime':
             case 'datetimecombo':


pgorod:
Ia there an Issue for this? I'm trying to understand what are the manifestations of the bug.

gpibarra:
It is calling the method and the second parameter "true" and tries to pass a third parameter to a user, but the method actually gets the user in the second parameter, thus mistakenly taking the user with "true".
This error is quickly visible with a static code analysis since the parameters have their types correctly in the method definition.

https://github.com/salesagility/SuiteCRM/blob/5ededf4227202b398df2487a18e930fa60180799/include/TimeDate.php#L697

serhiisamko091184:
Hello @gpibarra ,
thanks a lot for contributing to SuiteCRM project!

Regards,
Serhii

clemente-raposo (approved):

johnM2401 (approved):

Issue #9820, opened 2022-11-15
Email Template doesn't parse currency with email compose.


pstevens71:
#### Issue

When creating an email template that has currency values, the values are not being parsed as currency in the email.   The field values are not being run through the already defined_convertToType() function.

#### Expected Behavior

When currency fields are added to an email template, when the email is parsed it should format the currency as currency.
#### Actual Behavior

Instead, currency is outputted as a raw value without currency formatting.

#### Possible Fix

in  /modules/EmailTemplates/EmailTemplate.php add the following two lines:
 (after line 458)
`` //add this line to get the type of field
			$type = $focus->field_defs[$field_name]['type'];
// convert type function to convert to currency
			$value =  self::_convertToType($type, $value); ``` 

#### Steps to Reproduce

1. Create an email template
2. Create a field with a currency value
3. Add the field to the email template
4. Send the email
5. The currency will be formatted like xxx.xxx instead of $xxx.xx

#### Context

This makes outputted currency messy.

Probably a low priority, it's just a formatting issue.
#### Your Environment

* SuiteCRM Version used:  7.12.7
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Chrome
* Environment name and version (e.g. MySQL, PHP 7): PHP 8.0
* Operating System and version (e.g Ubuntu 16.04):  Linux

SuiteBot:
This issue has been mentioned on **SuiteCRM**. There might be relevant details there:

https://community.suitecrm.com/t/email-template-parser-not-parsing-currency/86844/14

pgorod:
If you want to create the PR yourself, it's pretty easy to do it from the GitHub website for single-file commits.

Just navigate to the correct file on the `hotfix` branch and hit Edit button. Make your changes and save, which creates the commit. Then it offers to create the PR for you, you just need to fill in the template and save.

In this case, you would start here:
https://github.com/salesagility/SuiteCRM/blob/hotfix/modules/EmailTemplates/EmailTemplate.php

Note that it's the `hotfix` branch I selected there.

PR #10277, opened 2023-12-07
#9820 -Update EmailTemplate.php

pstevens71:
Allows currency fields  to be parsed in email templates as currency on the final email.

The current $value just outputs raw value. Needs to be formatted as currency. There is already a function in the file to do that, just need to get the field type to use the function.

## Description

First lirst line just gets the field "type" from the array. The second field just runs the value through the already defined function _convertToType() which formats for currency if it is a currency field.

This is issue https://github.com/salesagility/SuiteCRM/issues/9820ine just gets the field "type" from the array. The second field just runs the value through the already defined function _convertToType() which formats for currency if it is a currency field.

This is issue https://github.com/salesagility/SuiteCRM/issues/9820
## Motivation and Context

It solves the problem of email templates with currency being formatted properly.
## How To Test This

I tested this with a currency field and also a text field and both work after the change and produce the desired result in the email template final formatting.

To test you can create an email template with a currency field and see it is not formatted as currency, apply the patch and then re-output the email and you'll see its formatted as currency.

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)

### Final checklist

- [ x] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ x] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.


chris001:
It should it be titled `Fix #9820 Email Template doesn't parse currency with email compose.`
The pattern is `Fix #(issue number) (exact same title of that issue)`.

serhiisamko091184:
Hello @pstevens71,

thanks for your contribution! 

Regards,
Serhii

johnM2401:
Hey @pstevens71 !

I've given this PR a test locally
However, it seems as though the Email template always formats the amount value as USD.
This might lead to confusion / wrong information being interpreted if the CRM is using multiple Currencies.

For example, if I create the following Email template for Opportunities:
image

I create a 2nd Currency for GBP, and set it on an Opportunity:
image

The Email comes through with the correct Currency selected, but the "Amount" field is rendered in USD:
image

<br>

Do you know if this is intended/expected for this PR?

Thanks!

pstevens71:
Hmm, it seems like the problem (in addition to this one)  is with the _convertToType() either not parsing the right currency or, not getting passed what it needs to do so.  I'll have a look.

pstevens71:
Ok having a deeper dive into the  _convertToType() function it will always convert it to the default currency specified in the locale (not the amount, but they formatting of the output).  It needs a condition to check the current record to see if a different currency is set in order to work for an alternate currency from the default locale.  I'll see if I can come up with something for that too.

pstevens71:
Grr, why does everything have to be a can of worms!  Ok so the amount in $opportunity_amount_usdollar is not actually the amount in USD, it's an unfortunate field name, it is actually the amount in the default currency specified in the locale settings (in my case CDN).  $opportunity_amount is the $opportunity_usdollar_amount converted to the currency selected in the opportuntiy.  So in my case, $CDN = 1, and $25,000 outputs as both $25,000 for both fields. If I change the opp to USD then the output is:   Amount in USD = $25,000 (which is actuall $25, 000 CDN) and the $opportunity_amount is now $18,250 (USD).   So the rule would have to be something like if the record contains amount_usdollar, then use the locale currency, if the current record contains a currency ID other than the locale, then use that instead for the field.   So the amount_usdollar should always output as the default locale and any other amounts (if they exist) should output in the currency of current record.  Ok let's see if I can do that!!!

PR #10322, opened 2024-01-19
Fix# 10321 Creation of Project with Template Causes 500 Error

pstevens71:
## Description

I believe this line was added in the wrong place as part of:
https://github.com/salesagility/SuiteCRM/commit/9aff7071bb86df7e06e2cf56aa3060fc00e3978a

It appears in the final code in line 613 not 330.

This causes the array to re-initialize and then subsequently causing a 500 error because the date in the second iteration is blank in the array and cannot be converted to a date and returns boolean false.

## Motivation and Context

This is necessary to fix because no projects can be created with templates with this bug.

## How To Test This

Try to create a project with a template with more than one task and try to save it.  It will result in 500 error.

Make the fix and it will work as expected.

## Types of changes

### Final checklist

- [x ] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [x ] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

### DIFF
--- a/modules/Project/Project.php
+++ b/modules/Project/Project.php
@@ -331,6 +331,7 @@ class Project extends SugarBean
         global $current_user;
         $db = DBManagerFactory::getInstance();
         $focus = $this;
+        $enddate_array = [];
 
         //--- check if project template is same or changed.
         $new_template_id = property_exists($focus, 'am_projecttemplates_project_1am_projecttemplates_ida') ?
@@ -610,7 +611,6 @@ class Project extends SugarBean
 
                 //----------------------------------
 
-                $enddate_array = [];
 
                 if ($count == '1') {
                     $project_task->date_start = $start;


serhiisamko091184:
Hello @pstevens71, 

Thanks  for your PR, 

I've managed to replicate an issue and get 500 error creating a project with a project template where two tasks are nested and after applying the current fix it was possible to get a proper result. 

At the same time having tested the same behavior with another template with 4 tasks, I had to wait for about 2-3 minutes and I got the same error: 

image

image

image

image

Are you experiencing similar issues depending on a number of tasks in the project template? 

Thanks in advance! 

Regards,
Serhii

pstevens71:
Thanks @serhiisamko091184 for testing.  I've tested with projects with many tasks and they worked for me.  Hmm it's interesting 4 tasks creates a problem.  I'll test on my end as well with exactly 4 tasks and see what happens.  Did all your tasks have start dates and end dates?  The issue get's created when the date is blank and it tries to convert a null value to a date.  The fix above removes the cause that resets the date back to null inside the loop, so more than one task causes an error.  There may be other problems here.  Probably an if statement would be helpful if NULL, do not process the date conversion and you won't get the 500 error.  But really we should figure out the root cause if it's other than the fix I've proposed.

serhiisamko091184:
Thank you for your quick reply @pstevens71, 

Yes, everything is OK with the dates. 
I moved to a fresh instance with testing and I have proper results applying this fix (projects are created without error and despite the number of tasks).

Thanks one more time! 

Regards,
Serhii

pstevens71:
I just did another test with 4 project tasks and it worked fine.  @serhiisamko091184 did you enter anything different other than the duration in the tasks?  Just trying to figure out what's going on.  Also, what 500 error are you getting?  Can you check the logs.  The only thing that stands out to me is your date format is different.  That might be a clue.

jack7anderson7 (approved):

johnM2401 (approved):
LGTM, have tested with a large amount of nested and non-nested tasks, at various date and time formats.
Hadn't noticed any 500s or Errors, all Tasks are created fine in the created Project.

Issue #10269, opened 2023-11-26
Email Sync Breaking on MIME Decoding


DBRenny:
## Description

Email Synchronization was breaking after authenticating with OAUTH. The root cause was identified in the handleMimeHeaderDecode function, specifically when handling the returned JSON and its subsequent decryption. The changes introduce checks for the variable types and proper logging to understand the flow, ensuring that the CRM seamlessly decodes MIME headers and handles email headers without encountering fatal errors.

## Motivation and Context

This change is crucial as the issue disrupts the normal flow of email synchronization, preventing emails from being properly imported and read in the CRM. By ensuring the variable is iterable before proceeding and adding comprehensive logging, we enhance the system's stability and reliability in processing emails, regardless of their encoding.
## How To Test This

1. Set up email synchronization in SuiteCRM.
2. Send and receive emails with various subject encodings.
3. Monitor the SuiteCRM log for any errors during the email import process.
4. Ensure that emails are properly imported and readable in the CRM without errors.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [x] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

This relates to the issue on suiteCRM-Core (8.x):
[https://github.com/salesagility/SuiteCRM-Core/pull/366](url)

DBRenny:
I've created a PR, but don't know how to best proceed. I am kind of new to GIT.

Please let me know immediate next steps.

I would love to get to the point of using the HELM charts with automatic testing to make sure core functions work when we stage the changes.

PR #10323, opened 2024-01-20
Fix #10172 Emails don't show subject  MIME headers #10269 Email sync …

### LINKED ISSUE #10172: Emails don't show subject, from, to, cc, bcc fields
#### Issue

Fresh installation of SuiteCRM 7.14

I tried find this issue in other tickets but i was not, please if it duplicate, please let me know to which issue in github.

If you set up emails => outbound + inbound  and when you load emails

It missing information like: **FROM, TO, CC, BCC, SUBJECT**

This is listview
email_issue1

This is detail view of non imported email
email_issue2

This is detail view of imported email
Here you can see that FROM, TO fields are filled. I assume that will also work for CC, BCC
email_issue3

#### Expected Behavior

Should work, because in DB i see this informations, when you import it.

#### Actual Behavior

See images. 
In every view is missing SUBJECT
In list view is missing SUBJECT. To field is filled
In detail view is missing all this fields.
In imported view is missing only SUBJECT

#### Steps to Reproduce

1. Connect email mail server
2. List Emails
3. Detail Email
4. Import Email

#### Your Environment

* SuiteCRM Version used: 7.14
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Firefox 117
* Environment name and version (e.g. MySQL, PHP 7): 5.7.42,   php 8.1.12
* Operating System and version (e.g Ubuntu 16.04): Ubuntu 22.04

chris001:
…breaking on MIME decoding.  
A commonly reported bug report on the forum: https://community.suitecrm.com/t/subject-line-missing-from-inbound-emails/88059 
Also fixes this issue https://github.com/salesagility/SuiteCRM-Core/issues/363
..and fixes https://github.com/salesagility/SuiteCRM-Core/issues/420
..all similar reports about the same bug - a lack of resilience in decoding the MIME headers, resulting in broken inbox emails.
Credit also to @DBRenny for putting together a first version of this: https://github.com/salesagility/SuiteCRM-Core/pull/366

## Motivation and Context

Inbound email was broken on Bitnami containers, and on any installation that is missing the `php-imap` extension.  [Bitnami intentionally leaves out the `php-imap` extension, because of the long time since it last received security fixes](https://github.com/bitnami/containers/issues/51877).

## How To Test This

Try inbound email on a bitnami container.  There should be no subjects, for servers that return MIME encoded headers (languages with accented characters outside of US-ASCII character set, e.g. French, German, etc).
Then from inside the container, get the patch with `wget` and apply the patch `git apply`, try inbound email again, and it should show email Subjects in the inbox.  Other MIME encoded headers may also display properly as `UTF-8`, such as `From`, `To`, `CC`, `Bcc`.

## Types of changes

### Final checklist

### DIFF
--- a/modules/InboundEmail/InboundEmail.php
+++ b/modules/InboundEmail/InboundEmail.php
@@ -4866,7 +4866,7 @@ class InboundEmail extends SugarBean
             $subjectDecoded = $this->getImap()->MimeHeaderDecode($subject); // returns array or string.
         } else {
             // imap_mime_header_decode() is not installed on bitnami docker container! Fall back to iconv_mime_decode().
-            if function_exists('iconv_mime_decode') {
+            if (function_exists('iconv_mime_decode')) {
                 $subjectDecoded = iconv_mime_decode($subject, 0, 'UTF-8');  // returns string or false.
                 if ($subjectDecoded == false) { // error occurred in iconv_mime_decode().
                     $subjectDecoded = $subject;  // possibly still mime encoded.
@@ -4874,10 +4874,6 @@ class InboundEmail extends SugarBean
             } else {  // iconv module not installed or enabled!
                 $subjectDecoded = $subject;
             }
-        }
-
-        if (is_string($subjectDecoded)) {  //should be plain US-ASCII, compatible with UTF-8.
-            return $subjetDecoded;
         }
 
         if (is_array($subjectDecoded)) {
@@ -4889,8 +4885,10 @@ class InboundEmail extends SugarBean
                     $ret .= $object->text;
                 }
             }
-            return $ret;
+            $subjectDecoded = $ret;
         }
+
+        return $subjectDecoded;  //should be plain US-ASCII, compatible with UTF-8.
     }
 
     /**
--- a/modules/InboundEmail/InboundEmail.php
+++ b/modules/InboundEmail/InboundEmail.php
@@ -4861,18 +4861,36 @@ class InboundEmail extends SugarBean
      */
     public function handleMimeHeaderDecode($subject)
     {
-        $subjectDecoded = $this->getImap()->MimeHeaderDecode($subject);
-
-        $ret = '';
-        foreach ($subjectDecoded as $object) {
-            if ($object->charset != 'default') {
-                $ret .= $this->handleCharsetTranslation($object->text, $object->charset);
-            } else {
-                $ret .= $object->text;
+        $subjectDecoded = '';
+        if (function_exists('imap_mime_header_decode')) {   // function_exists() should be moved to MimeHeaderDecode().
+            $subjectDecoded = $this->getImap()->MimeHeaderDecode($subject); // returns array or string.
+        } else {
+            // imap_mime_header_decode() is not installed on bitnami docker container! Fall back to iconv_mime_decode().
+            if function_exists('iconv_mime_decode') {
+                $subjectDecoded = iconv_mime_decode($subject, 0, 'UTF-8');  // returns string or false.
+                if ($subjectDecoded == false) { // error occurred in iconv_mime_decode().
+                    $subjectDecoded = $subject;  // possibly still mime encoded.
+                }
+            } else {  // iconv module not installed or enabled!
+                $subjectDecoded = $subject;
             }
         }
 
-        return $ret;
+        if (is_string($subjectDecoded)) {  //should be plain US-ASCII, compatible with UTF-8.
+            return $subjetDecoded;
+        }
+
+        if (is_array($subjectDecoded)) {
+            $ret = '';
+            foreach ($subjectDecoded as $object) {
+                if ($object->charset != 'default') {
+                    $ret .= $this->handleCharsetTranslation($object->text, $object->charset);
+                } else {
+                    $ret .= $object->text;
+                }
+            }
+            return $ret;
+        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


Porhoken:
Hello Chris, 
thank you for your efforts so far. Can you tell me if this fix is for a specific version? I tried to apply this patch on my bitnami/suitecrm Docker Container (Suite CRM Version 8.5.1) but this breaks my menus completely. All menu entries vanished, only the home button remains. The profilebutton leaves me with 'logout' as only option. I checked with an administrator user.
Have you tried this patch in version 8.x?
I installed the german translation also, but this should not break the menus in conjunction with your fix. But then I thought the same of the E-mail decoding module... ;)
I'd like to help here, if you can point me where to look at.

chris001:
You'd need either the `php-imap` or the `php-iconv` module loaded, as this calls a function in one of those modules to decode the MIME to UTF-8.

Porhoken:
So i checked my bitnami container. The iconv module is loaded, version 8.1.27. This happens to be the php version in the latest bitnami suitecrm image. I will try your patch today in a suitecrm v 7.13 container. See how that behaves.

Porhoken:
SOLVED: Looking in the actual php and mysql error logs, showed 

"syntax error, unexpected identifier "function_exists", expecting "(" in bitnami/suitecrm/modules/InboundEmail/InboundEmail.php on line 4869" 

which led me to missing brackets around the second "function_exists" call. I added brackets like this (function_exists()) and everthing works. These brackets exist in the first call of this function in line 4865.

I hope this helps someone, sorry for flooding the comment section.

jack7anderson7:
Hi @chris001 

Thank you for your PR. 

We have been analysing and testing this PR. If we are not mistaken this happens when the imap php module is not installed. Do you agree?

The reason we ask is due to other changes that would need to be made if we were to support not having the imap module.

Thanks, 
Jack

chris001:
Yes, and [this is why bitnami (owned by VMware), maintainer of the Suite docker containers, removed it](https://github.com/bitnami/containers/issues/51877):

"I'm sorry but we recently removed support for the **php-imap** module in our images due to the fact that it was not receiving any updates at all.  Its latest release dates from 2011, and being this remarkably outdated has a direct impact in the security of the images.

>    how to install the php-imap package?

Unfortunately, this can't be achieved as PHP itself needs to be compiled together with support for this module."

High security Suite installations may also opt to avoid using `php-imap`.

Alternatives:
1. This PR - uses `php-iconv`.
2. [PHP-Mailparse](https://www.php.net/manual/en/book.mailparse.php) which requires `php-mbstring`.
3. [mail-mime-parser](https://github.com/salesagility/SuiteCRM-Core/blob/0820b6acc28a04b6386dc860f15ded621b8add27/composer.json#L95), already in `composer.json`.

pgorod:
Google just told me there's also this one

Docs: https://www.php-imap.com/
Github: [webklex/php-imap](https://github.com/Webklex/php-imap)

How extensive is our dependency on the old php-imap? Which functions does it provide, how can we search our codebase for those dependencies?

jack7anderson7:
Hi @chris001 @pgorod 

Thank you for the suggestions.

@chris001 great, that matches what we were expecting. Now, in order to not rely on php imap module there are some more changes that need to be done. 

Both the changes on this PR and the mentioned changes should be done on the Imap2Handler. 

We are going to move this PR into test and we will try to bring it into the codebase but in the future we will have to move the code from this PR to Imap2Handler.

Regarding the suggested libs, and the way to implement these changes, the Imap2 lib is a wrapper around roundcube which is a well-known and battle-tested email client. Therefore, we are thinking of using what we can from the roundcube lib.

Any thoughts or suggestions on using the roundcube lib are welcome.  

Thanks, 
Jack

clemente-raposo (dismissed):

chris001:
Here's some [thoughts from another open source PHP application that faced the same challenge](https://github.com/nextcloud/user_external/issues/44), they moved first to the roundcube lib, then to `curl`.  Their PHP code using these alternative methods, is open and available for use!

johnM2401:
Hey @chris001 !
Thank you for your PR !

During testing and discussions with @clemente-raposo, we've noticed that ```imap_mime_header_decode``` can sometimes still be found, even when ```php-imap``` is disabled.

This results in the records rendering with encoded values, such as:
image


We've tried making the following amendment to your first if statement:
```
        if (function_exists('imap_mime_header_decode') && in_array('imap', get_loaded_extensions(), true)) {   // function_exists() should be moved to MimeHeaderDecode().
            $subjectDecoded = $this->getImap()->MimeHeaderDecode($subject); // returns array or string.
        } else {
```

(ie: adding: ```&& in_array('imap', get_loaded_extensions(), true) ``` )

<br>

Which appears to work in all tested scenarios

For example:
image

<br>


Would you like us to push this change, or would you like to make this change yourself?

Please let us know how you would like to proceed?

Thanks!

chris001:
Thanks @johnM2401 @clemente-raposo @jack7anderson7 for your amendment!
I've updated with this change!

johnM2401 (approved):
Thank you for this @chris001 !

I've given this a test locally and I'm now able to view and import Emails with php-imap extension not found / disabled.

e.g:
image

Have tried with Fastmail, RoundCube and Outlook Mail accounts and all appear to work as expected.

LGTM from a QA perspective

Thanks again!

clemente-raposo (approved):

PR #10331, opened 2024-01-24
Fix #10329 - 'Is null' condition does not work on workflows conditions for numeric fields

### LINKED ISSUE #10329: "Is null" condition does not work on workflows conditions for numeric fields
#### Issue
When a not null condition is used on a numeric field on a workflow condition, if the field is null the workflow does not execute.

#### Expected Behavior
The workflow should be executed when the condition is met (the field is null).

#### Actual Behavior
The workflow is not executed

#### Possible Fix
Take into account the field type when evaluating "is null" condition.

#### Steps to Reproduce
1. Add a numeric field to a module
2. Define a workflow on that module
3. Define a condition that requires that field to be null
4. Check that on save (and on the scheduler) the workflow is not executed

#### Context
We were trying to define some workflows to fire when no amount was specified

#### Your Environment
SuiteCRM Version used: 7.14.2
Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Chrome
Environment name and version (e.g. MySQL, PHP 7): PHP 7.4
Operating System and version (e.g Ubuntu 16.04): Ubuntu 22.04

SinergiaCRM:
## Description
When evaluating "is null" conditions on workflows conditions, a special case is added when the type is numeric.
When checking if a field is null it was being compared to '', bu tnull numeric types are setted to NULL, so the condition was not met.
The field type has been added to the function, and a special case for numeric fields hs been added.

## Motivation and Context
Is null check was not working for numeric fields.

## How To Test This
1. Add a numeric field to a module
2. Define a workflow on that module
3. Define a condition that requires that field to be null
4. Check that on save (and on the scheduler) the workflow is executed if and only if the field is null.

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOW_WorkFlow/AOW_WorkFlow.php
+++ b/modules/AOW_WorkFlow/AOW_WorkFlow.php
@@ -5,7 +5,7 @@
  * SugarCRM, Inc. Copyright (C) 2004-2013 SugarCRM Inc.
  *
  * SuiteCRM is an extension to SugarCRM Community Edition developed by SalesAgility Ltd.
- * Copyright (C) 2011 - 2023 SalesAgility Ltd.
+ * Copyright (C) 2011 - 2018 SalesAgility Ltd.
  *
  * This program is free software; you can redistribute it and/or modify it under
  * the terms of the GNU Affero General Public License version 3 as published by the
@@ -38,7 +38,6 @@
  * display the words "Powered by SugarCRM" and "Supercharged by SuiteCRM".
  */
 
-#[\AllowDynamicProperties]
 class AOW_WorkFlow extends Basic
 {
     public $new_schema = true;
@@ -68,7 +67,6 @@ class AOW_WorkFlow extends Basic
     public $run_when;
     public $flow_run_on;
     public $multiple_runs;
-    public static $doNotRunInSaveLogic = false;
 
     /**
      * return an SQL operator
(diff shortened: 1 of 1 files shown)


serhiisamko091184:
Hello @SinergiaCRM,

thanks for constantly contributing to the SuiteCRM project, 

would you be so kind to: 

1)  update the changes to the code base only related to this specific issue ('is null' condition):

image
Seems like a part of the committed changes represent another PR:  
https://github.com/salesagility/SuiteCRM/pull/9720

2) update the code to clear the comments across the file:
image
 
Like in this example: https://github.com/salesagility/SuiteCRM/pull/10032/files

Many thanks in advance! 

Regards,
Serhii

enricsinergia:
Hi @serhiisamko091184,

sorry, I used the file as we have it and didn't think there were any more changes.
I have already corrected the branch.

SinergiaCRM:
Hi @serhiisamko091184,

sorry, I used the file as we have it and didn't think there were any more changes.
I have already corrected the branch.

serhiisamko091184:
Hello @SinergiaCRM, 

great, thanks a lot for the quick reply and changes, 

one minor update, please: 
image

we'd like to reference a particular issue associated with the current contribution, like in this example: 

image

Many thanks in advance!

enricsinergia:
Hi @serhiisamko091184,
Commit message updated

serhiisamko091184:
Hello @enricsinergia! 

Thanks for the changes.

Regards,
Serhii

PR #10341, opened 2024-01-26
Fix #10339 - Inconsistent application of trim function on name & varchar fields

### LINKED ISSUE #10339: Inconsistent application of trim function on name & varchar fields
#### Issue
In SuiteCRM, an issue has been detected with the handling of varchar type fields, especially those for first and last names. These fields, when filled out through web forms, retain additional spaces at the beginning or end. This results in difficulties in correctly locating these records in searches. For example, if the first_name field includes additional spaces, it will not be properly found when searching by full_name if it is written with only one space between the first and last name.
Additionally, it has been observed that some SuiteCRM modules, like Invoices or Spots, as well as modules created in the Module Builder, maintain these spaces at the beginning or end when creating a new record via the edit view. In contrast, the original modules from SugarCRM do apply the trim function, removing these spaces.

#### Expected Behavior
It is expected that the trim method be uniformly applied to all name and last name type fields in the CRM, regardless of the module or method used for data entry (edit view, inline editing, import, web form, etc.). This should ensure that there are no additional spaces in these fields, facilitating the search and management of records.

#### Current Behavior
Currently, the name and last name fields that are received through forms do not apply the trim method, unlike what happens in the CRM's edit view. This leads to names containing extra spaces, which complicates the search for people in the system (for example, `Alvaro    Gonzalez` would not be properly found due to the extra spaces). Additionally, an uneven application of the trim method has been observed when saving records in different CRM modules, depending on the module and the method used.

#### Proposed Solution
It is proposed to implement the trim function at a central point in the CRM's code flow, ensuring its uniform application in all name and varchar fields of all modules, regardless of the data entry method. This implementation should reflect the practice already established in the original modules from SugarCRM, applied in the edit view.

#### Steps to Reproduce
1. Enter or edit records in different CRM modules using various methods (edit view, inline editing, import, web forms, etc.).
2. Observe the lack of application of the trim method in name and last name fields from forms.
3. Verify the inconsistent application of the trim method in different modules and saving methods.

#### Your Environment
* SuiteCRM Version used: 7.14.2
* Browser name and version: Chrome Versión 121.0.6167.85 (Build oficial) (64 bits)
* Environment name and version (e.g. MySQL, PHP 7): MariaDB 10.2.41
* Operating System and version (e.g Ubuntu 16.04): Kubuntu 22.04

SinergiaCRM:
- Closes #10339 
## Description

This Pull Request addresses issue [#10339](https://github.com/salesagility/SuiteCRM/issues) by implementing improvements in the `cleanBean()` function in `data/SugarBean.php`. It now executes `trim` on all "name" and "varchar" type fields during save operations. A check has been added to avoid using `trim` when the field value is `null` to prevent errors in PHP.

## Motivation and Context

This change is necessary to improve data consistency and cleanliness in text fields where unnecessary spaces can impact data quality and search efficiency.

## How To Test This

1. Create or edit a record in SuiteCRM with additional spaces in "name" or "varchar" type fields in various modules, including modules created using the Module Builder and through Web forms.
2. Save these records and verify that the additional spaces have been automatically removed.
3. Ensure that this functionality works consistently across standard modules, custom modules created via Module Builder, and records entered through web forms to ensure comprehensive testing.

## Types of changes

- [x] Bug fix (non-disruptive change that solves an issue)
- [ ] New feature (non-disruptive change that adds functionality)

### Final Checklist

- [x] My code follows the coding style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change in the documentation.

### DIFF
--- a/data/SugarBean.php
+++ b/data/SugarBean.php
@@ -2520,6 +2520,11 @@ class SugarBean
                     $type .= $def['dbType'];
                 }
 
+                // Trim name & varchar type values on save when the value is not null
+                if (isset($def['type']) && in_array($def['type'], ['name', 'varchar']) && !is_null($this->$key)) {
+                    $this->$key = trim($this->$key);
+                }
+
                 if (isset($def['type']) && ($def['type'] == 'html' || $def['type'] == 'longhtml')) {
                     $this->$key = purify_html($this->$key);
                 } elseif (


serhiisamko091184:
Hey @SinergiaCRM,

many thanks for contributing to SuiteCRM project!

Regards,
Serhii

clemente-raposo (approved):

johnM2401 (approved):
LGTM
Have had a sanity check of:

- Standard "Name" and Concatanated "Name" fields (ie: Accounts and Contacts) on both Core and Custom Modules
- Inline Edit
- Web-to-Lead 
- Workflow

All appear to now strip out Starting and Ending trailing spaces.

PR #10383, opened 2024-03-21
Fix #10382 - Fix issue with enum keys being used when setting values

### LINKED ISSUE #10382: Workflow sets enum key and not label
#### Issue
When a workflow creates or modifies a record and when the WF action is configured to set the field value to another when it is an enum it sets the key rather than the label

E.g.
Field name -> Field -> Field name(Opportunity Name -> Field -> Opportunity Type)

#### Expected Behavior
Unless the target field is also an enum then the enum Label should be used rather than they key
e.g.
Opportunity Name -> Field -> Sales Stage
Results in the opportunity name being set to "Closed Won"

#### Actual Behavior
Unless the target field is also an enum then the enum Label should be used rather than they key
e.g.
Opportunity Name -> Field -> Sales Stage
Results in the opportunity name being set to "closed_won"

#### Steps to Reproduce

1. Create workflow
2. Set WF to modify record using an enum 
3. See the workflow set the enum key to the value 
image

#### Context
Set incorrect value when workflow is executed

#### Your Environment

* SuiteCRM Version used: 7.14.3
* Environment name and version (e.g. MySQL, PHP 7): MySQL, PHP 8.2
* Operating System and version (e.g Ubuntu 16.04): 20.04

JackBuchanan:
Update how enums fields are handled when setting workflow values

## Description
Add extra case to switch statement to handle enums
When a varchar is being set to the value of an enum use the enum label
When a enum is being set to the value of an enum use the enum key

## Motivation and Context
Fixes issue with varchars being set with enum key values

## How To Test This
See #10382

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOW_Actions/actions/actionCreateRecord.php
+++ b/modules/AOW_Actions/actions/actionCreateRecord.php
@@ -244,6 +244,14 @@ class actionCreateRecord extends actionBase
                         $value = $bean->$fieldName;
                     }
                 break;
+                            case 'enum':
+                                if($record->field_defs[$field]["type"] === 'enum'){
+                                    //when setting enum field to another use key rather than label value
+                                    $value = $bean->$fieldName;
+                                }else{
+                                    $value = $app_list_strings[$data["options"]][$bean->$fieldName];
+                                }
+
                             default:
                                 $value = $bean->$fieldName;
                                 break;


serhiisamko091184:
Hi @JackBuchanan,

Thanks for your PR!

Regards,
Serhii

johnM2401 (approved):

johnM2401:
Hey  @JackBuchanan 
Apologies, I'd marked this Approved but have just found a possible issue, (and don't seem to have permissions to dismiss my approval review)

However:

It looks as though the value may be lost when trying to convert to another enum type, when set through workflow.

For example, in this gif, I try to port the value from a Dropdown field to a Multienum field.
However, the value is lost.

wfEnumtypechange

I believe this was possible before this change, so perhaps something that needs addressed.

Would you be able to investigate when you get a moment?
Thanks!

PR #10409, opened 2024-04-19
fix skip to last page if disable_count_query=true

cyb456:
Pressing the skip to last page button in subpanels would result in 9 results showing instead of the default configured ten. This only happens if disable_count_query in the config is set to true. This config option does not appear by default any more so this issue will almost never come up, but still can be found on google: https://community.suitecrm.com/t/slow-suitecrm-with-large-database/33848/3 The previous employee in my position had set this config option to true in a bad attempt to fix performance issues many years ago.

## How To Test This

set disable_count_query=true in a config option
create contacts named test01 through test30 for an account
press skip to last page button on contacts subpanel in that account
notice that test30 is not present
spend hours tracking down this bug just to realize you could have set disable_count_query=false

## Types of changes

### Final checklist

- [ x] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ x] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

### DIFF
--- a/data/SugarBean.php
+++ b/data/SugarBean.php
@@ -1462,10 +1462,10 @@ class SugarBean
                 //C.L. Bug 43535 - Use the $index value to set the $rows_found value here
                 $rows_found = isset($index) ? $index : $row_offset + count($list);
 
-                if (count($list) >= $limit) {
-                    array_pop($list);
-                }
                 if (!$toEnd) {
+                    if (count($list) >= $limit) {
+                        array_pop($list);
+                    }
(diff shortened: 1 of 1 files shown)


serhiisamko091184:
Hi @cyb456,

thanks for your contribution! 

Regards,
Serhii

jack7anderson7 (approved):

johnM2401 (approved):

PR #10428, opened 2024-05-15
Fix #10427 - APC does not work for SuiteCRM 7.14, PHP8

### LINKED ISSUE #10427: APC does not work for SuiteCRM 7.14, PHP8
### Issue

 The memory cache APC was available as APC-bc (backward compatibility) until PHP8 ( https://www.php.net/manual/de/apcu.installation.php )
After that only APCu is available. SuiteCRM exclusively uses apc functions (i.e. apc_store() ), but now has to use apcu_store() ...
As SuiteCRM 7.14.x is only compatible with PHP > 8 we can safely switch it to apcu, as noone could run apc anyway.

There is also someone in the forum in 2019 mentioning this ... as always, nothing got fixed https://community.suitecrm.com/t/apc-cache-not-working/69261

### Possible Fix

_No response_

```bash
1. Install php-apcu
2. See nothing gets cached in memory cache apcu
```

### Context

_No response_

### Version

7.14.3

Chrome

### Browser Version

_No response_

### Environment Information

PHP 8.2

debian

gunnicom:
See https://github.com/salesagility/SuiteCRM/issues/10427

### DIFF
--- a/include/SugarCache/SugarCacheAPC.php
+++ b/include/SugarCache/SugarCacheAPC.php
@@ -57,7 +57,7 @@ class SugarCacheAPC extends SugarCacheAbstract
             return false;
         }
 
-        if (function_exists("apc_store")
+        if (function_exists("apcu_store")
                 && empty($GLOBALS['sugar_config']['external_cache_disabled_apc'])) {
             return true;
         }
@@ -70,7 +70,7 @@ class SugarCacheAPC extends SugarCacheAbstract
      */
     protected function _setExternal($key, $value)
     {
-        apc_store($key, $value, $this->_expireTimeout);
+        apcu_store($key, $value, $this->_expireTimeout);
     }
 
     /**
@@ -78,7 +78,7 @@ class SugarCacheAPC extends SugarCacheAbstract
      */
     protected function _getExternal($key)
     {
-        $res = apc_fetch($key);
+        $res = apcu_fetch($key);
         if ($res === false) {
             return null;
         }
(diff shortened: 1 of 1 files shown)


simonr44:
Hi Gunnicom,

Thanks for contributing to SuiteCRM.

I have tested your changes and they work well.

Having looked at the code, there are some related changes that may be worthwhile.

- `apc_clear_cache` is used in `include/SugarCache/SugarCache.php` around lines 119
- `apc_delete_file` is used in `include/SugarCache/SugarCache.php` around lines 157

The check ` && ini_get('apc.stat') == 0` may no longer be required on the above also?

In addition it may be worth lowering the priority in  `include/SugarCache/SugarCacheAPC.php` line 49 
Changing `protected $_priority = 940;` to `protected $_priority = 890;` will ensure APCu is prioritised over other caches that may be available but not fully configured such as memcached.

gunnicom:
@simonr44 At first I thought about the file include/SugarCache/SugarCache.php, too, but on second look I assume, that that is only for OPcache part of APC, that vanished long ago. So the real solution would be to remove that part completely.

simonr44:
Agreed yes APCu no longer includes opcode caches, however clearing the APCu cache when this function is called may still have benefits. Perhaps the function name is no longer too relevant but it may still have value in one form or another.

gunnicom:
@simonr44 That could all be discussed in a different issue/PR. Keep this PR to fixing APCu functionality, and leave cleanup of old obsolete APC OPcache to a different PR.

simonr44:
There is a good chance we would look to bring these changes in together and test together so we'll make a new PR to cover those.

Thanks again for your contribution.

PR #10490, opened 2024-08-05
Fix #8838 - Change logo even if tmp folder contains periods

### LINKED ISSUE #8838: Can't verify uploaded images if tmp folder path contains folders' name with periods
We are using a virtual server from a server provider, where we host several SuiteCRM instances simultaneously (This worked fine with SugarCRM). This wasn't a problem until we realize that we couldn't upload a logo from the system settings UI. Trying to find out if the problem was related with file-permission, we saw the problem is related with folder's name containing periods. More especifically the issue happens in the file _include/utils.php_ function _verify_uploaded_image_ lines 5278-5300:

```
function verify_uploaded_image($path, $jpeg_only = false)
{
    $supportedExtensions = array('jpg' => 'image/jpeg', 'jpeg' => 'image/jpeg', 'tmp' => 'tmp');
    if (!$jpeg_only) {
        $supportedExtensions['png'] = 'image/png';
    }

    if (!file_exists($path) || !is_file($path)) {
        return false;
    }

    $img_size = getimagesize($path);
    $filetype = $img_size['mime'];
    $tmpArray = explode('.', $path);
    $ext = end($tmpArray);

    if (substr_count('..', $path) > 0 || ($ext !== $path && !isset($supportedExtensions[strtolower($ext)])) ||
            !in_array($filetype, array_values($supportedExtensions))
    ) {
        return false;
    }

    return verify_image_file($path, $jpeg_only);
}
```
The function will split the path using the period character, therefore if the path name has characters, this won't work. This only happens during the first verification, where the $ext and the $path variables should be the same (Please check the condition on line 5294).

Our server provider distributes the file-path hostings using their hosting name. The hosting name contain period therefore the folders name also contains periods. For example: 
Our hosting is alberto.sinergiacrm.org therefore the /tmp folder is placed in /usr/alberto.sinergiacrm.org/tmp/temporaryimagefilename
so the explode() and the end() function will result in org/tmp/temporaryimagefilename, the condition in line 5294 ($ext !== $path) will be met and the process stop with error.

Our server provider won't be able to change this temporary path.

#### Issue
If the full-path tmp folder contains periods. The system won't upload an image.

#### Expected Behavior
The image should be uploaded and set regardless the file-path of the /tmp folder of the hosting

#### Actual Behavior
The system setting UI will show an alert "The file format of the image must be JPEG or PNG. Upload a new file with the file extension .jpg or .png. " when trying to upload an image

#### Possible Fix
Check if the $ext variable contains a "/". If it does, it means it isn't an extension, therefore it should get the value of $path:
```
    $img_size = getimagesize($path);
    $filetype = $img_size['mime'];
    $tmpArray = explode('.', $path);
    $ext = end($tmpArray);
    // STIC 2020/08/06 : This condition will allow that the /tmp full-path contains periods
    if (strpos($ext, '/')) {
        $ext = $path;
    }
    // END STIC
    if (substr_count('..', $path) > 0 || ($ext !== $path && !isset($supportedExtensions[strtolower($ext)])) ||
            !in_array($filetype, array_values($supportedExtensions))
    ) {
        return false;
    }
```
#### Steps to Reproduce
1. Host a SuiteCRM instance where the /tmp folder is followed by a path containing periods. For example /test.test/tmp
2. Try to upload an image using the system setting UI.
3.
4.

#### Context
Changing the logo is a basic initial function for the usage of the tool.

#### Your Environment
SuiteCRM Version used: 7.11.15 (and SugarCRM 6.5)
Browser name and version: Chrome 83.0.4103 and Firefox 77.0.1 (64-bit)
Environment name and version: PHP 7.3, MySQL
Operating System and version: Debian 9

SinergiaCRM:
## Description
This PR checks if the variable $ext contains a "/". If so, it means it is not an extension, so it should get the value of $path

## Motivation and Context
A logo image can't be uploaded through Admin / System settings area if the tmp folder's name contains at least one period.

## How To Test This
1. Host a SuiteCRM instance where the tmp folder full path contains at least one period. For example /test.test/tmp
2. Try to upload an image using through the Admin / System settings area and checks it is working.

## Types of changes

### Final checklist


serhiisamko091184:
Thanks for the PR, @SinergiaCRM!

Regards,
Serhii

serhiisamko091184:
Hello @SinergiaCRM,

I'd like to ask about the suggested control structure. The strpos function might return different values - bool and non-bool. Some of them might be evaluated as FALSE, for instance, 0 or "". In case of returning 0 as an index of the position, it could be treated as FALSE. The logic of the fix might be scrambled. What do you think? 

Thanks in advance! 

Regards,
Serhii

SinergiaCRM:
> I'd like to ask about the suggested control structure. The strpos function might return different values - bool and non-bool. Some of them might be evaluated as FALSE, for instance, 0 or "". In case of returning 0 as an index of the position, it could be treated as FALSE. The logic of the fix might be scrambled. What do you think?

Hi @serhiisamko091184, you are right!

I see in [php.net](https://www.php.net/manual/en/function.strpos.php) that `strpos()` returns false if the needle was not found so I think we can solve it in the following way. 

```
if (strpos($ext, '/')) {
    $ext = $path;
}
```

I just uploaded the change to PR. 
Does this seem right to you?

PR #10527, opened 2024-09-16
Fix #10425 - Value of "currency_id" in custom table changes from NULL to an empty string when saving

### LINKED ISSUE #10425: Value of "currency_id" in custom table changes from NULL to an empty string when saving
#### Issue
If a module is configured with a custom currency field, an additional column called "currency_id" is also created.
When a new record is created this value is NULL.
When the record is updated this value is then changed to be an empty string.
If the currency field is used within a report with an aggregated function combined with grouping of another column, then the report shows duplicated records due to the differing values of NULL and an empty string.

#### Expected Behavior
The stored value of the "currency_id" should be consistent

#### Actual Behavior
For new records "currency_id" is NULL, once opened and resaved this value is then an empty string.

#### Steps to Reproduce
1. Add a currency field to a module
2. Create a new record and save it
3. Observe the value "currency_id" column in the custom field table is NULL
4. Create a second record
5. Open the second record and save it
6. Observe the value "currency_id" column in the custom field table is now an empty string

#### Context
Medium priority

#### Your Environment
* SuiteCRM Version used: 7.14.3
* Environment name and version (e.g. MySQL, PHP 7): PHP 8.1, MariaDB 10.11.6
* Operating System and version (e.g Ubuntu 16.04): Debian 12

serhiisamko091184:
## Description

See the issue: 
https://github.com/salesagility/SuiteCRM/issues/10425

## How To Test This

See the issue: 
https://github.com/salesagility/SuiteCRM/issues/10425

## Types of changes

### Final checklist

### DIFF
--- a/modules/DynamicFields/DynamicField.php
+++ b/modules/DynamicFields/DynamicField.php
@@ -5,7 +5,7 @@
  * SugarCRM, Inc. Copyright (C) 2004-2013 SugarCRM Inc.
  *
  * SuiteCRM is an extension to SugarCRM Community Edition developed by SalesAgility Ltd.
- * Copyright (C) 2011 - 2018 SalesAgility Ltd.
+ * Copyright (C) 2011 - 2024 SalesAgility Ltd.
  *
  * This program is free software; you can redistribute it and/or modify it under
  * the terms of the GNU Affero General Public License version 3 as published by the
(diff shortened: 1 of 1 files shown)


johnM2401 (dismissed):

gunnicom:
I would fix that globally by changing one line in D manager.php. I do not have my fix at hand because I am on vacation right now.
I think it is just adding a && strlen($data[$field])>0) to this line https://github.com/salesagility/SuiteCRM/blob/1a114a92e699d7ad5035518c31ee639fdc5cf825/include/database/DBManager.php#L546
Works for me since years and solves additional problems if for example you really need a unique column and so define it in the database table.

johnM2401 (approved):

jack7anderson7 (approved):

PR #10070, opened 2023-05-24
#10069 fix(Workflow) - create condition on related module date field

re8260:
I want to create condition on related module date field (Account-> Calls field:end date):
image

Before fix:
image

After fix:
image

## Description
Fix for https://github.com/salesagility/SuiteCRM/issues/10069

## Motivation and Context
Fix conditions on related module date field

## How To Test This
Use steps from issue https://github.com/salesagility/SuiteCRM/issues/10069

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [x] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change to the documentation.
- [x] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

PR #10534, opened 2024-09-30
Fix #10069 - Fix issue with workflow/report conditions not saving values when moving between modules

### LINKED ISSUE #10069: Workflow condition with related entity on date field type doesn't save the value correct
#### Issue
When you have condition on related entity on field of type date, condition will always fail because it save incorrect value and the date value will be removed from UI.
image
The selected date saved as string and can not be parsed letter
I will add example for Account and Call but it will work for any modules .
Don't check on fields Created and Modified date because it exists in all modules and will save correct
#### Expected Behavior
The value should be save as date

#### Actual Behavior
Date value should be saved as date

#### Possible Fix
Need to pas the related entity to function that parse value before save is

#### Steps to Reproduce

1. Crate new workflow on module Account (can be any other module) 
2. In condition select related module I will use Calls (can be any other module)  select date field,  TYPE=VALUE and pick a date 
image
3. Save workflow
image

#### Your Environment

* SuiteCRM Version used: Version 7.13.3

#### Context
Not possible to create conditions on date fields of related modules

JackBuchanan:
Add functionality to loop through relationship path to pass in correct module to fixUpFormatting

## Description
Add in new loop that goes through the relationship path (e.g. Accounts->Contacts->Calls) so that the correct module is passed into fixUpFormatting.

Previously the report/workflow base module is always passed into this function and if the relationship has a custom field that does not exist (e.g. Accounts->Contacts->Calls:custom_date) 

Based off of fix: https://github.com/salesagility/SuiteCRM/pull/10070 and expanded to cover reports and additional relationships

Credit to user https://github.com/re8260 for the initial fix.

## Motivation and Context
Fixes issues with date fields from related modules not saving

## How To Test This
Use steps from issue https://github.com/salesagility/SuiteCRM/issues/10069

## Types of changes

### Final checklist

### DIFF
--- a/modules/AOR_Conditions/AOR_Condition.php
+++ b/modules/AOR_Conditions/AOR_Condition.php
@@ -91,10 +91,11 @@ class AOR_Condition extends Basic
 
         $j = 0;
         foreach ((array)$postData as $i => $field) {
+            $condition_module = $_REQUEST['report_module'];
             if (!isset($post_data[$key . 'deleted'][$i])) {
                 LoggerManager::getLogger()->warn('AOR Condition trying to save lines but POST data does not contains the key "' . $key . 'deleted' . '" at index: ' . $i);
             }
-
+            $flow_bean = BeanFactory::newBean($_REQUEST['report_module']);
             if (isset($post_data[$key . 'deleted'][$i]) && $post_data[$key . 'deleted'][$i] == 1) {
                 $this->mark_deleted($post_data[$key . 'id'][$i]);
             } else {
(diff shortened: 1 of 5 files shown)


serhiisamko091184:
Hello @JackBuchanan! 

Thanks for your PR! 

Regards,
Serhii

johnM2401 (dismissed):
LGTM
Tried Date and Datetime fields on various relationships, including one 3-deep.
Tried on WF and Reports, with each Date and Time format.
image

PR #10240, opened 2023-10-17
Fix #10237 - Range validation in inline edition

SinergiaCRM:
- Closes #10237

## Description

The inline edition did not validate that the new value is within the range of a numerical field with a defined maximum and/or minimum.

The implemented solution has two parts:
- **controller.php**: fills in ```$validate_array``` the parameters for the range validation: the ```type``` as the validation type (```range```), ```max``` and ```min```
- **inlineEditing.js**: checks this validation type and for fields to validate as ```range```, call the defined function ```addToValidateRange```

## How To Test This

1. In Studio, create an integer field, with Min Value 1 and Max Value 15
2. Add the field to the detail and list view.
3. Verify that we can not add a value outside the defined range.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist

- [X] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change to the documentation.
- [X] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

PR #8858, opened 2020-08-28
Remove echo (old debugging code?)

pgorod:
## Description
Fixes this:

image

## How To Test This
1. Users module
2. Inline edit some field
3. See how the string `Fetched from:` appears in the field, followed by the id of the user to whom they report to.
4. Cringe

PHP Storm informs me that the only usage of this function, besides being called from User->Save, is from Test code.

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Final checklist
- [X] My code follows the code style of this project found [here](https://docs.suitecrm.com/community/contributing-code/coding-standards/).
- [ ] My change requires a change to the documentation.
- [X] I have read the [**How to Contribute**](https://docs.suitecrm.com/community/contributing-code/) guidelines.

PR #10537, opened 2024-10-04
Inline edit fixes and improvements

### LINKED ISSUE #8886: Inline edit triggers multiple saves, very slow
#### Issue
When using inline editing, there are two performance problems:

1. It uses two Ajax calls to the server at the start, when one would suffice (this is not a bug, just an unfortunate design)
2. It often fires more than one (sometimes a dozen!) Ajax calls saving the field, which is a slow operation. This is a bug! It's easy for a save operation to go over 10 seconds delay (which is unacceptable on a server with just one user and little data).

I've tracked it down to a call generated from a keypress event that fires more than once as long as the key is held down, even if just for a fraction of a second.

#### Expected Behavior
Inline edit should be quick and use just one Ajax call when starting, one when saving.

#### Possible Fix
In `include/InlineEditing/inlineEditing.js`, change this line

```
$(document).keypress(function(e) {
```
to 
```
$(document).off('keypress').on('keypress', function(e) {
```
This fixes the bug.

Then, to improve part 1. of the issue mentioned above, refactor these two calls to be done in just one round-trip to the server, joining the two results in a single payload:

```
//Do ajax call to retrieve the validation for the field.
var validation = getValidationRules(field, module, id);
//Do ajax call to retrieve the html elements of the field.
var html = loadFieldHTML(field, module, id);
```

I might provide these features as a PR as soon as I see that PR's are getting merged normally again. Right now I don't see the point in just adding to the list.

#### Your Environment
* SuiteCRM Version used: any recent version

### LINKED ISSUE #9818: Custom validation do not run on inline edit (7.12.7)

### LINKED ISSUE #10237: Inline edition do not validate ranges
#### Issue

Range validation does not work on an integer field in inline edition, in which a numerical range has been defined.
The problem occurs in the inline edition of any numerical field with a defined maximum and/or minimum: Inline edition does not validate that the new value is a value within the range.

However, it works correctly when editing from the edit view.

#### Expected Behavior

Validation should also be applied in inline edition.

#### Steps to Reproduce

1. In Studio, create an integer field, with Min Value 1 and Max Value 15
2. Add the field to the detail and list view and verify that we can add a value outside the defined range.
3. Verify that it works correctly in the editing view.

#### Your Environment

* SuiteCRM Version used: 7.14.0
* Browser name and version: Chrome Version 118.0.5993.70 (64-bit)
* Environment name and version: MySQL, PHP 7
* Operating System and version: Ubuntu 22.04.3

pgorod:
## Description
This PR...
- makes inline edit use only one back-end request instead of two, sometimes three. This improved speed is quite welcome, not just to allow users to move faster, but also to prevent UX errors such as when people redo or undo an action accidentally because of a delay. If you've used inline edit with a not-so-fast server, you know what I mean...
- moves a couple of auxiliary functions from Home controller to `inlineEditing.php`
- adds the standard validations, as defined in vardefs, to inline editing: `callback, range, isbefore, less, more, binarydep, comparison, in_array, verified`. Adds a new one, `isafter`.
- makes `inlineEditing.php` customizable
- fixes a few minor bugs

Closes #8886
Closes #9818
Closes #10237
Supersedes #8858
Supersedes #10240 

Thanks to @gunnicom and @SinergiaCRM for their code or comments in those previous PR's, they were very useful to me in this work.

## How To Test This

I suggest starting by adding this callback example that @gunnicom provided to the Opportunities vardefs, `name` field:

`modules/Opportunities/vardefs.php`
```
    'validation' => 
    array (
      'type' => 'callback',
      'callback' => 'function(formname,nameIndex){
                           var fieldvalue=$("#" + formname + " input[name=" + nameIndex + "]").val();
                           var returncode=true;
                           if(fieldvalue.trim()!=fieldvalue){
                               add_error_style(formname,nameIndex,"Space at the start or end of value not allowed " + nameIndex); 
                               returncode= false;
                           }
                           return returncode;
			}',
    ),
```
... then QR&R.

Now test with:
- random places throughout the app, doing inline edits
- number validation: test with Opportunities Probability (between 0 and 100)
- callback validation: test with Opportunities name adding spaces in the beginning or end
- int validation: test with Calls duration hours 

## Types of changes

### Final checklist

### DIFF
--- a/include/InlineEditing/InlineEditing.php
+++ b/include/InlineEditing/InlineEditing.php
@@ -81,7 +81,7 @@ function getEditFieldHTML($module, $fieldname, $aow_field, $view = 'EditView', $
             $vardef = $focus->getFieldDefinition($fieldname);
         }
 
-        $displayParams = array();
+        $displayParams = [];
         //$displayParams['formName'] = 'EditView';
 
         // if this is the id relation field, then don't have a pop-up selector.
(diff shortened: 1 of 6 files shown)


pgorod:
*Known Issue*: there's an edge case that doesn't work so well: in 'isbefore' validations, typically used when there are date ranges involved (start + end dates), the validation is only in effect in one of the two fields. In normal Edit views, both fields will be validated, so it's enough for one of them to throw the error.

But in inline edit, if you inline-edit the other one, no validation will take place, even if two relative dates are out of order. To fix this we would need also a `isafter` filter, and set validations on both fields. I added the code for that new validation type, but I didn't go and check all the vardefs where `isbefore` is used. I don't think this is worth much trouble, if somebody really cares about this, they can update their vardefs. 

Note that my PR doesn't introduce the problem, it was there before... I fix a lot of other things, though not this one. I provide means to facilitate fixing it, that's good enough for the time available to work on this... 👼

serhiisamko091184:
Hi @pgorod,

Thanks for your contribution! 

Regards,
Serhii

shubham-pawar:
@pgorod will this PR solved our below PHP error? 

```
PHP Fatal error:  Uncaught Error: Call to a member function format() on bool in suitecrm/include/InlineEditing/InlineEditing.php:472
Stack trace:
#0 /suitecrm/include/InlineEditing/InlineEditing.php(426): formatDisplayValue()
#1 /suitecrm/include/InlineEditing/InlineEditing.php(393): getDisplayValue()
#2 /suitecrm/modules/Home/controller.php(60): saveField()
#3 /suitecrm/include/MVC/Controller/SugarController.php(522): HomeController->action_saveHTMLField()
#4 /suitecrm/include/MVC/Controller/SugarController.php(491): SugarController->do_action()
#5 /suitecrm/include/MVC/Controller/SugarController.php(465): SugarController->handle_action()
#6 /suitecrm/include/MVC/Controller/SugarController.php(361): SugarController->process()
#7 /suitecrm/include/MVC/SugarApplication.php(101): SugarController->execute()
#8 /suitecrm/index.php(52): SugarApplication->execute()
#9 {main}
  thrown in /suitecrm/include/InlineEditing/InlineEditing.php on line 472
```

PR #10538, opened 2024-10-04
Fix #10243 - studio not saving field properties correctly

### LINKED ISSUE #10243: Studio - Fields reverting back to default
#### Issue

I'm having issues with the studio, while trying to edit fields on the Contacts module (This does not seem to be a module specific bug). When editing a standard field, such as phone_fax, it will sometimes reset back to its default values. If i for instance deletes the value in "Comment Text" and sets the required field to active and saves. This will sometimes success and sometimes not. In some cases the comment value will return to the field and sometimes not. Even if it succeeds, the entire field record will go back to default if saved once again.

#### Expected Behavior

The values of the forms should be set to what the user defines, not randomly shuffle on each safe.

#### Actual Behavior

After changing the values on the form and saving, the values will only in rare instances be saved as defined. Other times only some of the values will be saved. I've read through the logs, but it doesn't throw any error or warnings.

#### Possible Fix

I dont know the stack good enough, to know what is causing the bug.

#### Steps to Reproduce

1. Go to Admin -> Studio -> Contacts -> Fields -> phone_fax
2. Change the value of "Comment Text" to an empty value and set required field as true.
3. Click save
4. If the values are as expected, try to navigate away from the field and back again, now set default value as any kind of string and hit save.
5. The previous edited values should now be changed back to the default values.

#### Context

High priority - We cannot edit our dropdown values, which means it is displayed as empty, because the values doesnt match the options.

#### Your Environment

* SuiteCRM Version used: 8.4.1
* Browser name and version (e.g. Chrome Version 51.0.2704.63 (64-bit)): Chrome Version 117.0.5938.150 (Official Build) (64-bit)
* Environment name and version (e.g. MySQL, PHP 7): 8.1.24 and has been tested on 8.2
* Operating System and version (e.g Ubuntu 16.04): Debian GNU/Linux 11 (bullseye) and has also been tested on Ubuntu 20

pgorod:
## Description
Studio was not saving field properties correctly. This bug has been there for years...

## How To Test This
1. Enter Studio
1. Enter a module (such as `Cases`)
1. Click `Fields`
1. Enter a field  (such as `Description`) to see its properties:
 
image

5. Change one of the properties, 
6. Save
7. Exit that screen to the field list
8. Enter the same field again
9. Change a different property
10. Save
11. Exit and go back in
12. Both properties should be retained, instead of the first one you changed being reverted to a default.

The bug caused each subsequent save to throw away the previously saved non-default properties. I guess for a long time people work around this by just changing one property, or repeating the saves when they were mysteriously throw away...

During tests, it might be helpful to keep separate terminals watching these two files:
```bash
watch -n 1 cat custom/modules/Cases/Ext/Vardefs/vardefs.ext.php
watch -n 1 cat custom/Extension/modules/Cases/Ext/Vardefs/_override_sugarfield_description.php
```

Note that the `Save` action already includes a QR&R, so it is not necessary to do it separately. That's why the contents of those two files will appear in sync after each save.

There's a downside to this PR, although it is really not problematic. Some fields that were not changed, and still have their default values will be included in the saved files. This is because it is not trivial to know if a value is at its default - some of these things depend on configurable settings in other places in the system. So the conservative approach is taken: retain the values that were there when saved, even if it is not clear whether they are defaults and some redundancy occurs.

## Types of changes
- [-x] Bug fix (non-breaking change which fixes an issue)

### Final checklist

### DIFF
--- a/modules/ModuleBuilder/parsers/StandardField.php
+++ b/modules/ModuleBuilder/parsers/StandardField.php
@@ -8,7 +8,7 @@ if (!defined('sugarEntry') || !sugarEntry) {
  * SugarCRM, Inc. Copyright (C) 2004-2013 SugarCRM Inc.
  *
  * SuiteCRM is an extension to SugarCRM Community Edition developed by SalesAgility Ltd.
- * Copyright (C) 2011 - 2018 SalesAgility Ltd.
+ * Copyright (C) 2011 - 2024 SalesAgility Ltd.
  *
  * This program is free software; you can redistribute it and/or modify it under
  * the terms of the GNU Affero General Public License version 3 as published by the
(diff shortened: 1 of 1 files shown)


serhiisamko091184:
Hello @pgorod,

thanks for your PR! 

Regards,
Serhii

mattlorimer (approved):
LGTM 👍🏻

PR #10567, opened 2024-10-29
Fix #10566 - Incorrect date and time in the relationship with a person in the target lists

### LINKED ISSUE #10566: Incorrect date and time in the relationship with a person in the target lists
### Issue

When deleting a Target List that has related records, the value of the _date_modified_ field of the _prospect_lists_prospects_ table is modified, however the DB stores the local date, and not the UTC date and time as it should.

This happens because the field update is being done directly with SQL using the `NOW()` function.

### Possible Fix

Use the value of `UTC_TIMESTAMP()` from php instead of the `NOW()` function in:
https://github.com/salesagility/SuiteCRM/blob/5f8062ceca2cf4d1c365ba777e3080be8604e4e5/modules/ProspectLists/ProspectList.php#L430

```bash
1. Create a Target List
2. Add a person to the Target List
3. Check the date_modified field that appears in the prospect_lists_prospects table and note that the date and time is UTC.
4. Delete the Target List and note that the date_modified field has taken the local time and date
```

### Context

_No response_

### Version

7.14.4

Chrome

### Browser Version

_No response_

### Environment Information

PHP 7.4

Ubuntu 22.04

SinergiaCRM:
## Description

Fixing https://github.com/salesagility/SuiteCRM/issues/10566

When deleting a Target List that has related records, the value of the _date_modified_ field of the _prospect_lists_prospects_ table is modified, however the DB stores the local date, and not the UTC date and time as it should.

Solution:
The `NOW()` function that collected the local DB date and time from the _date_modified_ field has been changed to `UTC_TIMESTAMP()` that records it in UTC.

## Motivation and Context

Motivation: This happens because the field update is being done directly with SQL using the `NOW()` function.

## How To Test This

1. Create a Target List.
2. Add a person to the Target List.
3. Check the _date_modified_ field of the _prospect_lists_prospects_ table in DB and observe that the date and time is UTC.
4. Delete the Target List and verify that the _date_modified_ field has taken the UTC date and time.

## Types of changes

### Final checklist

### DIFF
--- a/modules/ProspectLists/ProspectList.php
+++ b/modules/ProspectLists/ProspectList.php
@@ -427,7 +427,7 @@ FROM prospect_lists_prospects plp
 
     public function mark_deleted($id)
     {
-        $query = "UPDATE prospect_lists_prospects SET deleted = 1, date_modified = NOW() WHERE prospect_list_id = '{$id}' AND deleted = 0";
+        $query = "UPDATE prospect_lists_prospects SET deleted = 1, date_modified = UTC_TIMESTAMP() WHERE prospect_list_id = '{$id}' AND deleted = 0";
         $this->db->query($query);
         return parent::mark_deleted($id);
     }


modules/ProspectLists/ProspectList.php
```diff
@@ -427,7 +427,7 @@ public function save($check_notify = false)
 
     public function mark_deleted($id)
     {
-        $query = "UPDATE prospect_lists_prospects SET deleted = 1, date_modified = NOW() WHERE prospect_list_id = '{$id}' AND deleted = 0";
+        $query = "UPDATE prospect_lists_prospects SET deleted = 1, date_modified = UTC_TIMESTAMP() WHERE prospect_list_id = '{$id}' AND deleted = 0";
```
mattlorimer, line 430: I don't believe this will work with MS SQL, so you will likely need to use the [DB convert](https://github.com/salesagility/SuiteCRM/blob/5f8062ceca2cf4d1c365ba777e3080be8604e4e5/include/database/DBManager.php#L3814) function here to support both engines and equally update that to use `UTC_TIMESTAMP` in MySQL , so this fix is applied globally when used

mattlorimer (changes requested):

PR #10686, opened 2025-06-15
Add IN and NOT IN operators to V8 API

kale1d0code:
Add IN and NOT IN operators to V8 API

## Description
Adds ```IN``` and ```NOT IN``` operators to V8 API
example useage within Filter Query String
```
filter[id][IN]=REDACTED,REDACTED,REDACTED,REDACTED,REDACTED,REDACTED,REDACTED,REDACTED
```
this change handles the case if the operator of the filter is either IN or NOT IN which is to check the value is a string, explode it and quote each value. The other operators behave as they done before. (I have changed how the quote wrapping is done as I don't how how much db->quote is doing to wrap values however I may adjust this after further investigation)

## Motivation and Context

I have built a GraphQL like API Client which consumes the SuiteCRM V8 API. when resolving relationships within the graph using the V8 relationship endpoint, the records returned do not contain relationship information; this requires an additional fetch. Without this change this graphing API client will need to make an additional fetch for each record within a relationship. The amount of fetching increases exponentially with the complexity of the graph it is resolving. This change allows the graphing API client to make one fetch for all records at once for a link to get the additional relationship information, whilst also allowing it to do other filtering for each tier of the graph.

The graphing API client doesn't internally use the NOT IN operator but this was included for other use cases (checking a value in a field is not in a list). The Addition of the IN and NOT IN operators significantly improves the performance of this graphing API client as well as enriches the filtering capabilities of the V8 API.

## How To Test This

execute a GET query to the SuiteCRM API which uses filters. and use this filter to get records which values are either in or not in the list you provide.

## Types of changes

### Final checklist

### DIFF
--- a/Api/V8/JsonApi/Repository/Filter.php
+++ b/Api/V8/JsonApi/Repository/Filter.php
@@ -13,6 +13,9 @@ class Filter
     public const OP_LTE = '<=';
     public const OP_LIKE = 'LIKE';
 
+    public const OP_IN = 'IN';
+    public const OP_NOT_IN = 'NOT IN';
+
     public const OP_AND = 'AND';
     public const OP_OR = 'OR';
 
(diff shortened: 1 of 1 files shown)


kale1d0code:
Hi this has been left over almost a year now.

is it likely that is will be part of the CRM system or am i going to have to make some kind of plugin and charge people for it?

PR #10783, opened 2026-03-04
fix(api): prevent invalid JSON when creating Users via v8 API

### LINKED ISSUE #9002: Invalid JSON response when creating user via V8 API
Hey everyone, I hope y'all enjoyed your holidays.
I want to create SuiteCRM users via the V8 api, but the response is invalid json and contains some generic validation error message.

#### Issue

Create a new User by calling:
```POST https://mysuitecrm.com/Api/V8/module```
```
{
  "data": {
    "type": "Users",
    "attributes": {
      "user_name": "mail@example.org",
      "first_name": "John",
      "last_name": "Doe",
      "external_auth_only": 1,
      "status": "Active",
      "email1": "mail@example.org",
      "email_addresses": "mail@example.org",
      "email_addresses_primary": "mail@example.org"
    }
  }
}
```

The user will be created, but the response looks like this:
```
{
    "data": {
        "type": "User",
        "id": "1acd4bbf-7f8c-4e6f-aa92-5fea6e799b74",
        "attributes": {...}
    }
}Sie müssen einen Benutzernamen und eine E-Mail Adresse angeben.
```

The response includes the error messages string in the end. It translate to "You must specify a username and an e-mail address".

#### Expected Behavior

There shouldn't be any clear-text error messages appended to the JSON

#### Actual Behavior
There is a generic error message string appended to the JSON response

#### Possible Fix
Not sure

#### Steps to Reproduce
See above. Create User via API

#### Context

It's a blocker for me, since the API response can't be parsed properly by my user creation script.

#### Your Environment

* SuiteCRM Version used: 7.11.15

MohamedMBG:
Fixes #9002 — Invalid JSON response when creating Users via the v8 API.

Problem

Creating a User through the v8 /Api/V8/module endpoint could return a response body that starts as valid JSON but then has extra plaintext appended (e.g. language strings like LBL_PROVIDE_USERNAME_AND_EMAIL or Access Denied). This breaks JSON parsing on API clients even though the user record is created successfully.

Root cause

During User->save() for new users, SuiteCRM may include modules/Users/GeneratePassword.php. That script historically behaves like a UI/AJAX entry point and uses echo to return status/messages. When it’s included during an API request, those echo outputs are written to stdout and end up in the API response body, corrupting the JSON.

Fix

Suppress stdout output from GeneratePassword.php when it’s executed in a non-direct context (i.e., included during a save/API flow). In that case, messages are logged instead of echoed, ensuring the v8 response remains valid JSON. Direct usage as an entry point/action continues to behave as before.

Tests

Adds a v8 API test that creates a User without user_hash (to exercise the original path) and asserts the response is valid JSON and parses cleanly.


shubham-pawar:
@MohamedMBG please sign the CLA.

PR #10789, opened 2026-03-22
Fix typo in timing global

martin-pettersson:
I'm not at all familiar with the code base but glossing through it this seems like a typo since I can't find another occurrence of this variable, but found a couple of references to `$GLOBALS['startTime']`.


shubham-pawar:
@martin-pettersson thank you!

PR #10869, opened 2026-06-29
Fix #10763 issue with SMTP credentials wrong after changing a field in outbound email settings

### LINKED ISSUE #10763: SMTP credentials wrong after changing a field in outbound email settings
### Issue

Without changing username or password (or anything important related to the SMTP connection), after re-saving the outbound mail record I can't send out mails anymore.

First it works after testing / sending the SMTP record.
With that:
Image
I assumed, I can change the name / signature without entering the password again.

However, after saving a new signature, the SMTP doesn't work anymore:

Image

```
3: Connection: opening to ssl://smtp.domain.com:465, timeout=300, options=array()
3: Connection: opened
2: SERVER -> CLIENT: 220 domain.com (usr106) Nemesis ESMTP Service ready

1: CLIENT -> SERVER: EHLO suitecrm892.demo

2: SERVER -> CLIENT: 250-web.de Hello suitecrm892.demo [89.244.94.37]
250-8BITMIME
250-AUTH PLAIN LOGIN
250 SIZE 141557760

1: CLIENT -> SERVER: AUTH LOGIN

2: SERVER -> CLIENT: 334 VXNlcm5hbWU6

1: CLIENT -> SERVER: ---obfuscated---
2: SERVER -> CLIENT: 334 UGFzc3dvcmQ6

1: CLIENT -> SERVER: ---obfuscated---
2: SERVER -> CLIENT: 535 Authentication credentials invalid

1: SMTP ERROR: Password command failed: 535 Authentication credentials invalid

3: SMTP Error: Could not authenticate.
1: CLIENT -> SERVER: QUIT

2: SERVER -> CLIENT: 221 domain.com Service closing transmission channel

3: Connection: closed
3: SMTP Error: Could not authenticate.
```

### Possible Fix

_No response_

```bash
1. I've setup an (SMTP) outbound email account and everything works fine.
2. Then I'll change sth. like the name of the record or the signature
3. Afterwards SMTP won't work anymore
```

### Context

_No response_

### Version

8.9.2

Chrome

### Browser Version

_No response_

### Environment Information

MariaDB, PHP8.3

Ubuntu 24

maku-92:
## Description

Fixes Outbound Email password bug. When re-saving a setting that is not the password in the Outbound Email settings, the password is still changed and the SMTP does not work anymore. See issue #10763

## Motivation and Context

Leaves the password unchanged when changing a different setting.

## How To Test This

1. Set up an Outbound Email account that requires a password
2. Go back and change a setting that is not the set password
3. Save
4. Send test email

## Types of changes

### Final checklist

### DIFF
--- a/modules/OutboundEmailAccounts/OutboundEmailAccounts.php
+++ b/modules/OutboundEmailAccounts/OutboundEmailAccounts.php
@@ -121,6 +121,7 @@ class OutboundEmailAccounts extends OutboundEmailAccounts_sugar
                 $this->mail_smtppass = $bean->mail_smtppass;
             }
         }
+
         $this->mail_smtppass = $this->mail_smtppass ? blowfishEncode(blowfishGetKey('OutBoundEmail'), $this->mail_smtppass) : null;
 
         if ($this->auth_type === 'basic') {
@@ -390,6 +391,11 @@ class OutboundEmailAccounts extends OutboundEmailAccounts_sugar
 
         foreach ($this->field_defs as $field => $field_def) {
             if (empty($field_def['display']) || $field_def['display'] !== 'writeonly') {
+                continue;
+            }
+
+            // Skip mail_smtppass field as it is handled separately handled in save() and retrieve() methods
+            if (($field_def['name'] ?? '') === 'mail_smtppass') {
                 continue;
             }
 


modules/OutboundEmailAccounts/OutboundEmailAccounts.php
```diff
@@ -121,7 +121,11 @@ public function save($check_notify = false)
                 $this->mail_smtppass = $bean->mail_smtppass;
             }
         }
-        $this->mail_smtppass = $this->mail_smtppass ? blowfishEncode(blowfishGetKey('OutBoundEmail'), $this->mail_smtppass) : null;
+
+        $fromPOST = isset($_POST['mail_smtppass']) && $_POST['mail_smtppass'] !== '';
+        if($fromPOST) {
+            $this->mail_smtppass = blowfishEncode(blowfishGetKey('OutBoundEmail'), $_POST['mail_smtppass']);
+        }
```
clemente-raposo: Hi @maku-92, its not a good practice to look into $_POST, $_REQUEST or $_GET with the bean save, it should be avoided ideally. As an alternative you can check the value against the previous value, that is available in the $fetched_row. If the current value is the same as the one on the fetched row, you can skip the enconding. 

An extra note, that in this scenario we cannot apply since this is using a specific blowfishkey, but it is also possible to use the `'db_encrypted' => true` could be set on the vardefs. However, this does not use the 'OutBoundEmail' blowfish key.

Plus the code above, is also working with mail_smtppass so we could re-write this to have a more unified code.

johnM2401 (dismissed):

johnM2401 (approved):

clemente-raposo (approved):

PR #10871, opened 2026-06-29
Improve sugar_mkdir error handling

pgorod:
## Enhanced error handling for directory creation

**TLDR; sugar_mkdir failures should be considered critical, and a proper message should _always_ be logged. Failures should be circumvented when possible.**

A good number of people showing up in the forums with broken initial installations is stumbling with the typical ownership-permissions errors. Although this problem is "classical", there are some new factors at play due to the change to Suite8 and to the modernization of PHP.

- Added fallback for permission errors which are appearing in modern Ubuntu's etc due to PHP-FPM hardening
- Improved logging in some early moments of installation when the normal log is not available yet or the log level is not sufficient. This really helps to prevent critical, yet silent failures that leave people clueless
- this new code is much better at showing permissions errors for what they are, instead of letting them creep up later as 
  - missing CSS
  - missing screen elements
  - during initial login as "Login credentials incorrect"

## Motivation and Context

I spent two days installing SuiteCRM on Ubuntu 26. I am a pro at installing SuiteCRM. If it takes me 2 days, most other people will likely just give up.

Granted, it took me 2 days because I wasn't just trying to get it to work, I was stubbornly trying to understand _why_ it wasn't working, connecting a debugger, making sure each problem I discovered was getting properly logged etc.

## How To Test This

This helper sugar_mkdir function is used all over the place, so it affects basically everything. But the two main test cases I would recommend looking at:

1. initial installation, in Ubuntu 26, with PHP-FPM, using a permissions scheme which includes a SetGID bit: `2775` or similar. 
2. in normal usage after already installed, delete the entire cache dir and see if recovers well: `sudo rm -rf /var/www/suitecrm/public/legacy/cache/`

You can also redo those two test cases but with broken permissions.

1. try an initial install with broken ownerships or permissions, check that you can see a relevant log message indicating this (might show up in `php_errors.log` or in `suitecrm.log` (v7) or in `logs/prod/prod.log` (v8)
2. in normal usage after already installed, delete the entire cache dir and recreate it manually with some incorrect ownerships, for example, so that it is not writable. Check that you can see a relevant log message indicating this.

## Types of changes

### Final checklist

### DIFF
--- a/include/utils/sugar_file_utils.php
+++ b/include/utils/sugar_file_utils.php
@@ -72,8 +72,26 @@ function sugar_mkdir($pathname, $mode = null, $recursive = false, $context = nul
     if (empty($mode)) {
         $mode = 0777;
     }
+
+    $result = false;
     if (empty($context)) {
-        $result = @mkdir($pathname, $mode, $recursive);
+        try {
+            $result = mkdir($pathname, $mode, $recursive);
+        } catch(Throwable $e) {
+            // for hardened PHP-FPM installs without CAP_FSETID, retry the mkdir
+            // but drop the special bits from the mode (e.g. 02775 is retried as 0775)
+            if (strpos($e->getMessage(), 'not permitted') !== false) {
+                $result = @mkdir($pathname, $mode & 0777, $recursive);
+            }
+        }
+        // mkdir() also emits E_WARNING on failure (not always catchable as exception),
+        // so check error_get_last() as a fallback for the same condition
+        if (!$result) {
+            $error = error_get_last();
+            if ($error !== null && strpos($error['message'], 'not permitted') !== false) {
+                $result = @mkdir($pathname, $mode & 0777, $recursive);
+            }
+        }
     } else {
         $result = @mkdir($pathname, $mode, $recursive, $context);
     }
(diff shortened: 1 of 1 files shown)


pgorod:
Have a look at how prevalent this problem is, especially since v8:

https://community.suitecrm.com/search?q=login%20credentials%20incorrect

Do you believe all those people fumbled their password creation and typing in their correct password? No!

The problem is that this error is a big mix of **all the different things that can make the login request to the back-end go wrong**:
- parts of the software are missing because cache dir was unwritable
- MariaDB is now really strange regarding initial admin passwords, so we can't login to connect to the DB
- connecting to DB works but maybe some weird encoding problem with strange characters in the password prevents SuiteCRM login
- ok, the password was actually incorrect.

pgorod:
### Fixed a case that wasn't handled well

👉🏼  Added commit to handle cases where **`$context`** parameter is passed.

### Always throw Exception?

An important discussion I was having with @clemente-raposo : it's probably a good idea to **always** throw the exception, not just send a message to `suitecrm.log`. This would mean that the app stops working when it can't write a new directory.

This is the relevant part of the code that we would change:
```php
    } else {
        $errorMessage = "Cannot create directory $pathname cannot be touched";
        if (is_null($GLOBALS['log'])) {
            throw new Exception("Error occurred in sugar_mkdir but the system doesn't have logger. Error message: \"$errorMessage\"");
        }
        $GLOBALS['log']->fatal($errorMessage);
    }

    return $result;
}
```

This would make SuiteCRM stricter in the way it handles improperly configured servers (which unfortunately are very common), and would make these misconfigurations more obvious, and break earlier. 

Some people might be surprised and be mad at us, even if this is for their own good. 😄 It's better to see a clear cause for fixing the system immediately, than getting parts of the software to break randomly and often without an obvious reason.

A way to mitigate these surprises could be to make this a two-step change:
- one release warns in the release notes that the change will happen, and the logging would also reflect that, so that people would see the warning in their logs ("if you don't fix these errors before v8.xx, your app will break when you upgrade")
- the next release makes the changes effective

----

For now, the PR does not move into the territory discussed here, so it should be ok to merge as-is without any serious consequences.

DJUserkent:
this is a great initiative @pgorod  - and would help so much to raise the profile of SuiteCRM as being: "well written code": to those users who are not really technical at file permissions level and all too easily can wrongly blame Suite

PR #10888, opened 2026-07-28
Closed pending review

javokhir-sec:


mattlorimer:
Do not post security fixes here  if you want to do anything security use the [security section](https://github.com/SuiteCRM/SuiteCRM/security)

